PR #167, opened 2024-06-03
ci: Use a custom httpbin instance in tests

### LINKED ISSUE #160: Run httpbin locally instead of using httpbin.org
(it's only used in tests)

janbuchar:
- closes #160


tests/unit/playwright_crawler/test_playwright_crawler.py
```diff
@@ -163,6 +165,6 @@ async def test_pre_navigation_hook() -> None:
     async def request_handler(_context: PlaywrightCrawlingContext) -> None:
         pass
 
-    await crawler.run(['https://example.com', 'https://httpbin.org'])
+    await crawler.run(['https://example.com', f'{httpbin}'])
```
vdusek, line 168: Couldn't we use just `str(httpbin)`?

tests/unit/browsers/test_playwright_browser_controller.py
```diff
@@ -41,7 +44,7 @@ async def test_initial_state(browser: Browser) -> None:
     assert controller.has_free_capacity
 
 
-async def test_open_and_close_page(controller: PlaywrightBrowserController, httpbin: str) -> None:
+async def test_open_and_close_page(controller: PlaywrightBrowserController, httpbin: URL) -> None:
     page = await controller.new_page()
     await page.goto(f'{httpbin}')
```
vdusek, line 49: could you please update it here to only `str(httpbin)` as well?

src/crawlee/http_crawler/_http_crawler.py
```diff
@@ -127,7 +127,11 @@ async def _handle_blocked_request(self, context: HttpCrawlingContext) -> AsyncGe
         if self._retry_on_blocked:
             status_code = context.http_response.status_code
 
-            if context.session and context.session.is_blocked_status_code(status_code=status_code):
+            if (
+                context.session
+                and status_code not in self._http_client._ignore_http_error_status_codes  # noqa: SLF001
+                and context.session.is_blocked_status_code(status_code=status_code)
+            ):
```
vdusek, line 134: Wondering whether this part should not look more like this:

```python
if context.session and context.session.is_blocked_status_code(
    status_code=status_code,
    additional_blocked_status_codes=self._http_client.additional_blocked_status_codes,
    ignore_http_error_status_codes=self._http_client.ignore_http_error_status_codes
):
```

which would require further changes as well, but since you're touching it here, I'm mentioning it ... Maybe do it in other PR. If I am not overlooking anything.

vdusek (approved):
Thanks for finishing this!

I have just one nit pick regarding the string conversion, and one suggestion for further improvement (maybe even a bug fix). But let's do it in a separate PR as it is not ci-related.

PR #207, opened 2024-06-21
feat: Maintain a global configuration instance

janbuchar:


vdusek (approved):
Just a thought on the naming...

I suggest using `get_global` instead of `get_global_configuration`. Since the method serves as constructor for the `Configuration` class - including "configuration" in the method name seems redundant. Leaving it up to you.

Otherwise, LGTM.

janbuchar:
> I suggest using `get_global` instead of `get_global_configuration`. Since the method serves as constructor for the `Configuration` class - including "configuration" in the method name seems redundant. Leaving it up to you.

I was ready to tell you that it's named that way because of JS crawlee, but there it's called `getGlobalConfig`. I'm lost :smile:

PR #210, opened 2024-06-24
fix: request order on resumed crawl

janbuchar:
This change makes https://github.com/apify/apify-sdk-python/blob/162ce1080d024fe2cf399534e8f960a584524232/tests/unit/actor/test_actor_memory_storage_e2e.py#L54 pass again.

The PR is a draft, it exists mostly so that I don't lose or forget this.


janbuchar:
#411 fixes this better

PR #222, opened 2024-06-24
feat: BasicCrawler.export_data helper

### LINKED ISSUE #211: `BasicCrawler.export_to` should redefine the `key` and `content_type` parameters
- the `key` parameter is mandatory, but the user has to browse the source code or wait for a crash to know that
- going blind, I tried passing `foo.csv` to `key` and got a file called `foo.csv.json`, which we may also want to handle somehow

janbuchar:
- closes #211


src/crawlee/storages/dataset.py
```diff
@@ -189,6 +190,32 @@ async def get_data(self, **kwargs: Unpack[GetDataKwargs]) -> DatasetItemsListPag
         # https://github.com/apify/apify-sdk-python/issues/140
         return await self._resource_client.list_items(**kwargs)
 
+    async def write_to(self, content_type: Literal['json', 'csv'], destination: TextIO) -> None:
```
vdusek, line 193: Is `write_to` named after the same method in Crawlee JS? Why not use `export_data` as on the crawler?

vdusek (changes requested):
LGTM, just naming and one more thing: the class `ExportToKwargs` - couldn't it be removed now?

src/crawlee/storages/dataset.py
janbuchar: No, that's mine. I imagine having `export_data` *and* `export_to` would be confusing. If you have a better name in mind, I'm open to suggestions.

janbuchar:
> the class `ExportToKwargs` - couldn't it be removed now?

No, it's still used in `Dataset.export_to`.

vdusek (approved):

PR #226, opened 2024-06-24
feat: add support decompress *br* response content

Mantisus:
### Description

Adds support for decompressing a brotli-compressed server response if an "Accept-Encoding" header was passed with the "br"  or "gzip, deflate, br" parameter and the server supports this type of compression.

Httpx provides the necessary functionality, but requires additional libraries to be installed to work correctly.


vdusek (approved):
Thanks for the contribution!

PR #229, opened 2024-06-25
feat: automatic logging setup

### LINKED ISSUE #214: Automate default logging configuration for crawlers
Currently, when a user runs their crawler, no log is printed, leaving users unaware of the crawler's progress and actions.

We have a log formatter already ready, but users need to configure it manually. Like this:

```python
import asyncio
import logging

from crawlee.beautifulsoup_crawler import BeautifulSoupCrawler, BeautifulSoupCrawlingContext
from crawlee.log_config import CrawleeLogFormatter

# Configuration the logging
handler = logging.StreamHandler()
handler.setFormatter(CrawleeLogFormatter(include_logger_name=True))
root_logger = logging.getLogger()
root_logger.setLevel(logging.INFO)
root_logger.addHandler(handler)


async def main() -> None:
    crawler = BeautifulSoupCrawler()

    async def request_handler(context: BeautifulSoupCrawlingContext) -> None:
        await context.enqueue_links()
        data = {
            'url': context.request.url,
            'title': context.soup.title.string,
        }
        await context.push_data(data)

    await crawler.run(['https://crawlee.dev'])


if __name__ == '__main__':
    asyncio.run(main())
```

Can we make this setup a default? Possible solution: importing any module from Crawlee should configure the root logger automatically.

janbuchar:
- closes #214

### DIFF
--- a/src/crawlee/basic_crawler/basic_crawler.py
+++ b/src/crawlee/basic_crawler/basic_crawler.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import asyncio
+import logging
 import signal
 import sys
 import tempfile
@@ -10,7 +11,6 @@ from collections.abc import AsyncGenerator, Awaitable, Sequence
 from contextlib import AsyncExitStack, suppress
 from datetime import timedelta
 from functools import partial
-from logging import getLogger
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, AsyncContextManager, Callable, Generic, Literal, Union, cast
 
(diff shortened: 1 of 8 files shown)


src/crawlee/basic_crawler/basic_crawler.py
```diff
@@ -679,6 +702,7 @@ async def __run_task_function(self) -> None:
             send_request=self._prepare_send_request_function(session, proxy_info),
             add_requests=result.add_requests,
             push_data=self._push_data,
+            log=self._logger,
```
janbuchar, line 705: @vdusek not sure about this - it may be nice to make a new logger with a name like `crawlee.request_handler`, but that doesn't match any module, so we'd break a convention.

vdusek: it will log request queue / request list otherwise - am I correct? if so, I'd not override it

vdusek (approved):
LGTM :rocket:

src/crawlee/basic_crawler/basic_crawler.py
janbuchar: more like `crawlee.basic_crawler.basic_crawler`, depending on your crawler

PR #231, opened 2024-06-25
fix: Selector handling in BeautifulSoupCrawler enqueue_links

### LINKED ISSUE #230: Bug enqueue_links functionality for beautifulsoup_crawler
The "selector" variable that the function takes is not a selector, but the name of the tag. Since find_all is used

find_all - does not support standard types of selectors (CSS, Xpath) and for filtering elements, it requires specifying target attributes using named variables or attrs. [Docs](https://beautiful-soup-4.readthedocs.io/en/latest/index.html?highlight=find_all#find-all)

Example for "a.navbar__item":
```python
soup.find_all("a", class_="navbar__item")
soup.find_all("a", attrs={"class": "navbar__item"})
```

For correct work with CSS selector, it is necessary to use - select. [Docs](https://beautiful-soup-4.readthedocs.io/en/latest/index.html?highlight=select#css-selectors)

```python
soup.select("a.navbar__item")
```

Therefore, with the current implementation of the function it is impossible to set a selector providing more accurate filtering of links

janbuchar:
- closes #230

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [0.0.7](../../releases/tag/v0.0.7) - Unreleased
+
+### Fixes
+
+- selector handling in `enqueue_links` in `BeautifulSoupCrawler`
+
 ## [0.0.6](../../releases/tag/v0.0.6) - Unreleased
 
 ### Adds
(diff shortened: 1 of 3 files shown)


vdusek (approved):
LGTM

PR #234, opened 2024-06-26
fix: handle blocked request

Mantisus:
### Description

Replaced `find` to `select_one` because `find` does not support CSS selectors, which caused incorrect behavior.

### Related issues

#230


janbuchar (changes requested):
Thank you for your contribution! Before we merge this, please run `make format` and update the changelog.

janbuchar (approved):

PR #237, opened 2024-06-26
feat: Project templates

### LINKED ISSUE #215: Project bootstrapping
- [ ] add a templates directory with cookiecutter projects - https://cookiecutter.readthedocs.io/en/stable/advanced/directories.html
- [ ] make a simple CLI so that `pipx run crawlee create [--template some-template-name] project-dir/` works
  - [ ] the CLI should be just a simple wrapper around `cookiecutter` that always uses the current master (for fetching templates)
  - [ ] it could be beneficial to use `inquirer`
  - available templates should be
    - [ ] ??? simple playwright project
    - [ ] ??? simple beautifulsoup project

janbuchar:
- closes #215

to test this:

```sh
cookiecutter gh:apify/crawlee-python --checkout project-templates --directory templates/beautifulsoup
```

or even better (the `--spec ...` part won't be necessary after merge)
```sh
pipx run --spec git+https://github.com/apify/crawlee-python.git@project-templates crawlee create
```


pyproject.toml
```diff
@@ -92,6 +95,9 @@ proxy-py = "^2.4.4"
 beautifulsoup = ["beautifulsoup4", "lxml", "html5lib"]
 playwright = ["playwright"]
 
+[tool.poetry.scripts]
+crawlee = "crawlee.cli:cli"
```
janbuchar, line 99: This should enable us to do `pipx run crawlee create` once this hits PyPI

pyproject.toml
```diff
@@ -64,6 +64,9 @@ python-dateutil = "^2.9.0"
 sortedcollections = "^2.1.0"
 typing-extensions = "^4.1.0"
 tldextract = "^5.1.2"
+cookiecutter = "^2.6.0"
+typer = {extras = ["all"], version = "^0.12.3"}
+inquirer = "^3.3.0"
```
janbuchar, line 69: This was added because JS crawlee uses the original version, but it doesn't have type stubs, so mypy complains. I think we can remove it if @B4nan agrees and just use whatever `typer` offers

src/crawlee/cli.py
```diff
@@ -0,0 +1,83 @@
...
+                inquirer.Text(
+                    'project_name',
+                    message='Name of the new project folder',
+                    validate=lambda _, it: len(it) > 0,
+                    ignore=project_name is not None,
+                ),
+                inquirer.List(
+                    'template',
+                    message='Please select the template for your new Crawlee project',
+                    choices=[(choice[0].upper() + choice[1:], choice) for choice in template_choices],
```
janbuchar, line 66: This is not as pretty as a handwritten description, but we don't have to maintain a manifest. It sounds like an OK tradeoff to me, but I'm willing to discuss this.

src/crawlee/cli.py
```diff
@@ -0,0 +1,83 @@
...
+    else:
+        template_choices = []
+
+    answers = (
+        inquirer.prompt(
+            [
+                inquirer.Text(
+                    'project_name',
+                    message='Name of the new project folder',
+                    validate=lambda _, it: len(it) > 0,
```
janbuchar, line 60: This prevents us from using the defaults in `cookiecutter.json`. Not sure if that's a problem.

janbuchar:
This is still pretty barebones (see my comments), but I'd prefer to get the CLI to PyPI as quickly as possible and iterate upon that.

pyproject.toml
janbuchar: Look like neither `typer` nor `rich` support selection from list, which is a shame. So we're stuck with `inquirer` for now.

B4nan (approved):
gave this a quick look and i would merge it asap, better have some templates than no templates at all, we can improve over time

PR #251, opened 2024-06-28
chore: use latest crawlee version in templates

vdusek:
so we do not have to do things like https://github.com/apify/crawlee-python/pull/247


janbuchar (approved):
Thanks

PR #262, opened 2024-07-02
docs: autogenerated api docs

barjin:
Ports the API docs generator from other Python packages (https://github.com/apify/apify-client-python). Includes small tweaks for `@property` decorators and dataclasses.

image


.github/workflows/docs.yml
```diff
@@ -29,6 +29,17 @@ jobs:
                     corepack enable
                     corepack prepare yarn@stable --activate
 
+            -   name: Set up Python
+                uses: actions/setup-python@v5
+                with:
+                    python-version: 3.8
```
vdusek, line 35: Why 3.8? Couldn't we use newer version?

vdusek (approved):
Good job, thanks! Just one question regarding the Python version for building docs.

B4nan:
I am struggling a bit with running this locally, had to wipe the lockfile to be able to install, and `make build-api-reference` is failing for me with this

```
ModuleNotFoundError: No module named 'apify_shared'
error: pathspec 'v' did not match any file(s) known to git
```

But if it works for @vdusek its a go from me.

barjin:
Ah good point @B4nan , I commented this in one place, but not in the other. This is (afaik) `apify-client` quirk that was there because of some reexports, probably not important for us (definitely not right now).

.github/workflows/docs.yml
barjin: again, just a copy-paste from `apify-client` 

I have no idea what the latest version is (or what might break here), so I just went with it 😅

B4nan:
Still getting this for the `yarn` install

image

Wiping it helps, and I can run the docs finally. It looks good for the initial version.

B4nan:
so its not just me, it is broken

PR #270, opened 2024-07-06
fix: crawlee local storage dir environment variable name

janbuchar:


tests/unit/memory_storage_client/test_memory_storage_client.py
```diff
@@ -71,20 +71,20 @@ async def test_persist_storage(tmp_path: Path) -> None:
 
 def test_persist_storage_set_to_false_via_string_env_var(monkeypatch: pytest.MonkeyPatch, tmp_path: Path) -> None:
     monkeypatch.setenv('CRAWLEE_PERSIST_STORAGE', 'false')
-    ms = MemoryStorageClient(Configuration(crawlee_local_storage_dir=str(tmp_path)))  # type: ignore
+    ms = MemoryStorageClient(Configuration.model_validate({'local_storage_dir': str(tmp_path)}))
```
B4nan, line 74: shouldnt this be also `crawlee_storage_dir`? or maybe just `storage_dir`? the "config option" should be just `storage_dir` i think (we don't have one in the JS version and only deal with this in the call sites)

PR #274, opened 2024-07-08
chore: update check changelog entry script

vdusek:


vdusek:
converting back to draft, gonna update the regex instead

B4nan (approved):

PR #320, opened 2024-07-17
doc: tiered proxy urls

### LINKED ISSUE #311: Document tiered proxies

janbuchar:
- closes #311

### DIFF
--- a/docs/guides/proxy_management.mdx
+++ b/docs/guides/proxy_management.mdx
@@ -86,7 +86,7 @@ When no `session_id` is provided, our proxy URLs are rotated round-robin.
 
 ### Tiered proxies
 
-When you use HTTP proxies in real world crawling scenarios, you have to decide which type of proxy to use to reach a sweetspot between cost efficiency and reliably avoiding blocking. Some websites may allow crawling with no proxy, on some you may get away with using datacenter proxies, which are cheap but likely to get blocked, and sometimes you need to use expensive residential proxies.
+When you use HTTP proxies in real world crawling scenarios, you have to decide which type of proxy to use to reach the sweet spot between cost efficiency and reliably avoiding blocking. Some websites may allow crawling with no proxy, on some you may get away with using datacenter proxies, which are cheap but likely to get blocked, and sometimes you need to use expensive residential proxies.
 
 To take the guesswork out of this process, Crawlee allows you to configure multiple tiers of proxy URLs. When crawling, it will automatically pick the left-most tier where it doesn't encounter blocking. If you organize your proxy server URLs in tiers so that the left-most tier contains the cheapest, least reliable ones and each higher tier contains more expensive, more reliable ones, you will get an optimal anti-blocking performance.
 
--- a/docs/guides/proxy_management.mdx
+++ b/docs/guides/proxy_management.mdx
@@ -21,6 +21,9 @@ import SessionPlaywrightSource from '!!raw-loader!./proxy_management_examples/se
 import InspectionBeautifulSoupSource from '!!raw-loader!./proxy_management_examples/inspecting_beautifulsoup.py';
 import InspectionPlaywrightSource from '!!raw-loader!./proxy_management_examples/inspecting_playwright.py';
 
+import TiersBeautifulSoupSource from '!!raw-loader!./proxy_management_examples/tiers_beautifulsoup.py';
+import TiersPlaywrightSource from '!!raw-loader!./proxy_management_examples/tiers_playwright.py';
+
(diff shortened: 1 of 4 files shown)


docs/guides/proxy_management.mdx
```diff
@@ -81,6 +84,25 @@ When no `session_id` is provided, our proxy URLs are rotated round-robin.
     </TabItem>
 </Tabs>
 
+### Tiered proxies
+
+When you use HTTP proxies in real world crawling scenarios, you have to decide which type of proxy to use to reach a sweetspot between cost efficiency and reliably avoiding blocking. Some websites may allow crawling with no proxy, on some you may get away with using datacenter proxies, which are cheap but likely to get blocked, and sometimes you need to use expensive residential proxies.
```
barjin, line 89:
```suggestion
When you use HTTP proxies in real world crawling scenarios, you have to decide which type of proxy to use to reach the sweet spot between cost efficiency and reliably avoiding blocking. Some websites may allow crawling with no proxy, on some you may get away with using datacenter proxies, which are cheap but likely to get blocked, and sometimes you need to use expensive residential proxies.
```

docs/guides/proxy_management.mdx
```diff
@@ -81,6 +84,25 @@ When no `session_id` is provided, our proxy URLs are rotated round-robin.
...
+
+To take the guesswork out of this process, Crawlee allows you to configure multiple tiers of proxy URLs. When crawling, it will automatically pick the left-most tier where it doesn't encounter blocking. If you organize your proxy server URLs in tiers so that the left-most tier contains the cheapest, least reliable ones and each higher tier contains more expensive, more reliable ones, you will get an optimal anti-blocking performance.
```
barjin, line 91: I'm fine with it either way, but I always thought of tiers as top-down layers - so the first one is the "lowest", while the last one is the "top tier" (this is afaik an actual English phrase).

docs/guides/proxy_management.mdx
```diff
@@ -81,6 +84,25 @@ When no `session_id` is provided, our proxy URLs are rotated round-robin.
...
+
```
barjin, line 92: Also, maybe mention that the proxies in the active tier are rotated round robin, the 2D array of URLs can get confusing (my fault 😅)

src/crawlee/proxy_configuration.py
barjin, line 1: You sure this should be here? just surprised to see actual code change in a docs PR, so making sure this is intended :)

barjin (approved):
More or less nits, lgtm otherwise 👍 🎉

src/crawlee/proxy_configuration.py
janbuchar: It is foul play, but I saw an unused property while adding a docblock and just removed it. We're pre-1.0, yolo.

vdusek (approved):
LGTM

docs/guides/proxy_management.mdx
```diff
@@ -81,6 +84,25 @@ When no `session_id` is provided, our proxy URLs are rotated round-robin.
     </TabItem>
 </Tabs>
 
+### Tiered proxies
+
+When you use HTTP proxies in real world crawling scenarios, you have to decide which type of proxy to use to reach the sweet spot between cost efficiency and reliably avoiding blocking. Some websites may allow crawling with no proxy, on some you may get away with using datacenter proxies, which are cheap but likely to get blocked, and sometimes you need to use expensive residential proxies.
```
B4nan, line 89: > which are cheap but likely to get blocked

maybe better say they are easily detected instead? idk

docs/guides/proxy_management.mdx
B4nan: yeah agreed

src/crawlee/proxy_configuration.py
```diff
@@ -113,9 +112,19 @@ def __init__(
...
     ) -> None:
-        self._configuration = configuration or Configuration.get_global_configuration()
+        """Initialize a proxy configuration object.
+
+        Exactly one of `proxy_urls`, `tiered_proxy_urls` or `new_url_function` must be specified.
+
+        Args:
+            proxy_urls: A list of URLs of proxies that will be rotated in a round-robin fashion
+            tiered_proxy_urls: A list of URL tiers (where a tier is a list of proxy URLs). Crawlers will automatically
+                try to use the left-most (lowest index) tier where blocking does not happen. The proxy URLs in
```
B4nan, line 123: this also uses left/right instead of top/bottom

docs/guides/proxy_management.mdx
janbuchar: okay, so higher index ~ higher tier? I'll probably explain the relation of high/low tier to the index in the array anyway

B4nan: yeah, higher is more expensive and better working, that way it feels relatively understandable :D

barjin (approved):
👍

PR #323, opened 2024-07-18
chore: add doc-related commands to contributing and makefile

vdusek:


CONTRIBUTING.md
```diff
@@ -100,6 +100,12 @@ Enable Corepack, which installs Yarn automatically:
 corepack enable
 ```
 
+Build the API reference:
+
+```sh
+./build_api_reference.sh
```
janbuchar, line 106:
```suggestion
make build-api-reference
```

janbuchar (approved):
LGTM

CONTRIBUTING.md
vdusek: it would not work as we are in website/ directory

janbuchar: Huh. Do we need to be there?

PR #325, opened 2024-07-18
feat: integrate proxies into `PlaywrightCrawler`

vdusek:
### Description

- feat: integrate proxies into `PlaywrightCrawler`

### Issues

- N/A

### Testing

- Reproduce it with the following:

```python
import asyncio

from crawlee.playwright_crawler import PlaywrightCrawler, PlaywrightCrawlingContext


async def main() -> None:
    proxy_configuration = ProxyConfiguration(proxy_urls=['http://USERNAME:PASSWORD@HOSTNAME:PORT'])
    crawler = PlaywrightCrawler(proxy_configuration=proxy_configuration)

    async def request_handler(context: PlaywrightCrawlingContext) -> None:
        data = {
            'url': context.request.url,
            'title': await context.page.title(),
            'content': await context.page.content(),
        }

        context.log.info(f'Extracted data: {data}')
        await context.push_data(data)

    await crawler.run(['http://httpbin.org/ip'])


if __name__ == '__main__':
    asyncio.run(main())
```

### Checklist

- [ ] CI passed


src/crawlee/browsers/playwright_browser_plugin.py
```diff
@@ -58,12 +58,12 @@ def browser_type(self) -> Literal['chromium', 'firefox', 'webkit']:
 
     @property
     @override
-    def browser_options(self) -> Mapping[str, Any]:
+    def browser_options(self) -> MutableMapping[str, Any]:
```
janbuchar, line 61: Making this mutable is almost the same as making the property public. Why does it need to be mutable in the first place?

src/crawlee/browsers/playwright_browser_controller.py
```diff
@@ -69,9 +71,21 @@ def is_browser_connected(self) -> bool:
...
+        self,
+        page_options: MutableMapping[str, Any] | None = None,
+        proxy_info: ProxyInfo | None = None,
+    ) -> Page:
         page_options = page_options or {}
 
+        # If "proxy_info" is provided and no proxy is already set in "page_options", configure the proxy.
+        if proxy_info and 'proxy' not in page_options:
+            page_options['proxy'] = {
+                'server': f'{proxy_info.scheme}://{proxy_info.hostname}:{proxy_info.port}',
```
janbuchar, line 84: We have multiple libraries that can compose URLs for us, why not use one of them? :slightly_smiling_face:

src/crawlee/browsers/playwright_browser_plugin.py
vdusek: Well, only `page_options` have to be mutable since we add the proxy info there. Therefore I updated both page & browser options. Shall I revert it for the `browser_options`?

janbuchar: Couldn't we just make a copy when passing the options to `new_page`? Mutating the page options inside the browser plugin may introduce bugs...

src/crawlee/browsers/playwright_browser_controller.py
vdusek: I'm not sure if using HTTPX outside of the `httpx_client` module is a good idea, especially for such trivial operations.

janbuchar: Valid point. Maybe there is similar functionality in pydantic, but I'm shooting blind here.

B4nan: this feels like bikeshedding to me, `f'{proxy_info.scheme}://{proxy_info.hostname}:{proxy_info.port}'` looks just fine, lets focus on more important stuff :]

janbuchar (approved):

PR #330, opened 2024-07-19
ci: autogenerate changelog

### LINKED ISSUE #18: Generate changelog from the commit messages
Generate CHANGELOG from the commit messages as we do in JS/TS projects.

Once this is solved for this repository, please create the same issue in the SDK, Client, and Shared Python repositories.

@vladfrangu suggested to explore https://github.com/orhun/git-cliff as a solution to this.

Edit: also check this discussion: https://github.com/apify/crawlee-py/pull/125#discussion_r1576067168

janbuchar:
- closes #18


janbuchar:
Made this a draft, I'll try to update this so that it works with workflow_dispatch

janbuchar:
@vdusek @B4nan I haven't run this yet, I imagine there will be quite some hotfixing. Also version bumps are still quite spaghettific.

.github/workflows/_publish_to_pypi.yaml
janbuchar, line 1: - The release artifacts seem bordeline useless - I hope nobody installs packages by downloading .whl files from github. If someone tried, dependency management would be hell. Are there any points against dropping them?
- It is nice that we make tags for beta releases, but shouldn't we make github pre-releases instead?

janbuchar: Github pre-releases can swamp the release list page - see https://github.com/vercel/next.js/releases. You can use the `prerelease:false` search filter, but still.

B4nan: I don't like to pollute anything else than the package manager for prereleases, I would remove the git tags as well (I did in some repositories already, e.g. in the JS monorepos we maintain), since I can't justify adding tags to every single commit to master.

.github/workflows/run_release.yaml
```diff
@@ -8,22 +8,22 @@ on:
...
         required: true
         type: choice
-        default: alpha
+        default: auto
         options:
-          - alpha
-          - beta
-          - final
+          - auto
+          - custom
```
B4nan, line 21: i would like to be able to trigger major/minor/patch specifically (just like we have it in the js crawlee) instead of relying on the git messages. some feature commits are not really worth a feature release (but you want them in the changelog and its just wrong to call them fixes just because of this)

.github/workflows/_publish_to_pypi.yaml
janbuchar: Sure, removing the git tags is also an option. We'd lose a bit of traceability (what changed between those betas?), but if we keep releasing often, nobody will care :slightly_smiling_face:

.github/workflows/run_release.yaml
janbuchar: ...it looks like `poetry version` could also work for us.

.github/workflows/_publish_to_pypi.yaml
B4nan: If every commit to master is a beta, it should be pretty clear what changed between those betas.

.github/workflows/run_release.yaml
B4nan: How is this about git cliff? You use that to generate changelog between two git tags or commit hashes I'd guess, or its not working that way? The option I am after would only let you decide the version bump manually instead of relying on the git messages.

.github/workflows/_publish_to_pypi.yaml
janbuchar: What I mean is, if you have `0.42.2b73` installed, how do you look up what changed between that and the current `0.42.2b91` if there are no tags?

.github/workflows/run_release.yaml
janbuchar: [This is how](https://github.com/apify/crawlee-python/pull/330/files#diff-4992bcb5677b9499cbec92fcb2bc14c4e0232be0e127673730e7a6582e594c3dR60-R66) - it has a flag that outputs the new semver based on commits since the last tag. I used it because it was handy. The `workflow_dispatch` can also accept a custom version number, which will be used directly.

.github/workflows/_publish_to_pypi.yaml
vdusek: > If every commit to master is a beta, it should be pretty clear what changed between those betas.

Clarification, it's not every commit, the chore/docs commits are ignored.

.github/workflows/run_release.yaml
```diff
@@ -36,20 +36,109 @@ jobs:
     name: Check whether to release
     if: |
       github.event_name == 'workflow_dispatch' ||
-      github.event_name == 'release' ||
       (
         github.event_name == 'push' &&
         !startsWith(github.event.head_commit.message, 'docs') &&
```
vdusek, line 41:
```suggestion
        !startsWith(github.event.head_commit.message, 'doc') &&
```

.github/workflows/run_release.yaml
B4nan: > Also, kinda off topic, but some people (such as @netmilk IIRC) would tell you that attaching emotional value to version numbers is wrong and you should just follow the semantic versioning spec like a robot. Every side of the argument has its points IMO.

It's not emotional value, if you change something that has no effect on the user facing API (and therefore on docs), that is not a feature release to me. Yet it could be a feature commit (think of a small improvement of something). Also being able to market feature releases is a good thing.

For me this is about being in control, not about emotions. If you commit one thing with a wrong prefix (and that happens all the time, remember the small breaking change with URLs from yesterday?), you are stuck with what you have in master.

vdusek (commented):
It'd be great to have [Check PR title](https://github.com/apify/crawlee/blob/master/.github/workflows/check-pr-title.yml) workflow here as well.

janbuchar:
> It'd be great to have [Check PR title](https://github.com/apify/crawlee/blob/master/.github/workflows/check-pr-title.yml) workflow here as well.

Sure, added.

.github/workflows/_publish_to_pypi.yaml
B4nan: > What I mean is, if you have 0.42.2b73 installed, how do you look up what changed between that and the current 0.42.2b91 if there are no tags?

You would check the git history, but I agree its a chore to find what commit triggered what version (since you need to go through the CI logs). For me this is not something I would optimize for, we don't want users to be on the beta releases (only if they don't want to wait for us to do a stable release). 

To me, it's just adding (a lot of) noise, for a very low value, but that does not mean I don't see any value - so if you guys want to keep the git tags there, let's keep them (and we can remove later if you change your mind, heh). I would skip the GH releases, that really feels too much - they would always contain a single item only.

> Clarification, it's not every commit, the chore/docs commits are ignored.

I don't see a problem with that either, we would ignore those commits in the changelog too.

.github/workflows/run_code_checks.yaml
```diff
@@ -23,10 +23,6 @@ jobs:
     name: Run unit tests
```
vdusek, line 23: I suppose "Check PR title" workflow should be added here.

.github/workflows/check_pr_title.yaml
```diff
@@ -0,0 +1,14 @@
...
+        types: [ opened, edited, synchronize ]
+
+jobs:
+    check_pr_title:
+        name: 'Check PR title'
+        runs-on: ubuntu-latest
+        steps:
+            -   uses: amannn/action-semantic-pull-request@v5.5.2
+                env:
+                    GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
vdusek, line 14: Sorry but could we stay with the 2 depth indentation like in other files?

.github/workflows/run_release.yaml
vdusek: I agree with @B4nan on this.

.github/workflows/run_release.yaml
B4nan: the commit prefix is called `docs:` not `doc:`

https://www.conventionalcommits.org/en/v1.0.0/#summary

vdusek (changes requested):
Also please remove `check-changelog-entry` from `.pre-commit-config.yaml`.

.github/workflows/run_release.yaml
vdusek: Oh, I based this on what we have in `cliff.toml`, in that case, it should be updated there from `doc` to `docs`.

.github/workflows/run_release.yaml
B4nan: > You can already set a custom version, can't that be used to achieve the same thing you could do with a patch/minor/major toggle?

Yes, it can be used for the same, but is less ergonomic. It means you need to check the version you are on, instead of just releasing the next bump based on the type. You add another possibility for the user to screw it up. I want something in between, not full auto, but not full custom. We are using this for quite some time in crawlee, and it feels really nice, so I would like to see it the same way - especially because we want to port this setup to other repositories in the long run.

> While we're at the topic, can we not fix incorrect commit types by push-forcing to master?

Force push to master is a big no-no, especially in open-source repositories. This can piss all the other maintainers off easily, seen that many times in nette community, as the author liked to do this a lot. You break their local versions since you change commit hashes.

There are exceptions to everything, but I can justify force pushing only for the very last commit done by yourself, within very short time window after they were pushed (lets say minutes).

janbuchar: > Yes, it can be used for the same, but is less ergonomic. It means you need to check the version you are on, instead of just releasing the next bump based on the type. You add another possibility for the user to screw it up. I want something in between, not full auto, but not full custom. We are using this for quite some time in crawlee, and it feels really nice, so I would like to see it the same way - especially because we want to port this setup to other repositories in the long run.

Sure, I'm in favor of that, but I'd prefer to do that after this PR is merged and tested. Autogenerated changelogs and triggering releases via workflow_dispatch are enough of a milestone IMO.

> Force push to master is a big no-no, especially in open-source repositories. This can piss all the other maintainers collaborators off easily, seen that many times in nette community, as the author liked to do this a lot. You break their local versions since you change commit hashes.

Haha, I remember that. I agree that the negative impact on contributors is not worth it.

.github/workflows/_publish_to_pypi.yaml
janbuchar: I'm fine with removing the beta tags - every simplification of that workflow counts.

janbuchar: Can I also not upload the release artifacts? (see the argument above)

.github/workflows/run_release.yaml
B4nan: Ok, I am fine merging this without it, that's fair. But I would like to see it implemented before next release.

.github/workflows/_publish_to_pypi.yaml
B4nan: You mean completely, including for stable releases? I guess its fine too, I also don't think people use them, and we can revert this if we hear from someone who would like to do so.

.github/workflows/run_release.yaml
janbuchar: I suspect that git-cliff only puts `^doc` in the default configuration to catch both `doc:` and `docs:` so that they please everyone :smile: 

The newly added PR title check should prevent any errors anyway, but I can update `cliff.toml` too.

.github/workflows/_publish_to_pypi.yaml
vdusek: Regarding the beta tags - I'm ok with both keeping/removing them, so do as you prefer.

PR #341, opened 2024-07-22
bug: minor log fix

souravjain540:
### Description

wrong command in the log: 'pip install crawlee[playwright]', it should be "pip install 'crawlee[playwright]'"

### Issues

- N/A

### Testing

- N/A

### Checklist

- [ ] Changes are described in the `CHANGELOG.md`
- [ ] CI passed

### DIFF
--- a/src/crawlee/beautifulsoup_crawler/__init__.py
+++ b/src/crawlee/beautifulsoup_crawler/__init__.py
@@ -3,8 +3,8 @@ try:
     from .types import BeautifulSoupCrawlingContext
 except ImportError as exc:
     raise ImportError(
-        'To import anything from this subpacakge, you need to install the "beautifulsoup" extra. '
-        'For example, if you use pip, run "pip install crawlee[beautifulsoup]".',
+        "To import anything from this subpacakge, you need to install the 'beautifulsoup' extra."
+        "For example, if you use pip, run `pip install 'crawlee[beautifulsoup]'`.",
     ) from exc
 
 __all__ = ['BeautifulSoupCrawler', 'BeautifulSoupCrawlingContext']
--- a/src/crawlee/beautifulsoup_crawler/__init__.py
+++ b/src/crawlee/beautifulsoup_crawler/__init__.py
@@ -3,7 +3,7 @@ try:
     from .types import BeautifulSoupCrawlingContext
 except ImportError as exc:
     raise ImportError(
-        "To import anything from this subpacakge, you need to install the 'beautifulsoup' extra."
+        "To import anything from this subpackage, you need to install the 'beautifulsoup' extra."
         "For example, if you use pip, run `pip install 'crawlee[beautifulsoup]'`.",
     ) from exc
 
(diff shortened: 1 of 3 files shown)


vdusek:
- I understand the reasoning behind using single quotes around the package name with the extra, as it helps with shell compatibility when dealing with brackets.
- The backticks... Since this is just a log line without any formatting, I'm unsure whether using backticks (`) isn't just less readable.

What do you think @janbuchar?

src/crawlee/beautifulsoup_crawler/__init__.py
```diff
@@ -3,8 +3,8 @@
     from .types import BeautifulSoupCrawlingContext
 except ImportError as exc:
     raise ImportError(
-        'To import anything from this subpacakge, you need to install the "beautifulsoup" extra. '
-        'For example, if you use pip, run "pip install crawlee[beautifulsoup]".',
+        "To import anything from this subpacakge, you need to install the 'beautifulsoup' extra."
```
janbuchar, line 6:
```suggestion
        "To import anything from this subpackage, you need to install the 'beautifulsoup' extra."
```

src/crawlee/browsers/__init__.py
```diff
@@ -4,8 +4,8 @@
     from .playwright_browser_plugin import PlaywrightBrowserPlugin
 except ImportError as exc:
     raise ImportError(
-        'To import anything from this subpacakge, you need to install the "playwright" extra. '
-        'For example, if you use pip, run "pip install crawlee[playwright]".',
+        "To import anything from this subpacakge, you need to install the 'playwright' extra."
```
janbuchar, line 7:
```suggestion
        "To import anything from this subpackage, you need to install the 'playwright' extra."
```

src/crawlee/playwright_crawler/__init__.py
```diff
@@ -3,8 +3,8 @@
     from .types import PlaywrightCrawlingContext
 except ImportError as exc:
     raise ImportError(
-        'To import anything from this subpacakge, you need to install the "playwright" extra. '
-        'For example, if you use pip, run "pip install crawlee[playwright]".',
+        "To import anything from this subpacakge, you need to install the 'playwright' extra."
```
janbuchar, line 6:
```suggestion
        "To import anything from this subpackage, you need to install the 'playwright' extra."
```

janbuchar (approved):
@vdusek I think the backticks are fine. We may want to use colors instead at some point, but there's no need to stall such a minor PR because of that.

B4nan:
hmm, we need the PR title check here. `bug:` prefix for any PR makes no sense unless you planned to introduce a new bug with the PR :D

janbuchar:
> hmm, we need the PR title check here. `bug:` prefix for any PR makes no sense unless you planned to introduce a new bug with the PR :D 

I mean, we can update the log generator to put bug: in the same bucket as fix:

souravjain540:
> hmm, we need the PR title check here. `bug:` prefix for any PR makes no sense unless you planned to introduce a new bug with the PR :D 

Hehehe, sorry! still used to the last company, there we were using "bug: for bug fixes", "feat: for new features" and similarly

B4nan:
> I mean, we can update the log generator to put bug: in the same bucket as fix:

or we can actually follow the specs and not use random prefixes like `bug:` :D

B4nan:
here is the PR check we use in other repositories

https://github.com/apify/crawlee/blob/master/.github/workflows/check-pr-title.yml

Issue #300, opened 2024-07-15
Add URL validation


vdusek:
Currently, for end-user "requests" params we define:

```python
requests: Sequence[str | BaseRequestData | Request]
```

which, unfortunately, matches a single-string request (e.g. `requests='https://crawlee.dev'`) as well. 

We could make it more strict and update it to just:
```python
requests: list[str | BaseRequestData | Request]
```

so static type checkers can let us know in case someone is trying to pass a single string request there.

Or maybe use some custom type for either list, tuple, or set.

Or, add some URL pattern-matching validation, see Pydantic's [Network Types](https://docs.pydantic.dev/latest/api/networks/).

PR #343, opened 2024-07-22
feat: add URL validation

vdusek:
### Description

- Add [HttpUrl](https://docs.pydantic.dev/latest/api/networks/#pydantic.networks.HttpUrl) Pydantic type for URLs - URL validation.
- A lot of tests had to be adjusted.
- Some changes regarding relative/absolute URLs had to be implemented as well.

### Issues

- Closes: #300

### Testing

- N/A

### Checklist

- [x] CI passed

### DIFF
--- a/src/crawlee/_utils/requests.py
+++ b/src/crawlee/_utils/requests.py
@@ -4,7 +4,9 @@ import re
 from base64 import b64encode
 from hashlib import sha256
 from logging import getLogger
-from urllib.parse import parse_qsl, urlencode, urlparse
+from urllib.parse import parse_qsl, urlencode, urljoin, urlparse
+
+from pydantic import HttpUrl
 
 from crawlee._utils.crypto import compute_short_hash
 
@@ -128,3 +130,17 @@ def compute_unique_key(
 
     # Return the normalized URL as the unique key.
     return normalized_url
+
+
+def is_url_absolute(url: str | HttpUrl) -> bool:
+    """Check if a URL is absolute."""
+    url = url if isinstance(url, str) else str(url)
+    return bool(urlparse(url).netloc)
+
+
+def make_url_absolute(base_url: str | HttpUrl, relative_url: str | HttpUrl) -> HttpUrl:
+    """Make a relative URL absolute by combining it with a base URL."""
+    base_url = base_url if isinstance(base_url, str) else str(base_url)
+    relative_url = relative_url if isinstance(relative_url, str) else str(relative_url)
+    absolute_url = urljoin(base_url, relative_url)
+    return HttpUrl(absolute_url)
--- a/src/crawlee/_utils/requests.py
+++ b/src/crawlee/_utils/requests.py
@@ -4,9 +4,7 @@ import re
 from base64 import b64encode
 from hashlib import sha256
 from logging import getLogger
-from urllib.parse import parse_qsl, urlencode, urljoin, urlparse
-
-from pydantic import HttpUrl
+from urllib.parse import parse_qsl, urlencode, urlparse
 
 from crawlee._utils.crypto import compute_short_hash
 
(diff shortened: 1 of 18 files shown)


src/crawlee/_utils/requests.py
```diff
@@ -128,3 +130,17 @@ def compute_unique_key(
...
+    url = url if isinstance(url, str) else str(url)
+    return bool(urlparse(url).netloc)
+
+
+def make_url_absolute(base_url: str | HttpUrl, relative_url: str | HttpUrl) -> HttpUrl:
+    """Make a relative URL absolute by combining it with a base URL."""
+    base_url = base_url if isinstance(base_url, str) else str(base_url)
+    relative_url = relative_url if isinstance(relative_url, str) else str(relative_url)
+    absolute_url = urljoin(base_url, relative_url)
+    return HttpUrl(absolute_url)
```
janbuchar, line 146: This could go into a separate `urls` module, I guess.

src/crawlee/beautifulsoup_crawler/beautifulsoup_crawler.py
```diff
@@ -129,8 +130,12 @@ async def enqueue_links(
                 if label is not None:
                     link_user_data.setdefault('label', label)
 
-                if (href := link.attrs.get('href')) is not None:
-                    requests.append(BaseRequestData.from_url(href.strip(), user_data=link_user_data))
+                if (url := link.attrs.get('href')) is not None:
+                    if not is_url_absolute(url):
+                        url = str(make_url_absolute(context.request.url, url))
+
+                    url = url.strip()
```
janbuchar, line 137: Could we do this before the absolutifying?

src/crawlee/models.py
```diff
@@ -19,7 +19,7 @@ class BaseRequestData(BaseModel):
 
     model_config = ConfigDict(populate_by_name=True)
 
-    url: Annotated[str, Field(min_length=1)]
+    url: Annotated[HttpUrl, Field()]
```
janbuchar, line 22: I'd wager the `Field()` annotation is redundant here. But I'm not sure myself if I'd prefer to keep it or not.

src/crawlee/models.py
```diff
@@ -71,6 +71,13 @@ def from_url(
 
         return result
 
+    def get_query_param_from_url(self, param: str) -> str | None:
```
janbuchar, line 74: Could we add a `default` parameter to this? It's against YAGNI, but I think it's still justifiable.

src/crawlee/playwright_crawler/playwright_crawler.py
```diff
@@ -126,15 +128,19 @@ async def enqueue_links(
                 elements = await crawlee_page.page.query_selector_all(selector)
 
                 for element in elements:
-                    href = await element.get_attribute('href')
+                    url = await element.get_attribute('href')
+
+                    if url:
+                        if not is_url_absolute(url):
+                            url = str(make_url_absolute(context.request.url, url))
+                        url = url.strip()
```
janbuchar, line 136: Like in `BeautifulSoupCrawler`, could we put this higher?

janbuchar (commented):
This must have been pretty tedious. I'm sure it's a step in the right direction though, good job. Please see my comments.

src/crawlee/_utils/requests.py
vdusek: sure, why not

src/crawlee/models.py
vdusek: pylance is against removing it, so let's leave it there

src/crawlee/beautifulsoup_crawler/beautifulsoup_crawler.py
vdusek: yep, it makes sense

src/crawlee/models.py
vdusek: hmm, what default would you add?

janbuchar: I mean something like what `dict.get` goes. By default, the default should be `None`

janbuchar (approved):

PR #349, opened 2024-07-23
doc: extract URL as string in examples

vdusek:


README.md
```diff
@@ -108,7 +108,7 @@ async def main() -> None:
 
         # Extract data from the page.
         data = {
-            'url': context.request.url,
+            'url': str(context.request.url),
```
B4nan, line 111: so was this a string before #343? if so, it feels quite breaking. also i am honestly not sure if its a good idea to have something else than a string URL in the context

janbuchar:
Hm... what happens if you don't stringify the URL?

vdusek:
@janbuchar 

> Hm... what happens if you don't stringify the URL?

Log:

```log
[crawlee.playwright_crawler.playwright_crawler] INFO  Extracted data: {'url': Url('https://apify.com/'), 'title': 'Apify: Full-stack web scraping and data extraction platform'}
```

Dataset (it's converted to string somewhere):
```json
{
  "url": "https://apify.com/",
  "title": "Apify: Full-stack web scraping and data extraction platform"
}
```

janbuchar:
Well, honestly that's better than what I expected. 

However, @B4nan also raises a valid point. We may want to keep the `AnyHttpUrl` for the `url` attribute so that we still have validation and add a `url_string` property :shrug:.

vdusek:
Closing, `context.request.url` should stay string.

PR #353, opened 2024-07-24
refactor: `context.request.url` is string

vdusek:
### Description

- `context.request.url` is a string, but URL validation is kept

### Issues

- N/A

### Testing

- Tests pass

### Checklist

- [x] CI passed

### DIFF
--- a/src/crawlee/_utils/urls.py
+++ b/src/crawlee/_utils/urls.py
@@ -1,19 +1,34 @@
 from __future__ import annotations
 
-from urllib.parse import urljoin, urlparse
+from urllib.parse import parse_qs, urljoin, urlparse
 
-from pydantic import HttpUrl
+from pydantic import AnyHttpUrl, TypeAdapter
 
 
-def is_url_absolute(url: str | HttpUrl) -> bool:
+def is_url_absolute(url: str) -> bool:
     """Check if a URL is absolute."""
-    url = url if isinstance(url, str) else str(url)
     return bool(urlparse(url).netloc)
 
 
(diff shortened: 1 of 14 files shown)


tests/unit/memory_storage_client/test_request_queue_client.py
```diff
@@ -119,7 +118,7 @@ async def test_get_record(request_queue_client: RequestQueueClient) -> None:
 
     request = await request_queue_client.get_request(processed_request.id)
     assert request is not None
-    assert request.url == request.url == HttpUrl(request_url)
+    assert request.url == request.url == request_url
```
janbuchar, line 121: Um, why do we need to test that `request.url` is equal to itself?

src/crawlee/_utils/urls.py
```diff
@@ -1,19 +1,34 @@
...
+    """Validate the given HTTP URL.
+
+    Raises:
+        pydantic.error_wrappers.ValidationError: If the URL is not valid.
+    """
+    if value is None:
+        return None
+    adapter = TypeAdapter(AnyHttpUrl)
+    adapter.validate_python(value)
+    return value
```
janbuchar, line 34:
```suggestion
_http_url_adapter = TypeAdapter(AnyHttpUrl)

def validate_http_url(value: str | None) -> str | None:
    """Validate the given HTTP URL.

    Raises:
        pydantic.error_wrappers.ValidationError: If the URL is not valid.
    """
    if value is not None:
        _http_url_adapter.validate_python(value)

    return value
```

It is wasteful to make a new `TypeAdapter` every time.

janbuchar (changes requested):

janbuchar (approved):
Nice!

PR #357, opened 2024-07-25
chore: add basic badges to readme

vdusek:
If we would like to add a test coverage badge, we would need to utilize some 3rd part service like [Codecov](https://codecov.io/) or [Coveralls](https://coveralls.io/).

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,21 @@
     <small>A web scraping and browser automation library</small>
 </h1>
 
+<p align=center>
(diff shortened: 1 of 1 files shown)


B4nan:
I was using coveralls for a few years, and I am not ready to step into that shithole again, their platform keeps failing randomly (breaking builds for no reason), and the UI/UX is weird/ugly at best. Moved to covecov some time ago, and its much better, I am fine with that.

Issue #351, opened 2024-07-24
How to reset crawlee URL cache?


GHOST-LOVE-YOU:
I want to use fastapi+crawlee to scrape the posts content of a forum. I hope that my crawler will re-run every time a post request is made, but actually, when I request a second time, the links requested before will not be requested again. I have learned two solutions from [this post](https://stackoverflow.com/questions/74709844/how-to-reset-crawlee-url-cache):

1. Set environment variables
I did this, but it didn't work

```python
@app.post("/crawl")
async def crawl(auth_data: AuthData):
    if not authenticate(auth_data.username, auth_data.password):
        raise HTTPException(status_code=401, detail="Unauthorized")
    
    if not crawler_semaphore.locked():
        async with crawler_semaphore:
            try:
                crawl_results.clear()
                await run_crawler()
                return {"status": "success", "data": crawl_results}
            except Exception as e:
                raise HTTPException(status_code=500, detail=str(e))
    else:
        raise HTTPException(status_code=429, detail="Crawler is already running. Please try again later.")
```

```python
async def run_crawler():
    config = Configuration(
        persist_storage=False,
        purge_on_start=True,
    )
    crawler = PlaywrightCrawler(
        max_requests_per_crawl=500,
        request_handler=router,
        configuration=config,
    )
    unique_key = str(uuid.uuid4())
    request = BaseRequestData(url='https://localhost:8888', unique_key=unique_key)
    
    await crawler.run([request])
```

2. Modify the unique_key for each request
I did this in baseUrl, it's cool and it really worked, but I don't know how to add unique_key for enqueue_links

```python
await context.enqueue_links(
    selector=f"a[href='{post['url']}']",
    label='DETAIL',
)
```
Thank you very much for your help!

janbuchar:
Hi, and thanks for your interest in Crawlee! Setting `persist_storage = False` and `purge_on_start = True` should be enough for the RequestQueue not to be persisted to the filesystem. Could you check what's in your `storage/request-queues/default` directory?

GHOST-LOVE-YOU:
Thank you for your quick response!
Below is the content of my storage/request-queues/default:
```
> tree storage/
storage/
├── key_value_stores
│   └── default
│       └── __metadata__.json
└── request_queues
    └── default
        ├── 11ZDDqZVzTDFo3c.json
        ├── 2I0wNHjqjrkrZ3G.json
        ├── __metadata__.json
        ├── NejkxElN87hwno9.json
        ├── ......
[...]

Issue #152, opened 2024-05-14
Implement purging request queue in repeated `BasicCrawler.run` calls


janbuchar:
See https://github.com/apify/crawlee/blob/2d5d443da5fa701b21aec003d4d84797882bc175/packages/basic-crawler/src/internals/basic-crawler.ts#L836-L845 for inspiration

PR #375, opened 2024-07-29
fix: pass crawler configuration to storages

janbuchar:
This makes sure that storages opened via `BasicCrawler.get_dataset` and the like will use the same configuration object as the crawler. While this is relevant to #351, it does not fully resolve it (see #152).


vdusek (approved):
LGTM

Issue #292, opened 2024-07-12
Add new HTTP client `curl-cffi`


siddiqkaithodu:
Since we use [HTTPX](https://pypi.org/project/httpx/) , Isn't better to use [curl_cffi](https://pypi.org/project/curl-cffi/) instead or use it an optional client that switches when the HTTPX gets defeated in the AntiBot war.

I am thinking about intelligently switching between multiple http clients. (The same as of proxy rotation)

vdusek:
Hi, thanks for the issue. I think we can split this one into two parts. First, we can (probably, needs further investigation) switch from `httpx` to `curl_cffi`, which seems better meet our use-case. The second part would be providing the end user an option to specify the HTTP client.

PR #387, opened 2024-08-01
feat: add new curl impersonate HTTP client

vdusek:
### Description

- Add a new curl impersonate HTTP client utilizing the [curl-cffi](https://pypi.org/project/curl-cffi/) package.
- Improve API docs of the HTTP clients and define public & private interfaces.
- I encountered a few bugs or "not great behaviour" of `curl-cffi`, opening issues:
  - https://github.com/yifeikong/curl_cffi/issues/360
  - https://github.com/yifeikong/curl_cffi/issues/361
- Because of the above bugs, I decided not to set curl impersonate client as default and stay with HTTPX client.
- I also had to move some general components from the `basic_crawler` module to the root of the package. Maybe it's not good, so I am open to other options on how to sort it out.

### Issues

- Closes: #292

### Testing

- New unit tests were written.
- Or check the example below utilizing curl impersonate client with BeautifulSoup crawler.

```python
import asyncio

from crawlee.beautifulsoup_crawler import BeautifulSoupCrawler, BeautifulSoupCrawlingContext
from crawlee.http_clients import CurlImpersonateHttpClient


async def main() -> None:
    proxy_configuration = ProxyConfiguration(
        proxy_urls=[
            'http://username:password@proxy.apify.com:8000',
        ],
    )

    crawler = BeautifulSoupCrawler(
        max_requests_per_crawl=10,
        proxy_configuration=proxy_configuration,
        http_client=CurlImpersonateHttpClient(),
    )

    async def request_handler(context: BeautifulSoupCrawlingContext) -> None:
        context.log.info(f'Processing {context.request.url}...')
        await context.enqueue_links()

        data = {
            'url': context.request.url,
            'title': context.soup.title.text if context.soup.title else '',
        }

        context.log.info(f'Extracted data: {data}')
        await context.push_data(data)

    await crawler.run(['https://apify.com', 'https://crawlee.dev/'])
    crawler.log.info('Finished crawling.')


if __name__ == '__main__':
    asyncio.run(main())
```

### TODO

- [ ] Before merging add better documentation of HTTP clients, an option of switching them, and implementing a new one.

### Checklist

- [x] CI passed


vdusek:
@siddiqkaithodu Hey, good point, I'm gonna expose kwargs for the `curl_cffi.requests.AsyncSession`.

src/crawlee/http_clients/base_http_client.py
```diff
@@ -2,71 +2,97 @@
...
 
     @property
     def status_code(self) -> int:
-        """HTTP status code of the response (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status)."""
+        """The HTTP status code received from the server."""
 
     @property
-    def headers(self) -> Headers:
-        """Response headers."""
+    def headers(self) -> dict[str, str]:
```
janbuchar, line 26: That's probably a good call. However, `httpx.Headers` has some subtle differences from `dict`, such as being case-insensitive. This may lead to some annoying bugs when switching HTTP clients. Could we make a `Mapping` subclass that lowercases the header names? Also this should probably be immutable...

src/crawlee/beautifulsoup_crawler/beautifulsoup_crawler.py
```diff
@@ -55,22 +56,27 @@ def __init__(
...
-            context.session,
-            context.proxy_info,
-            self._statistics,
+            request=context.request,
+            session=context.session,
+            proxy_info=context.proxy_info,
+            statistics=self._statistics,
+        )
+
+        self.log.info(
```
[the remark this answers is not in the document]
vdusek: This was not intended to add to the codebase. I'll remove it. Thanks.

src/crawlee/beautifulsoup_crawler/beautifulsoup_crawler.py
```diff
@@ -55,22 +56,27 @@ def __init__(
 
         kwargs.setdefault(
             'http_client',
-            HttpxClient(
+            HttpxHttpClient(
                 additional_http_error_status_codes=additional_http_error_status_codes,
                 ignore_http_error_status_codes=ignore_http_error_status_codes,
             ),
         )
+        self._http_client: BaseHttpClient  # since we set it in the line above, it should be initialized
```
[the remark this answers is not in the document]
vdusek: Oh, this should not be here, thanks.

src/crawlee/http_clients/httpx_http_client.py
```diff
@@ -39,18 +59,8 @@ async def handle_async_request(self, request: httpx.Request) -> httpx.Response:
...
-
-    if any(needle in str(error) for needle in ROTATE_PROXY_ERRORS):  # noqa: SIM103
-        return True
-
-    return False
-
-
-class HttpxClient(BaseHttpClient):
-    """A httpx-based HTTP client used for making HTTP calls in crawlers (`BasicCrawler` subclasses)."""
+class HttpxHttpClient(BaseHttpClient):
```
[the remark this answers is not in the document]
vdusek: I was just following the pattern of using "library name" + "HttpClient" as a suffix for consistency, even though it doesn't add much information in this particular case.

src/crawlee/http_clients/base_http_client.py
vdusek: I added a new type `HttpHeaders`.

pyproject.toml
```diff
@@ -50,6 +50,7 @@ aioshutil = "^1.3"
 beautifulsoup4 = { version = "^4.12.3", optional = true }
 colorama = "^0.4.6"
 cookiecutter = "^2.6.0"
+curl-cffi = "^0.7.1"
```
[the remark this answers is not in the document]
vdusek: `curl-cffi` is optional dependency ... I use the name `curl-impersonate` for it.

pyproject.toml
```diff
@@ -50,6 +50,7 @@ aioshutil = "^1.3"
 beautifulsoup4 = { version = "^4.12.3", optional = true }
 colorama = "^0.4.6"
 cookiecutter = "^2.6.0"
+curl-cffi = { version = "^0.7.1", optional = true }
```
janbuchar, line 53: Can we also suggest installing the extra on import error?

janbuchar (changes requested):

pyproject.toml
vdusek: it's there, isn't it? 

https://github.com/apify/crawlee-python/blob/16c67959af41b5043ccf91cf82167e8069f9d63e/src/crawlee/http_clients/__init__.py

or what do u have in mind?

janbuchar: Oh right, sorry!

janbuchar (approved):

PR #393, opened 2024-08-03
feat(playwright): `infinite_scroll` helper

janbuchar:
This implements a bare bones infinite scroll helper. It is also added into the crawling context.

### DIFF
--- a/cliff.toml
+++ b/cliff.toml
@@ -15,7 +15,7 @@ All notable changes to this project will be documented in this file.\n
 # https://keats.github.io/tera/docs/#introduction
 body = """
 {% if version %}\
-    ## [{{ version | trim_start_matches(pat="v") }}](<REPO>/releases/tag/{{ version }}) ({{ timestamp | date(format="%Y-%m-%d") }})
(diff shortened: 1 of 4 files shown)


barjin:
Why not [infinite_scroll](https://github.com/apify/crawlee/blob/master/packages/playwright-crawler/src/internals/utils/playwright-utils.ts#L396)?

(reading this on mobile, sorry if I'm missing something)

B4nan:
Yeah, I would stay consistent with the JS version.

janbuchar:
Sure thing, I'll change it

vdusek (approved):
LGTM, just we could probably make module `playwright_crawler/utils.py` private, what do u think?

janbuchar:
> LGTM, just we could probably make module `playwright_crawler/utils.py` private, what do u think?

Not sure. OG crawlee exposes this to the public, but it doesn't include it in the context.

B4nan:
It is included in the context

https://github.com/apify/crawlee/blob/master/packages/playwright-crawler/src/internals/utils/playwright-utils.ts#L752
https://crawlee.dev/api/playwright-crawler/interface/PlaywrightCrawlingContext#infiniteScroll

PR #407, opened 2024-08-06
ci: Add missing github token

janbuchar:


.github/workflows/run_release.yaml
```diff
@@ -43,6 +43,8 @@ jobs:
         !startsWith(github.event.head_commit.message, 'ci:')
       )
     runs-on: ubuntu-latest
+    env:
```
B4nan, line 46: maybe better move this to the top level (e.g. https://github.com/apify/crawlee/blob/master/.github/workflows/release.yml#L3)

PR #408, opened 2024-08-07
fix: Remove redundant log, fix format

janbuchar:


B4nan (approved):
can we add a test case for this maybe? so its clear how it looks

janbuchar:
> can we add a test case for this maybe? so its clear how it looks

We can, but should we? It sounds both flaky and fragile...

B4nan:
Well, I don't want any flaky tests, but you can mock everything out. My idea was a simple unit test where you log predefined stats and assert how they were logged.

B4nan (approved):
thats exactly what i meant 👍

Issue #399, opened 2024-08-05
Add support for launching Pw with user provided profile


sherpya:
### Which package is the feature request for? If unsure which one to select, leave blank

@crawlee/playwright (PlaywrightCrawler)

### Feature

Please add support for using user provided browser profile

### Motivation

a persistant profile is sometimes needed

### Ideal solution or implementation, and any additional constraints

my implementation (tested with chromium on linux)
```python
from pathlib import Path
from typing import override

from crawlee.browsers import PlaywrightBrowserPlugin, PlaywrightBrowserController, BrowserPool
from crawlee.playwright_crawler import PlaywrightCrawler

from app.crawler.routes import router

ARGS = []


class ScraperBrowserPlugin(PlaywrightBrowserPlugin):
    @override
    async def new_browser(self) -> PlaywrightBrowserController:
        if not self._playwright:
            raise RuntimeError('Playwright browser plugin is not initialized.')

        if 'user_data_dir' in self._browser_options:
            launch_fn = 'launch_persistent_context'
        else:
            launch_fn = 'launch'

        if self._browser_type == 'chromium':
            module = self._playwright.chromium
        elif self._browser_type == 'firefox':
            module = self._playwright.firefox
        elif self._browser_type == 'webkit':
            module = self._playwright.webkit
        else:
            raise ValueError(f'Invalid browser type: {self._browser_type}')

        launch = getattr(module, launch_fn)
        browser = await launch(**self._browser_options)

        return PlaywrightBrowserController(
            browser,
            max_open_pages_per_browser=self._max_open_pages_per_browser,
        )


def create_crawler(profile: Path, headless: bool = False):
    browser_options = {
        'args': ARGS,
        'headless': headless,
        'user_data_dir': profile
    }

    plugin = ScraperBrowserPlugin(browser_type='chromium', browser_options=browser_options)
    browser_pool = BrowserPool(plugins=[plugin])
    crawler = PlaywrightCrawler(browser_pool=browser_pool, request_handler=router)

    return crawler
````

B4nan:
Closing in favor of #963

PR #410, opened 2024-08-08
feat: support custom profile in playwright

sherpya:
if user_data_dir option is found in browser_option, then the launch function used is launch_persistent_context instead of launch and the user_data_dir option is passed to playwright

### Description

it makes possible to use custom user provided profile directory with playwright
without changing the api

### Issues

- Closes: #399

### Testing

```python
def create_crawler(profile: Path):
    browser_options = {
        'user_data_dir': profile
    }

    plugin = PlaywrightBrowserPlugin(browser_type='chromium', browser_options=browser_options)
    browser_pool = BrowserPool(plugins=[plugin])
    crawler = PlaywrightCrawler(browser_pool=browser_pool)

    return crawler
```

### Checklist

- [ ] CI passed

### DIFF
--- a/src/crawlee/browsers/playwright_browser_plugin.py
+++ b/src/crawlee/browsers/playwright_browser_plugin.py
@@ -92,14 +92,19 @@ class PlaywrightBrowserPlugin(BaseBrowserPlugin):
         if not self._playwright:
             raise RuntimeError('Playwright browser plugin is not initialized.')
 
-        if self._browser_type == 'chromium':
-            browser = await self._playwright.chromium.launch(**self._browser_options)
-        elif self._browser_type == 'firefox':
-            browser = await self._playwright.firefox.launch(**self._browser_options)
-        elif self._browser_type == 'webkit':
-            browser = await self._playwright.webkit.launch(**self._browser_options)
-        else:
-            raise ValueError(f'Invalid browser type: {self._browser_type}')
+        if self._browser_type not in ('chromium', 'firefox', 'webkit'):
+            raise ValueError(f'Invalid browser type: {self._browser_type}.')
+
+        module = getattr(self._playwright, self._browser_type, None)
+        if module is None:
+            raise ValueError(f'Invalid browser type: {self._browser_type}.')
+
+        # Determine whether to launch browser with persistent context or not
+        launch_fn = 'launch_persistent_context' if 'user_data_dir' in self._browser_options else 'launch'
+        launch = getattr(module, launch_fn, None)
+        if launch is None:
+            raise RuntimeError(f'Playwright {self._browser_type} browser module does not implement {launch_fn}.')
+        browser = await launch(**self._browser_options)
 
         return PlaywrightBrowserController(
             browser,
--- a/tests/unit/browsers/test_playwright_browser_plugin_with_profile.py
+++ b/tests/unit/browsers/test_playwright_browser_plugin_with_profile.py
@@ -1,0 +1,33 @@
+from __future__ import annotations
+
+from pathlib import Path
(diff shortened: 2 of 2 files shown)


sherpya:
it would be better to add a Literal check for the browser with typing.get_args, but this would unify the typevar used in __init__ so I think it's better to have a new PR

janbuchar:
Hello @sherpya and thanks for the contribution! We should have time to review it next week.

src/crawlee/browsers/playwright_browser_plugin.py
```diff
@@ -92,14 +92,18 @@ async def new_browser(self) -> PlaywrightBrowserController:
...
-        elif self._browser_type == 'webkit':
-            browser = await self._playwright.webkit.launch(**self._browser_options)
-        else:
-            raise ValueError(f'Invalid browser type: {self._browser_type}')
+        if self._browser_type not in ('chromium', 'firefox', 'webkit'):
+            raise ValueError(f'Invalid browser type: {self._browser_type}.')
+
+        module = getattr(self._playwright, self._browser_type, None)
+        if module is None:
+            raise ValueError(f'Unsupported browser type: {self._browser_type}.')
```
vdusek, line 100: Could we use the same error message as above?

```suggestion
            raise ValueError(f'Invalid browser type: {self._browser_type}.')
```

src/crawlee/browsers/playwright_browser_plugin.py
```diff
@@ -92,14 +92,18 @@ async def new_browser(self) -> PlaywrightBrowserController:
...
+
+        launch_fn = 'launch_persistent_context' if 'user_data_dir' in self._browser_options else 'launch'
```
vdusek, line 102: Maybe we could add a comment here explaining what's going on

```suggestion
        # Determine whether to launch browser with persistent context or not
        launch_fn = 'launch_persistent_context' if 'user_data_dir' in self._browser_options else 'launch'
```

vdusek (changes requested):
Seems good, thanks! Could you please also add a test covering `launch_persistent_context` case?

sherpya:
Hi, I can make requested changes, perhaps I have no idea how to make a test for a similiar feature, do I just need to see if the call work? Or I may check if somehow the profile is created in the given directory

janbuchar:
> Hi, I can make requested changes, perhaps I have no idea how to make a test for a similiar feature, do I just need to see if the call work? Or I may check if somehow the profile is created in the given directory

The bare minimum would be to just run it and make sure it doesn't crash :slightly_smiling_face: But if you could make a profile directory somewhere and find a way to check that it is actually being used by the headless browser, that would be awesome!

sherpya:
Hi, I've got a problem, launch with profile returns a `BrowserContext` instead of a `Browser` and browser_controller._browser is `<BrowserContext browser=None>`, `browser_controller.is_browser_connected` fails because of missing property in `_browser`

vdusek:
The fact that [launch_persistent_context](https://playwright.dev/python/docs/api/class-browsertype#browser-type-launch-persistent-context) returns a [BrowserContext](https://playwright.dev/python/docs/api/class-browsercontext) while the [launch](https://playwright.dev/python/docs/api/class-browsertype#browser-type-launch) returns a [Browser](https://playwright.dev/python/docs/api/class-browser) is a problem and our current abstraction is not designed to handle this difference. Addressing it would require deeper intervention into it, which I'm hesitant about.

Isn't there a way to extract a `Browser` instance from a `BrowserContext`? Alternatively, isn't there another approach to using a custom profile while still obtaining the `Browser` instance? I would rather go this way.

PR #411, opened 2024-08-08
fix: Dequeue items from RequestQueue in the correct order

janbuchar:


src/crawlee/storages/request_queue.py
```diff
@@ -97,7 +96,7 @@ def __init__(
         self._internal_timeout_seconds = 5 * 60
         self._assumed_total_count = 0
         self._assumed_handled_count = 0
-        self._queue_head_dict: OrderedDictType[str, str] = OrderedDict()
+        self._queue_head_dict: OrderedDict[str, str] = OrderedDict()
```
B4nan, line 99: so this line is the actual fix? it feels like you only changed the types, i thought that cannot have a runtime effect (just like with TS)?

or is it about the removed unknown label above?

janbuchar: 'tis but a draft... I just fixed a bit of weirdness I noticed in the file. Sadly the actual fix still evades me.

src/crawlee/storages/request_queue.py
```diff
@@ -472,7 +471,7 @@ async def get_handled_count(self) -> int:
     async def get_total_count(self) -> int:
         return self._assumed_total_count
 
-    async def ensure_head_is_non_empty(
+    async def _ensure_head_is_non_empty(
```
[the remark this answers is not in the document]
janbuchar: I know, but I'd consider it a bug for this to be public.

src/crawlee/storages/request_queue.py
```diff
@@ -31,6 +30,38 @@
 
 logger = getLogger(__name__)
 
+__all__ = ['RequestQueue']
+
+
+T = TypeVar('T')
+
+
+class BoundedSet(Generic[T]):
```
B4nan, line 39: maybe add a comment on why this is needed

B4nan (approved):
not that i would understand what this is doing completely 🙃

PR #425, opened 2024-08-12
refactor: Remove dependency on aiofiles and aioshutil

### LINKED ISSUE #412: Remove dependency on `aiofiles` and `aioshutil`
Under the hood, these don't do much more than `asyncio.to_thread` anyway.

janbuchar:
- closes #412

This is slightly uglier, but friendlier to type checkers and makes Crawlee's footprint a little smaller.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,8 +45,6 @@ keywords = [
 # https://github.com/apify/apify-sdk-python/pull/154.
 [tool.poetry.dependencies]
 python = "^3.9"
-aiofiles = "^23.2.1"
-aioshutil = "^1.3"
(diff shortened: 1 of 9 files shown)


B4nan (approved):
tests are passing, that's all i can evaluate here :]

one thing i would check here is if this works the same with recursion, e.g. when you try to create a nested folder without existing parent folders

janbuchar:
> one thing i would check here is if this works the same with recursion, e.g. when you try to create a nested folder without existing parent folders

`aiofiles` really is just a bunch of wrappers, and I checked that it really wraps the same functions that we replaced its usages with... It'll be fine.

Issue #417, opened 2024-08-09
URL Validation edge case - Protocol/Scheme relative URLs


aka-Ani:
It is possible that one encounters URLs that are protocol relative in HTML, like an `a` tag with href="//example.com". This is technically a valid URL, and essentially means it takes on whatever protocol that is used by the page that is in. 

The `is_url_absolute` function returns True for this case and while that may be correct, Pydantic doesn't know what to do with it and even the `AnyUrl` type requires a scheme (any scheme) to be present. So it throws a validation error. 

A potential fix is to use the `urllib.parse.urljoin` function regardless of whether the raw url is absolute or not. According to the [urllib docs](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin), "if _url_ is an absolute URL (that is, it starts with // or scheme://), the _url_’s hostname and/or scheme will be present in the result". So it was a full absolute url with its own different protocol and hostname, it stays that way. And if it is one of this protocol-relative kind of absolute url 😂 , it can take on the protocol of the base url it is being joined to.

I actually came across this because of a different but related issue - not sure if I should be creating a new one though. But essentially this URL validation fail shouldn't really cause a crawl to "crash", you'd expect just the one URL to not get queued in. But that is what I was experiencing. And it also stops the whole crawl job if it encounters a `tel:` link as well (also a Pydantic validation error). BUT curiously, if I forego using the `context.enqueue_links()` function and do the page parsing and request list building myself and then add it with `context.add_requests()`, it works! I mean the validation error still happens, but the crawl continues on with the other links. I basically copied the code from the enqueue_links implementation so I'm not sure why it would work here in the request handler function but not there, I'm guessing there's some difference in the error handling or something - but take this last theorizing piece as purely anecdotal, I haven't really investigated.

Anyway, sorry for the rambling - thanks for your work on this awesome project!

mattyjr007:
Hey @aka-Ani I am experience the same issues, i have tried resolving but it seems they only did thier implementations perfect of JS. please can you sure the code where you used the add_requests()

janbuchar:
Hi, and thanks for your interest in Crawlee! We appreciate the effort you put into investigating this. Could you provide us with a bunch of URL pairs (page URL + link), the expected behavior of add_requests and the actual results? 

Using that, it will be much easier for us to troubleshoot.

mattyjr007:
Hello Jan,

Thank you for reaching out.
When I try to crawl this website to crawl all the links on the website
[...]

PR #431, opened 2024-08-14
fix: Relative URLS supports & If not a URL, pass #417

black7375:
### Description

- If it starts with `//`, replace it with the same protocol.
- To avoid unsupported protocols, such as `mailto` and `tel`, we only allow writing `http` and `https`.

### Issues

- Closes: #417 

### Testing

- Add a unit test for cases that start with `//`.

### Checklist

- [x] CI passed

### DIFF
--- a/src/crawlee/_utils/urls.py
+++ b/src/crawlee/_utils/urls.py
@@ -7,7 +7,8 @@ from pydantic import AnyHttpUrl, TypeAdapter
 
 def is_url_absolute(url: str) -> bool:
     """Check if a URL is absolute."""
-    return bool(urlparse(url).netloc)
+    url_parsed = urlparse(url)
+    return bool(url_parsed.scheme) and bool(url_parsed.netloc)
 
 
 def convert_to_absolute_url(base_url: str, relative_url: str) -> str:
--- a/src/crawlee/beautifulsoup_crawler/beautifulsoup_crawler.py
+++ b/src/crawlee/beautifulsoup_crawler/beautifulsoup_crawler.py
@@ -5,6 +5,7 @@ import logging
 from typing import TYPE_CHECKING, Any, AsyncGenerator, Iterable, Literal
 
 from bs4 import BeautifulSoup, Tag
+from pydantic import ValidationError
 from typing_extensions import Unpack
 
 from crawlee._utils.blocked import RETRY_CSS_SELECTORS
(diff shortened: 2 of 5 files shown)


vdusek (changes requested):
Hey, great job on the implementation! 

I have just two suggestions:

1. **Refactoring URL logic**: To avoid repeating the same logic across multiple crawlers, we could move these functions to the `_utils/urls.py` module.

2. **URL concatenation**: I recommend using `urllib.parse.urljoin` instead of f-string concatenation.

`_utils/urls.py`:

```python
# ...

def is_url_absolute(url: str) -> bool:
    """Check if a URL is absolute."""
    # This case has to be handled explicitly.
    if url.startswith('//'):
        return False

    return bool(urlparse(url).netloc)


def is_http_url(url: str) -> bool:
    """Check if a URL is an HTTP/HTTPS."""
    return url.lower().startswith(('http://', 'https://'))


def convert_to_absolute_url(base_url: str, relative_url: str) -> str:
    """Convert a relative URL to an absolute URL using a base URL."""
    if relative_url.startswith('//'):
        protocol = urlparse(base_url).scheme
        return urljoin(protocol, relative_url)

    return urljoin(base_url, relative_url)

# ...
```    

`beautifulsoup_crawler/beautifulsoup_crawler.py`:

```python
# ...

                if (url := link.attrs.get('href')) is not None:
                    url = url.strip()

                    if not is_url_absolute(url):
                        url = convert_to_absolute_url(context.request.url, url)

                    if not is_http_url(url):
                        continue

# ...
```

janbuchar:
Please, could we have some unit tests before we merge this? 🙂

aka-Ani:
I agree this would "solve" the issue but it feels unnecessary to do string checks to verify things like `http` `https` or `//`. The urllib.parse library in addition to Pydantic models cover this functionality already. My suggestion would be to modify the crawlers to something more like this:
image

Essentially, relying on the urllib library to best construct a working absolute URL out of the href tags (it does this reliably for protocol relative URLs, as well as other relative URLs). And then instead of checking for specific scenarios to catch/pass, just handle any validation error Pydantic may raise. Since we are using the `AnyHttpUrl` Pydantic type, this should filter out all the unwanted URLs. 

I just think its cleaner approach vs introducing more (possibly redundant) functions into the codebase. But at the end of the day I suppose they both solve this specific scenario just fine, and I will leave it up to you guys 🙂

black7375:
1. Added a condition to `is_url_absolute` to check for URL scheme.
2. As suggested, it'll base `continue` on `try-except`.
3. Add a unit test for cases that start with `//`.

aka-Ani:
A couple things I noticed in the last set of changes:
1. Are you sure you should be handling a `ValueError` exception? I thought that line of code would raise a `ValidationError`
2. Maybe this is subjective, but I don't really see the purpose of checking `is_url_absolute` in these scenarios at all. The `urllib.parse.urljoin` function (reused as `convert_to_absolute_url` in utils) works just fine to construct a full/absolute url from the html extracted url and the base url. Taking the 7 examples from the url unit tests file (first 3 True, last 4 False), you can see that `urljoin` returns the same url if it was absolute or returns the joined url if it was not.
image
I think this would be the same output as doing the `if not is_url_absolute` then `convert_to_absolute_url`, but maybe it makes the code more readable or something, I'll let you decide.

black7375:
1. I think you're right. I pasted the content from the review, but I feel that `ValidationError` is more correct.
2. I'd like to approach this a bit defensively, as unexpected side effects might occur. However, as I don't have a strong opinion, I will follow @vdusek 's opinion in the follow-up review.

vdusek (approved):
LGTM, thank!

PR #447, opened 2024-08-20
chore: reschedule renovate bot

vdusek:


renovate.json
```diff
@@ -6,7 +6,7 @@
     "semanticCommits": "enabled",
     "lockFileMaintenance": {
         "enabled": true,
-        "schedule": ["before 2am"],
+        "schedule": ["before 1am on monday"],
```
B4nan, line 9: this should ideally run daily, and merge automatically (since we enable branch automerge). was this also problematic? maybe we just need to handle some permissions if it doesn't work, do you have a link to PR created for that? is it about flaky tests?

PR #466, opened 2024-08-27
ci: Skip running code checks on release

janbuchar:
Everything that is merged to master passed through code checks before, and they're needlessly long.

### DIFF
--- a/.github/workflows/run_release.yaml
+++ b/.github/workflows/run_release.yaml
@@ -26,10 +26,6 @@ on:
         default: ''
 
 jobs:
-  run_code_checks:
-    name: Run code checks
-    uses: ./.github/workflows/run_code_checks.yaml
-
   # This job determines if the conditions are met for a release to occur. It will proceed if triggered manually,
   # for any published release, or if the commit on push does not begin with "docs" or "chore".
   should_release:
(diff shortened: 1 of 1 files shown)


vdusek (approved):
Agreed

B4nan:
does this mean we no longer run tests on release? i would run at least some of them, this feels a bit dangerous. saying everything in master is checked sounds ok, but then you can push to master, skip CI checks, and stuff like that. we don't do that much, but it can happen. some time ago it also happened PR checks were less aggressive and they allowed some PRs in that ended up breaking the build. 

what does needlessly long mean and what takes the most time?

janbuchar:
> does this mean we no longer run tests on release?

Yeah... most of that was running all the tests after merging a PR, which triggers a pre-release to PyPI

> saying everything in master is checked sounds ok, but then you can push to master, skip CI checks, and stuff like that. we don't do that much, but it can happen. some time ago (in another repo) it also happened PR checks were less aggressive and they allowed some PRs in that ended up breaking the build.

Maybe we should disable merging PRs before code checks pass then? And use our eyes before triggering a release manually? Or we could of course check if `master` is green as the first step of the release workflow.

> what does needlessly long mean and what takes the most time?

5 minutes on a good day, 20 minutes divided over several retries when we get ratelimited by `httpbin` or encounter some windows-related flakiness... most of that are unit tests over 4 python versions times (windows, linux)

B4nan:
> Or we could of course check if master is green as the first step of the release workflow.

Yeah, this sounds like a good compromise, ideally run tests if the last commit is not green, and skip them if it they are (and also check if the last commit to master does not have `[skip ci]` in the commit message).

PR #474, opened 2024-08-30
chore(deps): Relax Playwright version constraint

janbuchar:
1.27 is the lowest version I'm comfortable supporting - it adds helper functions to locators that we may want to use in the future. It's ~2 years old, so it should be OK.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,7 @@ inquirer = ">=3.3.0"
 lxml = { version = ">=5.2.0", optional = true }
 more_itertools = ">=10.2.0"
 parsel = { version = ">=1.9.0", optional = true }
-playwright = { version = ">=1.43.0", optional = true }
+playwright = { version = ">=1.27.0", optional = true }
 psutil = ">=6.0.0"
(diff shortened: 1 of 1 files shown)


vdusek (approved):
:+1:

B4nan:
did someone asked us to do this? i wouldnt bother if not.

janbuchar:
I asked myself to do this :slightly_smiling_face: Without this, it's hard to build a docker image based on the Apify playwright images (version conflicts)

B4nan:
so our docker images use old playwright version? this should be all automated just like the js versions (so periodic image build for current pw version + few versions back too), cc @vladfrangu

janbuchar:
> so our docker images use old playwright version? 

Yes

> this should be all automated just like the js versions (so periodic image build for current pw version + few versions back too), cc @vladfrangu

Definitely a good idea. It doesn't prevent a template user from upgrading playwright and making a huge layer though.

B4nan:
also i can see we only have on pw image for python while we have 4 for node, i guess we should have the same setup for python?

PR #483, opened 2024-08-30
chore(deps): Relax pyee dependency version constraint

janbuchar:
This prevents dependency conflicts with playwright (requires 11.0.0)


vdusek (approved):
LGTM

PR #490, opened 2024-09-02
fix: JSON handling with Parsel

### LINKED ISSUE #488: Cannot use CSS on a Selector of type 'json' ( ParselCrawler )
Getting this error when using ParselCrawler.
`ValueError: Cannot use css on a Selector of type 'json'
`

### Reproduce:
**main.py**
```python
import asyncio

from crawlee.parsel_crawler import ParselCrawler

from crawler.routes import router


async def main() -> None:
    """The crawler entry point."""
    crawler = ParselCrawler(
        request_handler=router,
        max_requests_per_crawl=50,
    )

    await crawler.run(
        [
            'https://httpbin.org/ip',
        ]
    )


if __name__ == '__main__':
    asyncio.run(main())
```

**routes.py**
```python
from crawlee.parsel_crawler import ParselCrawlingContext
from crawlee.router import Router

router = Router[ParselCrawlingContext]()


@router.default_handler
async def default_handler(context: ParselCrawlingContext) -> None:
    """Default request handler."""
    print("hello")
```

janbuchar:
- closes #488

Thanks @Ehsan-U for reporting this!

### DIFF
--- a/src/crawlee/parsel_crawler/_parsel_crawler.py
+++ b/src/crawlee/parsel_crawler/_parsel_crawler.py
@@ -89,10 +89,12 @@ class ParselCrawler(BasicCrawler[ParselCrawlingContext]):
             if crawling_context.session and crawling_context.session.is_blocked_status_code(status_code=status_code):
                 raise SessionError(f'Assuming the session is blocked based on HTTP status code {status_code}')
 
+            parsel = crawling_context.selector
+
             matched_selectors = [
                 selector
                 for selector in RETRY_CSS_SELECTORS
-                if crawling_context.selector.css(selector).get() is not None
+                if parsel.type in ('html', 'xml') and parsel.css(selector).get() is not None
             ]
 
             if matched_selectors:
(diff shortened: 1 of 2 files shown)


vdusek (commented):
Good job. Just a couple of questions:

1) Shouldn't we handle this in BS crawler as well?

2) What about adding try-except block when parsing the response? (Same for BS crawler.)

```python
try:
    parsel_selector = await asyncio.to_thread(lambda: Selector(body=context.http_response.read()))
except ValueError as exc:
    raise ValueError(
        f'Failed to parse the response body. Ensure the response is parseble via Parsel: {exc}'
    ) from exc
```

janbuchar:
> 1. Shouldn't we handle this in BS crawler as well?

I'm not sure how to go about that. BS is super tolerant to malformed HTML:

```python
>>> from bs4 import BeautifulSoup
>>> BeautifulSoup('{"hello": "world"}')
<html><body><p>{"hello": "world"}</p></body></html>
```

So... let me propose employing the Ostrich algorithm yet again?

> 2. What about adding try-except block when parsing the response? (Same for BS crawler.)
> ```python
> try:
>     parsel_selector = await asyncio.to_thread(lambda: Selector(body=context.http_response.read()))
> except ValueError as exc:
>     raise ValueError(
>         f'Failed to parse the response body. Ensure the response is parseble via Parsel: {exc}'
>     ) from exc
> ```

Are you sure the original exception is not clear enough? This will make the traceback longer, which is not great for readability.

vdusek (approved):
LGTM then

PR #504, opened 2024-09-06
refactor: simplify type of request(s) args

vdusek:
### Description

- Simplify type of request(s) args to only string or Request.
- Should be only internal change, not breaking, as `BaseRequestData` was not exposed to the public in v0.3.

### Issues

- Closes: N/A

### Testing

- N/A

### Checklist

- [x] CI passed

### DIFF
--- a/src/crawlee/basic_crawler/_basic_crawler.py
+++ b/src/crawlee/basic_crawler/_basic_crawler.py
@@ -338,7 +338,7 @@ class BasicCrawler(Generic[TCrawlingContext]):
 
     async def run(
         self,
-        requests: Sequence[str | BaseRequestData | Request] | None = None,
+        requests: Sequence[str | Request] | None = None,
         *,
         purge_request_queue: bool = True,
     ) -> FinalStatistics:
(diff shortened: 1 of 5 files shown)


src/crawlee/basic_crawler/_basic_crawler.py
```diff
@@ -723,24 +723,26 @@ async def _commit_request_handler_result(
...
                     else:
                         base_url = call['base_url'] if call.get('base_url') else origin
                         absolute_url = convert_to_absolute_url(base_url, request)
-                        dst_request = BaseRequestData.from_url(absolute_url)
-                else:
-                    dst_request = request
+                        dst_request = Request.from_url(absolute_url)
+
+                # If the request is a BaseRequestData, convert it to Request object.
+                elif isinstance(request, BaseRequestData):
```
janbuchar, line 744: I don't mind if this is defensive, just... how does that happen?

janbuchar (approved):

src/crawlee/basic_crawler/_basic_crawler.py
vdusek: image

I don't know - just covering all the possible cases Pylance declares.

PR #508, opened 2024-09-09
fix: Correctly handle log level configuration

janbuchar:
- The `CRAWLEE_VERBOSE` and `CRAWLEE_LOG_LEVEL` variables are now respected by crawlee
- `CrawleeLogFormatter` doesn't alter the record being formatted anymore
- SDK counterpart coming soon

### DIFF
--- a/src/crawlee/_log_config.py
+++ b/src/crawlee/_log_config.py
@@ -47,6 +47,8 @@ def get_configured_log_level(configuration: Configuration) -> int:
             return logging.WARNING
         if configuration.log_level == 'ERROR':
             return logging.ERROR
+        if configuration.log_level == 'CRITICAL':
+            return logging.CRITICAL
 
         assert_never(configuration.log_level)
 
--- a/src/crawlee/_log_config.py
+++ b/src/crawlee/_log_config.py
@@ -2,11 +2,15 @@ from __future__ import annotations
 
 import json
 import logging
+import sys
 import textwrap
-import traceback
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 from colorama import Fore, Style, just_fix_windows_console
+from typing_extensions import assert_never
+
+if TYPE_CHECKING:
+    from crawlee.configuration import Configuration
 
 just_fix_windows_console()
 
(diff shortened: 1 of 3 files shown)


src/crawlee/configuration.py
```diff
@@ -51,14 +51,14 @@ class Configuration(BaseSettings):
     ] = False
 
     log_level: Annotated[
-        int,
+        Literal['DEBUG', 'INFO', 'WARNING', 'ERROR'],
```
vdusek, line 54: `NOTSET`? `CRITICAL`?

src/crawlee/basic_crawler/_basic_crawler.py
```diff
@@ -203,20 +203,14 @@ def __init__(
...
-            root_logger.addHandler(handler)
-            root_logger.setLevel(logging.INFO if not sys.flags.dev_mode else logging.DEBUG)
+            configure_logger(root_logger, self._configuration, remove_old_handlers=True)
 
             # Silence HTTPX logger
             httpx_logger = logging.getLogger('httpx')
-            httpx_logger.setLevel(logging.WARNING if not sys.flags.dev_mode else logging.INFO)
+            httpx_logger.setLevel(
+                logging.DEBUG if get_configured_log_level(self._configuration) <= logging.DEBUG else logging.WARNING
+            )
```
vdusek, line 213: Not so important, but why did you change this and how does it differ from the previous version?

src/crawlee/_log_config.py
```diff
@@ -31,6 +35,44 @@
...
+
+    if 'log_level' in configuration.model_fields_set:
+        if configuration.log_level == 'DEBUG':
+            return logging.DEBUG
+        if configuration.log_level == 'INFO':
+            return logging.INFO
+        if configuration.log_level == 'WARNING':
+            return logging.WARNING
+        if configuration.log_level == 'ERROR':
+            return logging.ERROR
```
vdusek, line 49: why not support all possible log levels?

vdusek (changes requested):

src/crawlee/configuration.py
janbuchar: These are what Javascript crawlee supports. Notset doesn't sound like a great idea to me here, but I'm open to adding `CRITICAL`.

src/crawlee/basic_crawler/_basic_crawler.py
janbuchar: It uses a helper function to determine the log level instead of repeating an incomplete condition all over the place. In the default case, it should behave the same as before.

src/crawlee/_log_config.py
janbuchar: Less is more :shrug: Same as the comment higher up, I'd consider adding `CRITICAL`. Am I missing something?

vdusek: ok, adding just `CRITICAL` si good, thanks

vdusek (approved):
just the CRITICAL thing, otherwise LGTM :rocket:

Issue #479, opened 2024-08-30
Create a new guide for result storages (`Dataset`, `KeyValueStore`)


vdusek:
- We should create a new documentation guide on how to work with result storages (`Dataset`, `KeyValueStore`).
- Inspiration: https://crawlee.dev/docs/guides/result-storage
- Check the structure of other guides - [docs/guides](https://github.com/apify/crawlee-python/tree/master/docs/guides), and try to render it using `make run-doc`.

MostlyKIGuess:
I want to work on this, @B4nan  @souravjain540 .  Assign me if it's still open, I want to work on #481  and this together, it will be convenient.

Atharva1723:
Can I work on this issue?

janbuchar:
> Can I work on this issue?

Yes, feel free to open a PR whenever you're ready.

saymedgm:
we need to add this to readme.md ?

janbuchar:
> we need to add this to readme.md ?

No, we want it added to the `docs/` folder so that it appears in the documentation website.

vdusek:
Check out the structure of other guides at https://github.com/apify/crawlee-python/tree/master/docs/guides.

saymedgm:
done i have raised pr

Manish-k723:
pls /assign , I am working on this.

MostlyKIGuess:
> pls /assign , I am working on this.
https://github.com/apify/crawlee-python/pull/585
this PR has already been raised

PR #515, opened 2024-09-11
feat: expose extended unique key when creating a new Request

vdusek:
### Description

When working on https://github.com/apify/crawlee-python/issues/479 I noticed a few minor problems in `RequestProvider`(s)  & `Request`. Mainly, depredated docstring in `RequestQueue.add_request`, and not exposing extended unique key functionality in the `Request.from_url`. This fixes it and also improves related docstrings.

### Issues

- Closes: N/A

### Testing

- N/A

### Checklist

- [ ] CI passed


src/crawlee/storages/_request_provider.py
```diff
@@ -55,6 +55,24 @@ async def mark_request_as_handled(self, request: Request) -> ProcessedRequest |
...
+
+        Args:
+            request: The request object (or its string representation) to be added to the provider.
+            forefront: Determines whether the request should be added to the beginning (if True) or the end (if False)
+                of the provider.
+
+        Returns:
+            Information about the request addition to the provider.
+        """
+
```
janbuchar, line 75: I think this was missing intentionally - the `RequestProvider` interface already has `add_requests_batched`, so what's the reason to also require a single-request version?

src/crawlee/_utils/requests.py
```diff
@@ -83,7 +83,7 @@ def normalize_url(url: str, *, keep_url_fragment: bool = False) -> str:
 def compute_unique_key(
     url: str,
     method: str = 'GET',
-    payload: bytes | None = None,
+    payload: str | None = None,
```
janbuchar, line 86: The helper is also used in the SDK - you may want to take `str | bytes | None` for better compatibility.

janbuchar (changes requested):

src/crawlee/storages/_request_provider.py
B4nan: both methods should be available, we do have `addRequests` in the js version on this level

https://github.com/apify/crawlee/blob/master/packages/core/src/storages/request_provider.ts#L107

in the long run (crawlee v4) we want to get rid of the duality between RQ v1/v2, so we might as well remove this abstraction. not sure how things are implemented in the python version.

janbuchar: Here, `RequestProvider` is just an interface for "something that BasicCrawler can pull requests from". Hopefully that will also be the case in the JS version one day... But here and now, I'd just not try to restore the parity with JS.

B4nan: hmm, and we dont use the `add_request` method internally anywhere now? 

i would say it should be here, in the end users can work with the RQ directly, not just via crawler.

janbuchar: I believe `RequestQueue` has `add_request` for convenience, but it's not required by the `RequestProvider` interface. So currently you can do `RequestsQueue.open().add_request()` and pass the same queue into the crawler via constructor. You can't do `crawler.get_request_provider().add_request()` though.

janbuchar: Maybe we could just add a default implementation of `add_request` to `RequestProvider` that would use the batched version?

janbuchar: @B4nan good point, I like that more.

janbuchar (approved):
LGTM

PR #529, opened 2024-09-16
ci: allow beta releases for chore commits

vdusek:


janbuchar (approved):

janbuchar:
@vdusek please try to come to an agreement with @B4nan on whether this is a good idea or not :slightly_smiling_face:

B4nan:
i wouldn't skip chore commits in general. on the other hand, i guess what happened is that docs related dependency updates triggered this, which we could improve in a better way - we could check git diff from the last commit, if it only contains changes from the `website` folder, then ignore it from publishing.

but it feels a bit like bikeshedding, i would just keep publishing on chores as well, it's not doing any harm, or is it?

Issue #402, opened 2024-08-06
Add header & fingerprint generator and integrate them into HTTPX client


vdusek:
- Implement a fingerprint generator for generating real-world HTTP headers.
- Integrate it into [HttpxHttpClient](https://github.com/apify/crawlee-python/blob/master/src/crawlee/http_clients/httpx.py).
- [fingerprint-generator](https://github.com/apify/fingerprint-suite/tree/master/packages/fingerprint-generator) in TS could serve as an inspiration.

janbuchar:
I believe that the generator will be needed for all clients, won't it?

vdusek:
> I believe that the generator will be needed for all clients, won't it?

`CurlImpersonate` has its own system, which should now work by default (when #473 is done).

So for now it will be used only by `HttpxHttpClient`.

B4nan:
fingerprinting for playwright has priority, this can wait (we could even skip it completely if the default based on curl impersonate wont need it)

vdusek:
> fingerprinting for playwright has priority, this can wait (we could even skip it completely if the default based on curl impersonate wont need it)

Header generator and fingerprint generator are prerequisites for a fingerprint injector (Playwright).

Issue #401, opened 2024-08-06
Mask Playwright's "headless" headers


vdusek:
### Description

- When running headless browsers with Playwright, certain HTTP headers (like User-Agent and Sec-Ch-Ua) reveal the browser is operating in headless mode by including the substring "headless". This makes it easy for anti-scraping systems to detect and block automated browsers.
- To prevent detection, these headers should be replaced with more realistic, browser-like values that mimic typical, non-headless browser behavior.
- Example of header values for headless Chromium:

```json
{
    "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7",
    "Accept-Encoding": "gzip, deflate, br, zstd",
    "Host": "httpbin.org",
    "Priority": "u=0, i",
    "Sec-Ch-Ua": "\"Chromium\";v=\"128\", \"Not;A=Brand\";v=\"24\", \"HeadlessChrome\";v=\"128\"",
    "Sec-Ch-Ua-Mobile": "?0",
    "Sec-Ch-Ua-Platform": "\"Linux\"",
    "Sec-Fetch-Dest": "document",
    "Sec-Fetch-Mode": "navigate",
    "Sec-Fetch-Site": "none",
    "Sec-Fetch-User": "?1",
    "Upgrade-Insecure-Requests": "1",
    "User-Agent": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/128.0.6613.18 Safari/537.36",
    "X-Amzn-Trace-Id": "Root=1-66d04117-141b301674c02e4e2136f1f1"
}
```

### Relevant links

- https://github.com/apify/fingerprint-suite/tree/master/packages/fingerprint-injector
- https://github.com/apify/crawlee/blob/v3.11.4/packages/browser-pool/src/abstract-classes/browser-plugin.ts#L20:L21
- https://github.com/apify/crawlee/blob/v3.11.4/packages/browser-pool/src/abstract-classes/browser-plugin.ts#L213

PR #530, opened 2024-09-16
feat: add header generator and integrate it into HTTPX client

vdusek:
### Description

- This is the first version of the header generator, providing common HTTP headers including user agent.
  - User-agent is picked randomly from a random pool of 1000 user agents from the Apify fingerprint dataset.
- This is integrated into the HTTPX client and will be further used in the Playwright fingerprint injector (#401).

### Issues

- Closes: #402 

### Testing

- New unit tests implemented.

### Checklist

- [x] CI passed


src/crawlee/http_clients/_httpx.py
```diff
@@ -215,6 +223,16 @@ def _get_client(self, proxy_url: str | None) -> httpx.AsyncClient:
 
         return self._client_by_proxy_url[proxy_url]
 
+    def _get_headers(self, explicit_headers: HttpHeaders | None) -> HttpHeaders | None:
```
janbuchar, line 226: The `get_` prefix doesn't seem fitting here - how about something like "combine"?

src/crawlee/http_clients/_httpx.py
```diff
@@ -85,6 +87,7 @@ def __init__(
         ignore_http_error_status_codes: Iterable[int] = (),
         http1: bool = True,
         http2: bool = True,
+        append_common_headers: bool = True,
```
janbuchar, line 90: I'd like to discuss making a header generator interface and passing it via constructor, instead of tying one specific implementation to the value of this parameter. What do you think?

vdusek: good idea, thanks

janbuchar (approved):
LGTM

Issue #495, opened 2024-09-03
`CurlImpersonateHttpClient` warning on Windows


janbuchar:
Using the `CurlImpersonateHttpClient` adds this warning message upon program start, which doesn't seem to be fixed if I add in the command that it asks for 

```
    asyncio.set_event_loop_policy(WindowsSelectorEventLoopPolicy())
    await crawler.run(["https://www.mtggoldfish.com/metagame/modern#paper"])
```

```
.venv\Lib\site-packages\curl_cffi\aio.py:137: RuntimeWarning:
    Proactor event loop does not implement add_reader family of methods required.
    Registering an additional selector thread for add_reader support.
    To avoid this warning use:
        asyncio.set_event_loop_policy(WindowsSelectorEventLoopPolicy())
```

_Originally posted by @MrTyton in https://github.com/apify/crawlee-python/issues/486#issuecomment-2325385889_

janbuchar:
@MrTyton Looks like you have to add `asyncio.set_event_loop_policy(WindowsSelectorEventLoopPolicy())` before `asyncio.run()`, i.e., before your event loop is created.

vdusek:
I believe we should document this somewhere, follow-up issues is ok

janbuchar:
> I believe we should document this somewhere, follow-up issue is ok

I'm not so sure about that, it's a limitation of a package that we integrate with and the warning tells you what to do pretty well. Maybe we can add a line about this into the docblock of `CurlImpersonateHttpClient`.

PR #538, opened 2024-09-20
feat: Improved project bootstrapping

janbuchar:
This adds a unified `crawler` template. The original `playwright` and `beautifulsoup` templates are kept for compatibility with older versions of the CLI.

The user is now prompted for package manager type (pip, poetry), crawler type, start URL and whether or not Apify integration should be set up.


janbuchar:
okay, this could use some testing now... plus there's a bunch of stuff to consider

- the pip thing is probably not very ergonomic and I'm not sure how to approach it well... should we create a .venv for users? should we have a `requirements.txt` and use it as a lockfile?
- does this work on windows?
- isn't the creation dialog too long now? maybe some of the options don't need the prompt and just CLI flags are enough
- should we include http client selection?

templates/crawler/cookiecutter.json
```diff
@@ -0,0 +1,12 @@
+{
+    "project_name": "crawlee-python-beautifulsoup-project",
```
vdusek, line 2: Is it correct?

templates/crawler/{{cookiecutter.project_name}}/README.md
```diff
@@ -0,0 +1,37 @@
+# {{cookiecutter.project_name}}
+
+Project skeleton generated by Crawlee (Beautifulsoup template).
```
vdusek, line 3: is it always BS?

src/crawlee/_cli.py
```diff
@@ -92,26 +114,66 @@ def _prompt_for_template() -> str:
...
+        help='The library that will be used for crawling in your crawler. If none is given, you will be prompted.',
     ),
-    template: Optional[str] = typer.Option(
+    package_manager: Optional[str] = typer.Option(
         default=None,
-        help='The template to be used to create the project. If none is given, you will be prompted.',
         show_default=False,
+        help='Package manager to be used in the new project. If none is given, you will be prompted.',
+    ),
+    start_url: Optional[str] = typer.Option(
```
vdusek, line 133: `Optional` and not `|` because of Typer, am I correct? thanks

vdusek (changes requested):
Nice, just a few comments

src/crawlee/_cli.py
janbuchar: Yes. Typer and Python 3.9, IIRC.

templates/crawler/cookiecutter.json
janbuchar: Good point. I'll probably just change it to my-crawler or something.

janbuchar:
Thanks for the review @vdusek! Do you have anything to add regarding those points for consideration?

vdusek:
> the pip thing is probably not very ergonomic and I'm not sure how to approach it well... should we create a .venv for users? should we have a requirements.txt and use it as a lockfile?

That is generally how pip works. I wouldn't add additional overhead here. Simply dumping the dependencies into `requirements.txt` should be enough. Managing the environment and installation should be left to the user - that's the standard approach with pip.

> isn't the creation dialog too long now? maybe some of the options don't need the prompt and just CLI flags are enough

IMO the current setup is fine. Also, all options have default values.

> should we include http client selection?

That would be a nice feature, but I wouldn't prioritize it at this moment. We have a guide covering that, which should be sufficient for now.

> we should probably scream and shout if poetry is not installed... or we could just disable the option in that case

I would suggest throwing an error. If a user wants to use Poetry, the option should be conditionally on Poetry being installed.

templates/crawler/{{cookiecutter.project_name}}/Dockerfile
vdusek, line 1: I know this is copied, but I suggest using Hadolint and improving it according to their suggestions.

We should also pin (or at least limit to < 2) the Poetry (mostly because of https://github.com/python-poetry/poetry/issues/3332).

templates/crawler/{{cookiecutter.project_name}}/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Usage
+
+{% if cookiecutter.package_manager == 'poetry' %}
```
vdusek, line 7: this leads to 2 empty lines

templates/crawler/{{cookiecutter.project_name}}/README.md
```diff
@@ -0,0 +1,37 @@
...
+To get started, ensure you have [Poetry](https://python-poetry.org/), a package and dependency management system, installed on your machine. We recommend installing it with the following command:
```
vdusek, line 8: could we wrap it to 120 line width?

vdusek (changes requested):

vdusek (changes requested):
the Pip versions does not work

janbuchar:
> > the pip thing is probably not very ergonomic and I'm not sure how to approach it well... should we create a .venv for users? should we have a requirements.txt and use it as a lockfile?
> 
> That is generally how pip works. I wouldn't add additional overhead here. Simply dumping the dependencies into `requirements.txt` should be enough. Managing the environment and installation should be left to the user - that's the standard approach with pip.

I generally agree. But how would you ensure that `pipx run crawlee create ...` installs the dependencies in a venv? Often people will want to keep the virtualenv in the project directory, which doesn't get created until you run `crawlee create`.

> > should we include http client selection?
> 
> That would be a nice feature, but I wouldn't prioritize it at this moment. We have a guide covering that, which should be sufficient for now.

It's a 15-minute adventure, the only problem is that it'd make the dialog longer.

janbuchar:
Relevant discussion - https://apify.slack.com/archives/C06ENBV1B9Q/p1728907202834259

templates/crawler/{{cookiecutter.project_name}}/README.md
janbuchar: This is consistent how we write our docs - one paragraph per line. Is there a reason to do it differently here?

vdusek: no, sorry, I don't know why I wrote it :smile:, this is correct

src/crawlee/_cli.py
```diff
@@ -1,21 +1,24 @@
...
 import typer
 from cookiecutter.main import cookiecutter  # type: ignore
 from inquirer.render.console import ConsoleRender  # type: ignore
 from rich.progress import Progress, SpinnerColumn, TextColumn
 
-TEMPLATE_LIST_URL = 'https://api.github.com/repos/apify/crawlee-python/contents/templates'
-
 cli = typer.Typer(no_args_is_help=True)
 
+cookiecutter_json = json.load((Path().parent.parent.parent / 'templates' / 'crawler' / 'cookiecutter.json').open())
```
janbuchar, line 16:
```suggestion
cookiecutter_json = json.load((Path(__name__).parent.parent.parent / 'templates' / 'crawler' / 'cookiecutter.json').open())
```

src/crawlee/_cli.py
```diff
@@ -1,21 +1,24 @@
...
 import typer
 from cookiecutter.main import cookiecutter  # type: ignore
 from inquirer.render.console import ConsoleRender  # type: ignore
 from rich.progress import Progress, SpinnerColumn, TextColumn
 
-TEMPLATE_LIST_URL = 'https://api.github.com/repos/apify/crawlee-python/contents/templates'
-
 cli = typer.Typer(no_args_is_help=True)
 
+cookiecutter_json = json.load((Path(__name__).parent.parent.parent / 'templates' / 'crawler' / 'cookiecutter.json').open())
```
janbuchar, line 16:
```suggestion
cookiecutter_json = json.load((Path(__file__).parent.parent.parent / 'templates' / 'crawler' / 'cookiecutter.json').open())
```

src/crawlee/project_template/{{cookiecutter.project_name}}/Dockerfile
```diff
@@ -0,0 +1,68 @@
+# First, specify the base Docker image.
+# You can see the Docker images from Apify at https://hub.docker.com/r/apify/.
+# You can also use any other image from Docker Hub.
+# % if cookiecutter.crawler_type == 'playwright'
+FROM apify/actor-python-playwright:3.12
+# % else
+FROM apify/actor-python:3.12
+# % endif
```
vdusek, line 8: Python 3.13?

src/crawlee/project_template/templates/main_playwright.py
```diff
@@ -0,0 +1,14 @@
...
+# % block import
+from crawlee.playwright_crawler import PlaywrightCrawler
+# % endblock
+
+# % block instantiation
+crawler = PlaywrightCrawler(
+    request_handler=router,
+    headless=True,
+    max_requests_per_crawl=50,
+    {{ self.http_client_instantiation() }}
```
vdusek, line 12: this is resulting in blank line:

```
        http_client=CurlImpersonateHttpClient(),

    )
```

src/crawlee/project_template/{{cookiecutter.project_name}}/{{cookiecutter.__package_name}}/__main__.py
```diff
@@ -0,0 +1,6 @@
+import asyncio
+
+from .main import main
+
+if __name__ == '__main__':
+    asyncio.run(main())
```
vdusek: since in this file it is not problematic, I would format it correctly
```suggestion
import asyncio

from .main import main


if __name__ == '__main__':
    asyncio.run(main())
```

vdusek (commented):
We provide http client option together with the PW crawler, is that what we want?

vdusek (commented):
how is this solving #495 please?

templates/crawler/{{cookiecutter.project_name}}/Dockerfile
vdusek: This still applies, I woudl rather add <2 to poetry version.

janbuchar:
> how is this solving #495 please?

See https://github.com/apify/crawlee-python/pull/538/commits/65edfe399ab77ffb38104908d7657fb27b2142d2 and https://github.com/apify/crawlee-python/issues/495#issuecomment-2511708789

templates/crawler/{{cookiecutter.project_name}}/Dockerfile
vdusek: @janbuchar

janbuchar:
> We provide http client option together with the PW crawler, is that what we want?

As long as we don't use `playwright.request` in `PlaywrightCrawlingContext.send_request`, it makes perfect sense. After that, it still might make sense in some niche cases, but the utility of giving that choice to users in project bootstrapping will be debatable.

vdusek:
> As long as we don't use playwright.request in PlaywrightCrawlingContext.send_request, it makes perfect sense. After that, it still might make sense in some niche cases, but the utility of giving that choice to users in project bootstrapping will be debatable.

I would not expose the option to the user, IMO it would results only to confusion. Net a request though.

templates/crawler/{{cookiecutter.project_name}}/Dockerfile
janbuchar: Hadolint probably won't work, the Dockerfile is templated. I'll see what I can do about the poetry version.

vdusek (approved):
I tested a lot of variants, seems to be working, good job!

Issue #94, opened 2024-04-08
Remove `json_` and `order_no` from `Request`


janbuchar:
The purpose of the fields is somewhat unclear, but it's certain that they don't belong to the `Request` class.

We should definitely explore the notion of an internal request in Crawlee and how it translates to the Python version.

janbuchar:
Also, the mechanism of determining an ID based on unique key should be exclusive to memory storage (internal request). So `Request.id` should only ever be populated by loading an API response. 

Otherwise, things like #623 will happen.

janbuchar:
Frankly, I don't think that there's any excuse for a request ID that's derived from a unique key. Maybe it has something to do with file path safety...

janbuchar:
Changes introduced in https://github.com/apify/crawlee-python/pull/683 are related to this - we do weird things to `Request` in `MemoryStorageClient` and `json_dumps` has `default=str`, which serializes `bytes` as `'b"asdf"'`, for example.

PR #540, opened 2024-09-23
fix: Improve `Request.user_data` serialization

### LINKED ISSUE #524: Implement/document a way how to pass information between handlers
I came across a situation where I scrape half of the item's data in the listing page handler and the other half in a handler taking care of the detail page. I think must be quite common case. I struggle to see how I pass the information down from one handler to another. See concrete example below:

```python
import re
import asyncio
from enum import StrEnum, auto

import click
from crawlee.beautifulsoup_crawler import (
    BeautifulSoupCrawler,
    BeautifulSoupCrawlingContext,
)
from crawlee.router import Router


LENGTH_RE = re.compile(r"(\d+)\s+min")


class Label(StrEnum):
    DETAIL = auto()


router = Router[BeautifulSoupCrawlingContext]()


@click.command()
def edison():
    asyncio.run(scrape())


async def scrape():
    crawler = BeautifulSoupCrawler(request_handler=router)
    await crawler.run(["https://edisonfilmhub.cz/program"])
    await crawler.export_data("edison.json", dataset_name="edison")


@router.default_handler
async def detault_handler(context: BeautifulSoupCrawlingContext):
    await context.enqueue_links(selector=".program_table .name a", label=Label.DETAIL)


@router.handler(Label.DETAIL)
async def detail_handler(context: BeautifulSoupCrawlingContext):
    context.log.info(f"Scraping {context.request.url}")

    description = context.soup.select_one(".filmy_page .desc3").text
    length_min = LENGTH_RE.search(description).group(1)
    # TODO get starts_at, then calculate ends_at

    await context.push_data(
        {
            "url": context.request.url,
            "title": context.soup.select_one(".filmy_page h1").text.strip(),
            "csfd_url": context.soup.select_one(".filmy_page .hrefs a")["href"],
        },
        dataset_name="edison",
    )
```

I need to scrape `starts_at` at the `default_handler`, then add more details to the item on the detail page, and calculate the `ends_at` time according to the length of the film. Even if I changed `enqueue_links` to something more delicate, how do I pass data from one request to another?

janbuchar:
- resolves #524

This adds validation to `Request.user_data` so that the user cannot pass in data that is not JSON-serializable.

### DIFF
--- a/src/crawlee/_request.py
+++ b/src/crawlee/_request.py
@@ -283,7 +283,7 @@ class Request(BaseRequestData):
         )
 
         if label is not None:
-            cast(UserData, request.user_data).label = label
+            request.user_data['label'] = label
 
         return request
 
--- a/src/crawlee/_request.py
+++ b/src/crawlee/_request.py
@@ -2,12 +2,22 @@
 
 from __future__ import annotations
 
+from collections.abc import Iterator, MutableMapping
 from datetime import datetime
 from decimal import Decimal
 from enum import Enum
-from typing import Annotated, Any
+from typing import Annotated, Any, cast
 
-from pydantic import BaseModel, BeforeValidator, ConfigDict, Field
+from pydantic import (
+    BaseModel,
+    BeforeValidator,
+    ConfigDict,
+    Field,
+    JsonValue,
+    PlainSerializer,
+    PlainValidator,
+    TypeAdapter,
+)
 from typing_extensions import Self
 
 from crawlee._types import EnqueueStrategy, HttpMethod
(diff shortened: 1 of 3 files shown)


src/crawlee/_request.py
```diff
@@ -286,32 +364,10 @@ def forefront(self) -> bool:
...
-    state: RequestState | None = None
-    """Describes the request's current lifecycle state."""
-
-    session_rotation_count: Annotated[int | None, Field(alias='sessionRotationCount')] = None
-
-    skip_navigation: Annotated[bool, Field(alias='skipNavigation')] = False
-
-    last_proxy_tier: Annotated[int | None, Field(alias='lastProxyTier')] = None
-
-    forefront: Annotated[bool, Field()] = False
```
janbuchar, line 317: This was just moved up

tests/unit/storages/test_request_queue.py
```diff
@@ -162,3 +163,49 @@ async def test_add_batched_requests(
...
+
+async def test_invalid_user_data_serialization() -> None:
+    with pytest.raises(ValidationError):
+        Request.from_url(
+            'https://crawlee.dev',
+            user_data={
+                'foo': datetime(year=2020, month=7, day=4, tzinfo=timezone.utc),
+                'bar': {datetime(year=2020, month=4, day=7, tzinfo=timezone.utc)},
+            },
+        )
```
janbuchar, line 176: Not sure where to put this one, but it is similar to the ones that follow...

vdusek: I believe we can move it to a separate file `test_request.py`.

vdusek (approved):
Really good job, thanks.

I just wonder whether we should not separate "user data" and "our Crawlee metadata". Currently, everything is stored in the `user_data` field. Let's say the `Request` would have 2 fields - `user_data` and `crawlee_data` / `crawlee_metadata`. Or is there any problem/drawback with this separation?

janbuchar:
> I just wonder whether we should not separate "user data" and "our Crawlee metadata". Currently, everything is stored in the `user_data` field. Let's say the `Request` would have 2 fields - `user_data` and `crawlee_data` / `crawlee_metadata`. Or is there any problem/drawback with this separation?

Something something platform :slightly_smiling_face: We need to persist this stuff when running on Apify, so it needs to be a part of `user_data` in the end. I guess we could separate the representation of "Crawlee Request" and "Payload for enqueue request API call" better, probably with/after #94.

B4nan:
> Something something platform 🙂 

Correct, the request fields are validated, we cant just add something the platform doesn't expect.

PR #542, opened 2024-09-24
fix!: merge payload and data fields of Request

### LINKED ISSUE #560: Unable to execute POST request with JSON payload
Example

```python
async def main() -> None:
    crawler = HttpCrawler()

    # Define the default request handler, which will be called for every request.
    async def request_handler(context: HttpCrawlingContext) -> None:
        context.log.info(f'Processing {context.request.url} ...')
        response = context.http_response.read().decode('utf-8')
        context.log.info(f'Response: {response}')  # To see the response in the logs.

    # Prepare a POST request to the form endpoint.
    request = Request.from_url(
        url='https://httpbin.org/post',
        method='POST',
        headers = {"content-type": "application/json"},
        data={
            'custname': 'John Doe',
            'custtel': '1234567890',
            'custemail': 'johndoe@example.com',
            'size': 'large',
            'topping': ['bacon', 'cheese', 'mushroom'],
            'delivery': '13:00',
            'comments': 'Please ring the doorbell upon arrival.',
        },
    )
    await crawler.run([request])
```

Current response format

```json
{
  "args": {}, 
  "data": "custname=John+Doe&custtel=1234567890&custemail=johndoe%40example.com&size=large&topping=bacon&topping=cheese&topping=mushroom&delivery=13%3A00&comments=Please+ring+the+doorbell+upon+arrival.", 
  "files": {}, 
  "form": {}, 
  "headers": {
    "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7", 
    "Accept-Encoding": "gzip, deflate, br", 
    "Accept-Language": "en-US,en;q=0.9", 
    "Content-Length": "190", 
    "Content-Type": "application/json", 
    "Host": "httpbin.org", 
    "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36", 
    "X-Amzn-Trace-Id": "Root=1-66fc90cf-11644d4f5483e1096211b721"
  }, 
  "json": null, 
  "origin": "91.240.96.149", 
  "url": "https://httpbin.org/post"
}

```

Expected response format

```json
{
  "args": {}, 
  "data": "{\"custname\": \"John Doe\", \"custtel\": \"1234567890\", \"custemail\": \"johndoe@example.com\", \"size\": \"large\", \"topping\": [\"bacon\", \"cheese\", \"mushroom\"], \"delivery\": \"13:00\", \"comments\": \"Please ring the doorbell upon arrival.\"}", 
  "files": {}, 
  "form": {}, 
  "headers": {
    "Accept": "*/*", 
    "Accept-Encoding": "gzip, deflate, br", 
    "Content-Length": "221", 
    "Content-Type": "application/json", 
    "Host": "httpbin.org", 
    "User-Agent": "python-httpx/0.27.2", 
    "X-Amzn-Trace-Id": "Root=1-66fc91c2-6db9989347fef25b150615e2"
  }, 
  "json": {
    "comments": "Please ring the doorbell upon arrival.", 
    "custemail": "johndoe@example.com", 
    "custname": "John Doe", 
    "custtel": "1234567890", 
    "delivery": "13:00", 
    "size": "large", 
    "topping": [
      "bacon", 
      "cheese", 
      "mushroom"
    ]
  }, 
  "origin": "91.240.96.149", 
  "url": "https://httpbin.org/post"
}
```

Both HTTPX and curl_impersonate allow creating a “POST” request with JSON payload, in two ways

```python
data={
    'custname': 'John Doe',
    'custtel': '1234567890',
    'custemail': 'johndoe@example.com',
    'size': 'large',
    'topping': ['bacon', 'cheese', 'mushroom'],
    'delivery': '13:00',
    'comments': 'Please ring the doorbell upon arrival.',
    }

response = httpx.post(url, json=data)

response = httpx.post(url, data=json.dumps(data))
```

But we can't reproduce this behavior in Crawlee, because the [`json` parameter is not passed when the request is created](https://github.com/apify/crawlee-python/blob/master/src/crawlee/http_clients/_httpx.py#L177) and the [`data` parameter cannot be a string](https://github.com/apify/crawlee-python/blob/master/src/crawlee/_request.py#L130)

vdusek:
### Description

- We had `data` and `payload` fields on the `Request` model.
  - `payload` was not being provided to the HTTP clients, only the `data` field.
- ~In this PR, I'm merging them together, keeping only the `data` field (use the same naming as `HTTPX` & `CurlImpersonate`).~
- In this PR, I'm merging them together, keeping only the `payload` field (use the same naming as in JS Crawlee).
- ~I also defined type aliases for HTTP data and HTTP query parameters and used them across the project.~
- ~Some struggle with Pydantic & serialization, but should be OK now...~

### Issues

- Reported by @honzajavorek on Slack
- Closes #560

### Testing

- The current set of unit tests should cover the changes.

### Checklist

- [x] CI passed

### DIFF
--- a/src/crawlee/_request.py
+++ b/src/crawlee/_request.py
@@ -133,7 +133,10 @@ class BaseRequestData(BaseModel):
     query_params: Annotated[HttpQueryParams, Field(alias='queryParams', default_factory=dict)] = {}
     """URL query parameters."""
 
-    payload: Annotated[HttpPayload, Field(default_factory=dict)] = {}
+    payload: Annotated[
+        HttpPayload | None,
+        PlainValidator(lambda value: None if value is None else str(value).encode('utf-8')),
+    ] = None
     """HTTP request payload."""
 
     user_data: Annotated[
@@ -181,7 +184,6 @@ class BaseRequestData(BaseModel):
         """Create a new `BaseRequestData` instance from a URL. See `Request.from_url` for more details."""
         headers = headers or HttpHeaders()
         query_params = query_params or {}
-        payload = payload or {}
 
         unique_key = unique_key or compute_unique_key(
             url,
@@ -287,7 +289,6 @@ class Request(BaseRequestData):
         """
         headers = headers or HttpHeaders()
         query_params = query_params or {}
-        payload = payload or {}
 
         unique_key = unique_key or compute_unique_key(
             url,
--- a/src/crawlee/_request.py
+++ b/src/crawlee/_request.py
@@ -133,10 +133,7 @@ class BaseRequestData(BaseModel):
     query_params: Annotated[HttpQueryParams, Field(alias='queryParams', default_factory=dict)] = {}
     """URL query parameters."""
 
-    payload: Annotated[
-        HttpPayload | None,
-        PlainValidator(lambda value: None if value is None else str(value).encode('utf-8')),
-    ] = None
+    payload: HttpPayload | None = None
     """HTTP request payload."""
 
     user_data: Annotated[
--- a/src/crawlee/_request.py
+++ b/src/crawlee/_request.py
@@ -188,6 +188,7 @@ class BaseRequestData(BaseModel):
         unique_key = unique_key or compute_unique_key(
             url,
             method=method,
+            headers=headers,
             payload=payload,
             keep_url_fragment=keep_url_fragment,
             use_extended_unique_key=use_extended_unique_key,
@@ -293,6 +294,7 @@ class Request(BaseRequestData):
         unique_key = unique_key or compute_unique_key(
             url,
             method=method,
+            headers=headers,
             payload=payload,
             keep_url_fragment=keep_url_fragment,
             use_extended_unique_key=use_extended_unique_key,
--- a/src/crawlee/_request.py
+++ b/src/crawlee/_request.py
@@ -127,15 +127,14 @@ class BaseRequestData(BaseModel):
     method: HttpMethod = 'GET'
     """HTTP request method."""
 
-    headers: Annotated[HttpHeaders, Field(default_factory=HttpHeaders())] = HttpHeaders()
+    headers: Annotated[HttpHeaders, Field(default_factory=HttpHeaders)] = HttpHeaders()
     """HTTP request headers."""
 
     query_params: Annotated[HttpQueryParams, Field(alias='queryParams', default_factory=dict)] = {}
     """URL query parameters."""
 
-    payload: HttpPayload | None = None
-
-    data: Annotated[dict[str, Any], Field(default_factory=dict)] = {}
+    payload: Annotated[HttpPayload, Field(default_factory=dict)] = {}
+    """HTTP request payload."""
 
     user_data: Annotated[
         dict[str, JsonSerializable],  # Internally, the model contains `UserData`, this is just for convenience
@@ -169,6 +168,8 @@ class BaseRequestData(BaseModel):
         url: str,
         *,
         method: HttpMethod = 'GET',
+        headers: HttpHeaders | None = None,
+        query_params: HttpQueryParams | None = None,
         payload: HttpPayload | None = None,
         label: str | None = None,
         unique_key: str | None = None,
@@ -178,6 +179,10 @@ class BaseRequestData(BaseModel):
         **kwargs: Any,
     ) -> Self:
         """Create a new `BaseRequestData` instance from a URL. See `Request.from_url` for more details."""
+        headers = headers or HttpHeaders()
+        query_params = query_params or {}
+        payload = payload or {}
+
         unique_key = unique_key or compute_unique_key(
             url,
             method=method,
@@ -193,6 +198,8 @@ class BaseRequestData(BaseModel):
             unique_key=unique_key,
             id=id,
             method=method,
+            headers=headers,
+            query_params=query_params,
             payload=payload,
             **kwargs,
         )
@@ -243,6 +250,8 @@ class Request(BaseRequestData):
         url: str,
         *,
         method: HttpMethod = 'GET',
+        headers: HttpHeaders | None = None,
+        query_params: HttpQueryParams | None = None,
         payload: HttpPayload | None = None,
         label: str | None = None,
         unique_key: str | None = None,
@@ -261,6 +270,8 @@ class Request(BaseRequestData):
         Args:
             url: The URL of the request.
             method: The HTTP method of the request.
+            headers: The HTTP headers of the request.
+            query_params: The query parameters of the URL.
             payload: The data to be sent as the request body. Typically used with 'POST' or 'PUT' requests.
             label: A custom label to differentiate between request types. This is stored in `user_data`, and it is
                 used for request routing (different requests go to different handlers).
@@ -274,6 +285,10 @@ class Request(BaseRequestData):
                 computation. This is only relevant when `unique_key` is not provided.
             **kwargs: Additional request properties.
         """
+        headers = headers or HttpHeaders()
+        query_params = query_params or {}
+        payload = payload or {}
+
         unique_key = unique_key or compute_unique_key(
             url,
             method=method,
@@ -289,6 +304,8 @@ class Request(BaseRequestData):
             unique_key=unique_key,
             id=id,
             method=method,
+            headers=headers,
+            query_params=query_params,
             payload=payload,
             **kwargs,
         )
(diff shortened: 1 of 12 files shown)


B4nan:
JS version have only `payload`, we should aim for compact with that in the first place.

honzajavorek:
I learned about `data` here, but I don't see it in changed files: https://crawlee.dev/python/docs/examples/fill-and-submit-web-form#preparing-a-post-request

honzajavorek:
It also says "Alternatively, you can send form data as URL parameters using the query_params argument." btw, not sure if that is something related or not.

vdusek:
> Also, since this is breaking, maybe we should wait, and merge it once we have more breaking changes than a simple one like this (or some feature worth the minor bump).

We will have the first version of fingerprints for Playwright soon :crossed_fingers:.

> I learned about data here, but I don't see it in changed files: https://crawlee.dev/python/docs/examples/fill-and-submit-web-form#preparing-a-post-request

Thanks. Of course, I forgot to update the docs.

src/crawlee/_types.py
```diff
@@ -26,6 +26,10 @@
 
 HttpMethod: TypeAlias = Literal['GET', 'HEAD', 'POST', 'PUT', 'DELETE', 'CONNECT', 'OPTIONS', 'TRACE', 'PATCH']
 
+HttpQueryParams: TypeAlias = dict[str, Any]
+
+HttpPayload: TypeAlias = dict[str, Any]
```
janbuchar, line 31: The `dict` is kinda surprising here - I'd expect `bytes` or something. How did this happen? :smile:

src/crawlee/_request.py
```diff
@@ -99,7 +99,7 @@ def __len__(self) -> int:
 class BaseRequestData(BaseModel):
     """Data needed to create a new crawling request."""
 
-    model_config = ConfigDict(populate_by_name=True)
+    model_config = ConfigDict(populate_by_name=True, arbitrary_types_allowed=True)
```
janbuchar, line 102: This is kinda scary. Maybe we could make `HttpHeaders` a `RootModel` instead (https://docs.pydantic.dev/dev/concepts/models/#rootmodel-and-custom-root-types)?

src/crawlee/_types.py
```diff
@@ -26,6 +26,10 @@
 
 HttpMethod: TypeAlias = Literal['GET', 'HEAD', 'POST', 'PUT', 'DELETE', 'CONNECT', 'OPTIONS', 'TRACE', 'PATCH']
 
+HttpQueryParams: TypeAlias = dict[str, Any]
```
janbuchar, line 29: I think the `Any` is too generous - `dict[str, str]` should be enough, right?

janbuchar (changes requested):
Nice! Just a couple of questions...

src/crawlee/_types.py
vdusek: The downstream libraries (HTTPX & CurlImpersonate) do the conversions. And both of them accept `RequestData = Mapping[str, Any]` type.

janbuchar: Right. I did some digging and this is what I found. `payload` is part of the `Request` schema on the API, and the API client requires it to be a string. It'll get serialized and passed to the API endpoint on platform. Of course, it also needs to be compatible with whatever the HTTP client accepts - that's where the actual processing gets done.

As a side note, I'm not sure what `data` was supposed to be, but it seems it's not needed...

src/crawlee/_request.py
vdusek: Ok, should be solved...

vdusek:
> Also, since this is breaking, maybe we should wait, and merge it once we have more breaking changes than a simple one like this (or some feature worth the minor bump).

I've separated the Curl impersonation fix and other minor updates into a separate PR (#543), allowing us to release a patch version for 0.3.

The `payload` & `data` merge and `HttpHeaders` thing will be resolved later...

vdusek:
A new model for `HttpHeaders` was also moved to a separate PR #544 and merged.

src/crawlee/_types.py
vdusek: payload is bytes type now

tests/unit/http_crawler/test_http_crawler.py
```diff
@@ -201,3 +206,37 @@ async def test_http_status_statistics(crawler: HttpCrawler, server: respx.MockRo
...
+    @crawler.router.default_handler
+    async def request_handler(context: HttpCrawlingContext) -> None:
+        response = json.loads(context.http_response.read())
+        # The httpbin.org/post endpoint returns the form data in the response.
+        assert response['form'] == form_data
+
+    request = Request.from_url(
+        url='https://httpbin.org/post',
+        method='POST',
+        payload=str(form_data).encode(),
```
janbuchar, line 239: Depending on your content type, you might want to use `urlencode.urlencode` or some library that will let you encode it as a multipart form.

janbuchar (changes requested):

tests/unit/http_crawler/test_http_crawler.py
```diff
@@ -201,3 +207,37 @@ async def test_http_status_statistics(crawler: HttpCrawler, server: respx.MockRo
...
+        'topping': '["bacon", "cheese", "mushroom"]',
+        'delivery': '13:00',
+        'comments': 'Please ring the doorbell upon arrival.',
+    }
+
+    @crawler.router.default_handler
+    async def request_handler(context: HttpCrawlingContext) -> None:
+        response = json.loads(context.http_response.read())
+        # The httpbin.org/post endpoint returns the form data in the response.
+        assert response['form'] == form_data
```
janbuchar, line 235: Could you also assert that the request_handler was called?

src/crawlee/_request.py
```diff
@@ -127,15 +127,17 @@ class BaseRequestData(BaseModel):
     method: HttpMethod = 'GET'
     """HTTP request method."""
 
-    headers: Annotated[HttpHeaders, Field(default_factory=HttpHeaders())] = HttpHeaders()
+    headers: Annotated[HttpHeaders, Field(default_factory=HttpHeaders)] = HttpHeaders()
     """HTTP request headers."""
 
     query_params: Annotated[HttpQueryParams, Field(alias='queryParams', default_factory=dict)] = {}
```
janbuchar, line 133: We should probably get rid of this one, or make it a property that changes the `url` field. Unfortunately, Apify API does not accept `queryParams`.

janbuchar (changes requested):

janbuchar:
Nice work @vdusek, and sorry for taking multiple rounds of review to pinpoint all issues.

src/crawlee/_request.py
vdusek: Yeah... Opened a new issue for it - #615.

tests/unit/http_crawler/test_http_crawler.py
vdusek: Oh, good catch.

janbuchar (commented):
Looking good, just a nit.

src/crawlee/_request.py
```diff
@@ -127,15 +127,14 @@ class BaseRequestData(BaseModel):
     method: HttpMethod = 'GET'
     """HTTP request method."""
 
-    headers: Annotated[HttpHeaders, Field(default_factory=HttpHeaders())] = HttpHeaders()
```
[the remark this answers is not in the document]
vdusek: I believe `Field(default_factory=HttpHeaders())` is actually used.

The `= HttpHeaders()` is just for the Mypy. Otherwise, it approaches it as a mandatory argument in the constructor.

tests/unit/http_crawler/test_http_crawler.py
```diff
@@ -201,3 +207,50 @@ async def test_http_status_statistics(crawler: HttpCrawler, server: respx.MockRo
...
+    responses = []
+
+    @crawler.router.default_handler
+    async def request_handler(context: HttpCrawlingContext) -> None:
+        response = json.loads(context.http_response.read())
+        # The httpbin.org/post endpoint returns the form data in the response.
+        responses.append(response)
+
+    request = Request.from_url(
+        url='https://httpbin.org/post',
```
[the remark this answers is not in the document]
vdusek: The `forms/post` endpoint shows you the web form. But you are submitting it using the "Submit order" button, which packs the data and sends them to the `https://httpbin.org/post` endpoint, if that makes sense.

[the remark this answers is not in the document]
janbuchar: My concern is, the `/post` endpoint just echoes whatever you send to it, right? So the whole form url-encoding and decoding becomes a pointless exercise, if I'm not missing anything.

janbuchar (approved):

PR #543, opened 2024-09-25
fix: adopt new version of curl-cffi

vdusek:
### Description

- Version 1.7.2 of `curl-cffi` introduces breaking changes.
- This update includes adopting the new version, adding type aliases for `Request` fields, and incorporating other minor changes from PR #542.

### Issues

- N/A (ad-hoc fix)

### Testing

- The current set of unit tests should cover the changes.

### Checklist

- [x] CI passed

### DIFF
--- a/src/crawlee/http_clients/curl_impersonate.py
+++ b/src/crawlee/http_clients/curl_impersonate.py
@@ -4,7 +4,8 @@ from typing import TYPE_CHECKING, Any, Optional
 
 try:
     from curl_cffi.requests import AsyncSession
-    from curl_cffi.requests.errors import RequestsError
+    from curl_cffi.requests.exceptions import ProxyError as CurlProxyError
+    from curl_cffi.requests.exceptions import RequestException as CurlRequestError
     from curl_cffi.requests.impersonate import BrowserType
 except ImportError as exc:
     raise ImportError(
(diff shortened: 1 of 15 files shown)


src/crawlee/http_clients/curl_impersonate.py
```diff
@@ -194,7 +195,7 @@ def _get_client(self, proxy_url: str | None) -> AsyncSession:
             # are set as default options.
             kwargs: dict[str, Any] = {
                 'proxy': proxy_url,
-                'impersonate': BrowserType.chrome,
+                'impersonate': BrowserType.chrome124,
```
janbuchar, line 198: Isn't there something like a "newest chrome" option?

janbuchar (approved):
LGTM, one thing to clarify

src/crawlee/http_clients/curl_impersonate.py
vdusek: Unfortunately, no. The `BrowserType.chrome` was the most recent option, but it has been removed. However, I discovered that there is a `DEFAULT_CHROME` constant that represents the latest version, so we can use that instead.

B4nan:
lol, so even in python world maintainers don't care about semver :D

janbuchar:
> lol, so even in python world maintainers don't care about semver :D

I don't think phenomenon is language-specific :smile:

PR #544, opened 2024-09-25
feat: add new model for `HttpHeaders`

vdusek:
### Description

- Unify the HTTP headers type across the project.
- ~I used just the type alias plus custom validator in the `Request` model to reach the same result (lowercase & sorted).~
  - ~Of course, now it is only in the `Request`, but I believe it is not a problem.~
- Take it out from #542.

### Issues

- N/A

### Testing

- N/A

### Checklist

- [x] CI passed

### DIFF
--- a/src/crawlee/_request.py
+++ b/src/crawlee/_request.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from collections.abc import Iterator, Mapping, MutableMapping
+from collections.abc import Iterator, MutableMapping
 from datetime import datetime
 from decimal import Decimal
 from enum import IntEnum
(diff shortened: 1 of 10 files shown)


src/crawlee/_request.py
```diff
@@ -119,7 +120,12 @@ class BaseRequestData(BaseModel):
     method: HttpMethod = 'GET'
     """HTTP request method."""
 
-    headers: Annotated[dict[str, str], Field(default_factory=dict)] = {}
+    headers: Annotated[
+        HttpHeaders,
+        # Normalize headers to lowercase keys and sort them.
+        PlainValidator(lambda value: normalize_headers(value)),
+        Field(default_factory={}),
```
janbuchar, line 127: Shouldn't `default_factory` be a callable? And do we really need it?

janbuchar (commented):
What about http client? That uses `HttpHeaders` as well and I'd expect normalization to work there.

src/crawlee/_request.py
vdusek: oh, good catch

src/crawlee/_request.py
```diff
@@ -119,7 +166,7 @@ class BaseRequestData(BaseModel):
     method: HttpMethod = 'GET'
     """HTTP request method."""
 
-    headers: Annotated[dict[str, str], Field(default_factory=dict)] = {}
+    headers: Annotated[HttpHeaders, Field(default_factory=HttpHeaders())] = HttpHeaders()
```
janbuchar, line 169: Suspicious default factory, do we really need both that and a regular default?

Also, doesn't this cause `Request` to serialize to `{..., "headers": {"headers": {...}}, ...}` now? Wouldn't a `RootModel` be more appropriate? :slightly_smiling_face:

src/crawlee/_request.py
```diff
@@ -25,6 +25,53 @@
...
 
 
+def _normalize_headers(headers: Mapping[str, str]) -> dict[str, str]:
+    """Converts all header keys to lowercase and returns them sorted by key."""
+    normalized_headers = {k.lower(): v for k, v in headers.items()}
+    sorted_headers = sorted(normalized_headers.items())
+    return dict(sorted_headers)
+
+
+class HttpHeaders(BaseModel, Mapping[str, str]):
```
janbuchar, line 35: Maybe we could move this to utils, types or something?

vdusek: I will leave it in the `_types` then.

src/crawlee/_request.py
vdusek: > Suspicious default factory, do we really need both that and a regular default?

yeah, mypy

> Also, doesn't this cause Request to serialize to {..., "headers": {"headers": {...}}, ...} now? Wouldn't a RootModel be more appropriate? 🙂

Thanks!

janbuchar (approved):
LGTM

Issue #551, opened 2024-09-29
Failing pydantic validation on request_dict 


cadlagtrader:
When running the httpx crawler on v0.3.7
My request_dict
```'{"url": "XXXXXX", "unique_key": "YYYYY", "method": "GET", "headers": {}, "query_params": {}, "data": {}, "user_data": {"__crawlee": {"state": "RequestState.ERROR_HANDLER"}}, "retry_count": 1, "no_retry": false, "id": "doOfa0tGsjBwT90"}'```

create the exception below when running ```Request.model_validate(request_dict)``` 
```pydantic_core._pydantic_core.ValidationError: 1 validation error for Request
user_data.__crawlee.state
  Input should be 0, 1, 2, 3, 4, 5, 6 or 7 [type=enum, input_value='RequestState.ERROR_HANDLER', input_type=str]
    For further information visit https://errors.pydantic.dev/2.9/v/enum
```

janbuchar:
Hello and thank you for your interest in Crawlee! Where does that request_dict come from? The state property is suspicious.

cadlagtrader:
Hello - the request_dict comes from a deserialised request from my crawlee storage. There is a bug in Crawlee enum serialisation/deserialisation through Pydantic (I am using v2.9.2, compatible with crawlee pyproject constraints)

I have modifed ```test_complex_user_data_serialization``` to highlight the issue: https://github.com/cadlagtrader/crawlee-python/blob/bug_with_pydantic/tests/unit/storages/test_request_queue.py

PR #556, opened 2024-09-30
fix: extend from `StrEnum` in `RequestState` to fix serialization

vdusek:
Closes: #551


janbuchar (changes requested):
1. let's consider if we want to keep this compatible with crawlee-js, where we store numbers
2. we could just use `IntEnum` or `(str, Enum)` - https://docs.pydantic.dev/dev/api/standard_library_types/#enum - I don't want to maintain duplicates of standard library code

vdusek:
> let's consider if we want to keep this compatible with crawlee-js, where we store numbers

I see a slight advantage in serializing it as something more readable, like `{"state": "Unprocessed"}`, instead of using just numbers, such as `{"state": 1}`.

> we could just use IntEnum or (str, Enum) - https://docs.pydantic.dev/dev/api/standard_library_types/#enum - I don't want to maintain duplicates of standard library code

Oh, that's good, thanks.

janbuchar:
> > let's consider if we want to keep this compatible with crawlee-js, where we store numbers
> 
> I see a slight advantage in serializing it as something more readable, like `{"state": "Unprocessed"}`, instead of using just numbers, such as `{"state": 1}`.

Yeah, I agree with that. I guess it boils down to whether we want the two formats to be compatible. Frankly, I don't see any advantage in that.

vdusek:
> Yeah, I agree with that. I guess it boils down to whether we want the two formats to be compatible. Frankly, I don't see any advantage in that.

Ok, let's use `IntEnum` then.

janbuchar:
> > Yeah, I agree with that. I guess it boils down to whether we want the two formats to be compatible. Frankly, I don't see any advantage in that.
> 
> Ok, let's use `IntEnum` then.

Just so we're clear, we're both OK with string values, but we're keeping the integers for the sake of compatibility, right?

janbuchar:
Could you also incorporate the test changes from https://github.com/cadlagtrader/crawlee-python/commit/02a5c6b9b35f0573068d878b4f4367a6a1d8fbf5?

vdusek:
> Could you also incorporate the test changes from https://github.com/cadlagtrader/crawlee-python/commit/02a5c6b9b35f0573068d878b4f4367a6a1d8fbf5?

I already implemented a new test covering this.

janbuchar:
> > Could you also incorporate the test changes from [cadlagtrader@02a5c6b](https://github.com/cadlagtrader/crawlee-python/commit/02a5c6b9b35f0573068d878b4f4367a6a1d8fbf5)?
> 
> I already implemented a new test covering this.

Yeah, I mean just to make it clear that the reported issue is *fixed* :smile:

B4nan:
We can change the enum in JS version to store numbers, no problems with that. Generally speaking, numbers will be more performant and less memory sensitive than strings, but we already store a lot of stuff in the request object that I doubt this can be measurable.

janbuchar:
> We can change the enum in JS version to store numbers, no problems with that. Generally speaking, numbers will be more performant and less memory sensitive than strings, but we already store a lot of stuff in the request object that I doubt this can be measurable.

image

...apparently, that's already the case. The performance impact is most likely non-existent, yes.

janbuchar (approved):

Issue #546, opened 2024-09-27
Error handler is not called for session errors


janbuchar:
See https://github.com/apify/crawlee/pull/2683

PR #557, opened 2024-09-30
fix: call `error_handler` for `SessionError`

vdusek:
Closes: #546

### DIFF
--- a/src/crawlee/basic_crawler/_basic_crawler.py
+++ b/src/crawlee/basic_crawler/_basic_crawler.py
@@ -865,6 +865,9 @@ class BasicCrawler(Generic[TCrawlingContext]):
             else:
                 self._logger.exception('Request failed and reached maximum retries', exc_info=session_error)
 
+                if self._error_handler:
+                    await self._error_handler(crawling_context, session_error)
+
                 await wait_for(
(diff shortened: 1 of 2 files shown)


src/crawlee/basic_crawler/_basic_crawler.py
```diff
@@ -865,6 +865,9 @@ async def __run_task_function(self) -> None:
             else:
                 self._logger.exception('Request failed and reached maximum retries', exc_info=session_error)
 
+                if self._error_handler:
+                    await self._error_handler(crawling_context, session_error)
+
```
janbuchar, line 870: This is different from what the JS version does - the error handler should be called before retries, not after all attempts have failed.

janbuchar (approved):

src/crawlee/basic_crawler/_basic_crawler.py
```diff
@@ -851,6 +851,9 @@ async def __run_task_function(self) -> None:
             if not crawling_context.session:
                 raise RuntimeError('SessionError raised in a crawling context without a session') from session_error
 
+            if self._error_handler:
+                await self._error_handler(crawling_context, session_error)
```
janbuchar, line 855: Now that I think of it, for regular (non-SessionError) failures we also handle the case when this throws as well.

PR #565, opened 2024-10-03
docs: homepage changes

souravjain540:
few changes on homepage for SEO as requested by marketing

### DIFF
--- a/website/src/pages/index.js
+++ b/website/src/pages/index.js
@@ -21,10 +21,7 @@ function Hero() {
                     <div className={clsx(styles.relative, 'row')}>
                         <div className="col">
                             <h1 className={styles.tagline}>
-                                Crawlee is a web<br /> scraping and browser<br /> automation library
-                            </h1>
-                            <h1 className={styles.tagline}>
-                                Crawlee is a <span>web<br /> scraping</span> and <span>browser<br /> automation</span> library
+                                Build your <span>Python<br /> web crawlers</span><br /> using <span>Crawlee</span> 
                             </h1>
                         </div>
                     </div>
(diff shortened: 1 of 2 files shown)


B4nan:
where is this coming from? are the changes proposed by someone else or are they coming from you? it feels weird to bend the homepage like this just because of SEO (and i have doubts it will have any positive effect, homepage is just one page, and adding the word "python" and "web" all over it won't make things better anyhow). i cant say i like the changes. 

also note that we plan to rework the homepage to be shared for both js and python, those changes will likely be overridden anyway. 

this should be somehow coordinated, we cant just change catchphrases on homepage because someone doesn't like some SEO report. not to mention some of those changes are going against what was there before. e.g. we already had a long title, it was changed to what we have now because it was too long. now you made it longer than before.

souravjain540:
@B4nan its from marketing, check here: https://apify.slack.com/archives/C03NESUMYQN/p1726825030526109?thread_ts=1718803289.053979&cid=C03NESUMYQN

website/src/pages/index.js
```diff
@@ -21,16 +21,13 @@ function Hero() {
                     <div className={clsx(styles.relative, 'row')}>
                         <div className="col">
                             <h1 className={styles.tagline}>
-                                Crawlee is a web<br /> scraping and browser<br /> automation library
-                            </h1>
-                            <h1 className={styles.tagline}>
-                                Crawlee is a <span>web<br /> scraping</span> and <span>browser<br /> automation</span> library
+                                Build your <span>Python<br /> web crawlers</span><br /> using <span>Crawlee</span>
```
B4nan, line 24: by removing the second h1 you disabled the gradient for the wrapped words, it looks quite ugly now
image

PR #569, opened 2024-10-04
docs: Add guide on how to avoid getting blocked

### LINKED ISSUE #481: Create a new guide about how to not get blocked
- We should create a new documentation guide about how to not get blocked.
- Inspiration: https://crawlee.dev/docs/guides/avoid-blocking
- This should be done once fingerprint-related issues are done (#401, #402, #549, ...).

akash47angadi:
### Description

This PR adds a new documentation guide on how to avoid getting blocked while scraping. The guide covers browser fingerprints, how to customize them, and best practices to reduce blocking.

### Issues

Fixes #481

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,38 +129,39 @@ The [`PlaywrightCrawler`](https://crawlee.dev/python/api/class/PlaywrightCrawler
 ```python
 import asyncio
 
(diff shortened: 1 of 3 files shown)


B4nan:
Same as with the other one, the PR is not doing anything else than changing the example code in our readme. The task is not about that at all.

Issue #514, opened 2024-09-10
Implement `KeyValueStore.get_public_url`


janbuchar:
- https://github.com/apify/crawlee/blob/master/packages/core/src/storages/key_value_store.ts#L457-L463 could serve as inspiration
  - we should consider also making this a method of the kv-store resource client and implementing it separately for memory storage and for platform storage in SDK
- when this is done, update the deployment/apify-platform docs page (there is commented out doc for this feature)

ravi-hash:
Hey will you assign this issue to me

naaa760:
Hello! @janbuchar  :)


I think we need to add some checks and better handling of errors. First, we should make sure that the key and **storeName** are valid before trying to generate a URL. Depending on whether we’re using memory storage or platform storage, we handle them differently: for memory storage, we can return a local URL **(like localhost)**, but we need to be aware that this might not be publicly accessible. For platform storage, we should create a URL that fits the platform’s format. If the **storageType** is unknown, we throw an error so the issue is clear. Wrapping everything in try-catch helps us catch any unexpected problems and log useful error messages, making the code more reliable.

janbuchar:
Hello @ravi-hash and @naaa760, we will accept a PR from whomever finishes it first, no need to have it assigned first.

PR #572, opened 2024-10-06
feat: Added get_public_url method to KeyValueStore

akshay11298:
### Description

Implement get_public_url method in KeyValueStore

- TODO

### Issues

- Closes: #514 

### Testing

- Unit tests added

### Checklist

- [ ] CI passed

### DIFF
--- a/src/crawlee/storages/_key_value_store.py
+++ b/src/crawlee/storages/_key_value_store.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import os
 from typing import TYPE_CHECKING, Any, AsyncIterator, TypeVar, overload
 
 from typing_extensions import override
@@ -157,3 +158,14 @@ class KeyValueStore(BaseStorage):
             return await self._resource_client.delete_record(key)
 
         return await self._resource_client.set_record(key, value, content_type)
+
+    def get_public_url(self, key: str) -> str:
+        """Get the public URL for the given key.
+
+        Args:
+            key: Key of the record for which URL is required
+        Returns:
+            The public URL for the given key.
+        """
+        name = self.name or self._configuration.default_key_value_store_id
+        return f'file://{os.getcwd()}/storage/key_value_stores/{name}/{key}'
--- a/tests/unit/storages/test_key_value_store.py
+++ b/tests/unit/storages/test_key_value_store.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import os
 from typing import AsyncGenerator
 
 import pytest
@@ -100,3 +101,9 @@ async def test_static_get_set_value(key_value_store: KeyValueStore) -> None:
     await key_value_store.set_value('test-static', 'static')
     value = await key_value_store.get_value('test-static')
     assert value == 'static'
+
+
+async def test_static_get_public_url(key_value_store: KeyValueStore) -> None:
+    await key_value_store.set_value('test-static', 'static')
+    public_url = key_value_store.get_public_url('test-static')
+    assert public_url == f'file://{os.getcwd()}/storage/key_value_stores/default/test-static'


janbuchar:
This looks like the right approach. You should use the `storage_dir` attribute of the configuration object instead of the current directory though.

akshay11298:
> This looks like the right approach. You should use the `storage_dir` attribute of the configuration object instead of the current directory though.

@janbuchar Thanks for the review and hint :) I have updated the PR with storage_dir

tests/unit/storages/test_key_value_store.py
```diff
@@ -100,3 +100,12 @@ async def test_static_get_set_value(key_value_store: KeyValueStore) -> None:
...
     assert value == 'static'
+
+
+async def test_static_get_public_url(key_value_store: KeyValueStore) -> None:
+    await key_value_store.set_value('test-static', 'static')
+    public_url = key_value_store.get_public_url('test-static')
+    assert (
+        public_url
+        == f'file://{key_value_store._configuration.storage_dir}/storage/key_value_stores/default/test-static'
+    )
```
janbuchar, line 111: This is not very useful. Could you perhaps try to open and read the URL instead and check if the content is actually `static`?

src/crawlee/storages/_key_value_store.py
```diff
@@ -157,3 +157,15 @@ async def set_value(
...
+        """Get the public URL for the given key.
+
+        Args:
+            key: Key of the record for which URL is required
+        Returns:
+            The public URL for the given key.
+        """
+        name = self.name or self._configuration.default_key_value_store_id
+        storage_dir = self._configuration.storage_dir
+        return f'file://{storage_dir}/storage/key_value_stores/{name}/{key}'
```
janbuchar, line 171: Issue #514 mentions that we should put the actual logic for determining the public url into the `KeyValueStoreClient` class (file `src/crawlee/memory_storage_client/_key_value_store_client.py`). And call the method from here. Can you please do that?

This will allow the Apify SDK to replace the functionality in a cleaner fashion.

janbuchar, line 171: I believe the `/storage` part is now redundant here.

janbuchar (changes requested):

src/crawlee/storages/_key_value_store.py
```diff
@@ -157,3 +157,15 @@ async def set_value(
...
 
         return await self._resource_client.set_record(key, value, content_type)
+
+    def get_public_url(self, key: str) -> str:
+        """Get the public URL for the given key.
+
+        Args:
+            key: Key of the record for which URL is required
+        Returns:
+            The public URL for the given key.
```
vdusek, line 167:
```suggestion
        Args:
            key: Key of the record for which URL is required

        Returns:
            The public URL for the given key.
```

vdusek (changes requested):

src/crawlee/storages/_key_value_store.py
akshay11298: Updated the spacing

src/crawlee/storages/_key_value_store.py
akshay11298: From test execution, I guess self._configuration.storage_dir doesn't contain /storage

akshay11298: Moved the implementation to client

tests/unit/storages/test_key_value_store.py
akshay11298: Since kv is created by a generator, it yields a different value each time. Can you please help me with how I can make that static?

janbuchar: I'm not sure I understand. Why can't you just do this?

```py
from urllib.parse import urlparse, unquote
from pathlib import Path
# ...

path = unquote(urlparse(public_url).path)
assert Path(path).read_text() == 'static'
```

src/crawlee/base_storage_client/_base_key_value_store_client.py
```diff
@@ -114,3 +115,16 @@ async def delete_record(self, key: str) -> None:
         Args:
             key: The key of the record which to delete
         """
+
+    @abstractmethod
+    def get_public_url(self, key: str, name: str | None, configuration: Configuration) -> str:
+        """Get the public URL for the given key.
+
+        Args:
+            name: Identifier for the key value store
```
janbuchar, line 124: This isn't necessary - the implementation should already know the name of the key-value store it is handling.

src/crawlee/base_storage_client/_base_key_value_store_client.py
```diff
@@ -114,3 +115,16 @@ async def delete_record(self, key: str) -> None:
...
+            key: Key of the record for which URL is required
+            configuration: Configuration of the key value store
```
janbuchar, line 126: This is also not necessary - you can already access this in memory storage clients and other implementations won't need it.

src/crawlee/memory_storage_client/_key_value_store_client.py
```diff
@@ -287,6 +288,12 @@ async def delete_record(self, key: str) -> None:
                 if self._memory_storage_client.persist_storage:
                     await existing_store_by_id.delete_persisted_record(record)
 
+    @override
+    def get_public_url(self, key: str, name: str | None, configuration: Configuration) -> str:
+        store_name = name or configuration.default_key_value_store_id
+        storage_dir = configuration.storage_dir
```
janbuchar, line 294:
```suggestion
        storage_dir = self._memory_storage_client.storage_dir
```

src/crawlee/memory_storage_client/_key_value_store_client.py
```diff
@@ -287,6 +288,12 @@ async def delete_record(self, key: str) -> None:
                 if self._memory_storage_client.persist_storage:
                     await existing_store_by_id.delete_persisted_record(record)
 
+    @override
+    def get_public_url(self, key: str, name: str | None, configuration: Configuration) -> str:
+        store_name = name or configuration.default_key_value_store_id
```
janbuchar, line 293:
```suggestion
        store_name = self.name or self.id
```

janbuchar (changes requested):

src/crawlee/base_storage_client/_base_key_value_store_client.py
akshay11298: Removed the unused param

tests/unit/storages/test_key_value_store.py
akshay11298: Sorry I mixed up the value static and path being static as in constant. I have updated the test case now

src/crawlee/storages/_key_value_store.py
akshay11298: @janbuchar Yes the /storage part was redundant. I have removed it. File extension was missing and that has been added now

src/crawlee/base_storage_client/_base_key_value_store_client.py
```diff
@@ -114,3 +114,14 @@ async def delete_record(self, key: str) -> None:
         Args:
             key: The key of the record which to delete
         """
+
+    @abstractmethod
+    def get_public_url(self, key: str) -> str:
```
janbuchar, line 119: Minor thing, but could you make this an `async` method? It's for forward compatibility in case some resource client needs to e.g., make a pre-signed S3 URL.

tests/unit/storages/test_key_value_store.py
```diff
@@ -100,3 +102,11 @@ async def test_static_get_set_value(key_value_store: KeyValueStore) -> None:
     await key_value_store.set_value('test-static', 'static')
     value = await key_value_store.get_value('test-static')
     assert value == 'static'
+
+
+async def test_static_get_public_url(key_value_store: KeyValueStore) -> None:
+    await key_value_store.set_value('test-static', 'static')
+    public_url = key_value_store.get_public_url('test-static')
+
+    path = unquote(urlparse(public_url).path)
```
janbuchar, line 111:
```suggestion
    path = unquote(urlparse(public_url).path)  # extract path from a 'file:///' URI
```

janbuchar:
@akshay11298 sorry for the broken CI. I confirm that code quality checks are passing on my machine.

src/crawlee/base_storage_client/_base_key_value_store_client.py
akshay11298: Updated the method to be async

src/crawlee/memory_storage_client/_key_value_store_client.py
```diff
@@ -287,6 +287,12 @@ async def delete_record(self, key: str) -> None:
                 if self._memory_storage_client.persist_storage:
                     await existing_store_by_id.delete_persisted_record(record)
 
+    @override
+    async def get_public_url(self, key: str) -> str:
+        store_name = self.name or self.id
+        storage_dir = self._memory_storage_client.storage_dir
+        return f'file://{storage_dir}/key_value_stores/{store_name}/{key}.txt'
```
janbuchar, line 294: This is... surprising. Since the tests appear to pass, I guess that's because of some weird implementation detail, but I don't think that `set_value` should automatically add the `.txt` suffix.

akshay11298: Ya not sure of the reason, but the file stored in my system also gets the .txt extension. I could not understand the reason of test failure when I wrote it without .txt, and then navigated to the directory to find it

image

janbuchar: @vdusek can you help out here? It looks like the key-value store adds a txt suffix to files with `content_type = None`

akshay11298: @janbuchar @vdusek  I took a quick look at it... Here the code determined the content type to be `text/plain`
https://github.com/apify/crawlee-python/blob/fb96524687c54f53f6f65b99f9574c1d777dff7a/src/crawlee/memory_storage_client/_key_value_store_client.py#L240-L246 
Afterwards at this point the extension for `text/plain` is determined to be `.txt` https://github.com/apify/crawlee-python/blob/fb96524687c54f53f6f65b99f9574c1d777dff7a/src/crawlee/_utils/file.py#L76
which is called from https://github.com/apify/crawlee-python/blob/fb96524687c54f53f6f65b99f9574c1d777dff7a/src/crawlee/memory_storage_client/_key_value_store_client.py#L260

And this is then eventually saved with `.txt` extension.

janbuchar: I don't see a good reason for keyvalue store to add any extension to the objects, but I'd like to hear your opinion @vdusek

vdusek: @akshay11298 Thanks for your investigations. I agree with @janbuchar on this - the KVS should not try to determine the extension and append it.

akshay11298: @vdusek Updated the code to not determine content_type if None is passed

janbuchar: I investigated the issue some more and decided that we probably shouldn't change the internal workings of the memory storage too much. Instead we can just do this:

```py
    @override
    async def get_public_url(self, key: str) -> str:
        store_name = self.name or self.id
        storage_dir = self._memory_storage_client.storage_dir
        record = await self._get_record_internal(key)

        return f'file://{storage_dir}/key_value_stores/{store_name}/{record.filename if record else key}'
```

It will use an internal methods that does a lookup based on the original key, and then you can return the filename used by the key-value store.

The extension handling code still seems unnecessary, but we'll leave it for now.

janbuchar: @akshay11298 would you mind reverting the latest commit and doing what I propose instead?

vdusek (approved):
LGTM

janbuchar (approved):

vdusek (changes requested):
The new test is not passing on Windows.

```
FAILED tests/unit/storages/test_key_value_store.py::test_static_get_public_url - FileNotFoundError: [Errno 2] No such file or directory: '\\key_value_stores\\default\\test-static.txt'
```

janbuchar:
@akshay11298 looks like there is an error on windows.
image

We probably need to make sure that the storage path is absolute in `get_public_url` and that it includes the drive letter and whatnot on windows.

akshay11298:
> @akshay11298 looks like there is an error on windows. image
> 
> We probably need to make sure that the storage path is absolute in `get_public_url` and that it includes the drive letter and whatnot on windows.

Wouldnt the self._memory_storage_client.storage_dir have drive letter as well? I am currently checking `existing_store_by_id.resource_directory` and `self._memory_storage_client.key_value_stores_directory` for drive letter

janbuchar:
> Edit: @janbuchar @vdusek After using `existing_store_by_id.resource_directory`, the error has changed to
> 
> ```
> FAILED tests/unit/storages/test_key_value_store.py::test_static_get_public_url - FileNotFoundError: [Errno 2] No such file or directory: '\\test-static.txt'
> ```
> 
> I am guessing the error is due to extension and not the drive letter. Can we somehow check what the file name is in case of windows?

The extension is correct - key value store still adds it, we just don't do it manually anymore. The problem is that it's missing the path to the storage dir. 

Do you have access to a windows machine?

akshay11298:
> > Edit: @janbuchar @vdusek After using `existing_store_by_id.resource_directory`, the error has changed to
> > ```
> > FAILED tests/unit/storages/test_key_value_store.py::test_static_get_public_url - FileNotFoundError: [Errno 2] No such file or directory: '\\test-static.txt'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I am guessing the error is due to extension and not the drive letter. Can we somehow check what the file name is in case of windows?
> 
> The extension is correct - key value store still adds it, we just don't do it manually anymore. The problem is that it's missing the path to the storage dir.
> 
> Do you have access to a windows machine?

Unfortunately I don't have access to a windows machine. I'll try to get access to one of it and let you know if it's possible

janbuchar:
> Unfortunately I don't have access to a windows machine. I'll try to get access to one of it and let you know if it's possible

If you don't have one, I believe we can troubleshoot this on Windows next week.

akshay11298:
@janbuchar @vdusek found the cause of the error. It was the slash used when creating the path manually. Used os.path.join to fix the error

akshay11298:
Now the test passes. urllib was returning netloc in windows and path in mac/linux

Windows - 
image

Mac -
image

tests/unit/storages/test_key_value_store.py
```diff
@@ -100,3 +102,14 @@ async def test_static_get_set_value(key_value_store: KeyValueStore) -> None:
...
+
+async def test_static_get_public_url(key_value_store: KeyValueStore) -> None:
+    await key_value_store.set_value('test-static', 'static')
+    public_url = await key_value_store.get_public_url('test-static')
+
+    url = urlparse(public_url)
+    path = url.netloc if url.netloc else url.path
+
+    f = await asyncio.to_thread(open, path, mode='r')
+    assert f.read() == 'static'
```
vdusek, line 115:
```suggestion
    with open(path, mode='r') as f:
        content = await asyncio.to_thread(f.read)
        assert content == 'static'
```

src/crawlee/base_storage_client/_base_key_value_store_client.py
```diff
@@ -114,3 +114,14 @@ async def delete_record(self, key: str) -> None:
         Args:
             key: The key of the record which to delete
         """
+
+    @abstractmethod
+    async def get_public_url(self, key: str) -> str:
+        """Get the public URL for the given key.
+
+        Args:
+            key: Key of the record for which URL is required
```
vdusek, line 123:
```suggestion
            key: Key of the record for which URL is required.
```

src/crawlee/storages/_key_value_store.py
```diff
@@ -157,3 +157,14 @@ async def set_value(
             return await self._resource_client.delete_record(key)
 
         return await self._resource_client.set_record(key, value, content_type)
+
+    async def get_public_url(self, key: str) -> str:
+        """Get the public URL for the given key.
+
+        Args:
+            key: Key of the record for which URL is required
```
vdusek, line 165:
```suggestion
            key: Key of the record for which URL is required.
```

tests/unit/storages/test_key_value_store.py
```diff
@@ -100,3 +102,15 @@ async def test_static_get_set_value(key_value_store: KeyValueStore) -> None:
...
+
+
+async def test_static_get_public_url(key_value_store: KeyValueStore) -> None:
+    await key_value_store.set_value('test-static', 'static')
+    public_url = await key_value_store.get_public_url('test-static')
+
+    url = urlparse(public_url)
+    path = url.netloc if url.netloc else url.path
+
+    with open(path, mode='r') as f:
```
vdusek, line 114:
```suggestion
    with open(path) as f:  # noqa: ASYNC230
```

vdusek (approved):
Thanks, @akshay11298, for solving it. I added one more test, `test_get_public_url_raises_for_non_existing_key,` and updated the method a bit so that it passes. LGTM.

src/crawlee/memory_storage_client/_key_value_store_client.py
```diff
@@ -287,6 +287,28 @@ async def delete_record(self, key: str) -> None:
...
+        if existing_store_by_id is None:
+            raise_on_non_existing_storage(StorageTypes.KEY_VALUE_STORE, self.id)
+
+        resource_dir = existing_store_by_id.resource_directory
+
+        record = await self._get_record_internal(key)
+        record_filename = record.filename if record else key
+
+        if record_filename is None:
+            raise ValueError('Record file name not found')
```
vdusek, line 307:
```suggestion
            raise ValueError(f'Record with key "{key}" was not found.')
```

src/crawlee/memory_storage_client/_key_value_store_client.py
```diff
@@ -287,6 +287,28 @@ async def delete_record(self, key: str) -> None:
...
+        resource_dir = existing_store_by_id.resource_directory
+
+        record = await self._get_record_internal(key)
+        record_filename = record.filename if record else key
+
+        if record_filename is None:
+            raise ValueError(f'Record with key "{key}" was not found.')
+
+        record_path = os.path.join(resource_dir, record_filename)
+        return f'file://{record_path}'
```
vdusek, line 310:
```suggestion
    @override
    async def get_public_url(self, key: str) -> str:
        existing_store_by_id = find_or_create_client_by_id_or_name_inner(
            resource_client_class=KeyValueStoreClient,
            memory_storage_client=self._memory_storage_client,
            id=self.id,
            name=self.name,
        )

        if existing_store_by_id is None:
            raise_on_non_existing_storage(StorageTypes.KEY_VALUE_STORE, self.id)

        record = await self._get_record_internal(key)

        if not record:
            raise ValueError(f'Record with key "{key}" was not found.')

        resource_dir = existing_store_by_id.resource_directory
        record_filename = self._filename_from_record(record)
        record_path = os.path.join(resource_dir, record_filename)
        return f'file://{record_path}'
```

src/crawlee/base_storage_client/_base_key_value_store_client.py
```diff
@@ -114,3 +114,16 @@ async def delete_record(self, key: str) -> None:
...
+    @abstractmethod
+    async def get_public_url(self, key: str) -> str:
+        """Get the public URL for the given key.
+
+        Args:
+            key: Key of the record for which URL is required.
+
+        Returns:
+            The public URL for the given key.
+        Raises:
```
vdusek, line 127:
```suggestion

        Raises:
```

janbuchar (approved):
LGTM

Issue #549, opened 2024-09-27
Add init script for Playwright browser context


vdusek:
- For now, let's use a data file containing fingerprints \(or at minimum user agents\) from the Apify fingerprint dataset.
- Use the init script from the [fingerprint suite](https://github.com/apify/fingerprint-suite/blob/master/packages/fingerprint-injector/src/utils.js).
- A fingerprint, selected based on parameters such as OS or browser type, should be passed to the init script.
- Load the JS init script from a file and convert it into a string for injection.
- Inject the script into the Playwright browser context using `BrowserContext.add_init_scripts()`.
- Possible inspiration:
    - [undetected-playwright-python](https://github.com/kaliiiiiiiiii/undetected-playwright-python),
    - \[playwright*stealth\]\(https://github.com/AtuboDad/playwright*stealth\).

Issue #284, opened 2024-07-10
Selenium crawler


codejunction:
The library is sick

It would be an beautiful addon if we can add an selenium crawler. 

Connecting to selenium webdriver or remote drivers

PR #576, opened 2024-10-06
docs: Add a new guide on how to avoid getting blocked

MostlyKIGuess:
### Description

-  Added 5 files, out of which 2 will aren't currently being used, when crawlee-python will complete puppeteer crawler, those can be used.
- Added additional information apart from the reference.

- Closes: #481 

### Testing

- Checked locally by running the website as given in the CONTRIBUTING.md guide.

image

image

image


MostlyKIGuess:
@janbuchar  @vdusek  Can you help me figure out add the operating system change, I am not sure if the current code works and there's no way to test it..

docs/guides/avoid_blocking_playwright.py
vdusek, line 1: Please move code files to `code/` dir with guide name - `avoid_getting_blocked` prefix. In the same way as we do it for other guides.

docs/guides/avoid_getting_blocked.mdx
```diff
@@ -0,0 +1,67 @@
+---
+id: avoid-getting-blocked
+title: Avoid getting blocked
+description: How to avoid getting blocked when scraping using crawlee-python.
+---
+
```
vdusek, line 6: Please add:

```mdx
import ApiLink from '@site/src/components/ApiLink';
```

and use it for all references of Crawlee components, like `PlaywrightCrawler`, `BrowserPool`, ...

docs/guides/avoid_getting_blocked.mdx
```diff
@@ -0,0 +1,67 @@
...
+On the contrary, sometimes we want to entirely disable the usage of browser fingerprints. This is easy to do with Crawlee too. All we have to do is set the `use_fingerprints` option to `False` when creating the `BrowserPool`:
+
+<Tabs groupId="avoid_blocking">
+    <TabItem value="playwright" label="PlaywrightCrawler" default>
+        <CodeBlock language="python">
+            {PlaywrightFingerprintsOffSource}
+        </CodeBlock>
+    </TabItem>
+   
+</Tabs>
```
vdusek, line 46: There is no need to use tabs for a single element. Please remove it. You can also mention that in Python we have only a single browser-based crawler - Playwright.

docs/guides/avoid_blocking_playwright_fingerprints_off.py
```diff
@@ -0,0 +1,16 @@
+from crawlee.browsers import BrowserPool
+from crawlee.playwright_crawler import PlaywrightCrawler
+
+# Create a browser pool with use_fingerprints set to False
+browser_pool = BrowserPool.with_default_plugin(
+    headless=True,
+    kwargs={
+        'use_fingerprints': False,
+    },
+)
```
vdusek, line 10: Additional kwargs can be provided directly.

```suggestion
browser_pool = BrowserPool.with_default_plugin(
    headless=True,
    use_fingerprints=False,
)
```

docs/guides/avoid_blocking_playwright.py
```diff
@@ -0,0 +1,28 @@
...
+                        'name': 'chromium',  # Or 'firefox', or 'webkit'
+                        'min_version': 96,
+                    },
+                ],
+                'devices': ['desktop'],  # Specify device types directly
+                'operating_systems': ['windows'],  # Specify OS types directly
+            },
+        },
+    },
+)
```
vdusek, line 23: As I wrote below, additional kwargs can be provided directly. But in this case, I'm not sure whether this is correct. Have you tried to execute it?

vdusek (changes requested):
Have you tried to execute the code samples?

docs/guides/avoid_getting_blocked.mdx
MostlyKIGuess: I kept tabs assuming we would later on add puppeteer.. but will do this!

docs/guides/avoid_getting_blocked.mdx
MostlyKIGuess: Thank you, will implement using this!

MostlyKIGuess:
@vdusek So if you suggest we can only keep those 3 options in the documentation, let me know what should be added, I have tweaked the source a little bit because it wasn't accepting plugin option in the above code

docs/guides/avoid_getting_blocked.mdx
vdusek: Puppeteer is not available for Python, unfortunately, see https://github.com/pyppeteer/pyppeteer:

> Attention: This repo is unmaintained and has been outside of minor changes for a long time. Please consider [playwright-python](https://github.com/microsoft/playwright-python) as an alternative.

janbuchar: But we may add Selenium if there's enough community traction (#284)

janbuchar: Not sure if that's enough of a reason to keep those tabs though :slightly_smiling_face:

MostlyKIGuess: Yeah I removed the tabs

docs/guides/avoid_blocking_playwright_fingerprints_off.py
MostlyKIGuess: for this it's not working directly so I added the extra code in the conversation as well and then there's a new issue , I have attached screenshots for it

MostlyKIGuess:
Hey @vdusek , can you please guide me on do we just keep the 3 options as I mentioned above or wait until features get implemented, I think keeping what we have along with additional tips would be a better option than nothing anyways

docs/guides/avoid_blocking_playwright_fingerprints_off.py
vdusek: Yes, this is not going to work, since `use_fingerprints` is the parameter of the `Plugin` and not the `BrowserPool`.

vdusek (changes requested):
The fingerprinting in Crawlee for Python is currently very limited. We have implemented only basics so far, see #401 and #402. The next step is #549. It means, you cannot just copy content from the JS guide.

Next steps...
- Write the guide only with current (limited) feature set regarding the blocking.
- Or wait for the fingerprinting to be completely implemented.

PR #582, opened 2024-10-08
ci: use python version instead of executable path

vdusek:
image


janbuchar:
Is this actually needed? The github actions runners are guaranteed to have at least some version of python installled, and that sounds good enough for this workflow.

janbuchar (approved):

Issue #249, opened 2024-06-28
Adaptive playwright crawler


janbuchar:
Add adaptive crawler similar to JS implementation.

Will be split in separate steps:

1. Adaptive crawler 
2. Rendering type predictor
3. Documentation, examples, templates

PR #584, opened 2024-10-10
feat: Key-value store context helpers

janbuchar:
- This adds a `get_key_value_store(id, name)` context helper to `BasicCrawlingContext`
- Also, push_data calls are held until the request handler terminates successfully (same as in JS version)
    - This is necessary for adaptive crawling (#249), among other things

### DIFF
--- a/src/crawlee/_types.py
+++ b/src/crawlee/_types.py
@@ -1,9 +1,20 @@
 from __future__ import annotations
 
 from collections.abc import Iterator, Mapping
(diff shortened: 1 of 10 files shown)


vdusek (commented):
TBH I don't see much value in the helper function on the context just for getting the instance of KVS, when I can use just `KeyValueStore.open(...)`, but I believe this is how things work in JS, am I right?

vdusek (approved):
LGTM then

Issue #469, opened 2024-08-27
Check CI status on `master` branch in release workflow


janbuchar:
More details in https://github.com/apify/crawlee-python/pull/466#issuecomment-2312331905

vdusek:
We should also publish a new version of documentation after release (and updated changelog).

janbuchar:
I managed to do this in https://github.com/apify/apify-client-js/pull/596 (https://github.com/apify/apify-client-js/actions/runs/11478924622), what do you think @B4nan @vdusek?

vdusek:
How does it work exactly? Does it verify if the "code checks" for the current commit are already passing, and if they are, it proceeds? If they are not, does it trigger them ?

janbuchar:
> How does it work exactly? Does it verify if the "code checks" for the current commit are already passing, and if they are, it proceeds? If they are not, does it trigger them ?

The checks are triggerred from a separate workflow that is started on push to master. The action periodically inspects them, fails if one of them fails, succeeds if they all succeed, and keeps waiting if they are not finished. Triggering the checks from the release workflow should not be necessary.

vdusek:
Thanks. It looks great then.

PR #597, opened 2024-10-16
ci: Use the release-metadata action

### LINKED ISSUE #598: Deprecated variables in git_cliff_core::changelog
```
 WARN  git_cliff_core::changelog > Variables ["commit.github", "commit.gitea", "commit.gitlab", "commit.bitbucket"] 
 are deprecated and will be removed in the future. Use `commit.remote` instead.
 ```
 This warning indicates that the variables currently used are deprecated. To avoid future issues, it is better to refactor the variables in config `cliff.toml`.

### PR DESCRIPTION
**Pull request by janbuchar**
- closes #598

### CODE REVIEW

File: .github/workflows/_publish_to_pypi.yaml [hunk 1]
```diff
@@ -19,7 +25,7 @@ jobs:
       id-token: write # Required for OIDC authentication.
     environment:
       name: pypi
-      url: https://pypi.org/project/crawlee
+      url: https://pypi.org/project/${{ inputs.package_name }}
 
     steps:
       - name: Checkout repository
```
janbuchar, line 31: It should be possible to configure the ref for this one, so that we have an up-to-date changelog

.github/workflows/_publish_to_pypi.yaml
janbuchar, line 1: This one should be movable to apify/workflows without further modifications... right?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,121 @@
...
+    name: Lint check
+    uses: apify/workflows/.github/workflows/python_lint_check.yaml@main
+
+  type_check:
+    name: Type check
+    uses: apify/workflows/.github/workflows/python_type_check.yaml@main
+
+  unit_tests:
+    name: Unit tests
+    uses: apify/workflows/.github/workflows/python_unit_tests.yaml@main
```
janbuchar, line 52: On a manual release, this really seems redundant. Also, see #469

.github/workflows/pre_release.yaml
janbuchar, line 1: I split pre-release into a separate workflow file because there isn't that much duplication (IMO), and it's better to have two workflows without too many conditions.

janbuchar: Also, I believe that this one could become a reusable workflow as well - the only crawlee-specific thing there is the pypi package name, right?

.github/workflows/_publish_to_pypi.yaml
vdusek: I would say so.

.github/workflows/release.yaml
vdusek: Currently, it is not redundant, since we don't have any additional trigger of code checks for the master (only for PRs).

.github/workflows/pre_release.yaml
vdusek: Agreed

.github/workflows/_publish_to_pypi.yaml
vdusek: Yeah, it looks like that

vdusek:
I'm thinking if `PYTHON_VERSION` couldn't be provided as an input parameter as well?

.github/workflows/pre_release.yaml
```diff
@@ -0,0 +1,90 @@
+name: Create a pre-release
+
+on:
+  # Trigger a beta version release (pre-release) on push to the master branch.
+  push:
+    branches:
+      - use-release-metadata-action # TODO
```
B4nan, line 7: how is this supposed to be working? canary builds should be done only on the default branch

.github/workflows/pre_release.yaml
B4nan: so the plan is to have JS and python flavors of those actions? because the contents look quite python specific now

.github/workflows/pre_release.yaml
```diff
@@ -0,0 +1,90 @@
...
+      - name: Update CHANGELOG.md
+        uses: DamianReeves/write-file-action@master
+        with:
+          path: CHANGELOG.md
+          write-mode: overwrite
+          contents: ${{ needs.release_metadata.outputs.changelog }}
+
+      - name: Commit changes
+        id: commit
+        uses: EndBug/add-and-commit@v9
```
B4nan, line 74: thinking of JS projects, we will need more control most likely, since this will also require running `npm install` or `yarn install` and committing the updated lockfile (which holds the current project version too, at least in NPM).

B4nan (commented):
few questions from my end.

also the changelog changes look wrong, there shouldn't be any, right?

CHANGELOG.md
```diff
@@ -6,227 +6,227 @@ All notable changes to this project will be documented in this file.
 
 ### 🚀 Features
 
-- Key-value store context helpers ([#584](https://github.com/apify/crawlee-python/pull/584)) ([fc15622](https://github.com/apify/crawlee-python/commit/fc156222c3747fc4cc7bd7666a21769845c7d0d5)) by [@janbuchar](https://github.com/janbuchar)
+- Key-value store context helpers ([#584](apify/crawlee-python/pull/584)) ([fc15622](<REPO>/commit/fc156222c3747fc4cc7bd7666a21769845c7d0d5)) by [@janbuchar](https://github.com/janbuchar)
```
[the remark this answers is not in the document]
janbuchar: yes, the postprocessor seems to not work - trying to fix this right now

.github/workflows/pre_release.yaml
janbuchar: that's so that I can test the workflow against this branch, I intend to change it to `master` before merging

vdusek (approved):
LGTM

PR #592, opened 2024-10-16
Render contributors in CHANGELOG.md

Telomelonia:
### Description
The current changelog cannot render the contributors as a list.
Currently, it stores the link to contributor's GH page. 

### Issues
- Getting error while running `git-cliff --output CHANGELOG.md`
Will attach error logs in the comments
(I have installed git-cliff)

- Closes: #468 

### Testing
`git-cliff --output CHANGELOG.md`

- TODO

### Checklist
- [ ] Test `git-cliff --output CHANGELOG.md`
- [ ] CI passed

PR #599, opened 2024-10-17
chore: don't add contributors as link to the changelog

orhun:
### Description

Hello! I'm the creator of `git-cliff` :wave:

GitHub does not add profile links to the releases when the usernames are being used as links. So they should be in the `@username` format instead of `[@username](...)`.
This PR updates the `git-cliff` config to address that.

This issue was brought to my attention in <https://github.com/orhun/git-cliff/discussions/917>

### Issues

closes https://github.com/orhun/git-cliff/discussions/917

### Testing

Not tested, but it will work :)

### Checklist

- [ ] CI passed


janbuchar:
Hi @orhun, nice to hear from you! Thanks for the contribution! A similar thing actually has been attempted in #592. Perhaps you could provide some feedback on that? The problem is that we want the release notes to contain the plain user handles, but changelog.md should have links to user profiles (current state).

orhun:
Ah, nice. I will follow-up on that PR then 👍🏼

Issue #427, opened 2024-08-13
Add support for `preNavigationHooks` in Playwright


viewv:
As the title, I want to know does the crawlee-python support the `preNavigationHooks` function?

janbuchar:
Hello and thanks for your interest in Crawlee for Python! Pre-navigation hooks are not currently supported. Before we implement this, would putting the hook code into the request handler help? Or do you need to do something before the page is loaded?

viewv:
> Hello and thanks for your interest in Crawlee for Python! Pre-navigation hooks are not currently supported. Before we implement this, would putting the hook code into the request handler help? Or do you need to do something before the page is loaded?

OK, thank you, I want to change some browser properties before the page loaded, so I think it will be really helpful to support  the `preNavigationHooks`.

Tallone:
I want to do something before the page loads, such as aborting when loading some images or CSS, Is there any trick available to achieve this as of now?

vdusek:
Implementation from Crawlee JS:
- [preNavigationHooks](https://crawlee.dev/api/playwright-crawler/interface/PlaywrightCrawlerOptions#preNavigationHooks)
- [postNavigationHooks](https://crawlee.dev/api/playwright-crawler/interface/PlaywrightCrawlerOptions#postNavigationHooks)

Yashwanth-Chandrakumar:
@vdusek /assign

janbuchar:
> @vdusek /assign

We don't assign issues for hacktoberfest. If you want to contribute, open a PR whenever you're ready. First mergeable one gets merged.

Yashwanth-Chandrakumar:
@janbuchar Thank Mate

Yashwanth-Chandrakumar:
@janbuchar i have created a PR to solve the issue please view it #602

PR #602, opened 2024-10-18
feat: Add support for `pre_navigation_hooks` in PlaywrightCrawler

Yashwanth-Chandrakumar:
### Description

This implementation adds support for preNavigationHooks in the PlaywrightCrawler

Use the new pre_navigation_hooks:

`async def example_hook(context: PlaywrightCrawlingContext, goto_options: dict) -> None:
    await context.page.evaluate("window.foo = 'bar';")
    goto_options['timeout'] = 60000  # Set a custom timeout for navigation

crawler = PlaywrightCrawler(
    # other options...
    pre_navigation_hooks=[example_hook]
)`

### Issues

[#427](https://github.com/apify/crawlee-python/issues/427)

- [ ] CI passed

### DIFF
--- a/src/crawlee/playwright_crawler/_playwright_crawler.py
+++ b/src/crawlee/playwright_crawler/_playwright_crawler.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, List, Callable, Awaitable
 
 from pydantic import ValidationError
 from typing_extensions import Unpack
@@ -18,10 +18,11 @@ from crawlee.playwright_crawler._utils import infinite_scroll
 
 if TYPE_CHECKING:
     from collections.abc import AsyncGenerator
-
     from crawlee._types import AddRequestsKwargs, BasicCrawlingContext
     from crawlee.browsers._types import BrowserType
+    from playwright.async_api import Page
 
+PlaywrightHook = Callable[[PlaywrightCrawlingContext, dict], Awaitable[None]]
 
 class PlaywrightCrawler(BasicCrawler[PlaywrightCrawlingContext]):
     """A crawler that leverages the [Playwright](https://playwright.dev/python/) browser automation library.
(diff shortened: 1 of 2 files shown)


Yashwanth-Chandrakumar:
@janbuchar Why is this not merged

janbuchar:
> @janbuchar Why is this not merged

1. because it was submitted just before the weekend
2. because we want to review the code first

We value your contribution, but please, keep your tone civil.

B4nan:
And let me add a third reason, there are no tests nor documentation for this new feature.

Also, this option should be available for all crawler types, not just playwright.

src/crawlee/playwright_crawler/_playwright_crawler.py
```diff
@@ -48,6 +49,7 @@ def __init__(
         browser_pool: BrowserPool | None = None,
         browser_type: BrowserType | None = None,
         headless: bool | None = None,
+        pre_navigation_hooks: List[PlaywrightHook] | None = None,
```
janbuchar, line 52: This is a perfectly valid way of passing the hooks, but have you considered adding a decorator instead/as well? We did a similar change for request handlers because Python does not have anonymous functions (that can contain more than one statement).

src/crawlee/playwright_crawler/_playwright_crawler.py
```diff
@@ -101,17 +101,37 @@ async def _make_http_request(
...
+                add_requests=context.add_requests,
+                send_request=context.send_request,
+                push_data=context.push_data,
+                proxy_info=context.proxy_info,
+                get_key_value_store=context.get_key_value_store,
+                log=context.log,
+                page=crawlee_page.page,
+                infinite_scroll=lambda: infinite_scroll(crawlee_page.page),
+                response=None,
+                enqueue_links=None,  # We'll set this later
```
janbuchar, line 121: This is a no-go. First, it won't pass the type checker. Second, nullable attributes lead to errors, mental overhead and hard to read code.

Instead, we should add one more parent class to `PlaywrightCrawlingContext` that only adds the `page` attribute. Then, we should split the `_make_http_request` into two stages - one that opens the page and executes prenavigation hooks, and second that does the navigation. The context pipeline (see line 78) should be responsible for chaining these.

Does that make sense?

janbuchar (changes requested):

Yashwanth-Chandrakumar:
> > @janbuchar Why is this not merged
> 
> 1. because it was submitted just before the weekend
> 2. because we want to review the code first
> 
> We value your contribution, but please, keep your tone civil.

Hi Mate I am really sorry for what is said i really did not mean it that way.

janbuchar:
Closing in favor of #631

PR #605, opened 2024-10-20
refactor: Refactor External Link Processor for Improved Readability and Modern Syntax

HarshMN2345:
Key Changes:
Replaced parse with new URL(): This modern approach is more reliable for parsing URLs, especially relative URLs. Removed import('unist-util-visit').then(...): Switched to a standard import, as there's no need for dynamic importing here. Error Handling: Added a try-catch block around the URL parsing to avoid runtime errors in case of invalid URLs. Cleaner Code: Simplified logic while maintaining the core behavior of processing internal vs. external links.

### DIFF
--- a/website/tools/utils/externalLink.js
+++ b/website/tools/utils/externalLink.js
@@ -1,38 +1,47 @@
-const { parse } = require('url');
-
-const visit = import('unist-util-visit').then((m) => m.visit);
+import { visit } from 'unist-util-visit';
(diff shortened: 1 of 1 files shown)


janbuchar:
Hello @HarshMN2345 and thanks for your contribution! Sadly, this does not resolve any of our issues and the value of the changes is debatable. It is usually a better idea to consult the issue tracker before making changes.

PR #621, opened 2024-10-25
feat: Add `always_enqueue` option to bypass URL deduplication

### LINKED ISSUE #547: Add `always_enqueue` option to `Request` for bypassing deduplication
- Add an `always_enqueue` option (or use a better name for it, but avoid negative terms) as an input parameter to the `Request.from_url` constructor. 
  - This will allow users to easily opt out of the request deduplication process.
- Implement the option as a convenient wrapper that generates a random `unique_key`, ensuring that each request is always enqueued and processed.
- Address edge cases where both `unique_key` and `always_enqueue=True` are provided.
- It should work in the same way as the `dont_filter` option in Scrapy ([docs](https://docs.scrapy.org/en/latest/topics/request-response.html)).

Rutam21:
### Description

This PR implements a new input parameter, always_enqueue, in the Request.from_url constructor. This feature allows users to bypass the request deduplication process, ensuring that each request is always enqueued and processed.

Key Changes:

- Added the always_enqueue parameter to the Request.from_url method.
- When always_enqueue is set to true, a random unique key is generated for each request, preventing it from being considered a duplicate.
- This enhancement provides users with a convenient way to enqueue requests without worrying about deduplication.

With this addition, users can seamlessly manage their requests while maintaining flexibility in their queuing strategy.

### Issues

Fixes #547 

### Testing

The modifications can be verified by making requests to the same URL. The operations should execute each time this is done without treating it as a duplicate request when `always_enqueue` is set to true.

### Checklist

- [x] CI passed

### DIFF
--- a/src/crawlee/_request.py
+++ b/src/crawlee/_request.py
@@ -306,6 +306,7 @@ class Request(BaseRequestData):
             use_extended_unique_key: Determines whether to include the HTTP method and payload in the `unique_key`
                 computation. This is only relevant when `unique_key` is not provided.
             always_enqueue: If set to `True`, the request will be enqueued even if it is already present in the queue.
+                Using this is not allowed when a custom `unique_key` is also provided and will result in a `ValueError`.
             **kwargs: Additional request properties.
         """
         if unique_key is not None and always_enqueue:
--- a/src/crawlee/_request.py
+++ b/src/crawlee/_request.py
@@ -324,7 +324,7 @@ class Request(BaseRequestData):
         )
         
         if always_enqueue:
-            unique_key += f'_{crypto_random_object_id()}'
+            unique_key = f'{unique_key}_{crypto_random_object_id()}'
 
         id = id or unique_key_to_request_id(unique_key)
 
--- a/src/crawlee/_request.py
+++ b/src/crawlee/_request.py
@@ -22,6 +22,7 @@ from typing_extensions import Self
 from crawlee._types import EnqueueStrategy, HttpHeaders, HttpMethod, HttpPayload, HttpQueryParams, JsonSerializable
 from crawlee._utils.requests import compute_unique_key, unique_key_to_request_id
 from crawlee._utils.urls import extract_query_params, validate_http_url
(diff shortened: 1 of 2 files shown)


Rutam21:
@vdusek I have reviewed this issue and your suggestions and implemented them in this PR. Please review and suggest changes, if any. Thanks.

Rutam21:
@vdusek I have reviewed this issue and your suggestions and implemented them in this PR. Please review and suggest changes, if any. Thanks.

cc; @janbuchar

src/crawlee/_request.py
```diff
@@ -316,6 +322,9 @@ def from_url(
             keep_url_fragment=keep_url_fragment,
             use_extended_unique_key=use_extended_unique_key,
         )
+        
+        if always_enqueue:
+            unique_key += f'_{crypto_random_object_id()}'
```
vdusek, line 327:
```suggestion
            unique_key = f'{unique_key}_{crypto_random_object_id()}'
```

vdusek (changes requested):
This looks quite good, but make sure the CI passes (lint check, type check, unit tests... all are failing),

Rutam21:
@vdusek Please run the CI checks again. I have pushed a commit to fix them all. Thanks.


cc: @souravjain540

src/crawlee/_request.py
```diff
@@ -303,8 +305,12 @@ def from_url(
                 the `unique_key` computation. This is only relevant when `unique_key` is not provided.
             use_extended_unique_key: Determines whether to include the HTTP method and payload in the `unique_key`
                 computation. This is only relevant when `unique_key` is not provided.
+            always_enqueue: If set to `True`, the request will be enqueued even if it is already present in the queue.
```
janbuchar, line 308:
```suggestion
            always_enqueue: If set to `True`, the request will be enqueued even if it is already present in the queue.
                Using this is not allowed when a custom `unique_key` is also provided and will result in a `ValueError`.
```

janbuchar (approved):
LGTM, thank you for your contribution!

PR #623, opened 2024-10-25
chore: change ValueError to warning in RQClient for ID mismatch

vdusek:


janbuchar (approved):
I guess this is fine for now, but we should definitely revisit this after resolving #94

PR #631, opened 2024-10-29
feat: Add `pre_navigation_hooks` to `PlaywrightCrawler`

Prathamesh010:
### Description

Add a new decorator for processing pre navigation hooks

Example Use:
```python
from crawlee.playwright_crawler import PlaywrightCrawler
from .routes import router

async def main() -> None:
    """The crawler entry point."""
    crawler = PlaywrightCrawler(
        request_handler=router,
        max_requests_per_crawl=50,
    )

    @crawler.pre_navigation_hook
    async def hooky(context) -> None:
        print(f'Hook1')

    @crawler.pre_navigation_hook
    async def hooky2(context) -> None:
        print(f'Hook2')

    await crawler.run(
        [
            'https://crawlee.dev',
        ]
    )
```

### Issues

- Closes: #427

### Checklist

- [x] CI passed

### DIFF
--- a/docs/examples/code/playwright_crawler.py
+++ b/docs/examples/code/playwright_crawler.py
@@ -1,6 +1,6 @@
 import asyncio
 
-from crawlee.playwright_crawler import PlaywrightCrawler, PlaywrightCrawlingContext
+from crawlee.playwright_crawler import PlaywrightCrawler, PlaywrightCrawlingContext, PlaywrightPreNavigationContext
 
 
 async def main() -> None:
@@ -46,6 +46,14 @@ async def main() -> None:
 
         # Find a link to the next page and enqueue it if it exists.
         await context.enqueue_links(selector='.morelink')
+
+    # Define a hook that will be called each time before navigating to a new URL.
+    # The hook receives a context parameter, providing access to the request and
+    # browser page among other things. In this example, we log the URL being
+    # navigated to.
+    @crawler.pre_navigation_hook
+    async def log_navigation_url(context: PlaywrightPreNavigationContext) -> None:
+        context.log.info(f'Navigating to {context.request.url} ...')
 
     # Run the crawler with the initial list of URLs.
     await crawler.run(['https://news.ycombinator.com/'])
--- a/src/crawlee/playwright_crawler/_playwright_crawler.py
+++ b/src/crawlee/playwright_crawler/_playwright_crawler.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Awaitable, Callable
 
 from pydantic import ValidationError
 from typing_extensions import Unpack
@@ -14,6 +14,7 @@ from crawlee.basic_crawler import BasicCrawler, BasicCrawlerOptions, ContextPipe
 from crawlee.browsers import BrowserPool
 from crawlee.errors import SessionError
 from crawlee.playwright_crawler._playwright_crawling_context import PlaywrightCrawlingContext
+from crawlee.playwright_crawler._playwright_pre_navigation_context import PlaywrightPreNavigationContext
 from crawlee.playwright_crawler._utils import infinite_scroll
 
 if TYPE_CHECKING:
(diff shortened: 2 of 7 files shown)


src/crawlee/playwright_crawler/_playwright_crawler.py
```diff
@@ -95,16 +96,42 @@ def __init__(
...
+            add_requests=context.add_requests,
+            send_request=context.send_request,
+            push_data=context.push_data,
+            proxy_info=context.proxy_info,
+            get_key_value_store=context.get_key_value_store,
+            log=context.log,
+            page=crawlee_page.page,
+        )
+
+    async def _navigate_and_make_http_request(
```
janbuchar, line 132: I'd just rename this one to `_navigate`.

janbuchar (changes requested):
Very cool! Could you please add a test of the functionality as well? Also, some documentation would be great, if you have the time.

janbuchar:
Just extending the [PlaywrightCrawler example](https://github.com/apify/crawlee-python/blob/master/docs/examples/playwright_crawler.mdx) should be enough for now.

docs/examples/code/playwright_crawler.py
```diff
@@ -47,6 +48,13 @@ async def request_handler(context: PlaywrightCrawlingContext) -> None:
         # Find a link to the next page and enqueue it if it exists.
         await context.enqueue_links(selector='.morelink')
 
+    # Define a hook that will be called each time before navigating to a new URL.
+    # The hook receives a context parameter, providing access to the request among
+    # other things. In this example, we log the URL being navigated to.
+    @crawler.pre_navigation_hook
+    async def log_navigation_url(context: BasicCrawlingContext) -> None:
```
janbuchar, line 55: While valid, `BasicCrawlingContext` does not provide `page` - let's update this to `BrowserPageContext`.

src/crawlee/playwright_crawler/_browser_page_context.py
```diff
@@ -0,0 +1,20 @@
...
+from typing import TYPE_CHECKING
+
+from crawlee._types import BasicCrawlingContext
+
+if TYPE_CHECKING:
+    from playwright.async_api import Page
+
+
+@dataclass(frozen=True)
+class BrowserPageContext(BasicCrawlingContext):
```
janbuchar, line 13: This should be exported from `crawlee.playwright_crawler` so that you can easily import it to type your hook parameters.

docs/examples/code/playwright_crawler.py
```diff
@@ -1,5 +1,6 @@
 import asyncio
 
+from crawlee._types import BasicCrawlingContext
```
janbuchar, line 3: We should not import from private subpackages in examples.

src/crawlee/playwright_crawler/_playwright_crawler.py
```diff
@@ -95,16 +96,39 @@ def __init__(
 
         # Compose the context pipeline with the Playwright-specific context enhancer.
         kwargs['_context_pipeline'] = (
-            ContextPipeline().compose(self._make_http_request).compose(self._handle_blocked_request)
+            ContextPipeline().compose(self._open_page).compose(self._navigate).compose(self._handle_blocked_request)
         )
         kwargs['_additional_context_managers'] = [self._browser_pool]
         kwargs.setdefault('_logger', logging.getLogger(__name__))
+        self.pre_navigation_hooks: list[Callable[[BasicCrawlingContext], Awaitable[None]]] = []
```
janbuchar, line 103: This should be private (prefixed with an underscore).

janbuchar (changes requested):
Almost there!

docs/examples/code/playwright_crawler.py
B4nan: you mean `BrowserCrawlingContext` i guess? that's how its called in the JS version

janbuchar: We don't have a `BrowserCrawlingContext` here. And `PlaywrightCrawlingContext` contains stuff that doesn't make sense in a pre-navigation hook, like `infinite_scroll`.

B4nan: well, it will contain things that are valid once we implement them, like `blockRequests` helper

you get pw context there, so the type should reflect that

janbuchar: I don't think we are in disagreement here. There is an inheritance chain like this: `BasicCrawlingContext` <- `BrowserPageContext` <- `PlaywrightCrawlingContext`. `BrowserPageContext` gets passed to pre-navigation hooks, `PlaywrightCrawlingContext` goes into the request handler.

If you're suggesting that `BrowserPageContext` could have a better name, that's a good point.

janbuchar: @Prathamesh010 better, but still not optimal... I'd either name it after the intended usage (in prenavigation hooks) or just try to convey the fact that there's less stuff than in regular `PlaywrightCrawlingContext`.

B4nan: afaik in JS we have this typed to `PlaywrightCrawlingContext`, do we really pass something else in the hooks? `PlaywrightCrawlingContext` is where we add the context helpers, so it should be that one, its not just about having the `page` property, its about the context helpers. 

i understand that some helpers might not make sense in here (maybe even all of these we currently have), and i am fine with dealing with this later (once we have some helpers that should be here, like the `blockRequests`)

the JS version has the browser crawler abstraction, not sure if its really needed here, given we will have just one browser crawler type (no puppeteer). i can imagine we will add the CDP crawler here too at some point, but that is a rather distant future

janbuchar: In JS, we mutate the context as it proceeds through the request handling process, so we promise the complete `PlaywrightCrawlingContext` in the `PlaywrightCrawler` pre-navigation hooks. In the Python port, we are trying to only pass whatever makes sense - hence the intermediate type.

B4nan: got it (and I surely like it too), so let's just try to find a better-suited name

B4nan: out of those, I'd say `PreNavigationContext` is the best one

janbuchar (approved):
LGTM, thank you!

PR #632, opened 2024-10-29
ci: Do not run pre-release in forks

janbuchar:
- see discussion in https://github.com/apify/apify-client-js/pull/596#discussion_r1814782521

### DIFF
--- a/.github/workflows/pre_release.yaml
+++ b/.github/workflows/pre_release.yaml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   release_metadata:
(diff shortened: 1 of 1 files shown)


B4nan (approved):
how come the whole condition is a string btw?

janbuchar:
> how come the whole condition is a string btw?

YAML programming! I guess the quotes are not really necessary, but better safe than sorry.

vdusek (approved):
LGTM

B4nan:
I guess they are not, we have quite complex conditions elsewhere without them, e.g. https://github.com/apify/crawlee/blob/master/.github/workflows/test-ci.yml#L17

PR #633, opened 2024-10-29
docs: Update input arg docstrings of `BasicCrawler` and `EventManager`

vdusek:
cc @barjin

### DIFF
--- a/src/crawlee/basic_crawler/_basic_crawler.py
+++ b/src/crawlee/basic_crawler/_basic_crawler.py
@@ -180,7 +180,31 @@ class BasicCrawler(Generic[TCrawlingContext]):
     ) -> None:
         """A default constructor.
 
-        See the `BasicCrawlerOptions` for a detailed description of the arguments.
+        Args:
+            request_provider: Provider for requests to be processed by the crawler.
+            request_handler: A callable responsible for handling requests.
+            http_client: HTTP client used by `BasicCrawlingContext.send_request` and the HTTP-based crawling.
+            concurrency_settings: Settings to fine-tune concurrency levels.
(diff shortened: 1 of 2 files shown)


src/crawlee/basic_crawler/_basic_crawler.py
janbuchar, line 1: I guess that by moving the attribute docs to `BasicCrawlerOptions`, we will lose argument docs in IDEs, right? Or does this still work normally with PyLance? I mean, if I do `ParselCrawler(http_client=|)`, will it show me "HTTP client used by `BasicCrawlingContext.send_request` and the HTTP-based crawling"? Did it do that before? :smile: 

Also, I'm not sure how to put together documentation (defined in `BasicCrawlerOptions`) and default values (defined in `BasicCrawler.__init__`) - is there something that could help @barjin with? Some decorator to link these or I don't know.

janbuchar (approved):

src/crawlee/basic_crawler/_basic_crawler.py
vdusek: > I guess that by moving the attribute docs to BasicCrawlerOptions, we will lose argument docs in IDEs, right? Or does this still work normally with PyLance? I mean, if I do ParselCrawler(http_client=|), will it show me "HTTP client used by BasicCrawlingContext.send_request and the HTTP-based crawling"? Did it do that before? 😄

Yeah... that's not good... So I don't know - shall we keep it duplicate (for now)?

janbuchar: I guess the duplication won't harm us. If it does actually improve DX in IDEs, that is.

Issue #460, opened 2024-08-26
Implement max crawl depth


vdusek:
- Implement "max crawl depth" / "crawling depth limit"
- See https://github.com/apify/crawlee-python/discussions/441
- The depth information should be stored in the `Request` (`user_data` -> `crawlee_data`)

janbuchar:
We should also make an issue in JS crawlee as well. Or am I missing a way to limit crawl depth in JS?

vdusek:
> We should also make an issue in JS crawlee as well. Or am I missing a way to limit crawl depth in JS?

AFAIK there is no such option.

29deepanshutyagi:
i want to work on this issue ,kindly assign me @B4nan ,

29deepanshutyagi:
.take

PR #637, opened 2024-10-31
feat: Add `max_crawl_depth` property

Prathamesh010:
### Description

- Implements "max crawl depth"

### Issues

- Closes: #460 

### Testing

- Added tests

### Checklist

- [ ] CI passed

### DIFF
--- a/src/crawlee/basic_crawler/_basic_crawler.py
+++ b/src/crawlee/basic_crawler/_basic_crawler.py
@@ -120,6 +120,10 @@ class BasicCrawlerOptions(TypedDict, Generic[TCrawlingContext]):
     configure_logging: NotRequired[bool]
     """If True, the crawler will set up logging infrastructure automatically."""
 
+    max_crawl_depth: NotRequired[int | None]
+    """Limits crawl depth from 0 (initial requests) up to the specified `max_crawl_depth`.
+    Requests at the maximum depth are processed, but no further links are enqueued."""
+
     _context_pipeline: NotRequired[ContextPipeline[TCrawlingContext]]
     """Enables extending the request lifecycle and modifying the crawling context. Intended for use by
     subclasses rather than direct instantiation of `BasicCrawler`."""
@@ -174,6 +178,7 @@ class BasicCrawler(Generic[TCrawlingContext]):
         statistics: Statistics | None = None,
         event_manager: EventManager | None = None,
         configure_logging: bool = True,
+        max_crawl_depth: int | None = None,
         _context_pipeline: ContextPipeline[TCrawlingContext] | None = None,
         _additional_context_managers: Sequence[AsyncContextManager] | None = None,
         _logger: logging.Logger | None = None,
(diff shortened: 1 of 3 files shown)


src/crawlee/basic_crawler/_basic_crawler.py
```diff
@@ -120,6 +120,9 @@ class BasicCrawlerOptions(TypedDict, Generic[TCrawlingContext]):
     configure_logging: NotRequired[bool]
     """If True, the crawler will set up logging infrastructure automatically."""
 
+    max_crawl_depth: NotRequired[int | None]
+    """Maximum crawl depth. If set, the crawler will stop crawling after reaching this depth."""
```
janbuchar, line 124: We should probably elaborate on the edge cases - for example, with `max_crawl_depth = 3`, do we process three or four "levels" of links? I'd assume that the start requests have `crawl_depth = 0`, and then we go all the way up to 3 and don't enqueue any further links, but it would be much better to have that stated explicitly in the docs.

src/crawlee/beautifulsoup_crawler/_beautifulsoup_crawler.py
```diff
@@ -181,6 +181,12 @@ async def enqueue_links(
         ) -> None:
             kwargs.setdefault('strategy', EnqueueStrategy.SAME_HOSTNAME)
 
+            if self._max_crawl_depth is not None and context.request.crawl_depth + 1 > self._max_crawl_depth:
+                context.log.info(
+                    f'Skipping enqueue_links for URL "{context.request.url}" due to the maximum crawl depth limit.'
+                )
+                return
+
```
janbuchar, line 189: 1. the repetition in all crawlers is not great
    - it is very easy to overlook this when implementing a new crawler
3. it is also possible to add requests to the queue via `context.add_requests`
    - ideally we should fill in default depth to the requests - it doesn't make much sense to exempt some requests from `max_crawl_depth`
    - if there is no ergonomic way to support max depth in `add_requests`, we should make it clear in the docs that the user has to handle crawl depth manually

src/crawlee/beautifulsoup_crawler/_beautifulsoup_crawler.py
```diff
@@ -191,6 +197,9 @@ async def enqueue_links(
                 if label is not None:
                     link_user_data.setdefault('label', label)
 
+                data = {'crawlDepth': context.request.crawl_depth + 1}
+                link_user_data.setdefault('__crawlee', CrawleeRequestData(**data))
```
janbuchar, line 201:
```suggestion
                link_user_data.crawlee = link_user_data.crawlee or CrawleeRequestData()
                link_user_data.crawlee.crawl_depth = context.request.crawl_depth + 1
```

I'd prefer not fiddling with field aliases if possible.

janbuchar (changes requested):
This is pretty great! I found just some minor issues.

src/crawlee/basic_crawler/_basic_crawler.py
janbuchar: Yes, what you propose is perfect.

src/crawlee/beautifulsoup_crawler/_beautifulsoup_crawler.py
janbuchar: Yes, I think that the check could go directly to `add_requests` in `BasicCrawler`. Is there any potential issue?

janbuchar (approved):

PR #638, opened 2024-10-31
ci: Improve release workflows

### LINKED ISSUE #469: Check CI status on `master` branch in release workflow

### LINKED ISSUE #706: Trigger docs build after updating the changelog
Currently the changelog is updated with a `[ci skip]` commits so it only gets incorporated in the docs after a delay.

janbuchar:
- closes #469 
- closes #706

### DIFF
--- a/.github/workflows/run_code_checks.yaml
+++ b/.github/workflows/run_code_checks.yaml
@@ -29,7 +29,7 @@ jobs:
     name: Unit tests
     uses: apify/workflows/.github/workflows/python_unit_tests.yaml@main
     secrets:
-      httpbin_url: https://janbuchar--httpbin.apify.actor?token=${{ secrets.APIFY_HTTPBIN_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/pre_release.yaml
```diff
@@ -24,21 +28,20 @@ jobs:
...
-    uses: apify/workflows/.github/workflows/python_type_check.yaml@main
-
-  unit_tests:
-    name: Unit tests
-    uses: apify/workflows/.github/workflows/python_unit_tests.yaml@main
+  wait_for_checks:
+    name: Wait for code checks to pass
+    runs-on: ubuntu-latest
+    steps:
+      - uses: lewagon/wait-on-check-action@v1.3.4
```
janbuchar, line 35: Note to self: This one works well in the `pre_release` case, but for `release`, we need a more robust way of checking the status of the target branch - if there is a [skip ci] commit at the top, it will be automatically considered green, which is not optimal.

janbuchar: Do we really? If the [skip ci] commit is a changelog update (it should be in 99% of cases), that only happens after successful code checks, so it should be safe to assume that is green as well.

vdusek (approved):
LGTM, thanks for including the explanatory comments

vdusek (approved):

PR #642, opened 2024-11-01
docs: add upgrading to v0.4 section

vdusek:


B4nan:
maybe we could have just one guide, if we don't plan to do bigger major releases just yet (so basically for all the 0.x releases)?

it feels a bit weird to have such a short page 🙃 i can imagine once we reach v1 we will do it much less often and the number of BCs will be bigger, so then it will make sense to have dedicated pages, but for the 0.x releases we will likely have just a few notes in there

just an idea, up to you guys, both is fine

vdusek:
image

docs/upgrading/upgrading_to_v0x.md
```diff
@@ -1,11 +1,28 @@
 ---
-id: upgrading-to-v03
-title: Upgrading to v0.3
+id: upgrading-to-v0x
+title: Upgrading to v0.x
```
B4nan, line 3: Maybe better this way?

```suggestion
id: upgrading
title: Upgrading guide
```

also we should probably add a redirect, the nginx config now lives in the apify-docs repo

https://github.com/apify/apify-docs/blob/master/nginx.conf#L312

janbuchar (approved):

PR #653, opened 2024-11-05
refactor: Simplify request queue implementation

janbuchar:
Follow up on the points raised in https://github.com/apify/crawlee/pull/2700

### DIFF
--- a/src/crawlee/storages/_request_queue.py
+++ b/src/crawlee/storages/_request_queue.py
@@ -5,7 +5,7 @@ from collections import OrderedDict
 from contextlib import suppress
 from datetime import datetime, timedelta, timezone
 from logging import getLogger
-from typing import TYPE_CHECKING, Any, Generic, TypedDict, TypeVar
+from typing import TYPE_CHECKING, Any, TypedDict, TypeVar
 
 from typing_extensions import override
 
@@ -29,30 +29,6 @@ if TYPE_CHECKING:
 logger = getLogger(__name__)
 
 T = TypeVar('T')
-
-
-class BoundedSet(Generic[T]):
-    """A simple set datastructure that removes the least recently accessed item when it reaches `max_length`."""
-
-    def __init__(self, max_length: int) -> None:
-        self._max_length = max_length
-        self._data = OrderedDict[T, object]()
-
-    def __contains__(self, item: T) -> bool:
-        found = item in self._data
-        if found:
-            self._data.move_to_end(item, last=True)
-        return found
-
-    def add(self, item: T) -> None:
-        self._data[item] = True
-        self._data.move_to_end(item)
-
-        if len(self._data) > self._max_length:
-            self._data.popitem(last=False)
-
-    def clear(self) -> None:
-        self._data.clear()
 
 
 class CachedRequest(TypedDict):
@@ -97,12 +73,6 @@ class RequestQueue(BaseStorage, RequestProvider):
     _MAX_CACHED_REQUESTS = 1_000_000
     """Maximum number of requests that can be cached."""
 
-    _RECENTLY_HANDLED_CACHE_SIZE = 1000
-    """Cache size for recently handled requests."""
-
-    _STORAGE_CONSISTENCY_DELAY = timedelta(seconds=3)
-    """Expected delay for storage to achieve consistency, guiding the timing of subsequent read operations."""
-
     def __init__(
         self,
         id: str,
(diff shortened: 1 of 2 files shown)


B4nan:
I was hoping we would first address this in the JS version where its actually costing us money, do we have some reported problems on the python side too?

janbuchar:
> I was hoping we would first address this in the JS version where its actually costing us money, do we have some reported problems on the python side too?

I get your angle, but I wanted to try it on code where the implementation isn't scattered between `RequestProvider` and `RequestQueue`. Plus, yeah, actors also get stuck with the Python SDK and the request queue is a prime suspect.

B4nan:
If that abstraction is problematic I am fine with removing it and leaving the base class acting like an interface mostly (we can't just make it one as that would be technically breaking), duplication is not a huge deal if we plan to drop the RQ v1 implementation anyway - but first we need to make sure v2 is working fine, we still get peeps from delivery moving back to v1 to deal with those weird issues...

janbuchar:
> If that abstraction is problematic I am fine with removing it and leaving the base class acting like an interface mostly (we can't just make it one as that would be technically breaking), duplication is not a huge deal if we plan to drop the RQ v1 implementation anyway - but first we need to make sure v2 is working fine, we still get peeps from delivery moving back to v1 to deal with those weird issues...

...and all of the above is why I chose to validate the points by Kuba in the python version :smile:

B4nan:
My issue is that not many people use the python version, and those issues reported by delivery are quite random, so hard to confirm something helps here. Maybe it will resolve the issue we have a repro for here, that's great, but it might be a different one than those where we actually bleed money (we usually refund people with stuck runs completely).

I am fine doing it first here, but let's not wait for some confirmation before trying to fix it in the JS version.

src/crawlee/storage_clients/_memory/_request_queue_client.py
```diff
@@ -210,8 +214,23 @@ async def list_head(self, *, limit: int | None = None) -> RequestQueueHead:
...
+            name=self.name,
+        )
+
+        if existing_queue_by_id is None:
+            raise_on_non_existing_storage(StorageTypes.REQUEST_QUEUE, self.id)
+
+        result = await self.list_head(limit=limit, skip_in_progress=True)
+
+        for item in result.items:
+            existing_queue_by_id._in_progress.add(item.id)  # noqa: SLF001
```
vdusek, line 230: Maybe add a read-only property `in_progress` to avoid type ignores?

src/crawlee/storages/_request_queue.py
```diff
@@ -326,31 +301,11 @@ async def fetch_next_request(self) -> Request | None:
         await self._ensure_head_is_non_empty()
 
         # We are likely done at this point.
-        if len(self._queue_head_dict) == 0:
-            return None
-
-        next_request_id, _ = self._queue_head_dict.popitem(last=False)  # ~removeFirst()
-
-        # This should never happen, but...
```
vdusek, line 334: haha, great we're getting rid of this :smile:

src/crawlee/storages/_request_queue.py
```diff
@@ -33,35 +33,12 @@
...
-    def add(self, item: T) -> None:
-        self._data[item] = True
-        self._data.move_to_end(item)
-
-        if len(self._data) > self._max_length:
-            self._data.popitem(last=False)
-
-    def clear(self) -> None:
-        self._data.clear()
-
```
vdusek, line 58: :+1:

src/crawlee/storages/_request_queue.py
```diff
@@ -100,13 +77,12 @@ class RequestQueue(Storage, RequestManager):
     _MAX_CACHED_REQUESTS = 1_000_000
     """Maximum number of requests that can be cached."""
 
-    _RECENTLY_HANDLED_CACHE_SIZE = 1000
-    """Cache size for recently handled requests."""
-
-    _STORAGE_CONSISTENCY_DELAY = timedelta(seconds=3)
-    """Expected delay for storage to achieve consistency, guiding the timing of subsequent read operations."""
```
vdusek, line 107: :+1:

vdusek (approved):
The commit type - maybe fix would be better? I believe this deserves to be in the changelog.

Other than that, it looks much better, although I cannot say I do understand everything.

src/crawlee/storage_clients/_memory/_request_queue_client.py
```diff
@@ -72,6 +72,8 @@ def __init__(
         self.file_operation_lock = asyncio.Lock()
         self._last_used_timestamp = Decimal(0)
 
+        self._in_progress = set[str]()
```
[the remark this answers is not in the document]
janbuchar: See https://github.com/apify/crawlee-python/pull/653/files/b4a55933ad0e610c21aef955dbb8b58f6c7a0f41#r1987259821

PR #661, opened 2024-11-06
fix: Fix total_size usage in memory size monitoring

janbuchar:
SDK passes max memory usage as `total_size`, and we don't actually use that field anyway, so let's just drop it


vdusek (approved):
If we really do not need it, then LGTM.

PR #665, opened 2024-11-07
chore: update identifiers & API docs in `_types` module

vdusek:
Docstrings & some (internal) name alignments.


janbuchar (approved):
Since this renames types, I'd probably make it a chore or something like that. LGTM otherwise

PR #666, opened 2024-11-07
fix: Fix unhandled ValueError in request handler result processing

janbuchar:
We only checked the limit when committing the result, when the request handler was already considered successful. I added a check that results in an error before the request proceeds.

### DIFF
--- a/src/crawlee/_types.py
+++ b/src/crawlee/_types.py
@@ -399,6 +399,10 @@ class RequestHandlerRunResult:
         **kwargs: Unpack[PushDataKwargs],
     ) -> None:
         """Track a call to the `push_data` context helper."""
+        from crawlee.storages._dataset import Dataset
+
+        await Dataset.check_and_serialize(data)
+
         self.push_data_calls.append(
             PushDataFunctionCall(
                 data=data,
--- a/tests/unit/basic_crawler/test_basic_crawler.py
+++ b/tests/unit/basic_crawler/test_basic_crawler.py
@@ -650,6 +650,7 @@ async def test_crawler_push_data_over_limit() -> None:
 
     @crawler.router.default_handler
     async def handler(context: BasicCrawlingContext) -> None:
+        # Push a roughly 15MB payload - this should be enough to break the 9MB limit
         await context.push_data({'hello': 'world' * 3 * 1024 * 1024})
 
     stats = await crawler.run(['http://example.tld/1'])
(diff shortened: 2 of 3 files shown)


tests/unit/basic_crawler/test_basic_crawler.py
```diff
@@ -645,6 +645,17 @@ async def handler(context: BasicCrawlingContext) -> None:
     assert exported_json_str == expected_json_str
 
 
+async def test_crawler_push_data_over_limit() -> None:
+    crawler = BasicCrawler()
+
+    @crawler.router.default_handler
+    async def handler(context: BasicCrawlingContext) -> None:
+        await context.push_data({'hello': 'world' * 3 * 1024 * 1024})
```
Pijukatel, line 653: Why these numbers?
 len(('world' * 3 * 1024 * 1024).encode('utf-8')) is 15728640

Shouldn't this be somehow linked to the 9 mb defined in https://github.com/apify/crawlee-python/blob/e26a32383244d3972d3bd7621e7ae7aabf9a39c9/src/crawlee/storages/_dataset.py#L185 ?

src/crawlee/_types.py
```diff
@@ -399,6 +399,10 @@ async def push_data(
         **kwargs: Unpack[PushDataKwargs],
     ) -> None:
         """Track a call to the `push_data` context helper."""
+        from crawlee.storages._dataset import Dataset
```
Pijukatel, line 402: What is the justification for not having import on top of the file?

tests/unit/basic_crawler/test_basic_crawler.py
janbuchar: The important thing is that it's longer than the limit. Relying on constants imported from the code that's being tested is not great. We may want to explain this in a comment.

src/crawlee/_types.py
janbuchar: It would introduce an import cycle.

tests/unit/basic_crawler/test_basic_crawler.py
Pijukatel: I would expect both the test and code to import this currently unnamed "9" constant from somewhere. Test would then use something that is "imported_9+1". Such test would be more self explanatory and more resistant to change.

vdusek: > We may want to explain this in a comment.

Yes please

vdusek (approved):
Just please add the explanatory comment that was already discussed. Otherwise LGTM.

Pijukatel (approved):

Issue #157, opened 2024-05-20
Polish up logging


janbuchar:
In the current state, we make a new logger in every module that needs to log something. There is `CrawleeLogFormatter`, which handles logging in the console.

- our loggers should default to the "info" level
- ??? the log format for Apify platform may be different (doesn't it ingest JSON? probably not, but I'm not sure)
- in JS Crawlee, some classes accept loggers instead of creating them - we may want to revisit that
- printing extra key-value information should now be possible via `kwargs` of `logging.info` and related methods - should we use that?

vdusek:
We should also review the debug logs to ensure they provide some value. Suppose a user switches from info to debug level. In that case, they should get some meaningful information (e.g., about sending this request, processing this response, storing this item into the dataset, ...) that could help them (currently it is quite random and maybe a bit of spam).

PR #673, opened 2024-11-09
feat: If custom logger is set, pass to the AutoScaledPool

aayusharyan:
### Description
Even if I use a custom logger object to set the logs in my Crawlee instance, for the AutoScaled logs, it still uses the default logger. Ideally, when I am passing it to Crawlee, then it should also be internally shared across other subclasses.

Changed the AutoScaledPool constructor to also accept logger instance, if not sent, then create a new logger.

### Issues
Can be loosely said to improve logging for (https://github.com/apify/crawlee-python/issues/157)

### Testing
Need a little help/time to figure out why `test_emit_system_info_event` is breaking.

### Checklist
- [ ] CI passed

### DIFF
--- a/src/crawlee/_autoscaling/autoscaled_pool.py
+++ b/src/crawlee/_autoscaling/autoscaled_pool.py
@@ -3,10 +3,10 @@
 from __future__ import annotations
 
 import asyncio
+import logging
 import math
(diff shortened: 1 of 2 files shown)


vdusek:
Closing due to inactivity.

PR #674, opened 2024-11-10
feat: Add BeautifulSoupParser type alias

Pijukatel:
### Description

To avoid repeating same Literal definitions.

### Checklist

- [x] CI passed

### DIFF
--- a/src/crawlee/beautifulsoup_crawler/__init__.py
+++ b/src/crawlee/beautifulsoup_crawler/__init__.py
@@ -1,5 +1,5 @@
 try:
-    from ._beautifulsoup_crawler import BeautifulSoupCrawler
+    from ._beautifulsoup_crawler import BeautifulSoupCrawler, BeautifulSoupParser
     from ._beautifulsoup_crawling_context import BeautifulSoupCrawlingContext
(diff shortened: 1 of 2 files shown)


janbuchar (commented):
Sure, this makes sense. Could you consider reexporting the new type so that you can import it without accessing private subpackages?

Pijukatel:
> Sure, this makes sense. Could you consider reexporting the new type so that you can import it without accessing private subpackages?

Sure. Just a tiny question. What is the established practice here with __all__? I know it is for star imports, but do you put all stuff inside or just some selected top level stuff?

janbuchar:
You should include only names that are part of the "public interface" of the package in `__all__`.

vdusek:
I will also add that the `__all__` field is there primarily because some Python tooling uses it to discover the package's public interface.

janbuchar:
@Pijukatel And I will add this - it's a nice read: https://mkdocstrings.github.io/griffe/guide/users/recommendations/public-apis/

janbuchar (approved):
LGTM

vdusek (approved):
LGTM

PR #680, opened 2024-11-11
fix: Dataset.iterate_items

Pijukatel:
Missing await to return AsyncIterator instead of coroutine.


janbuchar:
Awesome! Could you add a simple test case to make sure this doesn't happen again?

B4nan:
@Pijukatel one note from me, please don't rebase once you get a review (or don't rebase at all except for rebasing changes from master to your branch), reviews are harder if not impossible because of that, since all we see is a single commit. We squash merge PRs so they end up as a single commit in master, so doing this inside the PR is not needed and only complicates the review.

(it's more important for future work, for small PRs like this its surely not a big deal)

---

Also - tests, tests, tests. If you ask me, you should start with that and only then write some code to fix the problem :] They can be really small, or just an extension of some existing tests, but we need at least something.

janbuchar:
> It seems wrong to me to define an abstract method as synchronous and then override it with an asynchronous implementation.

The base method is annotated with an `AsyncIterator` return type, so it's still pretty async to me... Obligatory "mypy complains if we do it any other way" :smile: 

Of course, feel free to elaborate on this @Pijukatel.

vdusek:
I see, thanks... I probably prefer

```python
    raise NotImplementedError
    if False:
        yield 0
```

more (with appropriate comment), as the method declarations remain consistent.

vdusek (approved):
Anyway, approving, as this preference might be subjective.

In any case, please:
- Add a comment explaining the rationale for the inconsistency or the empty yield statement, ideally with a link to the [Mypy docs on async iterators](https://mypy.readthedocs.io/en/stable/more_types.html#asynchronous-iterators).
- Conventional commits: after specifying the commit type, include a verb to make the rendered message meaningful. The commit type itself is just for categorizing commits.

src/crawlee/base_storage_client/_base_dataset_client.py
```diff
@@ -122,6 +122,11 @@ async def iterate_items(
             An asynchronous iterator of dictionary objects, each representing a dataset item after applying
             the specified filters and transformations.
         """
+        # This syntax is to make mypy properly work with abstract AsyncIterator.
+        # https://mypy.readthedocs.io/en/stable/more_types.html#asynchronous-iterators
+        raise NotImplementedError
+        if False:
```
vdusek, line 128: Unfortunately, this will be necessary

```suggestion
        if False:  # type: ignore
```

Issue #669, opened 2024-11-08
Add poetry lock file and automate its updates


vdusek:
### Motivation

Even patch updates can break things. Such as the latest patch release of Ruff caused CI failures, see [apify-sdk-python/actions/runs/11742212377/job/32712478164](https://github.com/apify/apify-sdk-python/actions/runs/11742212377/job/32712478164).

### Potential concerns

Ensure that adding a `poetry.lock` file does not impact the Poetry build step. Dependencies in the main library should remain version-flexible to avoid imposing unnecessary constraints. The lock file will be used only for development and CI to maintain consistent environments.

### Realization

- Add poetry lock file.
- Configure Renovate bot to automatically update the lock file weekly, aligned with its dev package update process.
- Do this in all Crawlee & SDK & Client repositories.

PR #681, opened 2024-11-11
chore: add poetry lock file

vdusek:
### Description

- Add poetry lock file and rm it from git ignore.
- There should be no need to update the Renovate config, as it should look for `poetry.lock` by default, see [docs](https://docs.renovatebot.com/configuration-options/#lockfilemaintenance).

### Issues

- Closes: #669

### Testing

- N/A

### Checklist

- [x] CI passed


poetry.lock
```diff
@@ -0,0 +1,3322 @@
...
+
+[package.dependencies]
+apify-client = ">=1.8.1"
+apify-shared = ">=1.1.2"
+crawlee = ">=0.3.8"
+cryptography = ">=42.0.0"
+httpx = ">=0.27.0"
+lazy-object-proxy = ">=1.10.0"
+typing-extensions = ">=4.1.0"
+websockets = ">=10.0"
```
Pijukatel, line 66: Maybe version should be locked here as well to <14

poetry.lock
```diff
@@ -0,0 +1,3322 @@
...
+optional = true
+python-versions = "*"
+files = [
+    {file = "webencodings-0.5.1-py2.py3-none-any.whl", hash = "sha256:a0af1213f3c2226497a97e2b3aa01a7e4bee4f403f95be16fc9acd2947514a78"},
+    {file = "webencodings-0.5.1.tar.gz", hash = "sha256:b36a1c245f2d304965eb4e0a82848379241dc04b865afcc4aab16748587e1923"},
+]
+
+[[package]]
+name = "websockets"
+version = "14.0"
```
Pijukatel, line 3110: It goes for the 14.

poetry.lock
janbuchar: AFAIK this is just a transitive dependency of `apify` here, so no need to lock it manually - it should happen automatically when we update to apify 2.0.2.

Issue #191, opened 2024-06-18
Implement `use_state` context helper method


vdusek:
- https://crawlee.dev/api/core/function/useState

vdusek:
@Mantisus

PR #682, opened 2024-11-11
feat: Add `use_state` context method

Mantisus:
### Description

Add `use_state` context helper method with auto-save behavior inspired by TS's realization of `crawlee`.
Adds a `get_auto_saved_value` method for `kvs`. Working with internal cache and saving data on `PERSIST_STATE` event

(At this point, it's more of a gut check and synchronization with the team. It will probably require some more refinement)

### Issues

- Closes: #191 

### Testing

For correct testing, it is necessary to use event_manager in tests for KeyValueStore, therefore I have added a new fixture.

### Checklist

- [x] CI passed


tests/unit/basic_crawler/test_basic_crawler.py
```diff
@@ -671,6 +679,20 @@ async def handler(context: BasicCrawlingContext) -> None:
...
+
+    @crawler.router.default_handler
+    async def handler(context: BasicCrawlingContext) -> None:
+        await context.use_state('state', {'hello': 'world'})
+        await asyncio.sleep(0.1)
+
+    await crawler.run(['https://hello.world'])
+
+    store = await crawler.get_key_value_store()
+    assert (await store.get_value('state')) == {'hello': 'world'}
```
janbuchar, line 693: I'd suggest adding a more complex test as well, perhaps one that accesses the shared state in different request handler "runs"?

tests/unit/basic_crawler/test_basic_crawler.py
```diff
@@ -671,6 +679,20 @@ async def handler(context: BasicCrawlingContext) -> None:
     assert (await store.get_value('foo')) == 'bar'
 
 
+async def test_context_use_state(mock_event_manager: EventManager) -> None:
+    crawler = BasicCrawler()
+
+    @crawler.router.default_handler
+    async def handler(context: BasicCrawlingContext) -> None:
+        await context.use_state('state', {'hello': 'world'})
+        await asyncio.sleep(0.1)
```
janbuchar, line 688: Could the sleep go after `crawler.run` instead? I assume it's to ensure key-value store consistency.

src/crawlee/storages/_key_value_store.py
```diff
@@ -171,3 +178,39 @@ async def get_public_url(self, key: str) -> str:
             The public URL for the given key.
         """
         return await self._resource_client.get_public_url(key)
+
+    async def get_auto_saved_value(self, key: str, default_value: dict[str, Any] | None = None) -> dict[str, Any]:
```
janbuchar, line 182: Currently, crawlee-js does this in two separate ways. What you ported over is the "regular" one. Apart from that, we have adaptive crawling, where `useState` bypasses `get_auto_saved_value`. See https://github.com/apify/crawlee/blob/master/packages/playwright-crawler/src/internals/adaptive-playwright-crawler.ts for details. 

The general idea behind this is that we don't want to "save" the changed state before we know that the request handler succeeded.

We should probably discuss what would be the best way forward. Try to give the javascript adaptive crawler a glance, but if it doesn't make any sense, just say so so that we don't waste any time. Then we'll talk more :slightly_smiling_face:

janbuchar (commented):
Seems like the right way to go, see my comments please :slightly_smiling_face:

tests/unit/basic_crawler/test_basic_crawler.py
Mantisus: That's a great idea. So I found out that I didn't take into account that in different handlers we work with different `KeyValueStore` objects.

I moved the `_cache` to the class variable level. But maybe the best solution would be to make some kind of `CacheService`.

src/crawlee/storages/_key_value_store.py
Mantisus: I looked up adaptive-playwright-crawler, if I understand its workings correctly, it completely overrides the default `useState`

We could do the same thing when we need to implement this in Python. Given that this behavior is specific to one particular implementation of `crawler`.

tests/unit/basic_crawler/test_basic_crawler.py
janbuchar: Okay, would that mean that if I did this in production, state changes in the last request handler run might not be saved?

Mantisus: That's right.

I tested this with different `persist_state_interval` values for `EventManager`, without forced sleep the saves only occur at values less than 10 milliseconds.

So if we have to ensure that the state is persisted, all `Event.PERSIST_STATE` listeners should be forced run when the `EventManager` is closed.

src/crawlee/_types.py
```diff
@@ -300,6 +300,16 @@ def __call__(
     ) -> Coroutine[None, None, HttpResponse]: ...
 
 
+class UseStateFunction(Protocol):
+    """Type of a function for performing Use State."""
```
vdusek, line 304:
```suggestion
    """Type of a function for performing use state."""
```

src/crawlee/storages/_key_value_store.py
```diff
@@ -15,6 +17,7 @@
     from crawlee.configuration import Configuration
 
 T = TypeVar('T')
+DictT = TypeVar('DictT', bound=dict[str, Any])
```
vdusek, line 20: Sorry, but what is this good for?

src/crawlee/storages/_key_value_store.py
```diff
@@ -175,3 +182,56 @@ async def get_public_url(self, key: str) -> str:
             The public URL for the given key.
         """
         return await self._resource_client.get_public_url(key)
+
+    async def get_auto_saved_value(self, key: str, default_value: dict[str, Any] | None = None) -> dict[str, Any]:
+        """Gets a value from store that will be automatically saved on changes.
```
vdusek, line 187: Could we use KVS in the docs please? `store` could be easily mistaken for something else

```suggestion
        """Gets a value from KVS that will be automatically saved on changes.
```

src/crawlee/storages/_key_value_store.py
```diff
@@ -175,3 +182,56 @@ async def get_public_url(self, key: str) -> str:
...
+
+        Args:
+            key: Key of the record, to store the value.
+            default_value: value to be used if the record does not exist yet. Should be a dictionary
```
vdusek, line 191:
```suggestion
            default_value: Value to be used if the record does not exist yet. Should be a dictionary.
```

src/crawlee/storages/_key_value_store.py
```diff
@@ -175,3 +182,56 @@ async def get_public_url(self, key: str) -> str:
...
+
+        Returns:
+            Returns the value of the key
```
vdusek, line 194:
```suggestion
            Returns the value of the key.
```

src/crawlee/storages/_key_value_store.py
```diff
@@ -175,3 +182,56 @@ async def get_public_url(self, key: str) -> str:
...
+        if key in self._cache:
+            return self._cache[key]
+
+        value = await self.get_value(key, default_value)
+
+        self._cache[key] = value
+
+        self._ensure_persist_event()
+
+        return cast(dict[str, Any], value)
```
vdusek, line 207: I believe the cast is not necessary, correct?

The returned value of `get_value` is the same type of `default_value`, which is `dict[str, Any]`.

src/crawlee/basic_crawler/_basic_crawler.py
```diff
@@ -524,6 +526,14 @@ async def add_requests(
             wait_for_all_requests_to_be_added_timeout=wait_for_all_requests_to_be_added_timeout,
         )
 
+    async def _use_state(self, key: str, default_value: dict | None = None) -> dict:
```
vdusek, line 529: General dicts (`dict`) are used here, but more specific (`dict[str, Any]`) are used in the KVS. Is that intended?

vdusek (commented):
a few small things

src/crawlee/basic_crawler/_basic_crawler.py
Mantisus: Yes, it's a state `dict`, it is the `Any` in KVS `dict`

src/crawlee/storages/_key_value_store.py
Mantisus: Thank you for pointing that out to me.

A forgotten artefact that was left over from the implementation.

src/crawlee/basic_crawler/_basic_crawler.py
Mantisus: Thank you. Matched the types between KVS and the function

vdusek (approved):
Good job :slightly_smiling_face:

src/crawlee/basic_crawler/_basic_crawler.py
janbuchar: What is the reasoning behind using a plain `dict`? Also, we might want to use `dict[str, JsonSerializable]` - otherwise it might confuse users (like `Request.user_data` did).

src/crawlee/basic_crawler/_basic_crawler.py
```diff
@@ -524,6 +526,14 @@ async def add_requests(
...
+        store = await self.get_key_value_store()
+        return await store.get_auto_saved_value(key, default_value)
+
+    async def _save_persist_state(self) -> None:
```
janbuchar, line 533: I suggest we rename this method to `_save_crawler_state` and the one on the `KeyValueStore` to `persist_autosaved_values`.

src/crawlee/storages/_key_value_store.py
```diff
@@ -52,6 +54,10 @@ class KeyValueStore(BaseStorage):
     ```
     """
 
+    # Cache for persistent (auto-saved) values
+    _cache: ClassVar[dict[str, dict[str, Any]]] = {}
```
janbuchar, line 58: This way the cache will be shared between the default key-value store and its potential named counterparts, which means it will behave quite unpredictably if you use named key-value stores in your crawler.

I know you did this so that the cache would be shared between separate crawler instances, but that should work nonetheless - `KeyValueStore.open()` should always return the same instance.

src/crawlee/storages/_key_value_store.py
```diff
@@ -175,3 +181,61 @@ async def get_public_url(self, key: str) -> str:
...
+    def clear_cache(self) -> None:
+        """Clear cache with persistent values."""
+        self._cache.clear()
+
+    def drop_persist_state_event(self) -> None:
+        """Off event_manager listener and drop event status."""
+        if self._persist_state_event_started:
+            event_manager = service_container.get_event_manager()
+            event_manager.off(event=Event.PERSIST_STATE, listener=self._persist_save)
+        self._persist_state_event_started = False
```
janbuchar, line 236: I think these two methods should be protected?

janbuchar (changes requested):
It makes sense, but I have a couple of concerns :slightly_smiling_face:

janbuchar:
One thing we came up with @vdusek - we should mark the `use_state` helper as experimental. We may want to change the behavior later on, so let's make that clear.

src/crawlee/basic_crawler/_basic_crawler.py
Mantisus: I like the idea of using `dict[str, JsonSerializable]` it makes the signature and usage clearer.

src/crawlee/storages/_key_value_store.py
Mantisus: I was moving this cache to the class level. Because when using `self.get_key_value_value_value_store()` I was getting different instances of `KeyValueStore`. So different handlers had different cache state, this became noticeable after adding the `test_context_handlers_use_state` test.

But I don't see this behavior anymore and the test passes correctly, this is interesting.....

src/crawlee/_types.py
```diff
@@ -300,6 +300,21 @@ def __call__(
     ) -> Coroutine[None, None, HttpResponse]: ...
 
 
+class UseStateFunction(Protocol):
+    """Type of a function for performing use state.
+
+    Warning:
+    -------
+    This is an experimental feature. The behavior and interface may change in future versions.
+    """
```
vdusek, line 309: I believe this should be written like this following the Google style Python docstrings

```suggestion
    """Type of a function for performing use state.

    Warning:
        This is an experimental feature. The behavior and interface may change in future versions.
    """
```

tests/unit/basic_crawler/test_basic_crawler.py
```diff
@@ -20,15 +20,34 @@
...
+            yield event_manager
+
+
+@pytest.fixture
+async def key_value_store() -> AsyncGenerator[KeyValueStore, None]:
+    kvs = await KeyValueStore.open()
+    yield kvs
+    await kvs.drop()
+    kvs._clear_cache()
+    kvs._drop_persist_state_event()
```
janbuchar, line 49: Might be better to use a mock event manager and just discard it at the end of the test. That way the `KeyValueStore` doesn't have to deal with unregistering the listeners at all, and I think that's a good thing.

janbuchar (approved):
Seems legit, thanks!

Issue #668, opened 2024-11-07
Issues with payload in POST requests


francomanca93:
### Description

When using the `crawlee.http_crawler._http_crawler` module, I'm encountering a `HttpStatusCodeError` with a `404` status code. This error occurs during the `_make_http_request` operation and causes the crawler to reach the maximum number of retries.

The issue with the 404 error in the crawler seems to be related to the way i'm handling the request payload.

![image](https://github.com/user-attachments/assets/f5215110-0b64-4666-9300-975f6d6d1676)

### Steps to Reproduce

1. Initialize a new Crawlee project and set up an `HttpCrawler` with either the `CurlImpersonateHttpClient` or the `HttpxHttpClient`:
```py
import asyncio
import json
from crawlee import Request
from crawlee.http_crawler import HttpCrawler, HttpCrawlingContext
from crawlee.http_clients.curl_impersonate import CurlImpersonateHttpClient
from crawlee.http_clients import HttpxHttpClient

async def main() -> None:
    #http_client = CurlImpersonateHttpClient(
    #    persist_cookies_per_session=True,
    #)
    # Or use HttpxHttpClient:
    http_client = HttpxHttpClient(
        persist_cookies_per_session=True,
    )

    crawler = HttpCrawler(
        http_client=http_client,
        max_requests_per_crawl=20,
    )

    url = "https://www.viagogo.es/Entradas-Deportes/Futbol/Real-Madrid-C-F-Entradas/E-153769088"
    payload = {
        'ShowAllTickets': True,
        'HideDuplicateTicketsV2': False,
        'Quantity': 2,
        'IsInitialQuantityChange': False,
        'PageSize': 20,
        'CurrentPage': 2,
        'SortBy': 'NEWPRICE',
        'SortDirection': 0,
        'Sections': '',
        'Rows': '',
        'Seats': '',
        'SeatTypes': '',
        'TicketClasses': '',
        'ListingNotes': '',
        'PriceRange': '0,100',
        'InstantDelivery': False,
        'EstimatedFees': True,
        'BetterValueTickets': True,
        'PriceOption': '',
        'HasFlexiblePricing': False,
        'ExcludeSoldListings': False,
        'RemoveObstructedView': False,
        'NewListingsOnly': False,
        'PriceDropListingsOnly': False,
        'SelectBestListing': False,
        'ConciergeTickets': False,
        'Method': 'IndexSh'
    }
    payload_bytes = json.dumps(payload).encode()

[...]

PR #683, opened 2024-11-12
fix: Fix serialization payload in Request. Fix Docs for Post Request

Mantisus:
### Description

An incorrect serialization for `payload` is happening now. `payload` is represented as a byte string, but bytes cannot be serialized to json. Therefore, when loading and unloading from the queue (`_persist_single_request_to_storage`, `_json_to_request`) we get strings of format “b'test'”

Also corrected the documentation, as when filling forms the data should be passed not in json format and have the appropriate header.

### Issues

- Closes: #668 

### Testing

Updated `test_request_state_serialization` to also take `payload` into account when serializing and deserializing `Request`

### Checklist

- [x] CI passed

### DIFF
--- a/docs/examples/code/fill_and_submit_web_form_crawler.py
+++ b/docs/examples/code/fill_and_submit_web_form_crawler.py
@@ -1,5 +1,5 @@
 import asyncio
-import urllib.parse
+from urllib.parse import urlencode
 
 from crawlee import Request
 from crawlee.http_crawler import HttpCrawler, HttpCrawlingContext
@@ -20,7 +20,7 @@ async def main() -> None:
         url='https://httpbin.org/post',
         method='POST',
         headers={'content-type': 'application/x-www-form-urlencoded'},
-        payload=urllib.parse.urlencode(
+        payload=urlencode(
             {
                 'custname': 'John Doe',
                 'custtel': '1234567890',
--- a/docs/examples/code/fill_and_submit_web_form_crawler.py
+++ b/docs/examples/code/fill_and_submit_web_form_crawler.py
@@ -1,5 +1,5 @@
 import asyncio
-import json
+import urllib.parse
 
 from crawlee import Request
 from crawlee.http_crawler import HttpCrawler, HttpCrawlingContext
@@ -19,7 +19,8 @@ async def main() -> None:
     request = Request.from_url(
         url='https://httpbin.org/post',
         method='POST',
-        payload=json.dumps(
+        headers={'content-type': 'application/x-www-form-urlencoded'},
+        payload=urllib.parse.urlencode(
             {
                 'custname': 'John Doe',
                 'custtel': '1234567890',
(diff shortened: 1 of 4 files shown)


src/crawlee/_request.py
```diff
@@ -176,6 +178,20 @@ class BaseRequestData(BaseModel):
...
+        if isinstance(v, str):
+            v = v.encode()
+        return v
+
+    @field_serializer('payload')
+    @classmethod
+    def _payload_serializer(cls, v: bytes | None) -> str | None:
+        if isinstance(v, bytes):
+            return v.decode()
+        return None
```
vdusek, line 193: Could we do this directly on the field by using Pydantic's `Annotated`? Like we do for the other fields.

Something like this:

```python
    payload: Annotated[
        HttpPayload | None,
        BeforeValidator(lambda v: v.encode() if isinstance(v, str) else v),
        PlainSerializer(lambda v: v.decode() if isinstance(v, bytes) else None),
    ] = None
    """HTTP request payload."""
```

docs/examples/code/fill_and_submit_web_form_crawler.py
```diff
@@ -1,5 +1,5 @@
 import asyncio
-import json
+import urllib.parse
```
vdusek, line 2: A nitpick... Could you please use `from urllib.parse import urlencode` so that it's shorter in the code? Same for the other example. Thanks.

vdusek (changes requested):
I got just stylistic things... We already discussed it directly, but we should make sure this is going to work on the platform. Let me know once it is tested. Thank you.

janbuchar:
Thanks @Mantisus! Could you please add a test that shows that the bug has been fixed?

tests/unit/_memory_storage_client/test_request_queue_client.py
```diff
@@ -93,19 +93,20 @@ async def test_list_head(request_queue_client: RequestQueueClient) -> None:
...
     result = await request_queue_client.list_head()
-
     assert len(result.items) == 1
     assert result.items[0] == request
 
     got_request = await request_queue_client.get_request(request.id)
 
     assert request == got_request
+    assert request.payload == got_request.payload
+    assert got_request.payload != b"b'test'"
```
vdusek, line 109: I believe this should be enough, correct?

```suggestion
    assert request.payload == got_request.payload
```

vdusek (approved):
One last thing, otherwise LGTM, thanks.

src/crawlee/_request.py
Mantisus: Updated.

The code works correctly on Apify platform

janbuchar (commented):
I still don't see a test case that would prove that https://github.com/apify/crawlee-python/issues/668 is resolved. It might be necessary to test that on HttpClient level.

Mantisus:
> I still don't see a test case that would prove that #668 is resolved. It might be necessary to test that on HttpClient level.

The problem in #668  is due to the fact that after serialization-deserialization payload does not match, b “test” becomes b “b'test'”. We can add a test to check this case specifically.

But in essence, it is sufficient.

``python
assert request == got_request
```

All the other problems described in #668 are related to incorrectly created payload for the Request.

PR #685, opened 2024-11-12
fix: respect custom HTTP headers in `PlaywrightCrawler`

Mantisus:
### Description

add `Request.headers` support for `PlaywrightCrawler`. 

### Testing

The test uses `'https://httpbin.org/get' to ensure that the headers have been installed

### Checklist

- [x] CI passed

### DIFF
--- a/src/crawlee/playwright_crawler/_playwright_crawler.py
+++ b/src/crawlee/playwright_crawler/_playwright_crawler.py
@@ -149,6 +149,8 @@ class PlaywrightCrawler(BasicCrawler[PlaywrightCrawlingContext]):
         """
         async with context.page:
             # Navigate to the URL and get response.
+            if context.request.headers:
+                await context.page.set_extra_http_headers(context.request.headers.model_dump())
             response = await context.page.goto(context.request.url)
 
             if response is None:
--- a/src/crawlee/playwright_crawler/_playwright_crawler.py
+++ b/src/crawlee/playwright_crawler/_playwright_crawler.py
@@ -148,9 +148,9 @@ class PlaywrightCrawler(BasicCrawler[PlaywrightCrawlingContext]):
                 infinite_scroll).
         """
         async with context.page:
-            # Navigate to the URL and get response.
             if context.request.headers:
                 await context.page.set_extra_http_headers(context.request.headers.model_dump())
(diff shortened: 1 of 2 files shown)


src/crawlee/playwright_crawler/_playwright_crawler.py
```diff
@@ -149,6 +149,8 @@ async def _navigate(
         """
         async with context.page:
             # Navigate to the URL and get response.
+            if context.request.headers:
+                await context.page.set_extra_http_headers(context.request.headers.model_dump())
```
vdusek, line 153: The comment belongs to the next line.

```suggestion
            if context.request.headers:
                await context.page.set_extra_http_headers(context.request.headers.model_dump())
            # Navigate to the URL and get response.
```

tests/unit/playwright_crawler/test_playwright_crawler.py
```diff
@@ -133,6 +133,24 @@ async def request_handler(context: PlaywrightCrawlingContext) -> None:
...
+    async def request_handler(context: PlaywrightCrawlingContext) -> None:
+        response = await context.response.text()
+        response_headers = dict(json.loads(response)).get('headers', {})
+
+        for key, val in response_headers.items():
+            headers[key] = val
+
+    await crawler.run([Request.from_url('https://httpbin.org/get', headers=custom_headers)])
+    for key, value in custom_headers.items():
+        assert headers[key] == value
```
vdusek, line 151: I believe we could compare the dicts as a whole like this:

```suggestion
    assert headers == custom_headers
```

also could we use a better names, e.g. `request_headers` and `response_headers`? :slightly_smiling_face:

vdusek (commented):
Just a few details, otherwise it's great, thank you! :slightly_smiling_face:

tests/unit/playwright_crawler/test_playwright_crawler.py
Mantisus: Dictionary matching doesn't work here because `set_extra_http_headers` updates headers rather than completely replacing them. So `response_headers` will contain more headers than `request_headers`.

But yes, I think it's better to do a more explicit check and update the names )

vdusek (approved):
LGTM

Issue #539, opened 2024-09-23
Crawler doesn't respect `configuration` argument


tlinhart:
Consider this sample program:

```python
import asyncio

from crawlee.configuration import Configuration
from crawlee.parsel_crawler import ParselCrawler, ParselCrawlingContext


async def default_handler(context: ParselCrawlingContext) -> None:
    for category in context.selector.xpath(
        '//div[@class="side_categories"]//ul/li/ul/li/a'
    ):
        await context.push_data({"category": category.xpath("normalize-space()").get()})


async def main() -> None:
    config = Configuration(persist_storage=False, write_metadata=False)
    crawler = ParselCrawler(request_handler=default_handler, configuration=config)
    await crawler.run(["https://books.toscrape.com"])
    data = await crawler.get_data()
    print(data.items)


if __name__ == "__main__":
    asyncio.run(main())
```

The `configuration` argument given to `ParselCrawler` is not respected, during the run it creates the `./storage` directory and persist all the (meta)data. I have to work around it by overriding the global configuration likes this:

```python
import asyncio

from crawlee.configuration import Configuration
from crawlee.parsel_crawler import ParselCrawler, ParselCrawlingContext


async def default_handler(context: ParselCrawlingContext) -> None:
    for category in context.selector.xpath(
        '//div[@class="side_categories"]//ul/li/ul/li/a'
    ):
        await context.push_data({"category": category.xpath("normalize-space()").get()})


async def main() -> None:
    config = Configuration.get_global_configuration()
    config.persist_storage = False
    config.write_metadata = False
    crawler = ParselCrawler(request_handler=default_handler)
    await crawler.run(["https://books.toscrape.com"])
    data = await crawler.get_data()
    print(data.items)


if __name__ == "__main__":
    asyncio.run(main())
```

janbuchar:
Hello, and thanks for the reproduction! It seems that the problem is here:

https://github.com/apify/crawlee-python/blob/master/src/crawlee/storages/_creation_management.py#L122-L132

It looks like `service_container.get_storage_client` does not consider the adjusted `configuration`.

Issue #369, opened 2024-07-27
How can I disable cache completely?


1hachem:
I am trying to write a simple function to crawl a website and I don't want crawlee to cache anything (each time I call this function it will do everything from scratch).

here is my attempt so far, I tried with `persist_storage=False` and `purge_on_start=True` in the configuration, and with removing the storage directory entirely, but I keep getting either a concatenated result of all the requests or and empty result in case I delete the storage directory.
 
```python
async def main(
    website: str,
    include_links: list[str],
    exclude_links: list[str],
    depth: int = 5,
) -> str:
    crawler = BeautifulSoupCrawler(
        # Limit the crawl to max requests. Remove or increase it for crawling all links.
        max_requests_per_crawl=depth,
    )
    dataset = await Dataset.open(
        configuration=Configuration(
            persist_storage=False,
            purge_on_start=True,
        ),
    )

    # Define the default request handler, which will be called for every request.
    @crawler.router.default_handler
    async def request_handler(context: BeautifulSoupCrawlingContext) -> None:  # type: ignore
        # Extract data from the page.
        text = context.soup.get_text()

        await dataset.push_data({"content": text})

        # Enqueue all links found on the page.
        await context.enqueue_links(
            include=[Glob(url) for url in include_links],
            exclude=[Glob(url) for url in exclude_links],
        )

    # Run the crawler with the initial list of URLs.
    await crawler.run([website])
    data = await dataset.get_data()

    content = "\n".join([item["content"] for item in data.items])  # type: ignore

    return content
```

also is there a way to simple get the result of the crawl as a string, and not use `Dataset` ?

any help is appreciated 🤗 thank you in advance !

janbuchar:
Hello and thank you for your interest in Crawlee! This seems closely related to #351. Could you please re-check that you get an empty string if you run this after removing the storage directory? I can imagine getting an empty string on a second run without deleting the storage (because of both `persist_storage=False` and `purge_on_start` functioning incorrectly), but what you're describing sounds strange.

tlinhart:
After some debugging I found a workaround to avoid re-using the cache. Basically we have to ensure that each time the crawler runs it uses a different request queue, e.g. like this:

```
import uuid

...
config = Configuration.get_global_configuration()
config.default_request_queue_id = uuid.uuid4().hex
...
```

It would be great if we could actually disable caching at all but this works for now.

tlinhart:
[...]

Issue #670, opened 2024-11-08
Fix the usage of Configuration class


vdusek:
Our [`Configuration`](https://github.com/apify/crawlee-python/blob/master/src/crawlee/configuration.py) class currently contains many fields that are unused and/or undocumented. Review each field, decide whether it should be used, removed, or consolidated, and add missing doc strings.

janbuchar:
Thanks for bringing this up! If you could write out the attributes in question here, we can discuss the appropriate steps together.

vdusek:
The fields that are used: 
- `internal_timeout`
- `verbose_log` (but it just sets the DEBUG level, and we have another field exactly for that)
- `default_dataset_id`
- `default_key_value_store_id`
- `default_request_queue_id`
- `purge_on_start`
- `write_metadata`
- `persist_storage`
- `memory_mbytes`
- `available_memory_ratio`
- `storage_dir`

The fields that are not used:
- `default_browser_path`
- `disable_browser_sandbox`
- `persist_state_interval`
- `system_info_interval`
- `max_used_cpu_ratio`
- `chrome_executable_path`
- `headless`
- `xvfb`

PR #691, opened 2024-11-13
fix!: Refactor service usage to rely on `service_locator`

vdusek:
### Description

- The `service_container` module has been completely refactored. It introduces changes to its usage, and resulting in many changes across the code base.
- While it remains possible to pass "services" directly to components as before, components now rely on the `service_container` internally. They no longer store service instances themselves.
- A new `force` flag has been added to the `service_container`'s setters, which is especially useful for testing purposes.
- This also quite simplifies the `memory_storage_client`.
- We now have only `set_storage_client`, the same approach as for the `event_manager`. This is more flexible (allows more envs than just local & cloud). And in the SDK Actor, we can them based on the `is_at_home`.
- This is a breaking change, but it affects only the `service_container` interface.

### Open discussion

- [x] Should we go further and remove the option to pass configuration, event managers, or storage clients directly to components - requiring them to be set exclusively via the `service_container`? Thoughts are welcome.
  - No
- [x] Better name for `service_container`? `service_locator`?
  - `service_locator`

### Issues

- Closes: #699
- Closes: #539 
- Closes: #369 
- It also unlocks the:
  - #670,
  -  and maybe https://github.com/apify/apify-sdk-python/issues/324#issuecomment-2475786435.

### Testing

- Existing tests, including those covering the `service_container`, have been updated to reflect these changes.
- New tests covering the `MemoryStorageClient` respects the `Configuration`.

### Manual reproduction

- This code snippet demonstrates that the `Crawler` and `MemoryStorageClient` respects the custom `Configuration`. 
- Note: Some fields remain non-working for now. These will be addressed in a future PR, as this refactor is already quite big. However, with the new architecture, those updates should now be easy.

```python
import asyncio

from crawlee.http_crawler import HttpCrawler, HttpCrawlingContext
from crawlee.service_container import set_configuration


async def main() -> None:
    config = Configuration(persist_storage=False, write_metadata=False)
    set_configuration(config)  # or Crawler(config=config)
    crawler = HttpCrawler()

    async def request_handler(context: HttpCrawlingContext) -> None:
        context.log.info(f'Processing {context.request.url} ...')
        await context.push_data({'url': context.request.url})

    await crawler.run(['https://crawlee.dev/'])


if __name__ == '__main__':
    asyncio.run(main())
```

### Checklist

- [x] CI passed


src/crawlee/errors.py
```diff
@@ -89,13 +88,3 @@ def __init__(self, wrapped_exception: Exception, crawling_context: BasicCrawling
 @docs_group('Errors')
 class ContextPipelineInterruptedError(Exception):
     """May be thrown in the initialization phase of a middleware to signal that the request should not be processed."""
-
-
-@docs_group('Errors')
-class ServiceConflictError(RuntimeError):
```
janbuchar, line 95: Do you have any other plans to prevent services from being overwritten from multiple call sites? I think it might be useful to keep it.

src/crawlee/service_container.py
```diff
@@ -18,50 +16,51 @@
...
 StorageClientType = Literal['cloud', 'local']
 
 
-class _Services(TypedDict):
-    local_storage_client: NotRequired[BaseStorageClient]
-    cloud_storage_client: NotRequired[BaseStorageClient]
-    configuration: NotRequired[Configuration]
-    event_manager: NotRequired[EventManager]
+@dataclass
+class _Services:
```
janbuchar, line 30: It might be useful to pull the type out so that we can require it in places where we may want to override the default service container - e.g., `BasicCrawlerOptions`.

janbuchar (commented):
Thanks for kicking this off!

src/crawlee/configuration.py
```diff
@@ -229,20 +228,3 @@ class Configuration(BaseSettings):
...
-            service_container.set_configuration(cls())
-
-        global_instance = service_container.get_configuration()
-
-        if not isinstance(global_instance, cls):
-            raise TypeError(
-                f'Requested global configuration object of type {cls}, but {global_instance.__class__} was found'
-            )
-
-        return global_instance
```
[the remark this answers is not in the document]
vdusek: Is there any issue why we cannot use `service_container.get_configuration` from SDK?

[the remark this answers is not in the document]
janbuchar: backward compatibility

janbuchar:
Also https://github.com/apify/apify-sdk-python/issues/324#issuecomment-2475786435 is relevant here - if we could remove the need to initialize `BasicCrawler` under `async with Actor`, that would be great.

src/crawlee/errors.py
vdusek: `ServiceConflictError` is back

src/crawlee/service_container.py
vdusek: We do not need it in the end as the services can be still passed directly to the components.

src/crawlee/storages/_dataset.py
```diff
@@ -260,7 +241,7 @@ async def push_data(self, data: JsonSerializable, **kwargs: Unpack[PushDataKwarg
         # Handle singular items
         if not isinstance(data, list):
             items = await self.check_and_serialize(data)
-            return await self._resource_client.push_items(items, **kwargs)
+            return await self._resource_client.push_items(items, **kwargs)  # type: ignore[no-any-return] # Mypy is broken
```
vdusek, line 244: No idea what happened. Pyright has no problem with these. For Mypy I had to add these type ignores.

Pijukatel (approved):
Looks good to me, but maybe @janbuchar should approve as well.

src/crawlee/storages/_dataset.py
vdusek: Exactly, Pyright/Pylance correctly recognizes the type, but for some reason, Mypy assumes it's Any 🤷‍♂️.

janbuchar (changes requested):
Sorry for writing this on a Friday. Also, maybe this is the right approach, but I feel obligated to challenge it because it will have quite the impact. 

> * While it remains possible to pass "services" directly to components as before, components now rely on the `service_container` internally. They no longer store service instances themselves.
> * Should we go further and remove the option to pass configuration, event managers, or storage clients directly to components - requiring them to be set exclusively via the `service_container`? Thoughts are welcome.

1. Not storing instances is a good idea if we want to keep working with the library simple. It's a bit of a shame that we cannot enforce this somehow (apart from a custom linter I guess... correct me if I'm wrong).
2. Removing the option to pass services directly also sounds good - otherwise, we are allowing users to bypass the service container in unpredictable ways. **_But._**
3. We absolutely need a way to override some services :grin: ideally when instantiating the `BasicCrawler`. I imagine we could find some way to "fork" the service container and pass it down with some magic such as contextvars. We could also just pass the service container down explicitly. But before elaborating, I'd like to hear what you think about that - I'm sure you've seen some dead end approaches already. 

> * A new `force` flag has been added to the `service_container`'s setters, which is especially useful for testing purposes.

This seems a real mother of a footgun to me. For 99% of programmers, if something fails and they notice that there's a `force` flag, they will use it without as much as a thought. See my thoughts above - for testing, we should have a more robust mechanism of overriding stuff in the service container.

> * We now have only `set_storage_client`, the same approach as for the `event_manager`. This is more flexible (allows more envs than just local & cloud). And in the SDK Actor, we can them based on the `is_at_home`.

This breaks the SDK - see my comment on `_creation_management.py`. I think we should take some time to think about this after resolving the rest of the issues.

janbuchar:
One more note - we might want to utilize https://github.com/ionelmc/python-lazy-object-proxy like we do in the SDK. It would allow us to pass proxies to "components" that would reach for the actual dependency into the service container. This would make it easier to switch out dependencies on the fly, but perhaps it would be too magic.

tests/unit/test_service_container.py
```diff
@@ -1,92 +1,59 @@
...
+)
 
 
-async def test_get_storage_client() -> None:
-    storage_client = service_container.get_storage_client()
-    assert isinstance(storage_client, MemoryStorageClient)
+def test_configuration() -> None:
+    default_config = Configuration()
+    config = get_configuration()
+    assert config == default_config
```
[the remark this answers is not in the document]
vdusek: No.

```
In [1]: from crawlee.configuration import Configuration

In [2]: Configuration() == Configuration()
Out[2]: True

In [3]: Configuration() is Configuration()
Out[3]: False
```

vdusek:
Make sure SDK's `force_cloud` works.

Pijukatel:
> One more note - we might want to utilize https://github.com/ionelmc/python-lazy-object-proxy like we do in the SDK. It would allow us to pass proxies to "components" that would reach for the actual dependency into the service container. This would make it easier to switch out dependencies on the fly, but perhaps it would be too magic.

To me it seems like a similar functionality that https://docs.python.org/3/library/functools.html#functools.cached_property offers. Is there any benefit of using python-lazy-object-proxy over cached_property?

janbuchar:
> To me it seems like a similar functionality that https://docs.python.org/3/library/functools.html#functools.cached_property offers. Is there any benefit of using python-lazy-object-proxy over cached_property?

Yep, the proxy acts like a normal object and does the loading on demand, no function call required. See the Actor class for an example.

tests/unit/test_configuration.py
```diff
@@ -1,7 +1,67 @@
...
+
+
+async def test_storage_not_persisted_when_disabled(tmp_path: Path) -> None:
+    config = Configuration(
+        persist_storage=False,
+        write_metadata=False,
+        crawlee_storage_dir=str(tmp_path),  # type: ignore
+    )
+    storage_client = MemoryStorageClient(config)
+    set_storage_client(storage_client, force=True)
```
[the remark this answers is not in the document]
vdusek: Sorry, could you elaborate it a bit? I am not sure I know what you have in mind.

If you are referring to passing options to the `MemoryStorageClient`, you can now provide them directly:

```python
    storage_client = MemoryStorageClient(
        persist_storage=False,
        write_metadata=False,
        storage_dir=str(tmp_path),
    )
    set_storage_client(storage_client, force=True)
```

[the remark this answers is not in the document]
janbuchar: Yes, but in JS you can do this:

```ts
new BasicCrawler({
  configuration: new Configuration({
    storageClient: SQLiteStorageClientIDontKnow()
  })
})
```

...and it just works. I'm not advocating this exact API, it's really not that great. My point is, the `*Crawler` classes are basically the entrypoint to Crawlee. If we could have a way to tweak the "dependencies" using their constructors alone, it could make this functionality more discoverable... Which would be nice, given that users evidently struggle with the default filesystem-backed memory storage often :slightly_smiling_face:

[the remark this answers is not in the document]
vdusek: Sure, `BasicCrawler` will accept storage client as an input arg.

src/crawlee/configuration.py
```diff
@@ -233,17 +233,15 @@ class Configuration(BaseSettings):
...
+        """Retrieve the global instance of the configuration.
 
-        if service_container.get_configuration_if_set() is None:
-            service_container.set_configuration(cls())
+        TODO: Can we remove this?
+        """
+        from crawlee.service_container import get_configuration
 
-        global_instance = service_container.get_configuration()
+        cfg = get_configuration()
```
Pijukatel, line 242: Super tiny detail. In Previous file you called such variable config, but here cfg. So for consistency I would suggest to all it config or cfg everywhere.

vdusek: Makes sense. I will stay with `config` then.

janbuchar:
> Make sure SDK's `force_cloud` works.
> 
> Update: This is completely duable on the SDK level without breaking changes. E.g. for opening a request queue on the `Actor` level:
> 
> ```python
>     async def open_request_queue(
>         self,
>         *,
>         id: str | None = None,
>         name: str | None = None,
>         force_cloud: bool = False,
>     ) -> RequestQueue:
>         if force_cloud:
>             cloud_storage_client = ApifyStorageClient(...)
>             service_container.set_storage_client(cloud_storage_client)
> 
>         return await RequestQueue.open(id=id, name=name)
> ```

Really tho? In this snippet, you set a "global" storage client, which might have unexpected side effects (other storages picking up the cloud storage client even with `force_cloud=False`). You also need to do an `await`, so even if you clean up after yourself by reverting the storage client, there is still a chance that somebody gets a surprise cloud client while your coroutine is waiting for `RequestQueue` to be opened.

vdusek:
> Really tho? In this snippet, you set a "global" storage client, which might have unexpected side effects (other storages picking up the cloud storage client even with force_cloud=False). You also need to do an await, so even if you clean up after yourself by reverting the storage client, there is still a chance that somebody gets a surprise cloud client while your coroutine is waiting for RequestQueue to be opened.

Alright, the requirements are clarified now.

Probably the best approach would be to allow storages to optionally accept a custom (underlying) storage client.
- If a storage client is provided, it should be used directly without interacting with the global service locator.
- If no storage client is provided, the global client from the service locator will be used.

This ensures we can further support the `force_cloud` option in the SDK without serious pain. The `Actor` will internally maintain an instance of `ApifyStorageClient`.

tests/unit/_memory_storage_client/test_memory_storage_client.py
```diff
@@ -224,13 +224,15 @@ async def test_not_implemented_method(tmp_path: Path) -> None:
 async def test_default_storage_path_used(monkeypatch: pytest.MonkeyPatch) -> None:
     # We expect the default value to be used
     monkeypatch.delenv('CRAWLEE_STORAGE_DIR', raising=False)
+    service_locator.set_configuration(Configuration())
```
janbuchar, line 227: Is this necessary? Shouldn't the `service_locator` be empty at the start of the test?

tests/unit/basic_crawler/test_basic_crawler.py
```diff
@@ -796,31 +796,16 @@ async def handler(context: BasicCrawlingContext) -> None:
...
-    assert isinstance(request_provider, RequestQueue)
-    assert request_provider._configuration is configuration
-
-
 async def test_respects_no_persist_storage() -> None:
-    configuration = Configuration(persist_storage=False)
-    crawler = BasicCrawler(configuration=configuration)
+    config = Configuration(persist_storage=False)
+    service_locator.set_configuration(config)
+    crawler = BasicCrawler()
```
janbuchar, line 802: I think the previous usage was fine and also more in line with what a user might do.

tests/unit/conftest.py
```diff
@@ -10,62 +10,113 @@
...
+        A callable that cleans up the test environment.
+    """
+
+    def _cleanup_test_env() -> None:
+        # Reset the flags in the service locator to indicate no services are explicitly set.
+        service_locator._service_locator._configuration_was_set = False
+        service_locator._service_locator._storage_client_was_set = False
+        service_locator._service_locator._event_manager_was_set = False
+
+    return _cleanup_test_env
```
janbuchar, line 88: Can't we just inline this in `_isolate_test_environment`? Or do we use this one on its own somewhere?

tests/unit/conftest.py
```diff
@@ -10,62 +10,113 @@
...
+    """
+    prepare_test_env()
+    yield
+    cleanup_test_env()
 
 
 @pytest.fixture
 def memory_storage_client(tmp_path: Path) -> MemoryStorageClient:
-    cfg = Configuration(
+    return MemoryStorageClient(
```
janbuchar, line 115: Maybe not, but shouldn't we also plug this one into the `service_locator` here?

janbuchar (commented):
I think we're getting there - here's a bunch of comments from my first read.

src/crawlee/storages/_key_value_store.py
```diff
@@ -92,6 +86,9 @@ async def open(
     ) -> KeyValueStore:
         from crawlee.storages._creation_management import open_storage
 
+        configuration = configuration or service_locator.get_configuration()
```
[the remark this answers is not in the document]
vdusek: Completely agree, but let's do it as separate PR.

src/crawlee/memory_storage_client/_memory_storage_client.py
```diff
@@ -45,13 +46,49 @@ class MemoryStorageClient(BaseStorageClient):
...
+        self,
+        configuration: Configuration | None = None,
+        *,
+        write_metadata: bool | None = None,
+        persist_storage: bool | None = None,
+        storage_dir: str | None = None,
+        default_request_queue_id: str | None = None,
+        default_key_value_store_id: str | None = None,
+        default_dataset_id: str | None = None,
+    ) -> None:
```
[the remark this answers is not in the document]
vdusek: I believe this isn't possible because the config relies on values from the env vars. Therefore, we should use the config values as defaults and override them with explicit arguments when provided.

src/crawlee/storages/_key_value_store.py
```diff
@@ -52,19 +53,12 @@ class KeyValueStore(BaseStorage):
...
     """
 
-    def __init__(
-        self,
-        id: str,
-        name: str | None,
-        configuration: Configuration,
-        client: BaseStorageClient,
-    ) -> None:
+    def __init__(self, id: str, name: str | None, storage_client: BaseStorageClient) -> None:
```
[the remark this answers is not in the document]
vdusek: `BaseStorage` do not have any dunder init method.

tests/unit/conftest.py
vdusek: Currently, we only use `prepare_test_env` in some tests. However, this approach is consistent, and it's possible that we might need to use `cleanup_test_env` in the future as well. I don't see any issue in the current state.

tests/unit/conftest.py
vdusek: I don't think this is necessary, as this fixture is intended for testing the memory storage client (and its resource clients). Maybe it doesn't belong here.

janbuchar: Huh, probably not :slightly_smiling_face:

vdusek: I'd say I can move the fixture somewhere to `tests/memory_storage_client`

src/crawlee/memory_storage_client/_memory_storage_client.py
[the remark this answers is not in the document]
janbuchar: Maybe I wasn't clear enough. I mean something like this:

```py
def __init__(
    self,
    configuration: Configuration | None = None,
    *,
    write_metadata: bool,
    persist_storage: bool,
    storage_dir: str,
    default_request_queue_id: str,
    default_key_value_store_id: str,
    default_dataset_id: str,
) -> None:
    pass

def from_configuration(cls, configuration: Configuration | None = None) -> Self:
    return cls(
        write_metadata=configuration.write_metadata,
        persist_storage=configuration.persist_storage,
        storage_dir=configuration.storage_dir,
        default_request_queue_id=configuration.default_request_queue_id,
        default_key_value_store_id=configuration.default_key_value_store_id,
        default_dataset_id=configuration.default_dataset_id,
    )
```

...and if someone wants to override `Configuration` values with whatever, they can use `Configuration.model_copy(update={...})`. Now as I'm writing it, I don't really like it that much. Maybe we could add something like `Configuration.replace()` as a shortcut.

src/crawlee/storages/_key_value_store.py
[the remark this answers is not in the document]
janbuchar: Right, I mistook it with `open` :facepalm:

tests/unit/_memory_storage_client/test_memory_storage_client.py
janbuchar: And couldn't we remove the configuration from `service_locator` and let the test trigger the "defaulting" on its own?

tests/unit/conftest.py
janbuchar: Well, it's one more fixture to maintain and I'm just not very sure about it being useful on its own.

vdusek: Ok, let's keep only `cleanup_test_env` fixture.

tests/unit/conftest.py
```diff
@@ -22,50 +23,68 @@
...
+
+        # Set the environment variable for the local storage directory to the temporary path.
         monkeypatch.setenv('CRAWLEE_STORAGE_DIR', str(tmp_path))
 
-        # Reset services in crawlee.service_container
-        cast(dict, service_container._services).clear()
+        # Initialize services in the service container with default values.
+        service_locator.set_configuration(Configuration())
+        service_locator.set_storage_client(MemoryStorageClient())
+        service_locator.set_event_manager(LocalEventManager())
```
janbuchar, line 53: I don't see why we can't set this to `None`?

src/crawlee/service_locator.py
```diff
@@ -0,0 +1,147 @@
...
+    @property
+    def event_manager_was_set(self) -> bool:
+        return self._event_manager_was_set
+
+    @property
+    def storage_client_was_set(self) -> bool:
+        return self._storage_client_was_set
+
+
+_service_locator = _ServiceLocator()
```
janbuchar, line 81: We discussed that we could just re-export this as `crawlee.service_locator` and rename this module to `crawlee._service_locator`. That way, we won't need the functions below.

vdusek:
> What about https://github.com/apify/apify-sdk-python/issues/324 - supporting crawler instantiation outside of the async with Actor block? I suppose it's not possible with this setup, right?

Correct. Could we address it later in a separate PR?

janbuchar:
> > What about [apify/apify-sdk-python#324](https://github.com/apify/apify-sdk-python/issues/324) - supporting crawler instantiation outside of the async with Actor block? I suppose it's not possible with this setup, right?
> 
> Correct. Could we address it later in a separate PR?

Absolutely. I guess that it'll require some internal restructuring of `BasicCrawler`, but it shouldn't be breaking.

src/crawlee/service_locator.py
vdusek: Done, much better, thanks.

src/crawlee/memory_storage_client/_memory_storage_client.py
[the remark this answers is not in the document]
vdusek: Done

vdusek:
We should be good here, feel free to do the final look. I'm working on the SDK adoption now -> https://github.com/apify/apify-sdk-python/pull/355.

tests/unit/test_configuration.py
```diff
@@ -1,7 +1,74 @@
...
+        is service_locator.get_configuration()
+        is service_locator.get_configuration()
+    )
+
+    assert (
+        service_locator.get_configuration()
+        is service_locator.get_configuration()
+        is Configuration.get_global_configuration()
+        is Configuration.get_global_configuration()
+    )
```
janbuchar, line 32:
```suggestion

def test_global_configuration_works_reversed() -> None:
    assert (
        service_locator.get_configuration()
        is service_locator.get_configuration()
        is Configuration.get_global_configuration()
        is Configuration.get_global_configuration()
    )
```

PR #697, opened 2024-11-14
fix: Accept string payload in the Request constructor

vdusek:
- This is a follow-up to https://github.com/apify/crawlee-python/issues/668#issuecomment-2476392555.
- Let's accept `payload` as both str and bytes in the `Request.from_url` constructor.
- The field `Request.payload` has a validator that accepts both. With this change, we are consistent in the `from_url` constructor as well.
- We do a similar thing for the `Request.headers` as well.


Mantisus:
LGTM

PR #711, opened 2024-11-19
chore: Use runtime-evaluated-base-classes for ruff

Pijukatel:
This allows to remove ruff exceptions, without breaking the ruff check. Also some exceptions were not necessary in the first place, so fix them.

See: https://docs.astral.sh/ruff/settings/#lint_flake8-type-checking_runtime-evaluated-base-classes


vdusek (approved):
lgtm

PR #713, opened 2024-11-20
docs: update `@apify/docusaurus-plugin-typedoc-api`

barjin:
Updating the plugin allows us to use the new unified transformation pipeline instead of handling the API docs generation separately in each Python project. 

The new version also now renders the Typedoc objects in machine-readable elements on the page, so we can use the new `reexports` plugin option (more about that in a PR in `apify-sdk-python`).


Makefile
```diff
@@ -49,7 +49,7 @@ build-api-reference:
 	cd website && poetry run ./build_api_reference.sh
 
 build-docs:
-	cd website && corepack enable && yarn && yarn build
+	cd website && corepack enable && yarn && poetry run yarn build
 
 run-docs: build-api-reference
-	cd website && corepack enable && yarn && yarn start
+	cd website && corepack enable && yarn && poetry run yarn start
```
barjin, line 55: Since the plugin now runs `pydoc-markdown` (and other python tooling) automatically, we need to run the Docusaurus build with the correct venv using `poetry run` (hopefully this is how it works).

website/docusaurus.config.js
```diff
@@ -78,9 +100,11 @@ module.exports = {
                     excludeExternals: false,
                 },
                 sortSidebar: groupSort,
-                pathToCurrentVersionTypedocJSON: `${__dirname}/api-typedoc-generated.json`,
                 routeBasePath: 'api',
-                python: true,
+                pythonOptions: {
+                    pythonModulePath: path.join(__dirname, '../src/crawlee'),
+                    moduleShortcutsPath: path.join(__dirname, 'module_shortcuts.json'),
+                },
```
barjin, line 107: The new unified API docs generating pipeline - we don't need to generate the json ourselves, just supply the path to the project (and the `module_shortcuts.json` file, which we might want to remove / integrate later, further simplifying the interface).

B4nan (approved):

PR #714, opened 2024-11-20
fix: Fix init of context managers and context handling in `BasicCrawler`

vdusek:
### Problems

- The `EventManager` could be initialized multiple times, such as when running a `Crawler` wrapped inside an `Actor`.
- In `crawler.run`, the async context was entered and exited directly, which could lead to issues if the caller had already entered it. This scenario might occur when users provide their own instances of `BrowserPool`, `SessionPool`, `EventManager`, or others.

### Solution

- Address these issues by introducing an `active` flag to the existing context managers to prevent multiple initializations.
- Implement an `ensure_context` helper and apply it to methods where context management is required.
- Fix & improve tests to ensure these cases are covered.

### Others

- I added missing names to asyncio tasks.

### DIFF
--- a/src/crawlee/_autoscaling/snapshotter.py
+++ b/src/crawlee/_autoscaling/snapshotter.py
@@ -18,6 +18,7 @@ from crawlee._autoscaling.types import (
     Snapshot,
 )
 from crawlee._utils.byte_size import ByteSize
+from crawlee._utils.context import ensure_context
 from crawlee._utils.docs import docs_group
 from crawlee._utils.recurring_task import RecurringTask
 from crawlee.events._types import Event, EventSystemInfoData
@@ -114,6 +115,9 @@ class Snapshotter:
 
         self._timestamp_of_last_memory_warning: datetime = datetime.now(timezone.utc) - timedelta(hours=1)
 
+        # Flag to indicate the context state.
+        self._active = False
+
     @staticmethod
     def _get_sorted_list_by_created_at(input_list: list[T]) -> SortedList[T]:
         return SortedList(input_list, key=attrgetter('created_at'))
@@ -126,12 +130,22 @@ class Snapshotter:
         logger.info(f'Setting max_memory_size of this run to {max_memory_size}.')
         return max_memory_size
 
+    @property
+    def active(self) -> bool:
+        """Indicates whether the context is active."""
+        return self._active
+
     async def __aenter__(self) -> Snapshotter:
         """Starts capturing snapshots at configured intervals."""
-        self._event_manager.on(event=Event.SYSTEM_INFO, listener=self._snapshot_cpu)
-        self._event_manager.on(event=Event.SYSTEM_INFO, listener=self._snapshot_memory)
-        self._snapshot_event_loop_task.start()
-        self._snapshot_client_task.start()
+        if self._active:
+            logger.warning(f'The {self.__class__.__name__} is already active.')
+        else:
+            self._active = True
+            self._event_manager.on(event=Event.SYSTEM_INFO, listener=self._snapshot_cpu)
+            self._event_manager.on(event=Event.SYSTEM_INFO, listener=self._snapshot_memory)
+            self._snapshot_event_loop_task.start()
+            self._snapshot_client_task.start()
+
         return self
 
     async def __aexit__(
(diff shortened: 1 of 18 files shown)


vdusek:
> Is this for use case where we do something like ?
>
> with Actor:
>     await some_crawler.run(...)
>
> Maybe adding some tiny test for such usecase would be good.

Yes, well, if we would like to have such a test, it would make sense to do it in SDK, as it extends the Crawlee.

janbuchar (approved):
LGTM, I would just change the title to e.g., "Prevent accidental multiple initialization of event manager"

src/crawlee/_autoscaling/snapshotter.py
```diff
@@ -145,11 +159,16 @@ async def __aexit__(
...
-        await self._snapshot_event_loop_task.stop()
-        await self._snapshot_client_task.stop()
+        if self._active:
+            self._event_manager.off(event=Event.SYSTEM_INFO, listener=self._snapshot_cpu)
+            self._event_manager.off(event=Event.SYSTEM_INFO, listener=self._snapshot_memory)
+            await self._snapshot_event_loop_task.stop()
+            await self._snapshot_client_task.stop()
+            self._active = False
+        else:
+            logger.warning(f'The {self.__class__.__name__} is not active.')
```
Pijukatel, line 169: I think this should raise an Exception. Once someone gets here, then they did something wrong.

tests/unit/sessions/test_session_pool.py
```diff
@@ -162,3 +162,27 @@ async def test_session_pool_persist_and_restore(event_manager: EventManager, kvs
...
+    with pytest.raises(RuntimeError, match='SessionPool is not active.'):
+        await session_pool.get_session()
+
+    with pytest.raises(RuntimeError, match='SessionPool is not active.'):
+        await session_pool.get_session_by_id(session.id)
+
+    await session_pool.reset_store()
+
+    async with session_pool:
+        assert session_pool.active is True
```
Pijukatel, line 188: I think that this is can lead to unexpected situations. Edit following tests like this to see. You will be still in outer context, but session_pool.active will be false. 
This implementation let's consumers of these context managers enter many times, but once single context has been left, all other context levels are no longer active. 

I think it is good idea to give consumers of those context managers "active" flag, but nothing else. Consumers should safeguard usage of such contexts themselves. When you are in the context, you do not want(users can decide) to enter it again, but when you are "twice" in the context and you leave the inner context, you still want(users can decide) to be in the outer context. 

```suggestion
    async with session_pool:
        async with session_pool:
            pass
        assert session_pool.active is True
        
```

src/crawlee/events/_event_manager.py
```diff
@@ -95,11 +101,13 @@ async def __aexit__(
 
         This will stop listening for the events, and it will wait for all the event listeners to finish.
         """
-        await self.wait_for_all_listeners_to_complete(timeout=self._close_timeout)
-        self._event_emitter.remove_all_listeners()
-        self._listener_tasks.clear()
-        self._listeners_to_wrappers.clear()
-        await self._emit_persist_state_event_rec_task.stop()
+        if self._initialized:
```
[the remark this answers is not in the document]
vdusek: I think this is resolved as now we raise an exception in case of multiple enters/exits.

tests/unit/sessions/test_session_pool.py
vdusek: I think this is resolved as now we raise an exception in case of multiple enters/exits.

src/crawlee/_utils/context.py
```diff
@@ -0,0 +1,40 @@
+from __future__ import annotations
+
+import asyncio
+from functools import wraps
+from typing import Any, Callable, TypeVar
+
+T = TypeVar('T', bound=Callable[..., Any])
+
+
+def ensure_context(method: T) -> T:
```
janbuchar, line 10: It would probably be hard to do so without straying even further from how the JS implementation works, but it may be easier to just do something like

```py
class EventManager:
  async def __aenter__(self) -> EventManagerContext:
    pass
  # ... no methods that need to be called on an initialized event manager

class EventManagerContext:
  def emit():
    pass
  # and whatever else needs an initialized context manager
```

... and so on for everything with an initialization step. Honestly, even `BasicCrawler` could benefit from this - the tracking of active state is kinda hairy there as well.

janbuchar (approved):
This looks like a good enough solution for the time being, especially if it fixes the tests. Thanks!

(see my comment :slightly_smiling_face:)

src/crawlee/_utils/context.py
vdusek: This sounds interesting; however, it would mean a more significant change. I'd prefer to approach it as a potential second step following a broader discussion. I will merge this for now, as it works and addresses the desired goals. Feel free to open an issue, and we can continue there, or on Slack.

PR #716, opened 2024-11-21
fix: Pydantic 2.10.0 issues

Pijukatel:
Lock pydantic to version ~2.9.0
New fresh version 2.10.0 has many unresolved issues that break our codebase.


B4nan (approved):

vdusek (commented):
Why are you updating the lockfile? Add an upper bound to Pydantic should be enough, correct?

Issue #630, opened 2024-10-29
Add a custom URL type instead of rely on `httpx.URL`


vdusek:
- We use the `httpx.URL` in the [ProxyConfiguration](https://github.com/apify/crawlee-python/blob/master/src/crawlee/proxy_configuration.py), in the tests and maybe in other places as well.
- It kind of shot us into the foot in #618.
- ~I suggest replacing it with our custom data type (either the Pydantic model or data class).~
    - As Honza said, we could probably utilize some 3rd party library, [yarl](https://pypi.org/project/yarl/) seems like a good option.
- Also please find other occurrences of the `httpx.URL` in the Crawlee and replace them with our custom type.
  - This applies also to tests: `httpbin: URL`
- Keep in mind, that URLs in the `Request` model need to be serialized as strings.

janbuchar:
A Pydantic type wouldn't probably help us much, those are typically just for validation/serialization.

janbuchar:
We might also want to use something like https://github.com/aio-libs/yarl

vdusek:
@Mantisus

janbuchar:
I experimented with `yarl` in #167 and found an issue - if you do `URL("https://some.tld/?asdf=hjkl") / "path"`, you lose the query string. This is unfortunate.

`httpx.URL` does not exhibit this behavior. I did not find any other usable (and maintained) URL-manipulation packages.

Mantisus:
This is fixed in `yarl` 1.18.0, released today.

In this regard, I think `yarl` is optimal for this purpose

vdusek:
The issue https://github.com/aio-libs/yarl/issues/111 is from 2017, and was resolved just a few hours ago, with a release that dropped 30 minutes ago. It feels like destiny.

PR #729, opened 2024-11-22
refactor: change httpx.URL to yarl.URL

Mantisus:
### Description

- Replaces `httpx.URL` with `yarl.URL` in proxy_configuration and tests
- URL manipulation utilities updated to use `yarl.URL`
- Add support `yarl.URL` in `Request.from_url`

### Issues

- Closes: #630 

### Testing

Moving to `yarl.URL` for some `utils` required updating the expected behavior for existing tests.

Since some things allowed for `urllib.parse` are not allowed for `yarl` which tries to strictly adhere to RFC 3986

Example:

'https://crawlee.dev#fragment' is allowed for `urllib.parse`
But `yarl` converts this to 'https://crawlee.dev/#fragment', because `path` cannot be empty if there is a `query` or a `fragment`

### Checklist

- [ ] CI passed


src/crawlee/_utils/urls.py
```diff
@@ -1,25 +1,31 @@
...
 
 
 def convert_to_absolute_url(base_url: str, relative_url: str) -> str:
     """Convert a relative URL to an absolute URL using a base URL."""
-    return urljoin(base_url, relative_url)
+    return str(URL(base_url).join(URL(relative_url)))
 
 
-def extract_query_params(url: str) -> dict[str, list[str]]:
+def extract_query_params(url: str) -> MultiDictProxy[str]:
```
janbuchar, line 25: I'd argue this function could be removed - `extract_query_params(url)` does not seem to have an advantage over `URL(url).query` IMO.

tests/unit/conftest.py
janbuchar, line 1: Crazy idea, but couldn't we extend `yarl.URL` **only** for the `httpbin` fixture and override `with_path` and `__truediv__` so that it always uses `keep_query=True` and `keep_fragment=True`? It would make tests quite a bit more readable.

tests/unit/proxy_configuration/test_new_proxy_info.py
```diff
@@ -32,7 +31,7 @@ async def test_throws_on_invalid_new_url_function() -> None:
     with pytest.raises(ValueError) as exc:  # noqa: PT011
         await config.new_proxy_info(None, None, None)
 
-    assert isinstance(exc.value.__cause__, InvalidURL)
+    assert isinstance(exc.value.__cause__, ValueError)
```
janbuchar, line 34: I believe that check can be removed - the `pytest.raises` three lines up ensures the same thing.

janbuchar (commented):
This is super nice, thanks!

tests/unit/conftest.py
Mantisus: I was thinking about that too.

But unfortunately we can't do it, `yarl.URL` uses `__slots__` and the prohibition to create subclasses. As a result we can't create our own extended subclass or override methods of a created object.

https://github.com/aio-libs/yarl/blob/master/yarl/_url.py#L438

src/crawlee/_request.py
```diff
@@ -275,7 +278,7 @@ class Request(BaseRequestData):
     @classmethod
     def from_url(
         cls,
-        url: str,
+        url: str | URL,
```
vdusek, line 281: To be honest, I'm not a fan of this approach. It feels like a slippery slope—why not accept even more types and simply convert them to strings here? Unless there's a particular reason to accept `yarl.URL`, I would stick to accepting just strings and let users explicitly call it with the conversion like `Request.from_url(str(my_url))`.

src/crawlee/_utils/requests.py
```diff
@@ -55,33 +56,21 @@ def normalize_url(url: str, *, keep_url_fragment: bool = False) -> str:
         A string containing the normalized URL.
     """
     # Parse the URL
-    parsed_url = urlparse(url.strip())
-    search_params = dict(parse_qsl(parsed_url.query))  # Convert query to a dict
+    parsed_url = URL(url.strip())
```
vdusek, line 59: Is the `strip` call necessary?

src/crawlee/proxy_configuration.py
```diff
@@ -124,19 +123,19 @@ async def new_proxy_info(
         if url is None:
             return None
 
-        # httpx.URL port field is None for default ports
-        default_ports = {'http': 80, 'https': 443}
-        port = url.port or default_ports.get(url.scheme)
-        if port is None:
+        if url.port is None:
```
vdusek, line 126: Just to be sure, so `yarl` returns e.g. `80` for `http://crawlee.dev`, correct?

src/crawlee/proxy_configuration.py
```diff
@@ -124,19 +123,19 @@ async def new_proxy_info(
...
         info = ProxyInfo(
             url=str(url),
             scheme=url.scheme,
             hostname=url.host,
-            port=port,
-            username=url.username,
-            password=url.password,
+            port=url.port,
+            username=url.user if url.user else '',
+            password=url.password if url.password else '',
```
vdusek, line 138: not a request, but we can use this:

```suggestion
            username=url.user or '',
            password=url.password or '',
```

which is shorter

tests/unit/conftest.py
janbuchar: Unfortunate. Can we monkeypatch those methods in the `httpbin` fixture then? :smiling_imp:

tests/unit/_utils/test_requests.py
```diff
@@ -44,18 +44,18 @@ def test_unique_key_to_request_id_matches_known_values(unique_key: str, expected
 @pytest.mark.parametrize(
     ('url', 'expected_output', 'keep_url_fragment'),
     [
-        ('https://example.com/?utm_source=test&utm_medium=test&key=value', 'https://example.com?key=value', False),
+        ('https://example.com/?utm_source=test&utm_medium=test&key=value', 'https://example.com/?key=value', False),
```
vdusek, line 47: Do we have to add the slashes here?

vdusek (changes requested):
otherwise LGTM, thanks!

vdusek:
Also, I'm thinking about the PR type as I believe this should be in the changelog.

src/crawlee/_utils/requests.py
Mantisus: Yep, `yarl` does not remove spaces at the end of a string

src/crawlee/proxy_configuration.py
Mantisus: That's right.

80 for http://crawlee.dev
445 for http://crawlee.dev:445

tests/unit/_utils/test_requests.py
Mantisus: Yep, `yarl` does not allow `path` to be empty when parameters or fragment is present.

janbuchar:
> Also, I'm thinking about the PR type as I believe this should be in the changelog.

Well, the only user-facing change is that `Request.from_url` accepts a `yarl.URL`, and we may not want to keep that in the end...

But there is the additional dependency, I guess it would be fair to mention that... Not sure which conventional commit type suits that though :smile:

Mantisus:
> Not sure which conventional commit type suits that though

Maybe `chore!`?

vdusek:
> Maybe chore!?
The exclamation mark is for the breaking changes.

I would use `refactor` and we should consider whether or not to include them into changelogs. I would be for it.

tests/unit/basic_crawler/test_basic_crawler.py
```diff
@@ -536,9 +536,9 @@ async def handler(context: BasicCrawlingContext) -> None:
         seen_urls.append(context.request.url)
 
     start_urls = [
-        str(httpbin.copy_with(path='/1')),
-        str(httpbin.copy_with(path='/2')),
-        str(httpbin.copy_with(path='/3')),
+        str(httpbin / '/1'),
```
janbuchar, line 539: nit - the leading slashes in the path should not be necessary

tests/unit/conftest.py
```diff
@@ -69,7 +69,18 @@ def memory_storage_client(tmp_path: Path) -> MemoryStorageClient:
...
+    original_with_path = URL.with_path
+
+    def mock_with_path(self: URL, path: str, *, keep_query: bool = True, keep_fragment: bool = True) -> URL:
+        return original_with_path(self, path, keep_query=keep_query, keep_fragment=keep_fragment)
+
+    def mock_truediv(self: URL, path: str) -> URL:
+        return self.with_path(path)
+
+    monkeypatch.setattr(URL, 'with_path', mock_with_path)
+    monkeypatch.setattr(URL, '__truediv__', mock_truediv)
```
janbuchar, line 82: This potentially breaks other code that uses `yarl.URL`. I'd just patch the result url (`URL(os.environ.get('HTTPBIN_URL', 'https://httpbin.org'))`).

janbuchar (changes requested):

tests/unit/conftest.py
janbuchar: If it's too much trouble, then I guess we can take that risk :slightly_smiling_face: But in the end, it should be possible to just assign the methods to the resulting object.

janbuchar (approved):

Issue #719, opened 2024-11-21
Expose the `browser_options` to the PlaywrightCrawler


vdusek:
- In the same way as we do it for `headless` and `broser_type`.
- https://playwright.dev/python/docs/api/class-browsertype#browser-type-launch
- Also improve arg docs, add links to the PW docs

Mantisus:
It seems to me that in `PlaywrightBrowserPlugin`[https://crawlee.dev/python/api/class/PlaywrightBrowserPlugin] documentation it is enough to specify that `browser_options` supports all parameters supported by `Playwright launch`[https://playwright.dev/python/docs/api/class-browsertype#browser-type-launch].

Now it is not clear without looking at the source code. But adding all parameters to `PlaywrightCrawler` looks like an overhead to me.

PR #730, opened 2024-11-22
feat: Expose browser_options and page_options to PlaywrightCrawler

vdusek:
Closes: #719

### DIFF
--- a/src/crawlee/playwright_crawler/_playwright_crawler.py
+++ b/src/crawlee/playwright_crawler/_playwright_crawler.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING, Awaitable, Callable
+from typing import TYPE_CHECKING, Any, Awaitable, Callable, Mapping
 
 from pydantic import ValidationError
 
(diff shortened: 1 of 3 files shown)


src/crawlee/playwright_crawler/_playwright_crawler.py
```diff
@@ -80,20 +82,35 @@ def __init__(
...
+            browser_options: Keyword arguments to pass to the browser launch method.
+                This option should not be used if `browser_pool` is provided.
+            page_options: Keyword arguments to pass to the new page method.
+                This option should not be used if `browser_pool` is provided.
             headless: Whether to run the browser in headless mode.
                 This option should not be used if `browser_pool` is provided.
             kwargs: Additional keyword arguments to pass to the underlying `BasicCrawler`.
         """
         if browser_pool:
             # Raise an exception if browser_pool is provided together with headless or browser_type arguments.
```
Pijukatel, line 94: This comment was not updated to reflect change in the code.

src/crawlee/playwright_crawler/_playwright_crawler.py
```diff
@@ -80,20 +82,35 @@ def __init__(
...
-            if headless is not None or browser_type is not None:
+            if (
```
Pijukatel, line 95: With many conditions like this you could consider using **if any**(headless, browser_type, browser_options, page_options)

Pijukatel: I guess that:
 if any(param is not None for param in (headless, browser_type, browser_options, page_options))
insetad of 
if any(headless, browser_type, browser_options, page_options) 
is there because of mypy want's explisit is not none

Pijukatel (approved):

janbuchar (approved):
I'm not sure if I have enough context, but the code looks solid.

PR #757, opened 2024-11-28
fix: Remove pydantic constraint <2.10.0 and update timedelta validator, serializer type hints

Pijukatel:
Seems like new 2.10.0 version made some changes that made our type hints cause errors. On the other hand, it seems our type hints were too generic. Making them more specific fixes the issue in new pydantic versions.


src/crawlee/_utils/models.py
```diff
@@ -22,7 +22,9 @@ def _timedelta_to_ms(td: timedelta | None) -> Any:
 _number_parser = TypeAdapter(float)
 
 
-def _timedelta_from_ms(value: float | timedelta | Any | None, handler: Callable[[Any], Any]) -> Any:
+def _timedelta_from_ms(
+    value: float | timedelta | Any | None, handler: Callable[[Any], timedelta | None]
+) -> timedelta | None:
```
Pijukatel, line 27: Same as _timedelta_to_ms comment

src/crawlee/_utils/models.py
```diff
@@ -9,7 +9,7 @@
 """Utility types for Pydantic models."""
 
 
-def _timedelta_to_ms(td: timedelta | None) -> Any:
+def _timedelta_to_ms(td: timedelta | None) -> float | None:
```
Pijukatel, line 12: We could be more strict and require float, but I have a feeling, that None could be used for something connected with defaults somewhere in code???

vdusek: I guess we can consider using only floats.

However, in `StatisticsState` and `StatisticsPersistedState`, we have fields like this:

```python
request_min_duration: Annotated[timedelta_ms | None, Field(alias='requestMinDurationMillis')] = None
```

Make sure it won't break them.

Pijukatel: I tried adding some tests (but not sure whether I should keep them) and I think it works as expected. I am not really so sure why the previous implementation allowed **any** in so many places in this file, so I am little afraid there is some non obvious edge case I could not think of...

vdusek (approved):
LGTM, but let's wait for Honza's approval as well

tests/unit/_statistics/test_statistics_models.py
janbuchar, line 1: I appreciate testing `timedelta_ms`, but there's no need to pull `StatisticsState` with it - I'd just define a new model in the test file and test that.

src/crawlee/_utils/models.py
```diff
@@ -9,20 +9,17 @@
...
-        return td
-
     return int(round(td.total_seconds() * 1000))
 
 
 _number_parser = TypeAdapter(float)
 
 
-def _timedelta_from_ms(value: float | timedelta | Any | None, handler: Callable[[Any], Any]) -> Any:
+def _timedelta_from_ms(value: float | timedelta | Any | None, handler: Callable[[Any], timedelta]) -> timedelta | None:
```
janbuchar, line 22: Was this the problematic type hint? :slightly_smiling_face:

janbuchar (changes requested):
Just some minor things, otherwise it's great!

src/crawlee/_utils/models.py
Pijukatel: I think it is still Pydantic bug, somewhere down there. I did not spent effort on trying to figure it out once I saw it is working well with more specific model.
I can reproduce it if I use new Pydantic version and for example in the new tests I created I do not import **Any** in the test file. It will then fail on importing **Any** even if I do not use it in the test files and it is needed only in model file.

Pijukatel: It basically goes all the way to problem with the way annotations are evaluated: https://github.com/pydantic/pydantic/issues/2678

When you use **from __future__ import annotations** , it can break pydantic :D in those files

janbuchar: That sounds scary, future annotations should be the norm 😃 can you elaborate a bit about what the problem is?

src/crawlee/_utils/models.py
Pijukatel: Well that can also break pydantic. See new commit that allows | None, but has to use Union due to known issues that Pydantic can have with from __future__ import annotations and | symbol not being available for older Python versions in type hints...

pyproject.toml
```diff
@@ -161,6 +161,10 @@ indent-style = "space"
     "F841",   # Local variable {variable} is assigned to but never used
     "N999",   # Invalid module name
 ]
+"src/crawlee/_utils/models.py" = [
+    "FA100",  # In order to be able to use | instead of Union for type hints in Pythons<3.10 we need to do from __future__ import annotations. This import can break pydantic in unxpecetd ways.
+]
+
```
janbuchar, line 167: I'd rather put that at the start of the file... this seems kinda fragile.

tests/unit/_utils/test_timedelata_ms.py
janbuchar, line 1: While you're at it, could you also test infinity and some normal value?

janbuchar (changes requested):
Probably the last two nits :slightly_smiling_face:

src/crawlee/_utils/models.py
Pijukatel: I do not think I can elaborate. This requires knowledge I do not posses. Let's wait with this then and see if Pydantic people can explain it: https://github.com/pydantic/pydantic/issues/11004

tests/unit/_utils/test_timedelata_ms.py
Pijukatel: Done

pyproject.toml
Pijukatel: Pydantic people solved the issue I reported. So I will just wait for thew new release and then this will not longer be needed.

tests/unit/_utils/test_timedelata_ms.py
```diff
@@ -0,0 +1,35 @@
...
+    [
+        (1.0, timedelta(milliseconds=1), 1),
+        (1, timedelta(milliseconds=1), 1),
+        ('1', timedelta(milliseconds=1), 1),
+        (timedelta(milliseconds=1), timedelta(milliseconds=1), 1),
+        (3.01, timedelta(microseconds=3010), 3),
+        (3.5, timedelta(microseconds=3500), 4),
+        (3.99, timedelta(microseconds=3990), 4),
+        (None, None, None),
+        (float('inf'), timedelta(days=999999999, seconds=3600 * 24 - 1, microseconds=999999), float('inf')),
```
Pijukatel, line 27: I considered 3600 * 24 - 1 to be more readable than 86399. Since it is just a test, I avoided defining SECONDS_IN_HOUR and HOURS_IN_DAYS ....

janbuchar: Thanks for avoiding that :slightly_smiling_face: I usually do `timedelta(days=3, hours=8).total_seconds()` instead.

src/crawlee/_utils/models.py
janbuchar: Are you sure that the pipe syntax will work with Pydantic 3.9, for example?

Issue #687, opened 2024-11-12
`tieredProxyUrls` accept `null` for switching the proxy off


barjin:
This is a parity-tracking issue for Crawlee JS / Crawlee Python.

With https://github.com/apify/crawlee/pull/2743 , we enabled `ProxyConfiguration.tieredProxyUrls` to accept `null` as a command to switch the proxy off on that tier. In Crawlee JS, this was only a matter of widening the types and making the validation less strict - `ProxyConfiguration` was already prepared for `null` URLs (returning `null` from newUrlFunction - this seems to work in Crawlee for Python too).

PR #760, opened 2024-11-28
feat: Add possibility to use None as no proxy in tiered proxies

Pijukatel:
Using None in input lists for tiered proxy configuration will be interpreted as no proxy. This allows to have for example lowest tier without proxy.

Add tests for it.
Update docs.

- Closes: https://github.com/apify/crawlee-python/issues/687


src/crawlee/proxy_configuration.py
```diff
@@ -85,7 +85,7 @@ def __init__(
                 the proxy selection mechanism.
         """
         self._next_custom_url_index = 0
-        self._used_proxy_urls = dict[str, URL]()
+        self._used_proxy_urls = dict[str, Union[URL, None]]()
```
Pijukatel, line 88: Mypy was very aggressive when I wanted to use **|**  on this specific line. I got bullied into using Union

barjin:
Looking good to me 👏🏽  (I'm glad to see this really was as simple as in the JS version). Have you tested this locally with a crawler (i.e. won't other parts of Crawlee mind the `None` instead of a proxy URL?) Do we already support this case with `new_url_function` returning `None`?

Pijukatel:
> Looking good to me 👏🏽 (I'm glad to see this really was as simple as in the JS version). Have you tested this locally with a crawler (i.e. won't other parts of Crawlee mind the `None` instead of a proxy URL?) Do we already support this case with `new_url_function` returning `None`?

I tested running 
```
...
crawler = BeautifulSoupCrawler(
            proxy_configuration=ProxyConfiguration(tiered_proxy_urls=[[None, None], [None]]),
        )
await crawler.run(["https://crawlee.dev"])
...
```

I am not sure what you mean by this. Is there a usecase example?
"Do we already support this case with `new_url_function` returning `None`?"

Pijukatel:
> > Looking good to me 👏🏽 (I'm glad to see this really was as simple as in the JS version). Have you tested this locally with a crawler (i.e. won't other parts of Crawlee mind the `None` instead of a proxy URL?) Do we already support this case with `new_url_function` returning `None`?
> 
> I tested running
> 
> ```
> ...
> crawler = BeautifulSoupCrawler(
>             proxy_configuration=ProxyConfiguration(tiered_proxy_urls=[[None, None], [None]]),
>         )
> await crawler.run(["https://crawlee.dev"])
> ...
> ```
> 
> I am not sure what you mean by "**Do we already support this case with `new_url_function` returning `None`?"**. Is there a usecase example?

Ok I think I get it. I looked only at tiered proxies. Currently this will still not work:

```
crawler = BeautifulSoupCrawler( proxy_configuration=ProxyConfiguration(proxy_urls=[None, None]),)
await crawler.run(["https://crawlee.dev"])
```
I will make it work consistently for all possible ways of setting the proxies to None

barjin:
Right, sorry, I might have been confusing - in JS version, we support the `newUrlFunction` (alternative of [`new_url`](https://crawlee.dev/python/api/class/ProxyConfiguration#new_url)) returning `null` to turn the proxy off. I'm actually not sure if you can pass `null` in `proxyUrls` (although it would make sense).

obrazek

> I will make it work consistently for all possible ways of setting the proxies to None

Sounds good! (and hopefully not too much work 🤞🏽 )

src/crawlee/proxy_configuration.py
vdusek: Yeah, AFAIK in generics the pipe is still not possible

tests/unit/proxy_configuration/test_tiers.py
```diff
@@ -94,3 +94,43 @@ async def test_successful_request_makes_tier_go_down() -> None:
...
+    config = ProxyConfiguration(tiered_proxy_urls=tiered_proxy_urls)
+
+    # Calling `new_proxy_info` with the same request most probably means it's being retried
+    request_1 = Request(url='http://some.domain/abc', unique_key='1', id='1')
+
+    # No proxy used.
+    info = await config.new_proxy_info(None, request_1, None)
+    assert info is None
+
+    info = await config.new_proxy_info(None, request_1, None)
```
vdusek, line 114: Could you add a comment here explaining what is expected to happen? e.g., on the second call of the same request, the tier increases

(it can be either a comment or assert message)

tests/unit/proxy_configuration/test_tiers.py
```diff
@@ -94,3 +94,43 @@ async def test_successful_request_makes_tier_go_down() -> None:
...
+    info = await config.new_proxy_info(None, None, None)
+    assert info is None
+
+    info = await config.new_proxy_info(None, None, None)
+    assert info is not None
+    assert info.url == tiered_proxy_urls[0][1]
+
+    # No proxy used.
+    info = await config.new_proxy_info(None, None, None)
+    assert info is None
```
vdusek, line 136: As with the previous test, please document what is expected to happen, either in the assert message or as a comment.

vdusek (approved):
Just one note about the test documentation; otherwise, it looks good. Thanks.

tests/unit/proxy_configuration/test_tiers.py
Pijukatel: Done

tests/unit/proxy_configuration/test_tiers.py
Pijukatel: Done

PR #761, opened 2024-11-29
chore: Use specific rule codes when ignoring type issues

vdusek:
- Rm ignore of `PGH003` rule and update type ignores to be specific.
- Fix 2 asserts in the tests.
- Use underscore names for unused variables in tests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -121,7 +121,6 @@ ignore = [
     "G004",     # Logging statement uses f-string
     "ISC001",   # This rule may cause conflicts when used with the formatter
     "FIX",      # flake8-fixme
-    "PGH003",   # Use specific rule codes when ignoring type issues
     "PLR0911",  # Too many return statements
     "PLR0913",  # Too many arguments in function definition
     "PLR0915",  # Too many statements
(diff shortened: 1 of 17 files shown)


pyproject.toml
```diff
@@ -208,11 +204,13 @@ exclude = [
...
-[tool.mypy-scrapy]
-ignore_missing_imports = true
-
-[tool.mypy-sortedcollections]
-ignore_missing_imports = true
+[tool.basedpyright]
+pythonVersion = "3.9"
+typeCheckingMode = "standard"
+include = ["src", "tests", "docs"]
+ignore = [
```
Pijukatel, line 211: I tried removing this in another PR and it was ok. Not sure why it is ignored:

https://github.com/apify/crawlee-python/pull/760/commits/2721cc5cc7cce3e9ab478a2c7b1bdae7eb753dfc#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711

Pijukatel (approved):
One super tiny comment

pyproject.toml
vdusek: Let's remove it then, thanks.

janbuchar: That was mine :smile:

PR #763, opened 2024-11-29
chore: Remove unused ID arg from `BaseRequestData.from_url`

vdusek:
- `BaseRequestData` does not have the `id` attribute.
- It is not breaking, as 
  - `from_url` accepts `**kwargs`, 
  - and `BaseRequestData` is not exposed to the public.


janbuchar (approved):
Thanks!

PR #773, opened 2024-12-03
feat: Add metadata parameter to Request for additional context

CupOfGeo:
### Description
This PR allows you to hold metadata in the request object. the idea is that you can pull in from the context.request object later in the router. 

I could think of a few ways this would be used. In my case I'm scraping an api and I can't garnet the url / ids don't change so I want my own internal id that I want to pass along to when I persist the results. 

Someone else had a similar desire to this https://github.com/apify/crawlee-python/discussions/531 thread. 

Please let me know what y'all think. if you want me to add any more documentation for it anywhere. I have a solution that I can continue to use without this PR but i see this as a small thing I can do. Thanks for keeping it open source. 

### Testing

- TODO

### Checklist

- [ ] CI passed

### DIFF
--- a/src/crawlee/_request.py
+++ b/src/crawlee/_request.py
@@ -282,6 +282,7 @@ class Request(BaseRequestData):
         keep_url_fragment: bool = False,
         use_extended_unique_key: bool = False,
         always_enqueue: bool = False,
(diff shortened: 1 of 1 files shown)


CupOfGeo:
Nope thank you closing.

Issue #248, opened 2024-06-28
Sitemap-based request provider


janbuchar:
similar to what we're implementing in JS crawlee

vdusek:
@Mantisus this is resolved, am I correct?

Mantisus:
> @Mantisus this is resolved, am I correct?

Yep

Issue #99, opened 2024-04-10
Properly implement `RequestList`


janbuchar:
The current implementation is very basic and mostly serves for testing. We should make it more like https://github.com/apify/crawlee/blob/master/packages/core/src/storages/request_list.ts

it should

- support loading remote request lists in the background
- have persistent state

vdusek:
Closing, as this was addressed in https://github.com/apify/crawlee-python/pull/777.

janbuchar:
I beg to differ, it still doesn't have persistence

PR #777, opened 2024-12-03
feat!: Implement `RequestManagerTandem`, remove `add_request` from `RequestList`, accept any iterable in `RequestList` constructor

janbuchar:
> Tandem, or in tandem, is an arrangement in which two or more animals, machines, or people are lined up one behind another, all facing in the same direction.[[1]](https://en.wikipedia.org/wiki/Tandem#cite_note-OED-1) Tandem can also be used more generally to refer to any group of persons or objects working together, not necessarily in line.[[1]](https://en.wikipedia.org/wiki/Tandem#cite_note-OED-1)
(https://en.wikipedia.org/wiki/Tandem)

- Inspired by https://github.com/apify/crawlee/blob/4c95847d5cedd6514620ccab31d5b242ba76de80/packages/basic-crawler/src/internals/basic-crawler.ts#L1154-L1177 and related code in the same class
- In my opinion, it implements the feature more cleanly and without polluting `BasicCrawler` (...any further)
- The motivation for the feature is twofold:
    1. Apify Actor development - it is common that an Actor receives a `requestListSources` input from the user, which may be pretty complex (regexp-based extraction from remote URL lists), and which is usually parsed using `apify.RequestList.open`. At the same time, the Actor wants to use the built in `RequestQueue`.
    2. Sitemap parsing (#248) - similar to 1, but not coupled to the Apify platform - we want to read URLs from a sitemap in the background, but the URLs should go through the standard request queue

## Breaking changes

- `RequestList` does not support `.drop()`, `.reclaim_request()`, `.add_request()` and `add_requests_batched()` anymore
   - `RequestManagerTandem` with a `RequestQueue` should be used for this use case, `await list.to_tandem()` can be used as a shortcut
- The `RequestProvider` interface has been renamed to `RequestManager` and moved to the `crawlee.request_loaders` package
- `RequestList` has been moved to the `crawlee.request_loaders` package
- The `BasicCrawler.get_request_provider` method has been renamed to `BasicCrawler.get_request_manager` and it does not accept the `id` and `name` arguments anymore
- The `request_provider` parameter of `BasicCrawler.__init__` has been renamed to `request_manager`
 
## TODO
- [x] new tests
- [x] fix existing tests


src/crawlee/storages/_request_list.py
```diff
@@ -91,37 +85,3 @@ async def mark_request_as_handled(self, request: Request) -> None:
     @override
     async def get_handled_count(self) -> int:
         return self._handled_count
-
-    @override
-    async def add_request(
```
janbuchar, line 96: I'm pretty positive that removing `add_request` from `RequestList` is a good idea - the main purpose of the class is handling the `requestListSources` input from Apify anyway. Currently it is just an inefficient version of `RequestQueue`.

However, we may want to make a wrapper for `RequestSource` that just raises an error in `add_request`, so that crawlers that do not enqueue anything by design are possible. Thoughts?

src/crawlee/storages/_request_source.py
```diff
@@ -0,0 +1,84 @@
...
+
+        raise ValueError(f'Invalid request type: {type(request)}')
+
+    def _transform_requests(self, requests: Sequence[str | Request]) -> list[Request]:
+        """Transforms a list of request-like objects into a list of Request objects."""
+        processed_requests = dict[str, Request]()
+
+        for request in requests:
+            processed_request = self._transform_request(request)
+            processed_requests.setdefault(processed_request.unique_key, processed_request)
```
Pijukatel, line 82: I see that this was not changed, just moved. But why:
processed_requests.setdefault(processed_request.unique_key, processed_request)
and not just 
processed_requests[processed_request.unique_key]=processed_request
?

Does it mean, that the processed_request.**unique_key** is not so unique and can be repeated?

janbuchar: It may not be unique in the `requests` argument, so we deduplicate it like this.

tests/unit/storages/test_request_source_tandem.py
```diff
@@ -0,0 +1,59 @@
...
+        'http://b.com',
+        'http://c.com',
+    }
+
+    assert await tandem.get_handled_count() == 3
+    assert await request_list.get_handled_count() == 2
+    assert await request_queue.get_handled_count() == 3
+
+
+async def test_wait_for_read_only_source() -> None:
```
Pijukatel, line 35: Maybe add docstring about what are we testing here. To me it seems very similar to the previous test and I am not really sure what code change can break this test and not the previous one.

src/crawlee/basic_crawler/_basic_crawler.py
```diff
@@ -361,15 +361,10 @@ async def _get_proxy_info(self, request: Request, session: Session | None) -> Pr
...
-        self,
-        *,
-        id: str | None = None,
-        name: str | None = None,
-    ) -> RequestProvider:
+    async def get_request_provider(self) -> RequestProvider:
         """Return the configured request provider. If none is configured, open and return the default request queue."""
         if not self._request_provider:
-            self._request_provider = await RequestQueue.open(id=id, name=name, configuration=self._configuration)
+            self._request_provider = await RequestQueue.open(configuration=self._configuration)
```
vdusek, line 367: Why are we removing the `id` and `name` args?

src/crawlee/storages/__init__.py
```diff
@@ -2,5 +2,14 @@
 from ._key_value_store import KeyValueStore
```
vdusek, line 2: I am also wondering about the structure, taking into account the new inheritance chain. Perhaps it could look something like this:

```
$ ls -1 src/crawlee/storages/
_base_storage.py
_dataset.py  
_key_value_store.py        
_request_queue.py
```

```
$ ls -1 src/crawlee/request_sources/
_request_list.py      
_request_source_tandem.py
_request_provider.py  
_request_source.py
```

Opinions?

src/crawlee/storages/_request_source.py
```diff
@@ -0,0 +1,84 @@
...
+    Request providers are used to manage and provide access to a storage of crawling requests.
+
+    Key responsibilities:
+        - Fetching the next request to be processed.
+        - Reclaiming requests that failed during processing, allowing retries.
+        - Marking requests as successfully handled after processing.
+        - Managing state information such as the total and handled request counts.
+
+    Subclasses of `RequestSource` should provide specific implementations for each of the abstract methods.
+    """
```
vdusek, line 28: Request providers?

src/crawlee/storages/_request_source.py
```diff
@@ -0,0 +1,84 @@
...
+        if isinstance(request, Request):
+            return request
+
+        if isinstance(request, str):
+            return Request.from_url(request)
+
+        raise ValueError(f'Invalid request type: {type(request)}')
+
+    def _transform_requests(self, requests: Sequence[str | Request]) -> list[Request]:
+        """Transforms a list of request-like objects into a list of Request objects."""
```
vdusek, line 77:
```suggestion
        """Transforms a list of request-like objects into a list of `Request` objects."""
```

src/crawlee/storages/_request_source.py
```diff
@@ -0,0 +1,84 @@
+from __future__ import annotations
```
vdusek, line 1: You are talking about the request queue in the docstrings (I'm talking about the whole file)

src/crawlee/storages/_request_source_tandem.py
```diff
@@ -0,0 +1,109 @@
...
+class RequestSourceTandem(RequestProvider):
+    """Implements a tandem behaviour for a pair of `RequestSource` and `RequestProvider`.
+
+    In this scenario, the contents of the "source" get transferred into the "provider", allowing processing the requests
+    from both sources and also enqueueing new requests (not possible with plain `RequestSource`).
+    """
+
+    def __init__(self, request_source: RequestSource, request_provider: RequestProvider) -> None:
+        self._read_only_source = request_source
+        self._read_write_provider = request_provider
```
vdusek, line 31: I don't know, but the fact that we both inherit from `RequestProvider` and also accept it as an input argument suggests there might be a design issue. Isn't it?

src/crawlee/storages/_request_source_tandem.py
```diff
@@ -0,0 +1,109 @@
...
+
+    @property
+    @override
+    def name(self) -> str:
+        return ''
```
vdusek, line 36: so it does not implement the interface as a whole

tests/unit/basic_crawler/test_basic_crawler.py
```diff
@@ -43,8 +46,26 @@ async def handler(context: BasicCrawlingContext) -> None:
...
+        calls.add(context.request.url)
+
+    await crawler.run()
+
+    assert calls == {'http://a.com/', 'http://b.com/', 'http://c.com/', 'http://d.com', 'http://e.com'}
+
+
 async def test_processes_requests_from_run_args() -> None:
-    crawler = BasicCrawler(request_provider=RequestList())
+    crawler = BasicCrawler(request_provider=await RequestQueue.open())
```
vdusek, line 68: cannot we use crawler's default request provider in this case?

tests/unit/basic_crawler/test_basic_crawler.py
```diff
@@ -57,7 +78,7 @@ async def handler(context: BasicCrawlingContext) -> None:
 
 
 async def test_allows_multiple_run_calls() -> None:
-    crawler = BasicCrawler(request_provider=RequestList())
+    crawler = BasicCrawler(request_provider=await RequestQueue.open())
```
vdusek, line 81: cannot we use crawler's default request provider in this case?

tests/unit/basic_crawler/test_basic_crawler.py
```diff
@@ -78,7 +99,9 @@ async def handler(context: BasicCrawlingContext) -> None:
 
 
 async def test_retries_failed_requests() -> None:
-    crawler = BasicCrawler(request_provider=RequestList(['http://a.com/', 'http://b.com/', 'http://c.com/']))
+    request_queue = await RequestQueue.open()
+    await request_queue.add_requests_batched(['http://a.com/', 'http://b.com/', 'http://c.com/'])
+    crawler = BasicCrawler(request_provider=request_queue)
```
vdusek, line 104: Is it important for these tests to use explicit request provider? Can't we use just pass the requests to the `run`?

vdusek (commented):
Also, could you please add this to the new section in the "Request storage" guide?

src/crawlee/basic_crawler/_basic_crawler.py
janbuchar: Because those only apply to `RequestQueue`, and `self._request_provider` may contain any `RequestProvider`. Also if you do `await crawler.get_request_provider(); await crawler.get_request_provider(name="Frankie")`, the second call won't return a Frankie, which is confusing.

janbuchar: We could add a `get_request_queue` helper, but that may be even more confusing.

src/crawlee/storages/__init__.py
janbuchar: The new `request_sources` structure looks fine, I could do that :)

tests/unit/basic_crawler/test_basic_crawler.py
janbuchar: It is important in one or two of those tests, I believe. I'll see if I can change it in the other cases.

src/crawlee/storages/_request_source_tandem.py
janbuchar: Well, true. We could either remove `name` from the `RequestProvider` - it doesn't seem to be used anywhere - or we could just defer to the underlying request provider. I'm inclined towards 1.

src/crawlee/storages/_request_source_tandem.py
janbuchar: Well, what the `RequestSourceTandem` does is that it takes a `RequestProvider`, duct-tapes a `RequestSource` on top of it and then acts as a `RequestProvider` itself. I'm sure it's a known design pattern with a name and everything :smile:

vdusek: Yeah, the source of my concerns probably arise from the fact, that Python uses the same syntax for "implement an interface" and regular inheritance.

src/crawlee/basic_crawler/_basic_crawler.py
vdusek: Yeah, it makes sense. Either provide your own instance or you will get the default one (based on config).

src/crawlee/storages/_request_source_tandem.py
vdusek: Agreed.

src/crawlee/storages/_request_list.py
janbuchar: @vdusek a thumbs up is not enough here! :slightly_smiling_face: What part of the message do you agree with? If you think that a wrapper is a good idea, what would be a good name?

janbuchar:
> Also, could you please add this to the new section in the "Request storage" guide?

Done.

@vdusek @Pijukatel this is ready for review again.

B4nan:
don't forget to update the upgrading guide

Pijukatel (approved):

vdusek (commented):
I'm starting to like it, good job. Just a few more things...

src/crawlee/request_sources/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+from ._request_list import RequestList
+from ._request_provider import RequestProvider
+from ._request_source_tandem import RequestSourceTandem
+
+__all__ = [
+    'RequestList',
+    'RequestProvider',
+    'RequestSource',
+    'RequestSourceTandem',
+]
```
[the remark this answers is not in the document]
janbuchar: If we want to have `RequestSource` and `RequestSourceTandem` in the same package (and I believe we do), I can't think of any...

src/crawlee/request_sources/_request_list.py
```diff
@@ -17,7 +15,7 @@
 
 
 @docs_group('Classes')
-class RequestList(RequestProvider):
+class RequestList(RequestSource):
     """Represents a (potentially very large) list of URLs to crawl.
 
     Disclaimer: The `RequestList` class is in its early version and is not fully implemented. It is currently
```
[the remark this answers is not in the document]
janbuchar: I wondered as well, then decided to wait until #99 is done :slightly_smiling_face:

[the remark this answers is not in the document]
vdusek: okay

src/crawlee/request_sources/_request_source.py
```diff
@@ -0,0 +1,77 @@
...
+    @abstractmethod
+    async def mark_request_as_handled(self, request: Request) -> ProcessedRequest | None:
+        """Marks a request as handled after a successful processing (or after giving up retrying)."""
+
+    @abstractmethod
+    async def reclaim_request(self, request: Request, *, forefront: bool = False) -> ProcessedRequest | None:
+        """Reclaims a failed request back to the source, so that it can be returned for processing later again.
+
+        It is possible to modify the request data by supplying an updated request as a parameter.
+        """
```
janbuchar, line 53: Is there a good reason to keep this one here or should we move it to the `RequestProvider`? Since it requires support for updates to the `Request`, I'd prefer to see it in the provider. It would also make implementing a custom `RequestSource` easier.

vdusek: Hmm, that makes sense to some extent. On the other hand, I feel that `mark_request_as_handled` and `reclaim_request` should be at the same level. You either do one or the other when processing a request, correct? Or maybe both should be moved?

janbuchar: Well, if you don't need to support retries, you can do just fine with only `mark_request_as_handled`, and such use case makes some sense.

janbuchar: Fetching the next request and marking it as completed later makes it possible to implement a minimal request loader that can survive a migration, which will be a requirement for both the fully fledged `RequestList` and the `SitemapLoader` or whatever we decide to call it.

vdusek (commented):
Great, thanks! Only a few nits...

And before merging, could you please:
- Update the PR title to match the new version.
- Mention the breaking changes in the upgrading guide.
- We will also need to update the [Request storage](https://crawlee.dev/python/docs/guides/request-storage) and [Result storage](https://crawlee.dev/python/docs/guides/result-storage) guides to reflect these changes. It doesn't have to happen now, but could you at least open an issue for that?

vdusek (approved):
LGTM

Issue #759, opened 2024-11-28
autoscaled_pool.desired_concurrency_ratio is useless


Pijukatel:
It seems to me that autoscaled_pool.desired_concurrency_ratio is currently completely useless.

It is in init here:
https://github.com/apify/crawlee-python/blob/master/src/crawlee/_autoscaling/autoscaled_pool.py#L57

It is checked for bounds 0> desired_concurrency_ratio >=1
https://github.com/apify/crawlee-python/blob/master/src/crawlee/_autoscaling/autoscaled_pool.py#L97

It is used in only one place:
https://github.com/apify/crawlee-python/blob/master/src/crawlee/_autoscaling/autoscaled_pool.py#L198

And from there in condition here:
https://github.com/apify/crawlee-python/blob/master/src/crawlee/_autoscaling/autoscaled_pool.py#L202

That condition will **always be true for** currently runtime enforced values  **0> desired_concurrency_ratio >=1**:

self.current_concurrency >= math.floor(**self._desired_concurrency_ratio** * self.current_concurrency)

Did I miss something?

janbuchar:
Before we look deeper into this, did you cross-check this against the JS version? Does it do the same thing there?

Pijukatel:
I think it is exactly the same

[        const minCurrentConcurrency = Math.floor(this._desiredConcurrency * this.desiredConcurrencyRatio);](https://github.com/apify/crawlee/blob/30d0bf807f110b2bacf74d8099e9565edb509df7/packages/core/src/autoscaling/autoscaled_pool.ts#L597)

vikyw89:
In js concurrency setting works, in here it doesn't

Pijukatel:
Yes, in JS it is different. I read it incorrectly earlier. So it seems like there was a mistake when porting the code to Python.

JS:
const minCurrentConcurrency = Math.floor(**this._desiredConcurrency** * this.desiredConcurrencyRatio);

Python:
min_current_concurrency = math.floor(self._desired_concurrency_ratio * **self.current_concurrency**)

Pijukatel:
Yes, I will align the Python version with the JS version.

Issue #768, opened 2024-11-30
current concurrency is always 25 


vikyw89:
desired concurrency, minimum concurrency doesn't do anything

janbuchar:
Hello @vikyw89 and thanks for the bug report. Could you please supply a reproduction script and explain the expected results?

vdusek:
Relates #759

Pijukatel:
@vikyw89 we will soon merge  [#780](https://github.com/apify/crawlee-python/pull/780). I was not able to tell if it solves the issue you reported. Could you please provide more detailed description? How to reproduce the issue?

I see that under normal circumstances desired concurrency, minimum concurrency,current concurrency can all change, so I need to understand in what circumstances happened the issue you experienced.

PR #780, opened 2024-12-04
fix: Fix incorrect use of desired concurrency ratio

Pijukatel:
### Description
Concurrency ratio was not correctly used in Python version of autoscaled pool. Align it with the Javascript implementation of autoscaled pool.
Add test.
### Issues

- Closes: #759

### DIFF
--- a/tests/unit/_autoscaling/test_autoscaled_pool.py
+++ b/tests/unit/_autoscaling/test_autoscaled_pool.py
@@ -212,7 +212,7 @@ async def test_autoscales(system_status: SystemStatus | Mock) -> None:
 async def test_autoscales_uses_desired_concurrency_ratio(system_status: SystemStatus | Mock) -> None:
     """Test that desired concurrency ratio can limit desired concurrency.
 
-    This test crates situation where only one task is ready and then no other task is ever ready.
+    This test creates situation where only one task is ready and then no other task is ever ready.
     This creates situation where the system could scale up desired concurrency, but it will not do so because
     desired_concurrency_ratio=1 means that first the system would have to increase current concurrency to same number as
     desired concurrency and due to no other task ever being ready, it will never happen. Thus desired concurrency will
--- a/tests/unit/_autoscaling/test_autoscaled_pool.py
+++ b/tests/unit/_autoscaling/test_autoscaled_pool.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 from contextlib import suppress
 from datetime import datetime, timedelta, timezone
+from itertools import chain, repeat
 from typing import TYPE_CHECKING, TypeVar, cast
 from unittest.mock import Mock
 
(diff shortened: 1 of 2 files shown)


tests/unit/_autoscaling/test_autoscaled_pool.py
```diff
@@ -208,6 +209,60 @@ def get_historical_system_info() -> SystemInfo:
             await pool_run_task
 
 
+async def test_autoscales_uses_desired_concurrency_ratio(system_status: SystemStatus | Mock) -> None:
+    """Test that desired concurrency ratio can limit desired concurrency.
+
+    This test crates situation where only one task is ready and then no other task is ever ready.
```
vdusek, line 215:
```suggestion
    This test creates situation where only one task is ready and then no other task is ever ready.
```

tests/unit/_autoscaling/test_autoscaled_pool.py
```diff
@@ -208,6 +209,60 @@ def get_historical_system_info() -> SystemInfo:
...
+    stay 2 as was the initial setup, even though other conditions would allow the increase. (max_concurrency=4,
+    system being idle)."""
+
+    async def run() -> None:
+        await asyncio.sleep(0.1)
+
+    is_task_ready_iterator = chain([future(True)], repeat(future(False)))
+
+    def is_task_ready_function() -> Awaitable[bool]:
+        return next(is_task_ready_iterator)
```
vdusek, line 228: I know that `AutoscaledPool` requires this to be awaitable, but shouldn't it then be defined as async?

```suggestion
    async def is_task_ready_function() -> Awaitable[bool]:
        return next(is_task_ready_iterator)
```

vdusek (approved):
LGTM, just one typo & probably wrong test function declaration.

tests/unit/_autoscaling/test_autoscaled_pool.py
Pijukatel: No, but I was thinking the same when I first saw existing tests in this file.
This function returns [future](https://github.com/apify/crawlee-python/blob/27db2e4162324f40a3b5059dbc40f66d47843876/tests/unit/_autoscaling/test_autoscaled_pool.py#L30). So it is normal function, but the return value is awaitable. (Exactly what type hints require: `is_task_ready_function: Callable[[], Awaitable[bool]]`)

vdusek: Ah, thanks for the explanation.

vdusek (commented):
Let's wait for Honza's approve as well, since he implemented this and he knows the code well.

vdusek (commented):
One more thing, does this close https://github.com/apify/crawlee-python/issues/768 as well?

Pijukatel:
> One more thing, does this solve #768 as well? If so, add it please to the description with `closes`.

I was considering that. That issue is not well described so in the end I guessed, that it might be other problem. Might be solved by this, but I doubt it and without further details I will not be able to confirm it with certainty.
I have seen current concurrency, minimum concurrency and desired concurrency all changing, but user described it is not changing.

janbuchar (approved):

Issue #684, opened 2024-11-12
Integrate Camoufox into PlaywrightCrawler


Ehsan-U:
crawlee-python is great but often lacks stealth against modern anti-bot measures when using vanilla playwright. 
[Camoufox](https://github.com/daijro/camoufox) is fully [compatible](https://camoufox.com/python/usage/) with playwright, only have to change browser initialization. It [bypasses](https://camoufox.com/stealth/) all currently available anti-bot measures. It would be a great opportunity to make crawlee-python playwright stealthy.

janbuchar:
Hello @Ehsan-U, thank you for the suggestion, we will consider it!

Pijukatel:
I tried two approaches:

**Tight integration into crawlee:**
https://github.com/apify/crawlee-python/pull/782
**Pros:**
- Super easy to use.
**Cons:**
- Does not give much flexibility in customizing the Camoufox behavior. (Like turning off crawlee header generation and use Camoufox header generation)
- Additional dependency for crawlee (currently does not install on Python3.13)


**No integration, but creating example code that shows how to do it with current codebase:**
https://github.com/apify/crawlee-python/pull/789
**Pros:**
- No extra dependency for crawlee
- No need to modify existing code
- Big flexibility
**Cons**
-Not as easy to use as the first approach. But still straightforward with the example.

Ehsan-U:
It would be fantastic for Crawlee to introduce a plugin system similar to what Scrapy offers. For example, here's a plugin I developed for Camoufox that integrates seamlessly with Scrapy: [scrapy-camoufox](https://github.com/Ehsan-U/scrapy-camoufox). This approach proves highly effective, as it extends functionality without requiring any changes to Scrapy's core.

MarhyCZ:
Hey @Ehsan-U I stumbled upon your thread. I wanted to look into scrapy-camoufox. Have you taken the repo offline?

Ehsan-U:
> Hey @Ehsan-U I stumbled upon your thread. I wanted to look into scrapy-camoufox. Have you taken the repo offline?

Hi @MarhyCZ, Thanks for showing interest ! here is the repo: https://github.com/Ehsan-U/scrapy-camoufox

murilomulato:
Hey @Ehsan-U I wanted to look into scrapy-camoufox. Have you taken the repo offline?

PR #782, opened 2024-12-05
feat: Enable usage of Camoufox in PlaywrightCrawler

Pijukatel:
### Description
PlaywrightCrawler can use argument browser_type="camoufox" to use Camoufox as browser.
Camoufox is stealthy build of Firefox.

### Issues

- Closes: #684


Mantisus:
Hey!

This is going to be a great feature!

But I'm not sure if we should use `HeaderGenerator` with `camoufox`, they implement their own logic to work with fingerprint. 

I've tested it on several sites and it seems to work better without the `HeaderGenerator`.

Pijukatel:
> Hey!
> 
> This is going to be a great feature!
> 
> But I'm not sure if we should use `HeaderGenerator` with `camoufox`, they implement their own logic to work with fingerprint.
> 
> I've tested it on several sites and it seems to work better without the `HeaderGenerator`.

Hi, I think you are right. This PR was kind of just exploration. I will create another version that does not do any change to the crawlee code, but it will just add example code of how to integrate Camoufox into crawlee using custom BrowserPool. This option is already possible with current codebase and it gives users more control (including the headers generation). I like this option more as it does not add any other dependency to our codebase. 
We can than discuss pros and cons of both solutions and decide what would be better.

Pijukatel:
Closed in favor of: https://github.com/apify/crawlee-python/pull/789

PR #788, opened 2024-12-06
refactor!: remove `json_` and `order_no` from `Request`

Mantisus:
### Description

- Removing `json_` and `order_no` from `Request` which is a model of external API. Creating an internal model `InternalRequest` to work inside `_request_queue_client`
- Add new tests

### Issues

- Closes: #94 

### Checklist

- [x] CI passed


src/crawlee/base_storage_client/_models.py
```diff
@@ -230,3 +231,27 @@ class BatchRequestsOperationResponse(BaseModel):
...
+
+    id: str
+
+    unique_key: str
+
+    order_no: Decimal | None = None
+
+    handled_at: datetime | None
+
+    request: Request
```
janbuchar, line 248: If this could have an alias to `json_`, we wouldn't break compatibility with storage files from older versions, which is not a big deal, but still it would be nice.

src/crawlee/_request.py
```diff
@@ -256,18 +247,6 @@ class Request(BaseRequestData):
...
-    """Deprecated internal field, do not use it.
-
-    Should be removed as part of https://github.com/apify/crawlee-python/issues/94.
-    """
-
-    order_no: Decimal | None = None
-    """Deprecated internal field, do not use it.
-
-    Should be removed as part of https://github.com/apify/crawlee-python/issues/94.
-    """
```
vdusek, line 269: This is a breaking change as `Request` is part of public interface. Please add exclamation mark to the PR title. And also mention the fields removal in the upgrading the guide (it should not affect much people, as they were internals, but someone can touch them).

src/crawlee/base_storage_client/_models.py
```diff
@@ -230,3 +231,29 @@ class BatchRequestsOperationResponse(BaseModel):
...
+
+    id: str
+
+    unique_key: str
+
+    order_no: Decimal | None = None
+
+    handled_at: datetime | None
+
+    request: Annotated[Request, Field(alias='json_')]
```
vdusek, line 250: Could you add docstrings explaining the fields? Mostly what are `order_no` & `request` for. Also mention the purpose of the `json_` alias.

src/crawlee/memory_storage_client/_request_queue_client.py
```diff
@@ -15,7 +14,6 @@
 from crawlee._types import StorageTypes
 from crawlee._utils.crypto import crypto_random_object_id
 from crawlee._utils.data_processing import (
-    filter_out_none_values_recursively,
```
vdusek, line 18: Maybe we can get rid of this helper completely?

src/crawlee/memory_storage_client/_request_queue_client.py
```diff
@@ -200,16 +199,14 @@ async def list_head(self, *, limit: int | None = None) -> RequestQueueHead:
 
                 # Check that the request still exists and was not handled,
                 # in case something deleted it or marked it as handled concurrenctly
-                if request and request.order_no:
-                    requests.append(request)
-
-            items = [request for item in requests if (request := self._json_to_request(item.json_))]
+                if request and not request.handled_at:
+                    requests.append(request.to_request())
```
vdusek, line 203: Maybe we can use `internal_request` identifier to distinguish between `Request` and `InternalRequest`?

src/crawlee/memory_storage_client/_request_queue_client.py
```diff
@@ -240,39 +237,39 @@ async def add_request(
...
                 await existing_queue_by_id.update_timestamps(has_been_modified=False)
 
                 return ProcessedRequest(
-                    id=request_model.id,
-                    unique_key=request_model.unique_key,
+                    id=internal_request.id,
+                    unique_key=internal_request.unique_key,
                     was_already_present=True,
-                    was_already_handled=existing_request_with_id.order_no is None,
+                    was_already_handled=existing_request_with_id.handled_at,
```
vdusek, line 253: `handled_at` is `datetime` or `None`, but `was_already_handled` is boolean. And why this is not catched by type checker?

vdusek (changes requested):
Good job, it is much better now! And I have a few things...

src/crawlee/memory_storage_client/_request_queue_client.py
Mantisus: Deleted it. It really wasn't used anywhere else

src/crawlee/memory_storage_client/_request_queue_client.py
Mantisus: I wish I could say that after adding the type annotation the type checker started pointing out an error. But it just ignores it, and it's a mystery to me why.

src/crawlee/base_storage_client/_models.py
janbuchar: Is it unrealistic to add that (de)serialization logic though? 

If it can't be done, we would need to detect the case when a user tries to load an outdated memory storage file and throw an exception, which might be harder than ensuring compatibility.

src/crawlee/_utils/data_processing.py
```diff
@@ -10,32 +10,6 @@
     from crawlee._types import StorageTypes
 
 
-def filter_out_none_values_recursively(dictionary: dict, *, remove_empty_dicts: bool = False) -> dict | None:
```
janbuchar, line 13: Happy to see this one go :slightly_smiling_face:

src/crawlee/base_storage_client/_models.py
Mantisus: Makes sense. Updated

src/crawlee/_request.py
```diff
@@ -136,16 +135,8 @@ class BaseRequestData(BaseModel):
     headers: Annotated[HttpHeaders, Field(default_factory=HttpHeaders)] = HttpHeaders()
     """HTTP request headers."""
 
-    payload: Annotated[
-        HttpPayload | None,
-        BeforeValidator(lambda v: v.encode() if isinstance(v, str) else v),
```
janbuchar, line 141: Isn't the `BeforeValidator` still needed when loading an old request file?

src/crawlee/base_storage_client/_models.py
```diff
@@ -230,3 +232,33 @@ class BatchRequestsOperationResponse(BaseModel):
 
     processed_requests: Annotated[list[ProcessedRequest], Field(alias='processedRequests')]
     unprocessed_requests: Annotated[list[UnprocessedRequest], Field(alias='unprocessedRequests')]
+
+
+class InternalRequest(BaseModel):
+    """Represents an internal representation of queue request with additional metadata for ordering and storage."""
```
janbuchar, line 238:
```suggestion
    """Internal representation of a queue request with additional metadata for ordering and storage."""
```

src/crawlee/_request.py
Mantisus: This will work the same for old requests and new requests.

https://github.com/apify/crawlee-python/blob/c7cae423b3b09a5f69da1f6881565b179e014054/src/crawlee/memory_storage_client/_creation_management.py#L370

If it is an old file, we apply alias `json_` and deserialize the string.

I find it convenient to have this logic inside the model rather than in `_creation_management.py`

Mantisus: Good point. We need both so that payload is correctly serialized and deserialized to a file.
Updated the e2e test.

janbuchar (approved):
Nice one!

Issue #659, opened 2024-11-06
A helper function for simple HTML tag removal


janbuchar:
See https://github.com/apify/crawlee/blob/master/packages/utils/src/internals/cheerio.ts#L42-L85 for inspiration.

The Python version should probably use `BeautifulSoup` and tell the user to install `crawlee[beautifulsoup]` if they try to use it without dependencies.

Pijukatel:
I think that BS should already handle it using: `soup.get_text("\n", strip=True)`
https://www.crummy.com/software/BeautifulSoup/bs4/doc/#get-text . So maybe I should just wrap that in our utility function.

PR #792, opened 2024-12-09
feat: Add `html_to_text` helper function

Pijukatel:
### Description
Add simple html_to_text helper function. This generates new line separated plain text without tags.
Function is done to perform the same way as existing implementation in Javascript.
Add tests.

### Issues

- Closes: #659


Mantisus:
The `get_text` in `BeautifulSoup` will split the text with each tag encountered.

That is, for HTML of this format we will get rather strange results

```html
<!DOCTYPE html>
<html>
<head>
   <title>List with different elements</title>
</head>
<body>
   <ul>
       <li>This is a <a href="https://example.com">link</a> element</li>
       <li>This is a text with <b>bold</b> word</li>
       <li>This is an element<br>with line break</li>
   </ul>
</body>
</html>
```

Pijukatel:
> The `get_text` in `BeautifulSoup` will split the text with each tag encountered.
> 
> That is, for HTML of this format we will get rather strange results
> 
> ```
> <!DOCTYPE html>
> <html>
> <head>
>    <title>List with different elements</title>
> </head>
> <body>
>    <ul>
>        <li>This is a <a href="https://example.com">link</a> element</li>
>        <li>This is a text with <b>bold</b> word</li>
>        <li>This is an element<br>with line break</li>
>    </ul>
> </body>
> </html>
> ```

Yes, I thought about it. But am not so sure about requirements for our function. I can try to follow more closely the JS implementation.

Mantisus:
> Yes, I thought about it. But am not so sure about requirements for our function. I can try to follow more closely the JS implementation.

I'm not quite sure either

But just as a thought, since we always have `lxml` in our dependencies along with `beautifulsoup`.

You could try something like this:

```python
from lxml import html

text = html.fromstring(html_text).text_content().strip()
```

and regex for clear spaces

src/crawlee/_utils/html_to_text.py
```diff
@@ -0,0 +1,82 @@
...
+            if isinstance(page_element, (Tag, NavigableString)):
+                if isinstance(page_element, NavigableString):
+                    compr: str
+                    if isinstance(page_element.parent, Tag) and page_element.parent.name.lower() == 'pre':
+                        compr = page_element.get_text()
+                    else:
+                        # Compress white spaces outside of pre block
+                        compr = re.sub(r'\s+', ' ', page_element.get_text())
+                    # If text is empty or ends with a whitespace, don't add the leading whitespace or new line
+                    if (compr.startswith((' ', '\n'))) and re.search(r'(^|\s)$', text):
```
Pijukatel, line 61: JS version has only compr.startswith(' '), but I can't understand how it passed the test. In Python to pass exactly same test I had to do: compr.startswith((' ', '\n')

janbuchar:
Tagging @Mantisus for review - I believe you have the most field experience with beautifulsoup.

vdusek (commented):
I'm just suprised that `html_to_text` is in private `_utils` module, not exposed publicly, and not being used anywhere. Is that intended :slightly_smiling_face: ?

src/crawlee/_utils/html_to_text.py
```diff
@@ -0,0 +1,82 @@
...
+    'blockquote',
+    'dl',
+    'div',
+    'fieldset',
+    'form',
+    'table',
+    'tr',
+    'select',
+    'option',
+}
```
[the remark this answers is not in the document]
janbuchar: Perhaps a bit of investigation is in order then? Can you hunt down the person who implemented this? :smile:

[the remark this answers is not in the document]
Pijukatel: Well, that was six years ago: https://github.com/apify/apify-ts/commit/4c0e5e3e7377536a449bb7b205132348ad3b0fe9
If someone asks me about code I wrote six years ago, I would probably just punch the person.

src/crawlee/_utils/html_to_text.py
```diff
@@ -0,0 +1,82 @@
...
+                    pass
+                elif page_element.name.lower() == 'br':
+                    text += '\n'
+                elif page_element.name.lower() == 'td':
+                    _page_element_to_text(page_element.children)
+                    text += '\t'
+                else:
+                    # Block elements must be surrounded by newlines(unless beginning of text)
+                    is_block_tag = page_element.name.lower() in BLOCK_TAGS
+                    if is_block_tag and not re.search(r'(^|\n)$', text):
```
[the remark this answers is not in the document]
Pijukatel: Done

src/crawlee/_utils/html_to_text.py
```diff
@@ -0,0 +1,82 @@
...
+    'fieldset',
+    'form',
+    'table',
+    'tr',
+    'select',
+    'option',
+}
+
+
+def html_to_text(source: str | BeautifulSoup) -> str:
```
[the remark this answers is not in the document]
Pijukatel: I expose it in BScrawler. It is so BS-dependent that you anyway have to install crawlee[beautifulsoup] to be able to use it.
After checking Parsel or Playwright I realized, that I am not sure I would be able to implement this tree-based processing with those two packages alone.

Mantisus (approved):

vdusek (changes requested):
Why isn't this simply another helper method in `BeautifulSoupCrawler` or `BeautifulSoupCrawlingContext`, like the many others we already have? IMO this introduces a somewhat different "end-user interface" compared to what we've had so far.

Pijukatel:
> Why isn't this simply another helper method in `BeautifulSoupCrawler` or `BeautifulSoupCrawlingContext`, like the many others we already have? IMO this introduces a somewhat different "end-user interface" compared to what we've had so far.

I am not very sure where to put it. It is kind of general utility, that can be used by any Crawler, but due to it's implementation it has to be used only when BS is installed. I re-exported it from that BS-related init as that already contains all the BS based functionality.

src/crawlee/_utils/html_to_text.py
```diff
@@ -0,0 +1,86 @@
+from __future__ import annotations
```
vdusek, line 1: I would prefer moving this to e.g. `crawlee/beautifulsoup_crawler/_utils.py` to avoid the need for try-except imports here.

vdusek (commented):
Regarding exposing this on the context level, I was thinking about something like this:

```python
class BeautifulSoupCrawlingContext(ParsedHttpCrawlingContext[BeautifulSoup]):
    ...

    def html_to_text(self) -> str:
        """Converts HTML content to plain text."""
        return html_to_text(self.parsed_content)
```

Doesn't that make sense to you? @janbuchar @Pijukatel

src/crawlee/_utils/html_to_text.py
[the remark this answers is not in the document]
janbuchar: At the very least, please link that code from a comment above that element list :slightly_smiling_face:

janbuchar:
> btw. I tried to convert this funcionality to Parsel using LLMs and it seems fine:

I'd prefer to just keep one version of the function, but we can make it Parsel-based for all I care

src/crawlee/_utils/html_to_text.py
vdusek: on top of that, we have the PW's `infinite_scroll` helper there following the same structure

src/crawlee/beautifulsoup_crawler/_beautifulsoup_crawling_context.py
```diff
@@ -24,3 +25,7 @@ def soup(self) -> BeautifulSoup:
     def from_parsed_http_crawling_context(cls, context: ParsedHttpCrawlingContext[BeautifulSoup]) -> Self:
         """Convenience constructor that creates new context from existing `ParsedHttpCrawlingContext[BeautifulSoup]`."""
         return cls(**{field.name: getattr(context, field.name) for field in fields(context)})
+
+    def html_to_text(self) -> str:
+        """Converts parsed_content to newline-separated plain text without tags."""
```
vdusek, line 30:
```suggestion
        """Convert the parsed HTML content to newline-separated plain text without tags."""
```

src/crawlee/parsel_crawler/_parsel_crawling_context.py
```diff
@@ -24,3 +25,7 @@ def selector(self) -> Selector:
     def from_parsed_http_crawling_context(cls, context: ParsedHttpCrawlingContext[Selector]) -> Self:
         """Convenience constructor that creates new context from existing `ParsedHttpCrawlingContext[BeautifulSoup]`."""
         return cls(**{field.name: getattr(context, field.name) for field in fields(context)})
+
+    def html_to_text(self) -> str:
+        """Converts parsed_content to newline-separated plain text without tags."""
```
vdusek, line 30:
```suggestion
        """Convert the parsed HTML content to newline-separated plain text without tags."""
```

src/crawlee/_utils/html_to_text.py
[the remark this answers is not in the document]
Pijukatel: Done

src/crawlee/parsel_crawler/_utils.py
```diff
@@ -0,0 +1,63 @@
...
+    _ANY_CONSECUTIVE_WHITE_SPACES,
+    _EMPTY_OR_ENDS_WITH_ANY_WHITE_SPACE,
+    _EMPTY_OR_ENDS_WITH_NEW_LINE,
+    BLOCK_TAGS,
+    SKIP_TAGS,
+)
+
+
+def html_to_text(source: str | Selector) -> str:
+    """Converts markup string or `Selector` to newline-separated plain text without tags using Parsel."""
```
vdusek, line 17: I would use the same docstring title as for the BS variant. And maybe mention some specific details in the further description?

src/crawlee/beautifulsoup_crawler/_utils.py
```diff
@@ -0,0 +1,66 @@
...
+    BLOCK_TAGS,
+    SKIP_TAGS,
+)
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+
+def html_to_text(source: str | BeautifulSoup) -> str:
+    """Converts markup string or `BeautifulSoup` to newline separated plain text without tags using BeautifulSoup."""
```
vdusek, line 21: I believe Honza will want to expose this as a normal function as well. In that case, it will be great if you will document the args and return type as well and expose it to the API docs. (Same for the second variant.)

vdusek (commented):
Nice! More comments only regarding the docs.

src/crawlee/beautifulsoup_crawler/_utils.py
Pijukatel: I did it in src/crawlee/utils.py . But it is little weird with those implementation variants

vdusek: Oh, I missed that. Please avoid using a general `crawlee.utils` subpackage. Instead, let's place them under the `crawlee.beautifulsoup_crawler` and `crawlee.parsel_crawler` subpackages. See -> https://github.com/apify/crawlee-python/pull/792#discussion_r1886642768.

vdusek (approved):
LGTM

src/crawlee/_utils/html_to_text.py
```diff
@@ -0,0 +1,36 @@
+# This file contains shared constants used by different implementations of html_to_text function.
+from __future__ import annotations
+
+import re
+
+# Tags based on Javascript implementation of text_to_html from:
```
janbuchar, line 6:
```suggestion
# Tags based on Javascript implementation of html_to_text from:
```

PR #793, opened 2024-12-09
chore: Remove apify dependency

Pijukatel:
Apify is not really needed by crawlee. It is used only in example code and generated code templates. By removing apify we get rid of circular dependency between crawlee and apify.
If someone tries to create apify based project from template in crawlee version that is not supported by any version of apify, then it will fail to install and user will be confused.

This is done to unbloc dependency deadlock, that can happen with incompatible releases of crawlee and apify.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,14 +9,10 @@ clean:
 	rm -rf .mypy_cache .pytest_cache .ruff_cache build dist htmlcov .coverage
 
 install-dev:
-	poetry install --extras ci
-	poetry run pre-commit install
-	poetry run playwright install
-
-install-full:
 	poetry install --all-extras
 	poetry run pre-commit install
 	poetry run playwright install
+
 
 build:
 	poetry build --no-interaction -vv
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,6 @@ install-dev:
 	poetry run pre-commit install
 	poetry run playwright install
 
-
 build:
 	poetry build --no-interaction -vv
 
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,11 @@ clean:
 	rm -rf .mypy_cache .pytest_cache .ruff_cache build dist htmlcov .coverage
 
 install-dev:
+	poetry install --extras ci
+	poetry run pre-commit install
+	poetry run playwright install
+
+install-full:
 	poetry install --all-extras
 	poetry run pre-commit install
 	poetry run playwright install
--- a/docs/deployment/code/apify/crawler_as_actor_example.py
+++ b/docs/deployment/code/apify/crawler_as_actor_example.py
@@ -1,4 +1,4 @@
-from apify import Actor  # type: ignore[import-not-found]
+from apify import Actor
 
 from crawlee.beautifulsoup_crawler import BeautifulSoupCrawler, BeautifulSoupCrawlingContext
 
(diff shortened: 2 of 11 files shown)


Makefile
```diff
@@ -13,6 +13,7 @@ install-dev:
 	poetry run pre-commit install
 	poetry run playwright install
 
+
```
vdusek, line 16:
```suggestion

```

docs/deployment/code/apify/crawler_as_actor_example.py
```diff
@@ -1,4 +1,4 @@
-from apify import Actor
+from apify import Actor  # type: ignore[import-not-found]
```
vdusek, line 1: We should rather deal with this by adjusting the mypy settings in the `pyproject.toml`

src/crawlee/project_template/{{cookiecutter.project_name}}/requirements.txt
```diff
@@ -1,6 +1,6 @@
 # % set extras = [cookiecutter.crawler_type]
 # % if cookiecutter.enable_apify_integration
-# % do extras.append('apify')
+apify
```
vdusek, line 3: I believe `apify` should not be there as unconditional requirement

docs/deployment/code/apify/crawler_as_actor_example.py
janbuchar: Yup, adding comments like this to our examples is a no-go.

src/crawlee/project_template/{{cookiecutter.project_name}}/requirements.txt
janbuchar: More precisely, the template should not be affected by this change, I believe.

Pijukatel: It is still in the if block. See line above and below.

docs/deployment/code/apify/crawler_as_actor_example.py
Pijukatel: Done

pyproject.toml
```diff
@@ -205,6 +204,10 @@ warn_return_any = true
 warn_unreachable = true
 warn_unused_ignores = true
 
+[[tool.mypy.overrides]]
```
janbuchar, line 207: Any chance we could scope the override to docs only?

Pijukatel: I can't find the option to do both override of missing import of specific module and only on specific folder/pattern.

janbuchar: No big deal then :slightly_smiling_face:

janbuchar (approved):
LGTM

pyproject.toml
vdusek, line 207: Are the double brackets intentionally? What is their purpose? Thanks.

vdusek, line 207: Could you please add a comment explaining why it is here (because of docs)?

vdusek (changes requested):

pyproject.toml
Pijukatel: That is the syntax mypy wants:
https://github.com/python/mypy/blob/master/mypy/config_parser.py#L394

Pijukatel: Done

vdusek (approved):
LGTM

Issue #480, opened 2024-08-30
Create a new guide for session management


vdusek:
- We should create a new documentation guide on how to work with sessions (`SessionPool`).
- Inspiration: https://crawlee.dev/docs/guides/session-management

RaghavMangla:
Hi, I would like to work on this issue

vdusek:
@RaghavMangla We don't assign issues for hacktoberfest. If you want to work on this, open a PR. First mergeable one gets merged.

PR #795, opened 2024-12-09
docs: add session management guide

Mantisus:
### Description

- Add a session management guide

### Issues

- Closes: #480 

### Checklist

- [x] CI passed


Pijukatel (commented):
I think it is well described, but I am not so sure about explicit example for each crawler type. The examples are exactly the same from the session_management perspective and having so many same example codes will increase maintenance. So I would suggest to pick one Crawler type and demonstrate session handling only on that one. (This is weak suggestion)

vdusek:
Could you please rebase so that we can run the CI? (it should be unblocked now)

Mantisus:
> I think it is well described, but I am not so sure about explicit example for each crawler type. The examples are exactly the same from the session_management perspective and having so many same example codes will increase maintenance. So I would suggest to pick one Crawler type and demonstrate session handling only on that one. (This is weak suggestion)

I used the guide from the [JS version as a basis](https://crawlee.dev/docs/guides/session-management). The fact that they are the same may be useful for those who are switching from the JS version of crawlee to the python version.

vdusek (approved):
I did just a few minor adjustments, otherwise it is good, thanks!

Issue #755, opened 2024-11-27
Replace `page_options` with `context_options` for `PlaywrightBrowserPlugin`


Mantisus:
The `page_options` parameter worked when pages were created without using context:
https://playwright.dev/python/docs/api/class-browser#browser-new-page

Currently, we are working with context created in `PlaywrightBrowserController`. When creating a page from context, it doesn't accept any parameters:
https://playwright.dev/python/docs/api/class-browsercontext#browser-context-new-page

These parameters should now be passed when creating the context and therefore be renamed to `context_options`:
https://playwright.dev/python/docs/api/class-browser#browser-new-context

Issue #751, opened 2024-11-27
How to change `User Agent` in `PlaywrightCrawler`?


LeMoussel:
How can I change the user_agent in PlaywrightCrawler?

Here's what I tried:
```python
from crawlee.playwright_crawler import PlaywrightCrawler

crawler = PlaywrightCrawler(browser_options={'user_agent': "My User Agent"})
```
However, I encountered the following error: `BrowserType.launch() got an unexpected keyword argument 'user_agent'.`

LeMoussel:
As described in PR #753 , I specific [user_agent](https://playwright.dev/python/docs/api/class-browsertype#browser-type-launch-persistent-context-option-user-agent) to use but I still have an error:
```
File "c:\Users\pc\AppData\Local\Programs\Python\Python310\lib\site-packages\crawlee\playwright_crawler\_playwright_crawler.py", line 108, in __init__
    super().__init__(**kwargs)
TypeError: BasicCrawler.__init__() got an unexpected keyword argument 'browser_options'
```

janbuchar:
Hi @LeMoussel, the `browser_options` parameter is not yet released, so unless you are using a beta release, this works as expected. 

When are we planning to make a new release @vdusek?

Mantisus:
Hi @LeMoussel, the `user_agent` is not an option for `browser_options`. It must be passed in when the context is created. Watch out for #755

You can now set the `User-Agent` as the header for the request

But if your goal is to replace the default `User-Agent` in `Playwright` with something less conspicuous, note that this already happens automatically with [`HeaderGenerator`](https://crawlee.dev/python/api/class/HeaderGenerator)

LeMoussel:
It seems I'm missing something here. 🙂
When I set the User-Agent header for the request like this:
```python
    import json

    from crawlee import Request
    from crawlee.playwright_crawler import PlaywrightCrawler, PlaywrightCrawlingContext

    crawler = PlaywrightCrawler(
        max_requests_per_crawl=1,
    )

    @crawler.router.default_handler
    async def request_handler(context: PlaywrightCrawlingContext) -> None:
        context.log.info(f"Processing {context.request.url} ...")
        response = await context.response.text()
        data = json.loads(response)
        print(data['user-agent'])

    await crawler.run(
        [
            Request.from_url(
                url="https://httpbingo.org/user-agent",
                headers={"User-Agent": "Test User Agent"},
            )
        ]
    )
```
I get the following result:
```
[...]

PR #796, opened 2024-12-09
fix: fix `page_options` for `PlaywrightBrowserPlugin`

Mantisus:
### Description

- fix  `page_options` for `PlaywrightBrowserPlugin`

### Issues

- Closes: #755, #751 

### Testing

- Add test for check workability `page_options` in `PlaywrightBrowserPlugin`

### Checklist

- [ ] CI passed


Mantisus:
My initial thought was to change the name of `page_options` to `context_options`. However, I think this might confuse users who are not so familiar with `playwright`. Especially considering that the parameters taken by `new_context` and `new_page` in `playwright` are identical.

So I've updated the docstring to explicitly state that these parameters are passed when the context is created.

src/crawlee/browsers/_playwright_browser_controller.py
```diff
@@ -130,17 +129,20 @@ def _on_page_close(self, page: Page) -> None:
...
-    async def _create_browser_context(self, proxy_info: ProxyInfo | None = None) -> BrowserContext:
+    async def _create_browser_context(
+        self, page_options: Mapping[str, Any] | None = None, proxy_info: ProxyInfo | None = None
+    ) -> BrowserContext:
         """Create a new browser context with the specified proxy settings."""
         if self._header_generator:
             common_headers = self._header_generator.get_common_headers()
             sec_ch_ua_headers = self._header_generator.get_sec_ch_ua_headers(browser_type=self.browser_type)
             user_agent_header = self._header_generator.get_user_agent_header(browser_type=self.browser_type)
             extra_http_headers = dict(common_headers | sec_ch_ua_headers | user_agent_header)
```
Pijukatel, line 140: I think, that this extra_http_headers will never be used with current change. So either remove it or do some logic that decides when to use page_options['extra_http_headers'] and when extra_http_headers.

src/crawlee/browsers/_playwright_browser_controller.py
```diff
@@ -130,17 +129,20 @@ def _on_page_close(self, page: Page) -> None:
...
-            user_agent = user_agent_header.get('User-Agent')
```
Pijukatel, line 140: Just a question: Is user_agent expected to be part of page_options? Or why is it removed here?

src/crawlee/browsers/_playwright_browser_controller.py
```diff
@@ -98,13 +98,12 @@ async def new_page(
...
         if not self._browser_context:
-            self._browser_context = await self._create_browser_context(proxy_info)
+            self._browser_context = await self._create_browser_context(page_options, proxy_info)
 
         if not self.has_free_capacity:
             raise ValueError('Cannot open more pages in this browser.')
 
-        page_options = dict(page_options) if page_options else {}
-        page = await self._browser_context.new_page(**page_options)
+        page = await self._browser_context.new_page()
```
Pijukatel, line 106: Another question: Previously it was possible to pass arguments to self._browser_context.new_page(), now it is not possible. Was that not a valid use case?

src/crawlee/browsers/_playwright_browser_controller.py
Mantisus: user-agent is already set in `extra_http_headers` from `user_agent_header` headers, there is no need to set it twice.

src/crawlee/browsers/_playwright_browser_controller.py
Mantisus: Earlier to create a page we used to use [`browser.new_page()`](https://playwright.dev/python/docs/api/class-browser#browser-new-page) method which takes page parameters. But some time ago we switched to using context and we create a page using [`browser_context.new_page()`](https://playwright.dev/python/docs/api/class-browsercontext#browser-context-new-page) which doesn't take any parameters. Now the page opening parameters must be set when creating the context

tests/unit/browsers/test_browser_pool.py
```diff
@@ -151,3 +151,12 @@ async def test_methods_raise_error_when_not_active() -> None:
 
     async with browser_pool:
         assert browser_pool.active is True
+
+
+async def test_with_plugin_contains_page_options(httpbin: URL) -> None:
+    plugin = PlaywrightBrowserPlugin(page_options={'user_agent': 'My Best Uset-Agent'})
```
Pijukatel, line 157: Super tiny irrelevant typo: My Best Use**t**-Agent

Mantisus: Thank you for noticing that!

Pijukatel (approved):

src/crawlee/browsers/_playwright_browser_controller.py
vdusek, line 140: This was here to replace the default "headless"-something user agent. Why is it removed? Does it mean we're now using the default (headless) user agent?

Mantisus: We have a similar test

https://github.com/apify/crawlee-python/blob/master/tests/unit/playwright_crawler/test_playwright_crawler.py#L106

PR #800, opened 2024-12-10
feat: Expose event types, improve on/emit signature, allow parameterless listeners

### LINKED ISSUE #561: Expose possible event types from `crawlee._events`
The motivation is to simplify working with event data in custom listeners in `apify-sdk-python` - currently listener parameters cannot be typed without reaching into the private `crawlee._events` submodule.

See also https://github.com/apify/apify-sdk-python/pull/278#discussion_r1765091343

janbuchar:
- closes #561
- apify/apify-sdk-python#344


src/crawlee/events/_event_manager.py
```diff
@@ -125,22 +135,41 @@ async def __aexit__(
...
+    @overload
+    def on(self, *, event: Literal[Event.MIGRATING], listener: EventListener[EventMigratingData]) -> None: ...
+    @overload
+    def on(self, *, event: Literal[Event.ABORTING], listener: EventListener[EventAbortingData]) -> None: ...
+    @overload
+    def on(self, *, event: Literal[Event.EXIT], listener: EventListener[EventExitData]) -> None: ...
+    @overload
+    def on(self, *, event: Event, listener: EventListener[None]) -> None: ...
+
+    def on(self, *, event: Event, listener: EventListener[Any]) -> None:
```
Pijukatel, line 151: What is the point of keeping types on this non overloaded definition?
In documentation they keep this one without annotations:
https://docs.python.org/3/library/typing.html#typing.overload

If we keep such annotations, then it will probably limit the usefulness of all this overload definitions because this last version is so generic, it will accept almost anything.

src/crawlee/events/_event_manager.py
```diff
@@ -125,22 +135,41 @@ async def __aexit__(
...
+    @overload
+    def on(self, *, event: Literal[Event.SYSTEM_INFO], listener: EventListener[EventSystemInfoData]) -> None: ...
+    @overload
+    def on(self, *, event: Literal[Event.MIGRATING], listener: EventListener[EventMigratingData]) -> None: ...
+    @overload
+    def on(self, *, event: Literal[Event.ABORTING], listener: EventListener[EventAbortingData]) -> None: ...
+    @overload
+    def on(self, *, event: Literal[Event.EXIT], listener: EventListener[EventExitData]) -> None: ...
+    @overload
+    def on(self, *, event: Event, listener: EventListener[None]) -> None: ...
```
Pijukatel, line 149: Is this for user defined Events or why is Event not one of the Literal[Event....]

src/crawlee/events/_event_manager.py
Pijukatel: (btw. Same question for emit method)

src/crawlee/events/_event_manager.py
janbuchar: That should mean "anything not matched by the more specific overloads"

Pijukatel: Ok, I see that **Session pool events** and **Browser pool events** have no event data connected to them. But this overload allows even events with related event data to be called with EventListener[**None**]. For example call like this:

on(event: Literal[Event.EXIT], listener: EventListener[**None**])

Is that intentional?

src/crawlee/events/_event_manager.py
Pijukatel: I am still not sure I understand the use of  EventListener[Any] here. Isn't this too permissive? Shouldn't it be this instead(yes, it is ugly):

EventListener[EventSystemInfoData] | EventListener[EventSystemInfoData] | EventListener[EventMigratingData] | EventListener[EventAbortingData] | EventListener[EventExitData] | EventListener[None]

src/crawlee/events/_event_manager.py
janbuchar: Not exactly intentional, but also not a problem I believe. Who would make a listener that is annotated to accept a literal None?

src/crawlee/events/_event_manager.py
janbuchar: Well, the last method header (the one without @overload) should be a "combination" of all the overloads. Pyright doesn't let you leave out the annotations on it, so I chose to add them. Let me know if that's a problem.

janbuchar: Regarding the permissiveness, can you put together an example that shouldn't pass type check but passes?

src/crawlee/events/_event_manager.py
janbuchar: Well, `EventListener[None]` allows these two (plus their async counterparts):

```py
def listener(weird: None) -> None:  # No reason to do this
    print("Ding")

def parameterless_listener() -> None:  # The encouraged way
    print("Ding")
```

src/crawlee/events/_event_manager.py
Pijukatel: I am just surprised that for example this works for mypy:
```
...
def print_any(whatever: Any) -> None:
    print(whatever)
    
 self._event_manager.on(event=Event.SYSTEM_INFO, listener=print_any)
 ```
 On the other hand I am even more surprised that it works even when you do that explicit list of all options `EventListener[EventSystemInfoData] | EventListener[EventSystemInfoData] | EventListener[EventMigratingData] | EventListener[EventAbortingData] | EventListener[EventExitData] | EventListener[None] `. EventListener[Any] is not included!!!

File: src/crawlee/events/__init__.py [hunk 3]
```diff
@@ -1,4 +1,25 @@
...
+    'EventAbortingData',
+    'EventData',
+    'EventExitData',
+    'EventListener',
+    'EventManager',
+    'EventMigratingData',
+    'EventPersistStateData',
+    'EventSystemInfoData',
+    'LocalEventManager',
+]
```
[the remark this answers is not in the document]
janbuchar: Oh, good point. Maybe we could make a new docs group for the event payloads, huh?

tests/unit/events/test_event_manager.py
```diff
@@ -83,11 +89,36 @@ async def test_emit_event_with_no_listeners(
...
+    async_mock = MagicMock()
+
+    async def async_listener() -> None:
+        async_mock()
+
+    event_manager.on(event=Event.SYSTEM_INFO, listener=sync_listener)
+    event_manager.on(event=Event.SYSTEM_INFO, listener=async_listener)
+
+    event_manager.emit(event=Event.SYSTEM_INFO, event_data=event_system_info_data)
+    await asyncio.sleep(0.1)  # Allow some time for the event to be processed
```
Pijukatel, line 116: This probably will not be a problem in this case. But I would suggest to adopt a specific pattern in places where we use some time sleep(which we can´t rely on in our context) and have some expectation that should become true after some sleep time. 

Advantages of such pattern is that minimal time sleep can be very low. For example 0.01, but it can define large tolerance, like sleep 1s to make the test more robust.

I used it recently when I saw test that is locally stable to fail in CI. It could be abstracted and used in many places in our tests to avoid random CI failures and speed up tests on average.
https://github.com/apify/crawlee-python/pull/810/files

(Not strictly related to this change.)

janbuchar: Could you make a separate issue for that please?

Pijukatel: https://github.com/apify/crawlee-python/issues/824

src/crawlee/events/_types.py
```diff
@@ -2,15 +2,16 @@
...
 
 from pydantic import BaseModel, ConfigDict, Field
 
+from crawlee._utils.docs import docs_group
 from crawlee._utils.system import CpuInfo, MemoryUsageInfo
 
 
 class Event(str, Enum):
-    """Enum of all possible events that can be emitted."""
+    """Names of all possible events that can be emitted using an EventManager."""
```
vdusek, line 14:
```suggestion
    """Names of all possible events that can be emitted using an `EventManager`."""
```

B4nan:
If you wanted to close both of the issue listed in the description, you need to have `Closes #...` for both of them. The second issue is still opened now.

Issue #798, opened 2024-12-10
HTTPX `verify` field is not being propagated


zhouya2010:
Even when verify=False, it still reports a CERTIFICATE_VERIFY_FAILED error. The code is as follows:

```python
import asyncio

from crawlee.http_clients import HttpxHttpClient
from crawlee.parsel_crawler import ParselCrawler, ParselCrawlingContext
from crawlee.router import Router

parsel_router = Router[ParselCrawlingContext]()


@parsel_router.default_handler
async def request_handler(context: ParselCrawlingContext) -> None:
    context.log.info(f'Parsel processing {context.request.url} ...')

    # Enqueue all links found on the page.
    await context.enqueue_links()


async def main() -> None:
    http_client = HttpxHttpClient(verify=False)
    crawler = ParselCrawler(
        request_handler=parsel_router,
        http_client=http_client,
    )

    # Run the crawler with the initial list of URLs.
    await crawler.run(['https://www.shangbao.com.ph/more/scroll.html'])


if __name__ == '__main__':
    asyncio.run(main())
```

excetion as :

```
httpcore.ConnectError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1129)
```

Mantisus:
Hey @zhouya2010 thank you for your interest in the crawlee. It is a bug related to `verify` not being passed to `transport`.

vdusek:
You might want to report this issue upstream at - https://github.com/encode/httpx.

Issue #801, opened 2024-12-10
Is it possible to pass in a custom transport?


tleyden:
I'm trying to add response caching via [hishel transports](https://hishel.com/userguide/#using-the-transports), but am not seeing a way to customize the transport used by the Crawlee client as it is [created internally in _get_client()](https://github.com/apify/crawlee-python/blob/13bb4002c75ee906db3539404fa73fff825e83e4/src/crawlee/http_clients/_httpx.py#L213-L237):

```
    def _get_client(self, proxy_url: str | None) -> httpx.AsyncClient:
        """Helper to get a HTTP client for the given proxy URL.

        If the client for the given proxy URL doesn't exist, it will be created and stored.
        """
        if proxy_url not in self._client_by_proxy_url:
            # Prepare a default kwargs for the new client.
            kwargs: dict[str, Any] = {
                'transport': _HttpxTransport(
                    proxy=proxy_url,
                    http1=self._http1,
                    http2=self._http2,
                ),
                'proxy': proxy_url,
                'http1': self._http1,
                'http2': self._http2,
            }

            # Update the default kwargs with any additional user-provided kwargs.
            kwargs.update(self._async_client_kwargs)

            client = httpx.AsyncClient(**kwargs)
            self._client_by_proxy_url[proxy_url] = client

        return self._client_by_proxy_url[proxy_url]
```

Is there a way to customize the httpx client transport that I'm not seeing?

Or instead of using a 3rd party library, does Crawlee have a native method for storing long term persistent caches of responses?

Mantisus:
Hey @tleyden , thank you for your interest in crawlee.

Indeed, there's currently no way to pass a custom transport to `HttpxHttpClient`. I'm not entirely sure if we should add such capability. Since a transport can significantly affect client behavior, making it unpredictable, adding this feature requires careful consideration. Perhaps @janbuchar or @vdusek have some thoughts on this.

Regarding your proposed solution. Please note that it completely disables proxy usage. You might want to keep the current `_get_client` implementation and only modify the transport definition.

janbuchar:
> Indeed, there's currently no way to pass a custom transport to `HttpxHttpClient`. I'm not entirely sure if we should add such capability. Since a transport can significantly affect client behavior, making it unpredictable, adding this feature requires careful consideration. Perhaps @janbuchar or @vdusek have some thoughts on this.

[...]

PR #802, opened 2024-12-11
fix: pass `verify` in httpx client

Mantisus:
### Description

- fix for passing `verify` to `transport` in `HttpxHttpClient`

### Issues

- Closes: #798 

### Checklist

- [ ] CI passed

### DIFF
--- a/src/crawlee/http_clients/_httpx.py
+++ b/src/crawlee/http_clients/_httpx.py
@@ -16,6 +16,7 @@ from crawlee.sessions import Session
 
 if TYPE_CHECKING:
     from collections.abc import Iterable
+    from ssl import SSLContext
 
     from crawlee._types import HttpMethod, HttpPayload
     from crawlee.base_storage_client._models import Request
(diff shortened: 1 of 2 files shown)


Mantisus:
Since we all agree that we're not ready to add custom transport support #801, I think this PR is ready for review in its current form.

src/crawlee/http_clients/_httpx.py
```diff
@@ -101,6 +102,7 @@ def __init__(
         ignore_http_error_status_codes: Iterable[int] = (),
         http1: bool = True,
         http2: bool = True,
+        verify: str | bool | SSLContext = True,
```
janbuchar, line 105: I know that this is the name of the attribute in httpx, but maybe we could rename this to something like `verify_certificates` or `verify_ssl` for better clarity?

Mantisus: These names are indeed clearer.

But as a weak argument. The `requests` API, has become such a standard for HTTP clients in Python that any Python developer, will expect exactly `verify`.

janbuchar (approved):

Issue #651, opened 2024-11-05
Implement a way to stop crawler from the user function


Pijukatel:
Implement some way to stop crawler in obvious and controlled way from the user function. It should properly shutdown all resources and immediately stop crawler to send any requests. It should be mirroring [the JS version](https://github.com/apify/crawlee/blob/568c6556d79ce91654c8a715d1d1729d7d6ed8ef/packages/basic-crawler/src/internals/basic-crawler.ts#L1600).

**Use case:**
User wants to stop crawler from within the user function. 

**Example of current workarounds for user:**
1. Add flag at the beginning of the user function and shortcut user function evaluation.
if finished:
  return
...
Drawback: Currently queued requests are still being send,  but not processed.

2. Call some private internals:
await crawler._pool.abort()
Drawback: Internal. Remaining tasks will still finish.

3. Drop request provider
await request_provider.drop()
Drawback: Bunch of errors as existing tasks might still try to access request_provider()

  
**Example of how this is solved in scrapy:**
https://docs.scrapy.org/en/2.11/faq.html#how-can-i-instruct-a-spider-to-stop-itself

PR #807, opened 2024-12-11
feat: Add stop method to BasicCrawler

Pijukatel:
### Description

This adds the possibility to stop crawler from user defined handler function.
Updated docs, added example.
Added test.

### Issues

- Closes: #651

### DIFF
--- a/docs/examples/crawler_stop.mdx
+++ b/docs/examples/crawler_stop.mdx
@@ -1,6 +1,6 @@
 ---
 id: crawler-stop
-title: Crawler stop
+title: Stopping a Crawler with stop method
 ---
 
 import ApiLink from '@site/src/components/ApiLink';
--- a/docs/examples/crawler_stop.mdx
+++ b/docs/examples/crawler_stop.mdx
@@ -1,0 +1,15 @@
+---
+id: crawler-stop
+title: Crawler stop
+---
+
+import ApiLink from '@site/src/components/ApiLink';
+import CodeBlock from '@theme/CodeBlock';
+
+import BeautifulSoupExample from '!!raw-loader!./code/beautifulsoup_crawler_stop.py';
+
+This example demonstrates how to use `stop` method of <ApiLink to="class/BasicCrawler">`BasicCrawler`</ApiLink> to stop crawler once the crawler finds what it is looking for. This method is available to all crawlers that inherit from <ApiLink to="class/BasicCrawler">`BasicCrawler`</ApiLink> and in the example below it is shown on <ApiLink to="class/BeautifulSoupCrawler">`BeautifulSoupCrawler`</ApiLink>. Simply call `crawler.stop()` to stop the crawler. It will not continue to crawl through new requests. Requests that are already being concurrently processed are going to get finished. It is possible to call `stop` method with optional argument `reason` that is a string that will be used in logs and it can improve logs readability especially if you have multiple different conditions for triggering `stop`.
+
+<CodeBlock className="language-python">
+    {BeautifulSoupExample}
+</CodeBlock>
--- a/src/crawlee/basic_crawler/_basic_crawler.py
+++ b/src/crawlee/basic_crawler/_basic_crawler.py
@@ -947,7 +947,7 @@ class BasicCrawler(Generic[TCrawlingContext]):
         if self._unexpected_stop:
             self._logger.info(
                 'No new requests are allowed because crawler `stop` method was called. '
-                'Ongoing requests will be allowed to complete. '
+                'Ongoing requests will be allowed to complete.'
             )
             return False
 
--- a/src/crawlee/basic_crawler/_basic_crawler.py
+++ b/src/crawlee/basic_crawler/_basic_crawler.py
@@ -303,6 +303,9 @@ class BasicCrawler(Generic[TCrawlingContext]):
         self._failed = False
         self._abort_on_error = abort_on_error
 
+        self._unexpected_stop: bool = False
+        self._unexpected_stop_reason = ''
+
     @property
     def log(self) -> logging.Logger:
         """The logger used by the crawler."""
@@ -328,13 +331,26 @@ class BasicCrawler(Generic[TCrawlingContext]):
         """Statistics about the current (or last) crawler run."""
         return self._statistics
 
-    @property
-    def _max_requests_count_exceeded(self) -> bool:
-        """Whether the maximum number of requests to crawl has been reached."""
-        if self._max_requests_per_crawl is None:
-            return False
+    def stop(self, reason: str = 'Stop was called externally.') -> None:
+        """Set flag to stop crawler.
 
-        return self._statistics.state.requests_finished >= self._max_requests_per_crawl
+        This stops current crawler run regardless of whether all requests were finished.
+
+        Args:
+            reason: Reason for stopping that will be used in logs.
+        """
+        self._unexpected_stop_reason = reason
+        self._unexpected_stop = True
+
+    def _stop_if_max_requests_count_exceeded(self) -> None:
+        """Call `stop` when the maximum number of requests to crawl has been reached."""
+        if self._max_requests_per_crawl is None:
+            return
+
+        if self._statistics.state.requests_finished >= self._max_requests_per_crawl:
+            self.stop(
+                reason=f'The crawler has reached its limit of {self._max_requests_per_crawl} requests per crawl. '
+            )
 
     async def _get_session(self) -> Session | None:
         """If session pool is being used, try to take a session from it."""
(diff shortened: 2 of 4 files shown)


B4nan:
I thought we will use the same name as in JS version, was there some discussion around this?

Pijukatel:
> I thought we will use the same name as in JS version, was there some discussion around this?
Well, this is my proposal. Reasoning is following:

I looked at the JS version and it has .teardown(). To me it calling crawler.stop() vs crawler.teardown() is definitely better.
From user perspective I imagine that crawler.teardown() is kind of some internal stuff that I do not care about and that can be at some point called by crawler.stop(), but that to me seems like an internal detail, not a user exposed function to call in normal circumstances.

Also crawler.stop() does not really do any teardown. It just "sets conditions" that will make crawler stop on it's own, calling whatever teardowns and cleanups it normally calls. (I did not introduce any new mechanism, I just abstracted how it was already used in case of _max_requests_count_exceeded and made it available to the user.)

B4nan:
So we don't wait for the in-progress tasks in this new `stop` method? Maybe we should have both then, I think we need a teardown method that cleans up things properly (waits for the started tasks with a timeout).

Pijukatel:
> So we don't wait for the in-progress tasks in this new `stop` method? Maybe we should have both then, I think we need a teardown method that cleans up things properly (waits for the started tasks with a timeout).

Stop does following:
- stop processing of any **new** request, even if there are some request left in whatever RequestProvider we use, by forcing the __is_task_ready_function to return False.
- stop will force __is_finished_function to return True.
Those two actions make the AutoscaledPool to shut down in normal fashion with all it's normal teardown steps.

One example consequence of the "normal teardown" of AutoscaledPool is that AutoscaledPool will **wait for any still running tasks**.  https://github.com/apify/crawlee-python/blob/master/src/crawlee/_autoscaling/autoscaled_pool.py#L262

I don't really see usecase for the teardown method to be used externally. You can understand stop method as "stop the crawler now and do whatever teardowns you normally do when crawler finishes."

Pijukatel:
> > So we don't wait for the in-progress tasks in this new `stop` method? Maybe we should have both then, I think we need a teardown method that cleans up things properly (waits for the started tasks with a timeout).
> 
> Stop does following:
> 
>     * stop processing of any **new** request, even if there are some request left in whatever RequestProvider we use, by forcing the __is_task_ready_function to return False.
> 
>     * stop will force __is_finished_function to return True.
>       Those two actions make the AutoscaledPool to shut down in normal fashion with all it's normal teardown steps.
> 
> 
> One example consequence of the "normal teardown" of AutoscaledPool is that AutoscaledPool will **wait for any still running tasks**. https://github.com/apify/crawlee-python/blob/master/src/crawlee/_autoscaling/autoscaled_pool.py#L262
> 
> I don't really see usecase for the teardown method to be used externally. You can understand stop method as "stop the crawler now and do whatever teardowns you normally do when crawler finishes."

Maybe one difference to highlight:
In JS version **teardown** calls **autoscaledPool.abort()** . 
In **Python** the **stop** does not call **autoscaled_pool.abort()** instead it "starves autoscaled_pool to stop itself" by denying it new tasks and letting it know its job is finished.

src/crawlee/basic_crawler/_basic_crawler.py
```diff
@@ -303,6 +303,9 @@ def __init__(
         self._failed = False
         self._abort_on_error = abort_on_error
 
+        self._unexpected_stop: bool = False
```
vdusek, line 306: Isn't type inference enough here?

src/crawlee/basic_crawler/_basic_crawler.py
```diff
@@ -912,27 +929,25 @@ async def _commit_request_handler_result(
...
 
     async def __is_task_ready_function(self) -> bool:
-        if self._max_requests_count_exceeded:
+        self._stop_if_max_requests_count_exceeded()
+        if self._unexpected_stop:
             self._logger.info(
-                f'The crawler has reached its limit of {self._max_requests_per_crawl} requests per crawl. '
-                f'The crawler will soon shut down. Ongoing requests will be allowed to complete.'
+                'No new requests are allowed because crawler `stop` method was called. '
+                'Ongoing requests will be allowed to complete. '
```
vdusek, line 950:
```suggestion
                'Ongoing requests will be allowed to complete.'
```

tests/unit/basic_crawler/test_basic_crawler.py
```diff
@@ -941,3 +941,61 @@ async def handler(context: BasicCrawlingContext) -> None:
...
+
+    @crawler.router.default_handler
+    async def handler(context: BasicCrawlingContext) -> None:
+        processed_urls.append(context.request.url)
+
+        # This sleep ensures that first request is processed quickly and triggers stop() almost immediately.
+        # Second request will have some sleep time to make sure it is still being processed after crawler.stop() was
+        # called from the first request and so the crawler is already shutting down.
+        await asyncio.sleep(next(sleep_time_generator))
+        crawler.stop(reason=f'Stop called on  {context.request.url}')
```
vdusek, line 994:
```suggestion
        crawler.stop(reason=f'Stop called on {context.request.url}')
```

src/crawlee/basic_crawler/_basic_crawler.py
Pijukatel: Done

docs/examples/crawler_stop.mdx
```diff
@@ -0,0 +1,15 @@
+---
+id: crawler-stop
+title: Crawler stop
```
vdusek, line 3: Can we use something better? e.g. "Stopping a Crawler with stop method", or "Halt the crawling with stop method", ...

Pijukatel: Done

src/crawlee/basic_crawler/_basic_crawler.py
```diff
@@ -912,27 +927,25 @@ async def _commit_request_handler_result(
                 await store.set_value(key, value.content, value.content_type)
 
     async def __is_finished_function(self) -> bool:
+        self._stop_if_max_requests_count_exceeded()
```
MatousMarik, line 930: Hi, I understand the logic, but I don't like the names:
_is_finished calls _stop_if -> IMO "property getter" is stopping the crawler (by reading the names of the methods)

src/crawlee/basic_crawler/_basic_crawler.py
```diff
@@ -912,27 +927,25 @@ async def _commit_request_handler_result(
...
-            return True
-
         if self._abort_on_error and self._failed:
             return True
 
         return is_finished
 
     async def __is_task_ready_function(self) -> bool:
-        if self._max_requests_count_exceeded:
+        self._stop_if_max_requests_count_exceeded()
```
MatousMarik, line 944: Same as above

src/crawlee/basic_crawler/_basic_crawler.py
Pijukatel: This might be ugly name but it is as explicit as it can get. This is internal name so it is no big deal to change. Do you have preferred naming?

MatousMarik: Sorry, my idea is a bit different, but you don't have to consider it at all.

I will try to explain it:
I don't like the logic that previously if you wanted to know if you were stopped you checked all the relevant flags and properties, mainly _max_requests_count_exceeded. 
Now you are adding a new flag _unexpected_stop. So why don't you just check as before + _unexpected_stop. Why do you have to add call called _stop_if_something before each of those checks that would do the same thing as accessing the property _max_requests_count_exceeded and then checking only for "unexpected". 
As I see it you had 1 flag and wanted to add a second one, that is a different one but is used in the same decision. So instead of checking them both you just decided you rename flag 1 and set it also when you would set the flag 2.

I would say you did something like this:
i_am_dirty = flag 1
do_i_want_to_take_a_shower = decision: 
1. return i_am_dirty

now adding flag 2 = i_am_hot

the process would be =>
rename flag 1 to i_want_to_take_a_shower  # that is always set when i_am_hot would be and 
do_i_want_to_take_a_shower would change to:
1. if i_am_dirty_check: i_want_to_take_a_shower = true
2. return i_want_to_take_a_shower

Proposal:
- keep original property _max_requests_count_exceeded
- and also check for _unexpected_stop
- if you don't want to check both of them in each case you are deciding whether your current _unexpected_stop flag is set (while currently you are kind of doing that by 1. calling _stop_if 2. checking _unexpected_stop), you can create new single point of truth like _should_stop_flag, that would act as property that would check both _max_requests_count_exceeded and _unexpected_stop and possibly some other in the future

Sorry for this useless [might not even fit in nitpick category]...

Issue #756, opened 2024-11-27
Unable to configure retry behavior for status codes 400-499


Mantisus:
In `BasicCrawler._should_retry_request`, we always return `False` for status codes 400-499.

```python
    def _should_retry_request(self, context: BasicCrawlingContext, error: Exception) -> bool:
        if context.request.no_retry:
            return False

        # Do not retry on client errors.
        if isinstance(error, HttpStatusCodeError) and is_status_code_client_error(error.status_code):
            return False
```
The is_status_code_client_error function enforces this behavior:
```python
def is_status_code_client_error(value: int) -> bool:
    """Returns `True` for 4xx status codes, `False` otherwise."""
    return 400 <= value <= 499  # noqa: PLR2004
```

This makes it impossible to configure retry behavior for these codes.

vdusek:
Related to #708

monk3yd:
Having this issue with my spider, errors 403 and 406 from the target website are impossible to retry. Is there any way to bypass this issue for now? I instantiate a RequestQueue within the failed request handler but it doesn't seem to requeue the request, even if I change its unique_key

`    @crawler.failed_request_handler
    async def failed_request_handler(context: Union[ParselCrawlingContext, BasicCrawlingContext], error: Exception) -> None:
        context.log.error(f"failed_request_handler is processing {str(error)} for url {context.request.url} ...")

        if isinstance(error, HttpStatusCodeError):
            if context.session:
                context.log.error("retiring session ...")
                context.session.retire()

            timestamp = datetime.now().strftime('%Y%m%d%H%M%S')
            rq = await RequestQueue.open(name="rq")
            req = Request.from_url(url=context.request.url, label=context.request.label, unique_key=f"{context.request.unique_key}_retry_{timestamp}")
            await rq.add_request(req, forefront=True)

            context.log.error("requeue...")

            failed = {
                "id": context.request.id,
                "url": context.request.url,
                "error": str(error),
                "request": context.request.model_dump()
            }
            dataset = await Dataset.open(name="errors")
            await dataset.push_data(failed)
`

Mantisus:
Hey @monk3yd, regarding your `failed_request_handler`. I have a few questions about the `RequestQueue`:

1. Are you using the same named queue `RequestQueue.open(name="rq")` in your crawler, or do you only open it inside the handler?
2. Note that if you don't explicitly pass a named queue to the crawler, it uses the default queue (you can open it using just `RequestQueue.open()`).

[...]

PR #498, opened 2024-09-04
fix: http clients detect 4xx as errors by default

vdusek:
### Description

- Error codes between 400 and 599 are considered as errors by default.
- Before this change only codes between 500 and 599 were handled by default.

### Issues

- Closes: #496

### Testing

- I added a new test to test that 4xx status codes are not being retried.

### Checklist

- [x] CI passed

PR #812, opened 2024-12-13
fix: respect user configuration for work with status codes

### LINKED ISSUE #708: Refactor handling of `ignored_http_status_codes` and `SessionError`
- Technical debt introduced in https://github.com/apify/crawlee-python/pull/167
- Should probably be ported over to JS afterwards (v4?)

The reasoning behind the change was that some errors (such as 401) are automatically considered `SessionError` and http-based crawlers don't take `ignore_htttp_error_status_codes` into account for them. While uncommon, it should be possible to explicitly ignore any status code.

Mantisus:
### Description

Fixes HTTP status code configuration behavior:

- `additional_http_error_status_codes`: Properly triggers errors and retries when specified status codes are encountered
- `ignore_http_error_status_codes`: Correctly treats specified error status codes as successful responses

### Issues

- Closes: #756 
- closes #708

### DIFF
--- a/src/crawlee/abstract_http_crawler/_abstract_http_crawler.py
+++ b/src/crawlee/abstract_http_crawler/_abstract_http_crawler.py
@@ -204,13 +204,10 @@ class AbstractHttpCrawler(Generic[TCrawlingContext, TParseResult], BasicCrawler[
         """
         if self._retry_on_blocked:
             status_code = context.http_response.status_code
-
-            # TODO: refactor to avoid private member access
-            # https://github.com/apify/crawlee-python/issues/708
-            if (
-                context.session
-                and status_code not in self._http_client._ignore_http_error_status_codes  # noqa: SLF001
-                and context.session.is_blocked_status_code(status_code=status_code)
+            if context.session and context.session.is_blocked_status_code(
+                status_code=status_code,
+                additional_blocked_status_codes=self._http_client.additional_blocked_status_codes,
+                ignore_http_error_status_codes=self._http_client.ignore_http_error_status_codes,
             ):
                 raise SessionError(f'Assuming the session is blocked based on HTTP status code {status_code}')
             if blocked_info := self._parser.is_blocked(context.parsed_content):
--- a/src/crawlee/abstract_http_crawler/_abstract_http_crawler.py
+++ b/src/crawlee/abstract_http_crawler/_abstract_http_crawler.py
@@ -204,11 +204,7 @@ class AbstractHttpCrawler(Generic[TCrawlingContext, TParseResult], BasicCrawler[
         """
         if self._retry_on_blocked:
             status_code = context.http_response.status_code
-            if context.session and context.session.is_blocked_status_code(
-                status_code=status_code,
-                additional_blocked_status_codes=self._http_client.additional_blocked_status_codes,
-                ignore_http_error_status_codes=self._http_client.ignore_http_error_status_codes,
-            ):
+            if self._is_blocked_status_code(context.session, status_code):
                 raise SessionError(f'Assuming the session is blocked based on HTTP status code {status_code}')
             if blocked_info := self._parser.is_blocked(context.parsed_content):
                 raise SessionError(blocked_info.reason)
--- a/src/crawlee/abstract_http_crawler/_abstract_http_crawler.py
+++ b/src/crawlee/abstract_http_crawler/_abstract_http_crawler.py
@@ -204,7 +204,7 @@ class AbstractHttpCrawler(Generic[TCrawlingContext, TParseResult], BasicCrawler[
         """
         if self._retry_on_blocked:
             status_code = context.http_response.status_code
-            if self._is_blocked_status_code(context.session, status_code):
+            if self._is_session_blocked_status_code(context.session, status_code):
                 raise SessionError(f'Assuming the session is blocked based on HTTP status code {status_code}')
             if blocked_info := self._parser.is_blocked(context.parsed_content):
                 raise SessionError(blocked_info.reason)
--- a/src/crawlee/basic_crawler/_basic_crawler.py
+++ b/src/crawlee/basic_crawler/_basic_crawler.py
@@ -1074,7 +1074,16 @@ class BasicCrawler(Generic[TCrawlingContext]):
     async def __run_request_handler(self, context: BasicCrawlingContext) -> None:
         await self._context_pipeline(context, self.router)
 
-    def _is_blocked_status_code(self, session: Session | None, status_code: int) -> bool:
+    def _is_session_blocked_status_code(self, session: Session | None, status_code: int) -> bool:
+        """Check if the HTTP status code indicates that the session was blocked by the target website.
+
+        Args:
+            session: The session used for the request. If None, the method always returns False.
+            status_code: The HTTP status code to check.
+
+        Returns:
+            True if the status code indicates the session was blocked, False otherwise.
+        """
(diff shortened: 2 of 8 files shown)


Mantisus:
I have already discussed this with @janbuchar. This PR should not introduce any breaking changes. If `additional_http_error_status_codes` and `ignore_http_error_status_codes` are not specified, the previous behavior is preserved.

Status codes were previously discussed in PR: #498 

However, we have differences with the JS version of `crawlee`:

Python:
- 400-499 codes cause errors without retries

JS:  
- 400-499 codes are treated as successful
- Exceptions:
 - 401, 403, 429 - cause errors with retries
 - 406 - causes error without retries and ignores `ignoreHttpErrorStatusCodes` parameter

src/crawlee/sessions/_session.py
```diff
@@ -193,17 +193,22 @@ def is_blocked_status_code(
         self,
         *,
         status_code: int,
-        additional_blocked_status_codes: list[int] | None = None,
+        additional_blocked_status_codes: set[int] | None = None,
+        ignore_http_error_status_codes: set[int] | None = None,
```
janbuchar, line 197: Is there even a good reason for this method to be on the `Session` class? Maybe we could move it to the `BasicCrawler` instead.

src/crawlee/abstract_http_crawler/_abstract_http_crawler.py
```diff
@@ -204,13 +204,10 @@ async def _handle_blocked_request(
...
         if self._retry_on_blocked:
             status_code = context.http_response.status_code
-
-            # TODO: refactor to avoid private member access
-            # https://github.com/apify/crawlee-python/issues/708
-            if (
-                context.session
-                and status_code not in self._http_client._ignore_http_error_status_codes  # noqa: SLF001
-                and context.session.is_blocked_status_code(status_code=status_code)
+            if context.session and context.session.is_blocked_status_code(
```
janbuchar, line 207: Did you check the equivalent method in `PlaywrightCrawler` as well? Any chance we could deduplicate this?

janbuchar (commented):
Two questions, otherwise LGTM... much better!

vdusek (approved):
Just one last naming thing, otherwise LGTM.

src/crawlee/http_clients/_base.py
```diff
@@ -150,8 +150,19 @@ def _raise_for_error_status_code(
...
-            if include_error:
-                raise HttpStatusCodeError('Error status code (user-configured) returned.', status_code)
+        if include_error:
+            raise HttpStatusCodeError('Error status code (user-configured) returned.', status_code)
 
+        if is_status_code_client_error(status_code) and not exclude_error:
+            raise HttpClientStatusCodeError('Client error status code returned', status_code)
+
+        if is_status_code_server_error(status_code) and not exclude_error:
             raise HttpStatusCodeError('Error status code returned', status_code)
```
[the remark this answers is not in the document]
Mantisus: Updated

src/crawlee/sessions/_session.py
Mantisus: I'm not sure if this is a compelling reason. But if our public API allows configuring errors at the Session level through:

```python
Session(blocked_status_codes=[...])
```

and

```python
SessionPool(create_session_settings={'blocked_status_codes': [...]})
```

then I believe it makes sense to have this method at the Session class level.

src/crawlee/abstract_http_crawler/_abstract_http_crawler.py
Mantisus: This is possible after refactoring as @Pijukatel did with `AbstractHttpParser`. So for now, I've only added a general method for dealing with status codes.

src/crawlee/sessions/_session.py
janbuchar: Huh, yeah, I'm not sure if enabling this in `SessionPool` is in fact necessary, but let's not change it in this PR :smile:

src/crawlee/sessions/_session.py
```diff
@@ -193,17 +193,22 @@ def is_blocked_status_code(
...
+            ignore_http_error_status_codes: Optional status codes to allow suppression of
+            codes from `blocked_status_codes`.
 
         Returns:
             True if the session should be retired, False otherwise.
         """
-        blocked_status_codes = self._blocked_status_codes + (additional_blocked_status_codes or [])
+        blocked_status_codes = self._blocked_status_codes | (additional_blocked_status_codes or set())
+
+        blocked_status_codes = blocked_status_codes - (ignore_http_error_status_codes or set())
```
janbuchar, line 212: Not sure if I'm reading this right, but it seems to me that `ignore_http_error_status_codes` takes precedence over `additional_blocked_status_codes` here, but in `BaseHttpClient`, the `additional_http_error_status_codes` option takes precedence over `ignore_http_error_status_codes`.

This is an edge case for sure, but still, I'd like to keep things consistent.

janbuchar:
> > However, we have differences with the JS version of crawlee:
> > Python:
> > 400-499 codes cause errors without retries
> > JS:
> > 400-499 codes are treated as successful
> > Exceptions:
> > 401, 403, 429 - cause errors with retries
> > 406 - causes error without retries and ignores ignoreHttpErrorStatusCodes parameter
> 
> Any ideas how we want to align it? I believe we do not have issue for that, correct? cc @janbuchar

I believe https://github.com/apify/crawlee/issues/812 tracks this. I also think that what we do in this PR is the correct way to handle things. The "success on 4xx" in JS is a crutch that prevented retries on such errors, and we can do that more cleanly now (in both implementations).

src/crawlee/abstract_http_crawler/_abstract_http_crawler.py
janbuchar: Nice, thank you!

src/crawlee/sessions/_session.py
Mantisus: You're right.

Updated the code to make `additional_http_error_status_codes` take precedence in both cases, which matches the behavior in the JS version

src/crawlee/basic_crawler/_basic_crawler.py
```diff
@@ -1074,3 +1073,10 @@ async def __run_task_function(self) -> None:
 
     async def __run_request_handler(self, context: BasicCrawlingContext) -> None:
         await self._context_pipeline(context, self.router)
+
+    def _is_blocked_status_code(self, session: Session | None, status_code: int) -> bool:
```
janbuchar, line 1077: This one deserves a docblock.

janbuchar (commented):
One more thing - it looks like there are no tests for this behavior - could you add some? test_http_crawler would be an appropriate place I guess.

tests/unit/http_crawler/test_http_crawler.py
```diff
@@ -136,19 +136,48 @@ async def test_handles_redirects(
     assert server['html_endpoint'].called
 
 
+@pytest.mark.parametrize(
+    ('additional_http_error_status_codes', 'ignore_http_error_status_codes', 'expected_number_error'),
+    [
+        ([], [], 1),
+        ([403], [], 3),
+        ([], [403], 0),
+        ([403], [403], 3),
```
janbuchar, line 145: So this is not considered a `SessionError`, not before this PR and not after?

Mantisus: That's right

janbuchar (approved):

PR #823, opened 2024-12-16
docs: document inherited symbols properly

barjin:
Bumps the version of `@apify/docusaurus-plugin-typedoc-api` to use the patches from https://github.com/apify/docusaurus-plugin-typedoc-api/pull/34

obrazek

This means that the documentation correctly resolves the `(grand)+parent `-inherited symbols.


vdusek (approved):
thanks

PR #827, opened 2024-12-17
refactor!: Rename PwPreNavContext to PwPreNavCrawlingContext

vdusek:
- Rename `PlaywrightPreNavigationContext` to `PlaywrightPreNavigationCrawlingContext`.
- Of course, this is really long. Do we want to make it shorter, e.g.: `PlaywrightPreNavCrawlingContext`?
- Or make shorter all the crawling contexts... e.g. use just `CrawlContext`, (`PlaywrightPreNavCrawlContext`)
- Opinions?

### DIFF
--- a/docs/examples/code/playwright_crawler.py
+++ b/docs/examples/code/playwright_crawler.py
@@ -3,7 +3,7 @@ import asyncio
 from crawlee.playwright_crawler import (
     PlaywrightCrawler,
     PlaywrightCrawlingContext,
-    PlaywrightPreNavigationCrawlingContext,
+    PlaywrightPreNavCrawlingContext,
 )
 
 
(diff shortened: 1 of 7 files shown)


janbuchar:
I'm not sure if the consistency is worth the BC break. Also, yes, the new name does seem long, and inventing new abbreviations such as `CrawlContext` doesn't make much sense to me. That's my opinion, feel free to try and change it :smile:

vdusek:
@janbuchar And what about `PlaywrightPreNavCrawlingContext`? At least a few chars shorter. The issue is that all other "crawling contexts" are consistently named as `CrawlingContext`, except for this one. Additionally, the term "context" is used for other purposes, such as launch context, browser context, ...

vdusek:
True, but the same applies to the basic crawling context :slightly_smiling_face:. This was just an idea to straighten things out for now. The `PwPreNavCC` was written by an external developer, and we (or at least I) overlooked the absence of "crawling" in it.

vdusek:
Agreed. But for this PR and the v0.5 release, let's focus only on resolving the `PlaywrightPreNavCrawlingContext` issue. The broader discussion about simplifying these names (`...CrawlContext`, `...CContext`, `...Context`, etc.) should be postponed for later, it will be more difficult and we will also have to discuss that with JS-guys, as this naming should be aligned.

Pijukatel (approved):

Issue #764, opened 2024-11-29
Update the crawlers structure


vdusek:
Place all the crawlers into a common directory named `crawlers`:

```
$ tree src/crawlee/crawlers/
src/crawlee/crawlers//
├── basic_crawler/
│   ├── _basic_crawler.py
│   ├── _context_pipeline.py
│   └── __init__.py
├── beautifulsoup_crawler/
│   ├── _beautifulsoup_crawler.py
│   ├── _beautifulsoup_crawling_context.py
│   ├── _beautifulsoup_parser.py
│   └── __init__.py
├── http_crawler/
│   ├── _http_crawler.py
│   ├── _http_crawling_context.py
│   └── __init__.py
├── parsel_crawler/
│   ├── __init__.py
│   ├── _parsel_crawler.py
│   ├── _parsel_crawling_context.py
│   └── _parsel_parser.py
├── playwright_crawler/
│   ├── __init__.py
│   ├── _playwright_crawler.py
│   ├── _playwright_crawling_context.py
│   ├── _playwright_pre_navigation_context.py
│   └── _utils.py
├── static_content_crawler/
│   ├── __init__.py
│   ├── _static_content_crawler.py
│   ├── _static_content_parser.py
│   └── _static_crawling_context.py
└── __init__.py
```

with `__init__.py`:

```python
from .basic_crawler import BasicCrawler, BasicCrawlingContext
from .beautifulsoup_crawler import BeautifulSoupCrawler, BeautifulSoupCrawlingContext
from .http_crawler import HttpCrawler, HttpCrawlingContext
from .parsel_crawler import ParselCrawler, ParselCrawlingContext
from .playwright_crawler import PlaywrightCrawler, PlaywrightCrawlingContext
from .static_content_crawler import StaticContentCrawler, StaticContentCrawlingContext
```

To be able to do imports:
```python
from crawlee.crawlers import HttpCrawler, HttpCrawlingContext
```

janbuchar:
I'm afraid that the `__init__.py` file you propose here won't work (without some black magic) - the individual crawlers have optional dependencies and if you don't have those installed, there will be mayhem.

PR #828, opened 2024-12-17
refactor!: Update the crawlers & storage clients structure

vdusek:
## Description

Update the dir structure of crawlers & storage clients, as discussed earlier on the Slack.

I decided to export nothing on the 2nd level because of the extras & it would also be pretty huge (taking into account we have also models there).

E.g. for BS crawler:
```diff
- from crawlee.beautifulsoup_crawler import BeautifulSoupCrawler, BeautifulSoupCrawlingContext
+ from crawlee.crawlers import BeautifulSoupCrawler, BeautifulSoupCrawlingContext
```

Or for memory storage client:
```diff
- from memory_storage_client import MemoryStorageClient
+ from storage_clients import MemoryStorageClient
```

This should be generally more aligned with the concepts of Crawlee. Of course, quite a breaking change though. Better to do it now than later.

This will not be applied to the JS version because sub-pkgs like `PlaywrightCrawler` are its own package.

## Issue

- Closes: #764

### Crawlers & CrawlingContexts

- All crawler and crawling context classes have been consolidated into a single sub-package called `crawlers`.
- The affected classes include: `AbstractHttpCrawler`, `AbstractHttpParser`, `BasicCrawler`, `BasicCrawlerOptions`, `BasicCrawlingContext`, `BeautifulSoupCrawler`, `BeautifulSoupCrawlingContext`, `BeautifulSoupParserType`, `ContextPipeline`, `HttpCrawler`, `HttpCrawlerOptions`, `HttpCrawlingContext`, `HttpCrawlingResult`, `ParsedHttpCrawlingContext`, `ParselCrawler`, `ParselCrawlingContext`, `PlaywrightCrawler`, `PlaywrightCrawlingContext`, `PlaywrightPreNavCrawlingContext`.

Example update:
```diff
- from crawlee.beautifulsoup_crawler import BeautifulSoupCrawler, BeautifulSoupCrawlingContext
+ from crawlee.crawlers import BeautifulSoupCrawler, BeautifulSoupCrawlingContext
```

### Storage clients

- All storage client classes have been moved into a single sub-package called `storage_clients`.
- The affected classes include: `MemoryStorageClient`, `BaseStorageClient`.

Example update:
```diff
- from crawlee.memory_storage_client import MemoryStorageClient
+ from crawlee.storage_clients import MemoryStorageClient
```

### CurlImpersonateHttpClient

- The `CurlImpersonateHttpClient` changed its import location.

Example update:
```diff
- from crawlee.http_clients.curl_impersonate import CurlImpersonateHttpClient
+ from crawlee.http_clients import CurlImpersonateHttpClient
```


vdusek:
I will update the upgrading guide once we definitely agreed on this. cc @pijukatel @janbuchar

Pijukatel (approved):

janbuchar:
> I decided to export nothing on the 2nd level because of the extras & it would also be pretty huge (taking into account we have also models there).

I respect the reasoning behind this, but doing `from crawlee.crawlers import PlaywrightCrawler, PlaywrightCrawlingContext` would be pretty sweet. What would it take to make that possible? :grin: I'm also asking because importing from deeply nested packages could make users feel like they're reaching into the internals of the library... which is not the case.

barjin:
Once this is done, make sure to update the Actor templates over at https://github.com/apify/actor-templates/ :) I see the Crawlee version isn't pinned there (it probably shouldn't, given the early development stage), so this will break it.

janbuchar (changes requested):
- there is a failing test for importing parsel crawler
- the `crawler` template (the new, universal one) doesn't seem to be updated?

janbuchar (approved):

PR #829, opened 2024-12-18
feat: Integrate browserforge fingerprints

Pijukatel:
### Description
Integrate `browserforge` package and use it's fingerprint and header generation capabilities to enable fingerprint generation in PlaywrightCrawler

**How can users use it?**
Closely following JS implementation would lead to:
```python
crawler = PlaywrightCrawler(
    headless=True,
    browser_pool_options={
        'use_fingerprints': True,
        'fingerprint_generator_options': {
            'browsers': 'firefox',
            'os': 'android',
            'screen': Screen(
                min_width=min_width, max_width=max_width, min_height=min_height, max_height=max_height
            ),
        },
    },
)
```

I suggest this instead:
```python
from crawlee.fingerprint_suite import DefaultFingerprintGenerator, HeaderGeneratorOptions, ScreenOptions

fingerprint_generator = DefaultFingerprintGenerator(
    header_options=HeaderGeneratorOptions(browsers=['firefox'], operating_systems=['android']),
    screen=ScreenOptions(min_width=min_width, max_width=max_width, min_height=min_height, max_height=max_height),
)

crawler = PlaywrightCrawler(
    headless=True,
    fingerprint_generator=fingerprint_generator
)
```
Reasoning: 

- FingerprintGeneratorOptions are too complex to be represented by nested dicts - not very readable.
- avoid having arguments `use_fingerprints` and `fingerprint_generator_options`. Instead have argument `fingerprint_generator` that is implicit for `use_fingerprints`, it contains all the `fingerprint_generator_options` and it also allows custom implementations of generator to be passed.
- `DefaultFingerprintGenerator` is very convenient for us, as we can set it to `browserforge adapter` for now and change it later.

**Currently we do not have our own injector, so we have to use browser forge injector.** That means that temporarily we have to accept `Fingerprint` definition from `browserforge` as output that is expected from `AbstractFingerprintGenerator.generate()`.
It is explicitly mentioned there as something that is subject to change. 

Why doing it like this?
This change is already usable as it is and brings most value now. Common user can use some sort of fingerprint generation offered by `browserforge`.
Adding our own type `Fingerprint` and our own injector is some additional effort and has almost no value for external users and can be part of future refactoring.  This refactoring would affect only very advanced users that created their own custom fingerprint generator (quite advanced for experimental feature that is not yet even available. So purely theoretical.)

### Issues
- Will be documented later by https://github.com/apify/crawlee-python/issues/481 , which already has decent PR https://github.com/apify/crawlee-python/pull/576
- Closes: #549


barjin:
Good job! Why don't we use [browserforge's `injector` module](https://github.com/daijro/browserforge/blob/main/browserforge/injectors/playwright/injector.py) for this? Do we need more control over the fingerprint than `browserforge` can offer us?

src/crawlee/playwright_crawler/_playwright_crawler.py
```diff
@@ -168,9 +172,7 @@ async def _navigate(
         async with context.page:
             if context.request.headers:
                 await context.page.set_extra_http_headers(context.request.headers.model_dump())
-            # Navigate to the URL and get response.
-            response = await context.page.goto(context.request.url)
-
+            response = await context.page.goto(context.request.url, wait_until='domcontentloaded')
```
barjin, line 175: By default, Playwright waits until the `load` event (i.e. until the HTML has loaded and all the subresources - images, stylesheets - loaded too). 

I don't have a strong opinion here; I'm just curious - did this make a change in some behavior during the development? 👀

Pijukatel: No, I forgot it there when I was testing stuff. It is not needed. Good catch.

Pijukatel:
> Good job! Why don't we use [browserforge's `injector` module](https://github.com/daijro/browserforge/blob/main/browserforge/injectors/playwright/injector.py) for this? Do we need more control over the fingerprint than `browserforge` can offer us?

I was not aware of that. Wow, that guy did everything. I will try to use that injector,

src/crawlee/browsers/_playwright_browser_controller.py
```diff
@@ -152,8 +143,45 @@ async def _create_browser_context(self, proxy_info: ProxyInfo | None = None) ->
             else None
         )
 
-        return await self._browser.new_context(
+        if self._use_fingerprints:
```
Pijukatel, line 146: (This can be simpler if we decide to rely purely on browserforge and remove our own header generation. )

src/crawlee/browsers/_playwright_browser_controller.py
```diff
@@ -152,8 +143,45 @@ async def _create_browser_context(self, proxy_info: ProxyInfo | None = None) ->
...
             extra_http_headers=extra_http_headers,
             proxy=proxy,
         )
+
+    def _get_browser_context(self) -> BrowserContext:
+        if not self._browser_context:
+            raise RuntimeError('Browser context was not set yet.')
+        return self._browser_context
+
+    @override
```
Pijukatel, line 174: I am confused why code below is shown as changed. There was no change to this in any commit in this PR

Pijukatel:
> > 2. How can users use it?
> 
> Why not expose the entire class directly, instead of creating an additional one just for the options and exposing that?
> 

At least for the HeaderGenaratorOptions that I think is necessary to avoid users mixing various header generators with fingerprint generators. HeaderGenerator is internal to FingerprintGenerator to ensure that generated fingerprint and headers are consistent.

FingerprintGeneratorOptions can be  removed and the inputs defined in abstract class instead.

Pijukatel:
> > 4. Keep our own header generation or rely on browserforge?
> 
> I would completely remove our implementation of `HeaderGenerator` and rely on the one from `browserforge` to avoid mixing the two.

I agree, but that step I would rather take in separate follow up PR to not make this PR too large.

src/crawlee/fingerprint_suite/_fingerprint_generator.py
```diff
@@ -0,0 +1,47 @@
...
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from browserforge.fingerprints import Fingerprint
+
+    from crawlee.fingerprint_suite._types import HeaderGeneratorOptions, ScreenOptions
+
+
+class AbstractFingerprintGenerator(ABC):
```
janbuchar, line 12: Considering https://github.com/apify/crawlee-python/issues/906, could we drop the `Abstract` prefix?

src/crawlee/fingerprint_suite/_fingerprint_generator.py
```diff
@@ -0,0 +1,47 @@
...
+    @abstractmethod
+    def __init__(
```
janbuchar, line 14: You usually leave implementors to define any `__init__` they need, is there any specific reasoning behind requiring it to look like this?

janbuchar (commented):
I don't know this part of the codebase that well, so I found just some minor design nitpicks

src/crawlee/fingerprint_suite/_fingerprint_generator.py
Pijukatel: Ok. Well initially I thought we will pass pre-defined specific options object to `__init__`, but that is not needed.

src/crawlee/fingerprint_suite/_fingerprint_generator.py
Pijukatel: Ok

janbuchar (approved):
Looks good to me, but I'd prefer to wait for @vdusek's approval as well

vdusek:
> I agree, but that step I would rather take in separate follow up PR to not make this PR too large.

Do I understand it correctly that it will not require any changes on the interface?

Also, could you open an issue for it?

src/crawlee/browsers/_browser_pool.py
```diff
@@ -23,6 +23,7 @@
     from types import TracebackType
 
     from crawlee.browsers._base_browser_plugin import BaseBrowserPlugin
+    from crawlee.fingerprint_suite._fingerprint_generator import FingerprintGenerator
```
vdusek, line 26:
```suggestion
    from crawlee.fingerprint_suite import FingerprintGenerator
```

src/crawlee/browsers/_playwright_browser_controller.py
```diff
@@ -18,6 +19,7 @@
 
     from playwright.async_api import Browser
 
+    from crawlee.fingerprint_suite._fingerprint_generator import FingerprintGenerator
```
vdusek, line 22:
```suggestion
    from crawlee.fingerprint_suite import FingerprintGenerator
```

vdusek (commented):
> Will be documented later by https://github.com/apify/crawlee-python/issues/481 , which already has decent PR https://github.com/apify/crawlee-python/pull/576

I would prefer to include the documentation along with this PR. It could provide a clearer understanding and demonstrate how it is being used. This is one of the reasons we keep the documentation in the same repository as the source code.

Pijukatel:
> > I agree, but that step I would rather take in separate follow up PR to not make this PR too large.
> 
> Do I understand it correctly that it will not require any changes on the interface?
> 
> Also, could you open an issue for it?

Yes, I think it should be without any interface changes.
https://github.com/apify/crawlee-python/issues/937

src/crawlee/fingerprint_suite/_fingerprint_generator.py
```diff
@@ -0,0 +1,19 @@
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from browserforge.fingerprints import Fingerprint
+
+
+class FingerprintGenerator(ABC):
```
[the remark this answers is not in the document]
Pijukatel: Added.

src/crawlee/fingerprint_suite/_browserforge_adapter.py
```diff
@@ -0,0 +1,72 @@
...
+from browserforge.fingerprints import Screen
+from typing_extensions import override
+
+from ._fingerprint_generator import FingerprintGenerator
+
+if TYPE_CHECKING:
+    from ._types import HeaderGeneratorOptions, ScreenOptions
+
+
+class BrowserforgeFingerprintGenerator(FingerprintGenerator):
```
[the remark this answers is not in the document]
Pijukatel: Added.

Pijukatel:
> > Will be documented later by #481 , which already has decent PR #576
> 
> I would prefer to include the documentation along with this PR. It could provide a clearer understanding and demonstrate how it is being used. This is one of the reasons we keep the documentation in the same repository as the source code.
> 
> Otherwise it's great.

I added example code + page.

I guess I should also make Python variant of this doc in other repo: https://docs.apify.com/academy/anti-scraping/mitigation/generating-fingerprints

src/crawlee/fingerprint_suite/__init__.py
```diff
@@ -1 +1,4 @@
+from ._browserforge_adapter import BrowserforgeFingerprintGenerator as DefaultFingerprintGenerator
```
vdusek, line 1: Why use the `DefaultFingerprintGenerator` alias? I think we can stick with `Browserforge...`. (Sorry, but since this is a public interface, I want to ensure we have a well chosen name.)

src/crawlee/fingerprint_suite/_types.py
```diff
@@ -0,0 +1,51 @@
...
+
+from pydantic import BaseModel, Field
+
+SupportedOperatingSystems = Literal['windows', 'macos', 'linux', 'android', 'ios']
+SupportedDevices = Literal['desktop', 'mobile']
+SupportedHttpVersion = Literal['1', '2']
+SupportedBrowserType = Literal['chromium', 'firefox', 'webkit', 'edge']
+
+
+class ScreenOptions(BaseModel):
```
vdusek, line 13: Missing docs decorator.

src/crawlee/fingerprint_suite/_types.py
```diff
@@ -0,0 +1,51 @@
...
+    max_width: Annotated[float | None, Field(alias='maxWidth')] = None
+    min_height: Annotated[float | None, Field(alias='minHeight')] = None
+    max_height: Annotated[float | None, Field(alias='maxHeight')] = None
+
+    class Config:
+        extra = 'forbid'
+        populate_by_name = True
+
+
+class HeaderGeneratorOptions(BaseModel):
```
vdusek, line 26: Missing docs decorator.

src/crawlee/fingerprint_suite/_types.py
```diff
@@ -0,0 +1,51 @@
...
+SupportedBrowserType = Literal['chromium', 'firefox', 'webkit', 'edge']
+
+
+class ScreenOptions(BaseModel):
+    """Defines the screen constrains for the fingerprint generator."""
+
+    min_width: Annotated[float | None, Field(alias='minWidth')] = None
+    max_width: Annotated[float | None, Field(alias='maxWidth')] = None
+    min_height: Annotated[float | None, Field(alias='minHeight')] = None
+    max_height: Annotated[float | None, Field(alias='maxHeight')] = None
```
vdusek, line 19: Missing fields docstrings - I know their meaning is quite obvious, but I would prefer to have them because of the API docs rendering.

src/crawlee/fingerprint_suite/_fingerprint_generator.py
```diff
@@ -0,0 +1,23 @@
...
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING
+
+from crawlee._utils.docs import docs_group
+
+if TYPE_CHECKING:
+    from browserforge.fingerprints import Fingerprint
+
+
+@docs_group('Classes')
```
vdusek, line 12: This is abstract class.

src/crawlee/fingerprint_suite/_browserforge_adapter.py
```diff
@@ -0,0 +1,77 @@
...
+
+from ._fingerprint_generator import FingerprintGenerator
+
+if TYPE_CHECKING:
+    from ._types import HeaderGeneratorOptions, ScreenOptions
+
+
+@docs_group('Classes')
+class BrowserforgeFingerprintGenerator(FingerprintGenerator):
+    """`FingerprintGenerator` adapter for fingerprint generator from `browserforge`."""
```
vdusek, line 21: Please add link to the `browserforge` project.

vdusek (commented):
A few more comments, mostly regarding the docs.

Also this will be further documented in the new "avoid getting blocked" guide, am I correct?

src/crawlee/fingerprint_suite/__init__.py
Pijukatel: That is based on the discussion in the team where it was mentioned that we will use browserforge for now, and potentially replace it in the future. I do not insist on this specific name, but I would like to have an implementation independent name. 
`DefaultFingerprintGenerator` is supposed to communicate something like: "This is some fingerprint generator whose internals are implementation detail and can change."
Since [we kind of decided](https://github.com/apify/crawlee-python/issues/906#issuecomment-2592944217) that `FingerprintGenerator` is reserved for the abstract class we have to come up with some name for the default implementation of the abstract one.  

This way if we switch to another fingerprint generator we just change input to this export alias and user's won't have to do anything.

Pijukatel:
> A few more comments, mostly regarding the docs.
> 
> Also this will be further documented in the new "avoid getting blocked" guide, am I correct?

Yes. I will do it within https://github.com/apify/crawlee-python/issues/481.
This is just pre-requisite for that as mentioned in the description.

src/crawlee/fingerprint_suite/_browserforge_adapter.py
Pijukatel: Done

src/crawlee/fingerprint_suite/_fingerprint_generator.py
Pijukatel: Done

src/crawlee/fingerprint_suite/_types.py
Pijukatel: Done

src/crawlee/fingerprint_suite/_types.py
Pijukatel: Done

src/crawlee/fingerprint_suite/_types.py
Pijukatel: Done

src/crawlee/fingerprint_suite/__init__.py
vdusek: Yeah, that makes sense, thanks for the explanation.

vdusek (approved):
Good job :+1: .

Issue #628, opened 2024-10-29
Migrate from `poetry` & `pipx` to `uv` & `uvx`


janbuchar:
- `uvx crawlee create` is a bit sexier than `pipx run crawlee create`
- is `uv` as widely available as `pipx`? Windows support?
- any other pitfalls?

matt-hayden:
`uv tool install crawlee` adds a link in `~/.local/bin` to a sandboxed installation.

vdusek:
I am closing #59 as a duplicate since it is closely related to this issue.

TL;DR: Migrate from `poetry` & `pipx` to `uv` & `uvx`.

### Proposed changes

- Use `uv` and `uvx` for:
  - Project bootstrapping
  - Project management
  - Package installation in CI
  - Any other applicable use cases (to be determined)

### Next steps

- Do the research to identify any potential drawbacks or limitations before migrating.

vdusek:
Currently `uv publish` is experimental:

```
warning: `uv publish` is experimental and may change without warning
```

vdusek:
From @janbuchar:

> If we decide to mention uvx as well, the corresponding command would be uvx --with crawlee[cli] crawlee create my-crawler, which is... unsexy. See https://github.com/astral-sh/uv/issues/6296 - the --with flag may become unnecessary in the future.

Mantisus:
> From @janbuchar:
> 
> > If we decide to mention uvx as well, the corresponding command would be uvx --with crawlee[cli] crawlee create my-crawler, which is... unsexy. See [astral-sh/uv#6296](https://github.com/astral-sh/uv/issues/6296) - the --with flag may become unnecessary in the future.

`uvx crawlee create my-crawler` works, `--with crawlee[cli]` is not necessary

janbuchar:
> `uvx crawlee create my-crawler` works, `--with crawlee[cli]` is not necessary

Yep, we're considering adding the `cli` extra in the future so that the generated project doesn't have to depend on `cookiecutter`, `typer` and so on. See #703.

> Windows supports

Great to hear that :slightly_smiling_face:

Mantisus:
> Yep, we're considering adding the `cli` extra in the future so that the generated project doesn't have to depend on `cookiecutter`, `typer` and so on. See #703.

Oh, got it. But it seems to me that for uvx support it is easier to output `cli` dependencies as [script dependencies](https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies)

[...]

PR #852, opened 2024-12-31
chore: migrate to uv

Mantisus:
### Description

- Migration from poetry to uv

### Issues

- Closes: #628


Mantisus:
An experiment to test `uv` instead of `poetry`.

Perhaps a basis for future migration. But I agree with @vdusek that it's still raw. But quite promising.

vdusek:
Hi @Mantisus, thank you. Closing this in a favour of #994.

Issue #856, opened 2025-01-03
Align use_state with JS implementation


Pijukatel:
This is pre-requisite for adaptive crawler : https://github.com/apify/crawlee-python/issues/249

**use_state** should ensure access to global state without race conditions. Currently there can be race conditions when two parallel handlers try to access  same value at the same time. Both can get it from kvs and that results in two separate references. Instead both should always get the same reference. So there should be lock that ensures no race conditions.

use_state in JS implementation has hard-coded key.

PR #857, opened 2025-01-03
feat!: Update `use_state` side efefcts to go through `RequestHandlerRunResult`

Pijukatel:
### Description
Update use_state side efefcts to go through RequestHandlerRunResult
Hardcode key for use state. (Same key hardcoded in JS implementation.)
Update tests.

### Issues

- Closes: #856

### DIFF
--- a/src/crawlee/_types.py
+++ b/src/crawlee/_types.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from collections.abc import Iterator, Mapping
+from copy import deepcopy
 from dataclasses import dataclass
 from enum import Enum
 from typing import TYPE_CHECKING, Annotated, Any, Literal, Optional, Protocol, TypeVar, Union, cast, overload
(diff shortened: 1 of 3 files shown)


Mantisus:
I like your solution.

But in it, we go from saving `State` in `KVS` by time interval with `Event.PERSIST_STATE` to saving after each request when `State` has been changed in the handler. I hypothesize that this may not be desirable on the Apify platform, as it would cause too many write-read operations in `KVS` in some cases.

Perhaps @janbuchar will correct me

Pijukatel:
> I like your solution.
> 
> But in it, we go from saving `State` in `KVS` by time interval with `Event.PERSIST_STATE` to saving after each request when `State` has been changed in the handler. I hypothesize that this may not be desirable on the Apify platform, as it would cause too many write-read operations in `KVS` in some cases.
> 
> Perhaps @janbuchar will correct me

I also missed one crucial point about use_state. It is supposed to give access to global state without race conditions in parallel handlers. This is mutually exclusive with isolating state in RequestHandlerRunResult and committing it later - that leads to race conditions. So I will drop this, but I will most likely add some more tests to document/ensure no race conditions happen when using use_state.

PR #868, opened 2025-01-06
fix: Avoid `use_state` race conditions. Remove key argument to `use_state`

Pijukatel:
### Description
Align Python implementation of `use_state` wirth JS

Add lock when accessing the cache or kvs to prevent race conditions.
Add unit tests for race conditions.

### Breaking

Remove key argument from `use_state` and hard code it to `BasicCrawler.CRAWLEE_STATE_KEY`. 
(Change in experimental feature is not marked as hard breaking with `!`.) 

### Issues

- Closes: #856

### DIFF
--- a/tests/unit/crawlers/_basic/test_basic_crawler.py
+++ b/tests/unit/crawlers/_basic/test_basic_crawler.py
@@ -1056,17 +1056,46 @@ async def test_context_use_state_race_condition_in_handlers(key_value_store: Key
     crawler = BasicCrawler()
     store = await crawler.get_key_value_store()
     await store.set_value(BasicCrawler.CRAWLEE_STATE_KEY, {'counter': 0})
-    handler_barrier = asyncio.Barrier(2)
+
+    # Asyncio.Barrier introduced in Python 3.11. Code has to be compatible with Python 3.9.
+    # Use asyncio.Event as workaround.
+    both_handlers_got_state = asyncio.Event()
+    both_handlers_incremented_counter = asyncio.Event()
+
+    get_value_barrier = 0
+    increment_barrier = 0
+
+    async def set_events() -> None:
+        while True:
+            if get_value_barrier == 2:
+                both_handlers_got_state.set()
+            if increment_barrier == 2:
+                both_handlers_incremented_counter.set()
+            if both_handlers_got_state.is_set() and both_handlers_incremented_counter.is_set():
+                return
+            await asyncio.sleep(0.01)
+
+    set_events_task = asyncio.create_task(set_events())
 
     @crawler.router.default_handler
     async def handler(context: BasicCrawlingContext) -> None:
+        nonlocal increment_barrier
+        nonlocal get_value_barrier
+
         state = cast(dict[str, int], await context.use_state())
-        await handler_barrier.wait()  # Block until both handlers get the state.
+        # Block until both handlers get the state.
+        get_value_barrier += 1
+        await both_handlers_got_state.wait()
+
         state['counter'] += 1
-        await handler_barrier.wait()  # Block until both handlers increment state.
+        # Block until both handlers increment state.
+        increment_barrier += 1
+        await both_handlers_incremented_counter.wait()
 
     await crawler.run(['https://crawlee.dev/', 'https://crawlee.dev/docs/quick-start'])
+    await set_events_task
 
     store = await crawler.get_key_value_store()
-
+    # Ensure that local state is pushed back to kvs.
+    await store.persist_autosaved_values()
     assert (await store.get_value(BasicCrawler.CRAWLEE_STATE_KEY))['counter'] == 2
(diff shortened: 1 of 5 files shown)


tests/unit/crawlers/_basic/test_basic_crawler.py
```diff
@@ -1046,3 +1046,27 @@ async def handler(context: BasicCrawlingContext) -> None:
...
+
+async def test_context_use_state_race_condition_in_handlers(key_value_store: KeyValueStore) -> None:
+    """Two parallel handlers increment global variable obtained by `use_state` method.
+
+    Result should be incremented by 2.
+    Method `use_state` must be implemented in a way that prevents race conditions in such scenario."""
+    crawler = BasicCrawler()
+    store = await crawler.get_key_value_store()
+    await store.set_value(BasicCrawler.CRAWLEE_STATE_KEY, {'counter': 0})
+    handler_barrier = asyncio.Barrier(2)
```
Pijukatel, line 1059: TODO: Rewrite the test to not use Barrier (added in Python 3.11 and so it fails in older Python versions)

Pijukatel: It is so ugly without Barrier :-(

janbuchar:
May not be related, but I observed flaky behavior in state handling [here](https://github.com/apify/crawlee-python/actions/runs/12652472480/job/35255185577?pr=873)

Pijukatel:
> May not be related, but I observed flaky behavior in state handling [here](https://github.com/apify/crawlee-python/actions/runs/12652472480/job/35255185577?pr=873)

I will check it. It is supposed to be autosaved when specific event happens. It is supposed to happen every 50ms, so 1s seemed like a long enough reserve. Maybe there is some hard block in some cases and no amount of reserve time is enough?
https://github.com/apify/crawlee-python/blob/master/tests/unit/storages/test_key_value_store.py#L29

Mantisus (approved):
LGTM

tests/unit/storages/test_key_value_store.py
```diff
@@ -178,3 +179,29 @@ async def autosaved_within_deadline(key: str, expected_value: dict[str, str]) ->
...
+    async def delayed_get_value(key: str, default_value: None) -> None:
+        await asyncio.sleep(next(sleep_time_iterator))
+        return await KeyValueStore.get_value(key_value_store, key=key, default_value=default_value)
+
+    async def increment_counter() -> None:
+        state = cast(dict[str, int], await key_value_store.get_auto_saved_value('state'))
+        state['counter'] += 1
+
+    with patch.object(key_value_store, 'get_value', delayed_get_value):
+        tasks = [asyncio.create_task(increment_counter()) for _ in range(2)]
```
janbuchar, line 204: nit - I'd probably just make a two-item array here

tests/unit/storages/test_key_value_store.py
```diff
@@ -178,3 +179,29 @@ async def autosaved_within_deadline(key: str, expected_value: dict[str, str]) ->
 
     value['hello'] = 'new_world'
     assert await autosaved_within_deadline(key=key_name, expected_value={'hello': 'new_world'})
+
+
+async def test_get_auto_saved_value_auto_save_race_conditions(key_value_store: KeyValueStore) -> None:
```
janbuchar, line 184: Does this reliably reproduce the error if you remove the lock?

tests/unit/crawlers/_basic/test_basic_crawler.py
```diff
@@ -1046,3 +1046,56 @@ async def handler(context: BasicCrawlingContext) -> None:
...
+    """Two parallel handlers increment global variable obtained by `use_state` method.
+
+    Result should be incremented by 2.
+    Method `use_state` must be implemented in a way that prevents race conditions in such scenario."""
+    crawler = BasicCrawler()
+    store = await crawler.get_key_value_store()
+    await store.set_value(BasicCrawler.CRAWLEE_STATE_KEY, {'counter': 0})
+
+    # Asyncio.Barrier introduced in Python 3.11. Code has to be compatible with Python 3.9.
+    # Use asyncio.Event as workaround.
```
janbuchar, line 1061: Feel free to use `Barrier` and skip the test on older pythons.

janbuchar (commented):
Seems legit, but I want to re-read the tests after my comments are resolved.

tests/unit/storages/test_key_value_store.py
Pijukatel: I tried stress testing it 1000x times and it failed with old implementation all the time. But I acknowledge that it is not guaranteed and in theory it can pass. Same as in reality if such delay in kvs access occurs, it will most likely end in race condition, but in theoretical case it does not have to.
I did try coming up with test without such weakness, but I failed to find better approach.

This test should not have any false positives and should not be flaky, but has theoretical chance for false negative. In CI we run the same test for various Python implementations which makes already purely theoretical chance completely negligible.

bigcartelo:
您发给我的邮件已收到，谢谢。

tests/unit/storages/test_key_value_store.py
Pijukatel: ok

janbuchar (approved):
LGTM

Issue #625, opened 2024-10-29
Make identifiers clickable in the rendered API documentation


vdusek:
- Make identifiers of classes, functions, methods, etc., clickable in the generated API docs.
- https://github.com/apify/crawlee-python/pull/624#discussion_r1816976508

barjin:
(Marginally) related to https://github.com/apify/crawlee/issues/2717 - we might want to use something like e.g. JSDoc's [`{@link Dataset}`](https://jsdoc.app/tags-inline-link). In the best case, we'd like to render clickable links both in IDEs and on the web.

I'm not sure whether this aligns with Python project documentation standards; it might need more research.

barjin:
Revisiting this - is there a Pythonic way of linking other symbols in docstrings? Something like 

```python
class A:
    pass

class B:
    """Very similar to {@link A}"""
    pass

```

In JS, this would give us nice clickable links in IDE **and** the docs. Actually, with the current tooling, this gives us clickable links **in the rendered docs** with Python as well (but viewing the tooltip in VSCode renders the tag verbatim).

barjin:
Upon further research, the best way forward is IMO adding markdown links to `crawlee.dev/python/...` to the relevant spots in the docstrings. Most major IDEs have support for markdown rendering in the symbol tooltips, and so does our documentation engine.

As a bonus, setting the link to the actual web documentation will IMO help the end users more than linking to the symbol in the library (opening the Crawlee source file in their IDE).

Unassigning this issue and moving back to New Issues, as the scope IMO changed (now it's just about adding those links manually to the docstrings). Let's discuss this again on the next sync.

PR #872, opened 2025-01-06
feat: Add AdaptivePlaywrightCrawler

Pijukatel:
Add AdaptivePlaywrightCrawler, based on [JS implementation](https://github.com/apify/crawlee/blob/master/packages/playwright-crawler/src/internals/adaptive-playwright-crawler.ts#L211)

### Issues
https://github.com/apify/crawlee-python/issues/249

### Purpose:
Adaptive crawler can choose to crawl page with either static crawler(like BeautifulSoupCrawler or ParselCrawler) or browser-based PlaywrightCrawler.

This is done to theoretically benefit from both approaches - use faster static crawler where it is possible and fallback to slower browser-based crawler where needed.

### Implementation
Python implementation differs from JS implementation mainly in the way adaptive crawler performs the browser crawling or the static crawling. 

JS implementation explicitly defines the code that is used for static/browser crawling - this makes it more simple and easier to understand, but it is some code duplication(code that already exists in other crawlers) and also it does drop some features that exist on those crawlers, like pre-navigation hooks.

Python implementation does not define explicitly code for static/browser crawling. It instead delegates this task to "sub crawlers" (instance of implementation of AbstractHttpCrawler for static crawling and instance of PlaywrightCrawler.)
This avoids code duplication and supports more features and control over the static/browser crawling, but it adds complexity of connecting all the components together.

"Sub crawlers" are directly used only in `__init__` method of top crawler to create their context pipelines. These are then used by top crawler to perform single request crawls when needed. This approach re-uses existing code from other crawlers and fully preserve their context pipeline handling.

**Crawling contexts**
Since `AdaptivePlaywrightCrawler` can crawl page in two different manners the contexts can also look differently. I selected approach where there is unified context for both crawl types(static/browser) and this unified context contains all the attributes that are found both on static context and browser context, but browser based attributes will raise  `AdaptiveContextError(RuntimeError)` when accessed during static crawling. It is impossible to know in advance which mode of crawling will be used, so this forces users to use `try-catch` pattern when working with these unified contexts.

For user defined pre-navigation hooks there is `AdaptivePlaywrightPreNavCrawlingContext` (very nice name!) that has `page` property that can raise `AdaptiveContextError`.

For user defined request handler there is `AdaptivePlaywrightCrawlingContext` that has `page`, `infinite_scroll` and `response` properties (all copied from PlaywrightCrawlingContext) that can raise `AdaptiveContextError`.

### Missing in this PR
**Documentation**
Documentation, examples and template will be added as last step in follow up PRs. The reason is that this feature should be documented only when it is in state that is ready to be used by users. Updating documentation too early would lead to users to try to use incomplete feature.

**Rendering type predictor**
JS implementation uses statistical regression based predictor to decide what crawl type should be used. This predictor is not implemented in this PR and is replaced by dummy version - `RandomRenderingTypePredictor`. 
Proper predictor will be implemented in follow up PR.

**Helper methods on adaptive context**
JS implementation adds [several helper methods](https://github.com/apify/crawlee/blob/master/packages/playwright-crawler/src/internals/adaptive-playwright-crawler.ts#L124) on adaptive context. Since adaptive crawler in JS is not generic  with respect to static crawling, it is straight forward to implement it with the only option Cheerio. 
If same was to be done in Python it would require more handling of generics as it would basically require to wrap some sort of generic "select" method from whatever generic parser is used and it's return value would have to be another generic parameter.
This change is standalone and it is better to do it separately.

### Parity changes to JS code
Some improvements could be ported back to JS implementation. These are improvements and issues are summarized in https://github.com/apify/crawlee/issues/2798


src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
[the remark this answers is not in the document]
Pijukatel: That is just composing the HttpCrawlerOptions so that you can access only the additional HttpCrawlerOptions without the BasicCrawlerOptions

Image it as:  _HttpCrawlerOptions = HttpCrawlerOptions - BasicCrawlerOptions  (I do not think any operation like this is supported by TypedDicts so this is workaround for it.)

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,316 @@
...
+
+import asyncio
+import logging
+from contextlib import AsyncExitStack
+from copy import deepcopy
+from logging import getLogger
+from random import random
+from typing import TYPE_CHECKING, Any
+
+from IPython.core.completer import TypedDict
```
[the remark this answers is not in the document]
Pijukatel: I have to be more careful about IDE import suggestions.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,316 @@
...
+
+    from crawlee import Request
+    from crawlee.browsers import BrowserPool
+    from crawlee.browsers._types import BrowserType
+    from crawlee.crawlers._basic._basic_crawler import _BasicCrawlerOptions
+    from crawlee.statistics import FinalStatistics
+
+
+@docs_group('Data structures')
+class _BeautifulsoupCrawlerAdditionalOptions(_HttpCrawlerOptions):
```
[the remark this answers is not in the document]
Pijukatel: Well, I can keep it as implementation detail, but then the context will not have `soup` attribute and have just the generic `parsed_content`.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,316 @@
...
+
+                if bs_run.result and self.result_comparator(bs_run.result, pw_run.result):
+                    detection_result = 'static'
+                else:
+                    detection_result = 'client only'
+
+                context.log.debug(f'Detected rendering type {detection_result} for {context.request.url}')
+                self.rendering_type_predictor.store_result(context.request.url, context.request.label, detection_result)
+
+    async def commit_result(self, result: RequestHandlerRunResult, context: BasicCrawlingContext) -> None:
```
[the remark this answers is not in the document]
Pijukatel: This is constructing the result that will be used in _commit_request_handler_result, but it is also already committing the kvs changes directly due to the nature of how they are stored in the result.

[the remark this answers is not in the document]
Pijukatel: I added short docstring and aligned the kvs handling with the rest of the result handling.

So all that happens in this method is populating the final result object that will be used by the `_commit_request_handler_result`

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,316 @@
...
+        async with AsyncExitStack() as exit_stack:
+            for context in contexts_to_enter:
+                await exit_stack.enter_async_context(context)
+            return await super().run(requests=requests, purge_request_queue=purge_request_queue)
+
+        # AsyncExitStack can in theory swallow exceptions and so the return might not execute.
+        # https://github.com/python/mypy/issues/7726
+        raise RuntimeError('FinalStatistics not created.')
+
+    # Can't use override as mypy does not like it for double underscore private method.
```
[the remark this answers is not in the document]
Pijukatel: Crawlers are now used mainly to create two context pipelines and they are thrown away after `__init__`.
(References to those crawlers still exist in those context pipelines, but that is hidden.)

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,316 @@
...
+            # If statistics already specified by user, create AdaptivePlaywrightCrawlerStatistics from it.
+            statistics = AdaptivePlaywrightCrawlerStatistics.from_statistics(statistics=kwargs['statistics'])
+        else:
+            statistics = AdaptivePlaywrightCrawlerStatistics()
+        kwargs['statistics'] = statistics
+
+        # self.statistics is hard coded in BasicCrawler to Statistics, so even when we save children class in it, mypy
+        # will complain about using child-specific methods. Save same object to another attribute so that
+        # AdaptivePlaywrightCrawlerStatistics specific methods can be access in "type safe manner".
+        self.adaptive_statistics = statistics
```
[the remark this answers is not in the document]
Pijukatel: I changed it to save only adaptive crawling related stuff: `self.predictor_state`
This way basic_crawler deals with the Statistics
and AdaptiveCrawler deals with the extra that is in predictor_state.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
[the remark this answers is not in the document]
Pijukatel: It is now generic. Making it work with generic `AbstractHttpCrawler` makes it internally more complex, but it gives more flexibility in usage.

External facing code for common use cases should be on the same level of complexity as before.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,316 @@
...
+@docs_group('Data structures')
+class _BeautifulsoupCrawlerAdditionalOptions(_HttpCrawlerOptions):
+    """Additional options that can be specified for BeautifulsoupCrawler."""
+
+    parser: NotRequired[BeautifulSoupParserType]
+    """Parser type used by BeautifulSoup."""
+
+
+@docs_group('Data structures')
+class _PlaywrightCrawlerAdditionalOptions(TypedDict):
```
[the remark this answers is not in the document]
Pijukatel: Moved

src/crawlee/crawlers/_adaptive_playwright/__init__.py
```diff
@@ -0,0 +1,11 @@
+from crawlee.crawlers._adaptive_playwright._adaptive_playwright_crawler import AdaptivePlaywrightCrawler
+from crawlee.crawlers._adaptive_playwright._adaptive_playwright_crawling_context import (
+    AdaptivePlaywrightCrawlingContext,
+)
+
+__all__ = [
+    'AdaptivePlaywrightCrawler',
+    'AdaptivePlaywrightCrawlingContext',
+    'HttpCrawlerOptions',
+    'ParsedHttpCrawlingContext',
```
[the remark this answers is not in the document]
Pijukatel: Done

Pijukatel:
What is missing: Documentation, example codes, template.
(Most of that should be probably added only after we have full predictor done as well, so that people are not asking about documented, but incomplete feature.)

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,456 @@
...
+
+        if should_detect_rendering_type:
+            # Save copy of global state from `use_state` before it can be mutated by browser crawl.
+            # This copy will be used in the static crawl to make sure they both run with same conditions and to
+            # avoid static crawl to modify the state.
+            # (This static crawl is performed only to evaluate rendering type detection.)
+            kvs = await context.get_key_value_store()
+            default_value = dict[str, JsonSerializable]()
+            old_state: dict[str, JsonSerializable] = await kvs.get_value(BasicCrawler.CRAWLEE_STATE_KEY, default_value)
+            old_state_copy = deepcopy(old_state)
```
janbuchar, line 406: This may run in parallel - we need to watch out for race conditions. I know that the JS version might not handle them optimally either, but we should try harder here.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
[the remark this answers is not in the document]
janbuchar: Cool! The private method override is still here though :slightly_smiling_face:. Maybe it is really necessary, but there are some potential issues. First one that comes to mind is that in case of a detection, the two request handler runs will share the `reqeust_handler_timeout`, which is unexpected.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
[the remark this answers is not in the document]
janbuchar: Hm, I think I understand it now. I think we should rename the method so that it's less similar to `commit_request_handler_result`.

janbuchar (commented):
I like the direction where this is going. Here's a bunch of comments :slightly_smiling_face:

src/crawlee/crawlers/_adaptive_playwright/_rendering_type_predictor.py
```diff
@@ -0,0 +1,36 @@
...
+
+class RenderingTypePredictor(ABC):
+    @abstractmethod
+    def predict(self, url: str, label: str | None) -> RenderingTypePrediction: ...
+
+    @abstractmethod
+    def store_result(self, url: str, label: str | None, crawl_type: RenderingType) -> None: ...
+
+
+class DefaultRenderingTypePredictor(RenderingTypePredictor):
```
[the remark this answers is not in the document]
Pijukatel: I would call it `SoonToBeDeletedRenderingTypePredictor`, but your name is better.

src/crawlee/crawlers/_adaptive_playwright/_rendering_type_predictor.py
```diff
@@ -0,0 +1,36 @@
...
+    rendering_type: RenderingType
+    detection_probability_recommendation: float
+
+
+class RenderingTypePredictor(ABC):
+    @abstractmethod
+    def predict(self, url: str, label: str | None) -> RenderingTypePrediction: ...
+
+    @abstractmethod
+    def store_result(self, url: str, label: str | None, crawl_type: RenderingType) -> None: ...
```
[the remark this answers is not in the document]
Pijukatel: 100% agree. Predictor can decide what to pick from it. No need to limit it in advance.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,456 @@
...
+        Args:
+            requests: The requests to be enqueued before the crawler starts.
+            purge_request_queue: If this is `True` and the crawler is not being run for the first time, the default
+                request queue will be purged.
+        """
+        contexts_to_enter = [
+            cm
+            for cm in self._static_context_pipeline.needed_contexts + self._pw_context_pipeline.needed_contexts
+            if cm and getattr(cm, 'active', False) is False
+        ]
```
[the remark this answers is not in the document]
Pijukatel: I removed dummy statistics, by make them "active" in init already. So only browser_pool remains.

We should still check if it is already active or not . Too bad [async nullcontext](https://docs.python.org/3/library/contextlib.html#contextlib.nullcontext) is not available in Python3.9.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,456 @@
...
+        *,
+        static_parser: AbstractHttpParser[TStaticParseResult],
+        rendering_type_predictor: RenderingTypePredictor | None = None,
+        result_checker: Callable[[RequestHandlerRunResult], bool] | None = None,
+        result_comparator: Callable[[RequestHandlerRunResult, RequestHandlerRunResult], bool] | None = None,
+        static_crawler_specific_kwargs: _HttpCrawlerOptions | None = None,
+        playwright_crawler_specific_kwargs: PlaywrightCrawlerAdditionalOptions | None = None,
+        **kwargs: Unpack[_BasicCrawlerOptions],
+    ) -> None:
+        """A default constructor.
```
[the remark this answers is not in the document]
Pijukatel: Ok

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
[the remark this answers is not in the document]
Pijukatel: Not sure if mean this, but I guess it could be problem as the top handler has same timeout as delegated handlers. And in some uncommon situations the first delegated handler could use up all the time and the second would not get chance to run as the top handler would time out. So that could be solved by just setting the timeout of delegated handlers to some fraction of the top handler. Something like "delegated_timeout" = 0.45 * "top_handler_timeout"

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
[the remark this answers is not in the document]
janbuchar: Discussed offline - we should be fine if we just make the `state_model` parameter mandatory

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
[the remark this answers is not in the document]
janbuchar: We decided to move the `wait_for` around `__run_request_handler` into the method itself and make it protected instead of private

janbuchar (commented):
I'll have to read this one more time, but it looks alright!

src/crawlee/statistics/_statistics.py
```diff
@@ -100,11 +99,35 @@ def __init__(
 
         self._log_message = log_message
         self._periodic_message_logger = periodic_message_logger or logger
+        self._log_interval = log_interval
```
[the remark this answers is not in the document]
Pijukatel: I know. My intention for storing it was for possibility to create copies of such object. 
I tried different approach now and that makes it possible to completely delete AdaptivePlaywrightCrawlerStatistics
https://github.com/apify/crawlee-python/pull/872/commits/f219453f258c027620d30e296f8cd5aefe873b98

vdusek (commented):
Am I correct that there is only the `docs/examples/code/adaptive_playwright_crawler.py` code example file? No MDX? You might forgot to include it :slightly_smiling_face: .

Pijukatel:
> Am I correct that there is only the `docs/examples/code/adaptive_playwright_crawler.py` code example file? No MDX? You might forgot to include it 🙂 . And maybe AdaptiveCrawler would deserve a whole guide and not just an example?

That is on purpose and justification is written in PR description -> Missing in this PR -> Documentation

src/crawlee/crawlers/_playwright/_playwright_crawler.py
```diff
@@ -285,3 +287,35 @@ def pre_navigation_hook(self, hook: Callable[[PlaywrightPreNavCrawlingContext],
             hook: A coroutine function to be called before each navigation.
         """
         self._pre_navigation_hooks.append(hook)
+
+
+@docs_group('Data structures')
+class PlaywrightCrawlerAdditionalOptions(TypedDict):
```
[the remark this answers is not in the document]
Pijukatel: Yes, they look as expected. The composed options contain all the stuff in the docs.

docs/examples/code/adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,55 @@
...
+            some_locator = context.page.locator('div').first
+            # Code that will be executed only in Playwright crawl.
+            # Trying to access `context.page` in static crawl will throw `AdaptiveContextError`.
+
+            await some_locator.wait_for()
+            # Do stuff with locator...
+            context.log.info(f'Playwright processing of: {context.request.url} ...')
+        except AdaptiveContextError:
+            # Code that will be executed in only in static crawl
+            context.log.info(f'Static processing of: {context.request.url} ...')
```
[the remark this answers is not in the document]
Pijukatel: As of now this example is more of a starting point for the reviewers and not really for end users given the fact that it is not mentioned in docs. I agree with the changes you suggest, but I would prefer to do them once the context helpers are in place, so that the example is matching the current state of the code.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
```diff
@@ -0,0 +1,131 @@
...
+        context_kwargs = {field.name: getattr(context, field.name) for field in fields(context)}
+        # Remove playwright specific attributes and pass them as private instead to be available as property.
+        context_kwargs['_response'] = context_kwargs.pop('response')
+        context_kwargs['_page'] = context_kwargs.pop('page')
+        context_kwargs['_infinite_scroll'] = context_kwargs.pop('infinite_scroll')
+        # This might not be always available.
+        protocol_guess = await context_kwargs['_page'].evaluate('() => performance.getEntries()[0].nextHopProtocol')
+        http_response = await _HttpResponse.from_playwright_response(
+            response=context.response, protocol=protocol_guess or ''
+        )
```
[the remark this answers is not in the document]
Pijukatel: PlaywrightCrawlingContext does not have HttpResponse like object, so it does not need to deal with this. 
AdaptivePlaywrightCrawlingContext on the other hand has to implement all stuff required by ParsedHttpCrawlingContext and that requires this HttpResponse like object. That is why the factory method is doing it.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
```diff
@@ -0,0 +1,131 @@
...
+class AdaptiveContextError(RuntimeError):
+    pass
+
+
+TStaticParseResult = TypeVar('TStaticParseResult')
+
+
+@dataclass(frozen=True)
+@docs_group('Data structures')
+class AdaptivePlaywrightCrawlingContext(Generic[TStaticParseResult], ParsedHttpCrawlingContext[TStaticParseResult]):
```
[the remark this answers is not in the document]
Pijukatel: Done

src/crawlee/crawlers/_adaptive_playwright/_result_comparator.py
```diff
@@ -0,0 +1,45 @@
...
+
+    return (
+        (result_1.push_data_calls == result_2.push_data_calls)
+        and (result_1.add_requests_calls == result_2.add_requests_calls)
+        and (result_1.key_value_store_changes == result_2.key_value_store_changes)
+    )
+
+
+def push_data_only_comparator(result_1: RequestHandlerRunResult, result_2: RequestHandlerRunResult) -> bool:
+    """Compare results by comparing their push dara calls. Ignore other parts of results in comparison."""
```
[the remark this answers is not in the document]
Pijukatel: Done

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,439 @@
...
+            result_checker=result_checker,
+            result_comparator=result_comparator,
+            static_parser=ParselParser(),
+            static_crawler_specific_kwargs=static_crawler_specific_kwargs,
+            playwright_crawler_specific_kwargs=playwright_crawler_specific_kwargs,
+            statistics=adaptive_statistics,
+            **kwargs,
+        )
+
+    async def crawl_one_with(
```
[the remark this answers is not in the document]
Pijukatel: Done

tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,471 @@
...
+
+
+@pytest.mark.parametrize(
+    ('expected_pw_count', 'expected_static_count', 'rendering_types', 'detection_probability_recommendation'),
+    [
+        pytest.param(0, 2, cycle(['static']), cycle([0]), id='Static only'),
+        pytest.param(2, 0, cycle(['client only']), cycle([0]), id='Client only'),
+        pytest.param(1, 1, cycle(['static', 'client only']), cycle([0]), id='Mixed'),
+        pytest.param(2, 2, cycle(['static', 'client only']), cycle([1]), id='Enforced rendering type detection'),
+    ],
```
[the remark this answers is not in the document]
janbuchar: The only reason is so that the individual parameters could be keyworded.  This way, it's pretty hard to tell what each parameter means.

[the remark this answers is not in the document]
Pijukatel: I disagree with that. In **reasonably sized parameter set** the description of parameters is close to the parameters.
This is standard use, see examples in the documentation: https://docs.pytest.org/en/stable/example/parametrize.html

I think your point makes sense with very large parameter sets. The larger the parameter set is, the more sense it makes. I do not think 4x4 parameter set is large.

Let someone else arbitrate this. @vdusek could you please share your opinion on the matter?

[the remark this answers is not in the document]
janbuchar: I mean, when I look at the decorator, I have no idea which number is expected static and which is the other one 🙂 and looking it up in the parameter list is not that comfortable.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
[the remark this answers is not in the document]
janbuchar: Makes sense, but I cannot help but wonder if `PlaywrightCrawlingContext` could not benefit from an `HttpResponse` object as well, simply to make the interfaces more similar to each other. Then, switching from `AdaptivePlaywrightCrawler` to the good old `PlaywrightCrawler` would be easier.

I'm not suggesting you do it in this PR though, so if others think that this makes sense, I can make a new issue out of this. @vdusek?

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
```diff
@@ -0,0 +1,148 @@
...
+
+        Raises `AdaptiveContextError` if accessed during static crawling.
+        """
+        if self._page is not None:
+            return self._page
+        raise AdaptiveContextError('Page is not crawled with PlaywrightCrawler.')
+
+    @classmethod
+    def from_pre_navigation_contexts(
+        cls, context: AdaptivePlaywrightPreNavCrawlingContext | BasicCrawlingContext
```
janbuchar, line 119: Is there a case where we create an adaptive pre-navigation context from another instance of an adaptive pre-navigation context, i.e., from itself?

tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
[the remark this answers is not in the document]
vdusek: > The only reason is so that the individual parameters could be keyworded. This way, it's pretty hard to tell what each parameter means.

I agree here with Honza. Seeing something like `0, 2, cycle(['static']), cycle([0])` doesn't provide much clarity about what each parameter means. While it's true that the tuple above includes the argument names and we also have the `id` which should somehow summarize it ... Well, I would slightly prefer the dataclasses, however, not strong preference.

src/crawlee/crawlers/_adaptive_playwright/_result_comparator.py
```diff
@@ -0,0 +1,45 @@
...
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+    from crawlee._types import RequestHandlerRunResult
+
+
+@dataclass(frozen=True)
+class SubCrawlerRun:
```
janbuchar, line 13: Weird that this is declared here and not used. Also, maybe `exception` could be made part of the `RequestHandlerRunResult`?

tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,470 @@
...
+        proxy_info: ProxyInfo | None = None,
+    ) -> CrawleePage:
+        crawlee_page = await super().new_page(page_id=page_id, browser_plugin=browser_plugin, proxy_info=proxy_info)
+        await crawlee_page.page.route(
+            '**/*',
+            lambda route: route.fulfill(
+                status=200, content_type='text/plain', body='<!DOCTYPE html><html><body>What a body!</body></html>'
+            ),
+        )
+        return crawlee_page
```
janbuchar, line 74: Not a suggestion for this PR, but it would be absolutely awesome if we could make an universal http interceptor for testing that combines `respx` and this playwright route thing. The best thing is that it probably wouldn't even be that hard, it may be enough to delegate to httpx from the `page.route` callback.

tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,470 @@
...
+
+    assert pw_handler_count == expected_pw_count
+    assert pw_hook_count == expected_pw_count
+
+    assert static_handler_count == expected_static_count
+    assert static_hook_count == expected_static_count
+
+
+async def test_adaptive_crawling_parcel(test_urls: list[str]) -> None:
+    """Top level test for parcel. Only one argument combination. (The rest of code is tested with bs variant.)"""
```
janbuchar, line 164: Could the static content parser also be a parameter? I believe that it could actually make the code shorter...

tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,470 @@
...
+            context.page  # noqa:B018 Intentionally "useless expression". Can trigger exception.
+            context.request.user_data['data'] = 'pw'
+        except AdaptiveContextError:
+            context.request.user_data['data'] = 'bs'
+
+    await crawler.run(test_urls[:1])
+    # Check that pre nav hooks does not influence each other
+    assert user_data_in_pre_nav_hook == [None, None]
+    # Check that pre nav hooks can modify context
+    assert user_data_in_handler == ['pw', 'bs']
```
janbuchar, line 224:
```suggestion
    # Check that repeated pre nav hook invocations do not influence each other while probing
    assert user_data_in_pre_nav_hook == [None, None]
    # Check that the request handler sees changes to user data done by pre nav hooks
    assert user_data_in_handler == ['pw', 'bs']
```

tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,470 @@
...
+            # page is available only if it was crawled by PlaywrightCrawler.
+            context.page  # noqa:B018 Intentionally "useless expression". Can trigger exception.
+            await context.push_data({'handler': 'pw'})
+        except AdaptiveContextError:
+            await context.push_data({'handler': 'bs'})
+
+    await crawler.run(test_urls[:1])
+
+    dataset = await crawler.get_dataset()
+    items = [item async for item in dataset.iterate_items()]
```
janbuchar, line 249: Why not use `get_data` directly?

tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,470 @@
...
+
+    # Enforced rendering type detection will trigger both sub crawlers, but only pw crawler result is saved.
+    assert items == [{'handler': 'pw'}]
+
+
+@pytest.mark.parametrize(
+    ('pw_saved_data', 'static_saved_data', 'expected_result_renderingl_type'),
```
janbuchar, line 256: tpyo

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,441 @@
...
+        """Call that will be used by the top crawler to run through the pipeline."""
+
+        async def from_pipeline_to_top_router(context: TStaticCrawlingContext) -> None:
+            adaptive_crawling_context = AdaptivePlaywrightCrawlingContext.from_parsed_http_crawling_context(context)
+            await self.top_router(adaptive_crawling_context)
+
+        return self.pipeline(top_context, from_pipeline_to_top_router)
+
+    def __str__(self) -> str:
+        return 'Static context pipeline'
```
janbuchar, line 129: I think these two are not really necessary anymore. The pipelines and the static parser could be stored directly in the `AdaptivePlaywrightCrawler` and the router is already there. This feels like too much indirection, combined with the `_run_subcrawler_pipeline` function and the `_crawl_one_with` method - I'd prefer to have everything here inside one of these two.

janbuchar (commented):
Okay, I finally managed to read this one through again :slightly_smiling_face: It needs some cleaning up, but everything important looks correct - nice job!

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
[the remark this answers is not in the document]
Pijukatel: That would have to be done carefully to avoid confusion and redundant information. `PlaywrightCrawlingContext`  already has Playwright specific  `Response`
https://playwright.dev/python/docs/api/class-response

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
Pijukatel: That is a mistake. It should be PlaywrightPreNavCrawlingContext, but that method actually works for anything that is BasicCrawlingContext, so that is why mypy did not complain. That type hint is actually redundant, I will keep just BasicCrawlingContext

tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
Pijukatel: Yes, I agree. When doing this I already though about our existing tests for PlaywrightCrawler. 
I drafted it into this issue https://github.com/apify/crawlee-python/issues/929
Please feel free top modify and rephrase it.

tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
Pijukatel: I want to test mainly the AdaptiveCrawelers created from the factory methods as those will be the ones used by the users. Objects created through `__init__` (which allow the parametrization through the content_parser) are also tested as consequence as all factory methods are calling it. 

You could parametrize the tests for both variant of the adaptive crawler. Initially I tried all tests to be parametrized for both bs and parcel variant. But that just inflates the amount of tests for I believe close to no benefit.
So this is more of a test strategy. Test all on one implementation - bs variant and do just one basic test on the other parcel variant.

tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
Pijukatel: Thanks. Fixed.

tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
Pijukatel: I wasn't even aware of it. Thanks.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
```diff
@@ -0,0 +1,148 @@
...
+        headers = HttpHeaders(response.headers)
+        status_code = response.status
+        # Can't find this anywhere in PlayWright, but some headers can include information about protocol.
+        # In firefox for example: 'x-firefox-spdy'
+        # Might be also obtained by executing JS code in browser: performance.getEntries()[0].nextHopProtocol
+        # Response headers capitalization not respecting http1.1 Pascal case. Always lower case in PlayWright.
+        http_version = protocol
+        _content = await response.body()
+
+        return cls(http_version=http_version, status_code=status_code, headers=headers, _content=_content)
```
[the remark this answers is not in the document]
Pijukatel: Renamed and added docs to be more explicit.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
Pijukatel: Ok, moved all to the crawler and replaced by factory method to create sub crawler pipeline call.

src/crawlee/crawlers/_adaptive_playwright/_result_comparator.py
Pijukatel: I prefer to use this composition and not modifying `RequestHandlerRunResult` since that one is used by other code where the exception would not make any sense.
I moved to where it is used.

tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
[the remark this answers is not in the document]
Pijukatel: Ok, I made the change. 
Here you can compare:
https://github.com/apify/crawlee-python/pull/872/commits/4f316b50d95599248f84f395f0eecccd6cd37b62

docs/examples/code/adaptive_playwright_crawler.py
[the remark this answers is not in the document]
janbuchar: I get that, but I believe that even though that example is not actually accessible from anywhere, it shouldn't contain code that goes against how the adaptive crawler should be used. I'd prefer to update that file now and then once more after we have those context helpers in place.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,410 @@
...
+    )
+    from crawlee.crawlers._basic._basic_crawler import _BasicCrawlerOptions
+    from crawlee.crawlers._playwright._playwright_crawler import _PlaywrightCrawlerAdditionalOptions
+
+
+TStaticParseResult = TypeVar('TStaticParseResult')
+TStaticCrawlingContext = TypeVar('TStaticCrawlingContext', bound=ParsedHttpCrawlingContext)
+
+
+class _NoActiveStatistics(Statistics):
```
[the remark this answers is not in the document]
Pijukatel: ok

tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
Pijukatel: ok

docs/examples/code/adaptive_playwright_crawler.py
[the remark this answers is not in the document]
Pijukatel: Ok, but unless we split pre-navigation hooks to two different functions I do not think we can avoid try-catch there.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1109,8 +1132,14 @@ async def __run_task_function(self) -> None:
             )
             raise
 
-    async def __run_request_handler(self, context: BasicCrawlingContext) -> None:
-        await self._context_pipeline(context, self.router)
+    async def _run_request_handler(self, context: BasicCrawlingContext) -> None:
```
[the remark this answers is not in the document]
Pijukatel: Yes, I can do that, but I am worried about how much sense it would make outside of AdaptiveCrawler. 

Getting rid of one extra method that makes the AdaptiveCrawler behavior different from BasicCrawler by changing parent in a way that does make sense only from the child perspective. 

If the child needed behavior can be contained within the child (as it is now), then I would prefer to not make change to parent.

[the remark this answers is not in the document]
janbuchar: Well, the function is called `_run_request_handler`, so if it returns a `RequestHandlerResult`, I think nobody will be surprised.

The problematic part is the part where it would also accept the result - I'm suggesting that because creating the result in `_run_request_handler` would require us to also create the crawling context in there, and that might be a bit too much - we want to keep most of the functionality in `__run_task_function` so that there it can't be overriden in weird ways.

My gripe with `_push_result_to_context` is that it contains non-trivial logic that might even incur a performance hit just to copy the contents of a `RequestHandlerResult` into another empty one.

So, I'm pretty sure that returning the result from `_run_request_handler` is the right thing to do, but we can explore alternatives to also passing it into the function. Thoughts?

docs/examples/code/adaptive_playwright_crawler.py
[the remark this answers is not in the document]
janbuchar: Or unless we use the new `block_requests` context helper there and just have it do nothing in a static crawling context. Would that work?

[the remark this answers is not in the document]
Pijukatel: But that covers only some specific use cases. There can be any code in those hooks. So to cover general case and have 1 to 1 matching behavior with both PlaywrightCrawler and whatever variant of ParsedHttpCrawler, I still se just two options: Try-catch or separate hooks.

src/crawlee/crawlers/_basic/_basic_crawler.py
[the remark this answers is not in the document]
Pijukatel: _run_request_handler that accepts results and returns results and sometimes mutates them (all crawlers except adaptive), but does not have to(adaptive, that creates new results and has no use for the input results). I am pretty sure that it makes the code worse.

To justify this change by "might even incur a performance hit" is quite a premature optimization.  We should deal with this hypothetical performance hit if it proves to be an issue.

docs/examples/code/adaptive_playwright_crawler.py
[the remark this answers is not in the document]
janbuchar: Again, the point of the adaptive crawler is that you shouldn't need to make separate branches for browser/static. Is there any operation that you can do in a prenav hook in the static context but not in the browser? If not (I can't think of any, but that's not saying anything), then we can just provide `block_requests` (a possibly common use case) and `page` for all the other usage scenarios (but it will force browser-based crawling when used).

src/crawlee/crawlers/_basic/_basic_crawler.py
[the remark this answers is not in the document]
janbuchar: Well, in the "default" `BasicCrawler`, it's what, three lines of a difference? Assign the return value of `_run_request_handler` to `result`, and one added parameter and a return statement in `BasicCrawler._run_request_handler`. It does add some cognitive overhead, but it's not huge by any means. If you want to look for a way how to create the result directly in `_run_request_handler` and bind it to the context later on, we can do that as well.

The potential performance hit of the current solution is not the deciding factor, but launching a bunch of coroutines, one for every dataset item, enqueued link and modified key-value store item will be more expensive than just returning an object that we already have.

The important thing for me, however, is that "run two context pipelines, pick the result of one and let `BasicCrawler.__run_task_function` continue as usual" is more understandable (maybe even more elegant) than "pick the result of one context pipeline and replay it on the original crawling context so that the results look the same".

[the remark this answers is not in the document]
Pijukatel: I did the change and it makes a lot of sense in the context of AdpativePlaywrightCrawler, but in BasicCrawler it looks kind of weird.

docs/examples/code/adaptive_playwright_crawler.py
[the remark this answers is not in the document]
Pijukatel: Ok, I tried some compromise solution.

**@crawler.pre_navigation_hook**
This decorator registers hooks for both static and playwright sub crawlers  -> if anyone tries to access **context.page**, it will raise an exception. (The exception text points directly to the example below)


**@crawler.pre_navigation_hook(playwright_only=True)** 
This decorator registers hooks only playwright sub crawler  -> safe to access **context.page**



@crawler.pre_navigation_hook(playwright_only=True)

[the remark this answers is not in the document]
janbuchar: Cool, this is fine by me

src/crawlee/crawlers/_basic/_basic_crawler.py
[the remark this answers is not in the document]
janbuchar: I agree. The context and the result are intertwined in weird-ish ways and I'm not sure if a better solution exists. I contemplated making a map of `context -> result` pairs on the `BasicCrawler` that you could update in the `_run_request_handler` override - that would remove the need for the additional parameter and the return value as well. Not sure if there are any significant downsides.

janbuchar (approved):
LGTM.

Pijukatel:
I have to integrate properly `block_requests` changes from master

src/crawlee/crawlers/_basic/_basic_crawler.py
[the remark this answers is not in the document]
Pijukatel: I tried the suggested map approach. Seems better to me.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -290,6 +291,9 @@ def __init__(
         self._failed_request_handler: FailedRequestHandler[TCrawlingContext | BasicCrawlingContext] | None = None
         self._abort_on_error = abort_on_error
 
+        # Context of each request with matching result.
```
janbuchar, line 294: Can you please elaborate on the intended usage? Something like "inheritors can use this to override the result of individual request handler runs in `_run_request_handler`"?

Also I'm not sure about the naming, I guess I'd prefer something that mentions request handlers... but this one also works for me.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -908,9 +912,9 @@ async def send_request(
 
         return send_request
 
-    async def _commit_request_handler_result(
-        self, context: BasicCrawlingContext, result: RequestHandlerRunResult
-    ) -> None:
+    async def _commit_request_handler_result(self, context: BasicCrawlingContext) -> None:
```
janbuchar, line 915: Let's add a docblock here that explains that the actual data to be committed is taken from the `_context_result_map`.

src/crawlee/crawlers/_basic/_basic_crawler.py
Pijukatel: Done.

I thought about longer names, but it gets worse. Especially for map it is very convenient to use just two words to avoid any misinterpretation based on how the reader "parses" the name.

src/crawlee/crawlers/_basic/_basic_crawler.py
Pijukatel: Done

src/crawlee/crawlers/_adaptive_playwright/__init__.py
```diff
@@ -0,0 +1,20 @@
+try:
+    from ._rendering_type_predictor import RenderingTypePredictor
+except ImportError as exc:
+    raise ImportError(
```
janbuchar, line 4: Thanks for remembering this! We should also add a re-export to `crawlee.crawlers.__init__` and handle import errors consistently with the rest.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -886,9 +913,10 @@ async def send_request(
...
+        """Commit request handler result for the input `context`. Result is stored in `_context_result_map`."""
```
janbuchar, line 917:
```suggestion
        """Commit request handler result for the input `context`. Result is taken from `_context_result_map`."""
```

src/crawlee/crawlers/_adaptive_playwright/__init__.py
Pijukatel: Done.

(Btw. all doc and examples details will be finalized in the PR with context helpers as it is more convenient to write it for complete feature)

janbuchar (approved):

src/crawlee/crawlers/_adaptive_playwright/_result_comparator.py
```diff
@@ -0,0 +1,38 @@
...
+    # https://sdk.apify.com/docs/guides/getting-started?__hsfp=1136113150&__hssc=7591405.1.1735494277124&__hstc=7591405.e2b9302ed00c5bfaee3a870166792181.1735494277124.1735494277124.1735494277124.1
+
+    return (
+        (result_1.push_data_calls == result_2.push_data_calls)
+        and (result_1.add_requests_calls == result_2.add_requests_calls)
+        and (result_1.key_value_store_changes == result_2.key_value_store_changes)
+    )
+
+
+def push_data_only_comparator(result_1: RequestHandlerRunResult, result_2: RequestHandlerRunResult) -> bool:
```
vdusek, line 36: why not rather method on `RequestHandlerRunResult`?

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,425 @@
...
+    AdaptivePlaywrightPreNavCrawlingContext,
+)
+from crawlee.crawlers._adaptive_playwright._rendering_type_predictor import (
+    DefaultRenderingTypePredictor,
+    RenderingType,
+    RenderingTypePredictor,
+)
+from crawlee.crawlers._adaptive_playwright._result_comparator import (
+    create_default_comparator,
+)
```
vdusek, line 42: maybe it could be more readable if we use relative imports here, since they are from the same subpackage

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler_statistics.py
```diff
@@ -0,0 +1,17 @@
...
+from crawlee.statistics import StatisticsState
+
+
+@docs_group('Data structures')
+class AdaptivePlaywrightCrawlerStatisticState(StatisticsState):
+    model_config = ConfigDict(populate_by_name=True, ser_json_inf_nan='constants')
+
+    http_only_request_handler_runs: Annotated[int, Field(alias='http_only_request_handler_runs')] = 0
+    browser_request_handler_runs: Annotated[int, Field(alias='browser_request_handler_runs')] = 0
+    rendering_type_mispredictions: Annotated[int, Field(alias='rendering_type_mispredictions')] = 0
```
vdusek, line 17: Missing docstrings.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
```diff
@@ -0,0 +1,167 @@
...
+class AdaptiveContextError(RuntimeError):
+    pass
+
+
+TStaticParseResult = TypeVar('TStaticParseResult')
+
+
+@dataclass(frozen=True)
+@docs_group('Data structures')
+class AdaptivePlaywrightCrawlingContext(Generic[TStaticParseResult], ParsedHttpCrawlingContext[TStaticParseResult]):
```
vdusek, line 35: missing docstring

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
```diff
@@ -0,0 +1,167 @@
...
+@dataclass(frozen=True)
+@docs_group('Data structures')
+class AdaptivePlaywrightPreNavCrawlingContext(BasicCrawlingContext):
+    """This is just wrapper around BasicCrawlingContext or AdaptivePlaywrightCrawlingContext.
+
+    Trying to access `page` on this context will raise AdaptiveContextError if wrapped context is BasicCrawlingContext.
+    """
+
+    _page: Page | None = None
+    block_requests: BlockRequestsFunction | None = None
```
vdusek, line 111: missing docstring

src/crawlee/crawlers/_adaptive_playwright/_rendering_type_predictor.py
```diff
@@ -11,18 +11,21 @@
...
 
 from crawlee import Request
+from crawlee._utils.docs import docs_group
 
 UrlComponents = list[str]
 RenderingType = Literal['static', 'client only']
 FeatureVector = tuple[float, float]
 
 
+@docs_group('Data structures')
```
vdusek, line 21: missing docstrings

src/crawlee/crawlers/_adaptive_playwright/_rendering_type_predictor.py
```diff
@@ -11,18 +11,21 @@
...
 @dataclass(frozen=True)
 class RenderingTypePrediction:
     rendering_type: RenderingType
     detection_probability_recommendation: float
 
 
+@docs_group('Classes')
```
vdusek, line 28: missing docstrings

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
```diff
@@ -0,0 +1,167 @@
...
+        return self._content
+
+    @classmethod
+    async def from_playwright_response(cls, response: Response, protocol: str) -> Self:
+        headers = HttpHeaders(response.headers)
+        status_code = response.status
+        # Can't find this anywhere in PlayWright, but some headers can include information about protocol.
+        # In firefox for example: 'x-firefox-spdy'
+        # Might be also obtained by executing JS code in browser: performance.getEntries()[0].nextHopProtocol
+        # Response headers capitalization not respecting http1.1 Pascal case. Always lower case in PlayWright.
```
vdusek, line 163: This seems more a note from a code review rather than documentation meant to help someone understand the code, could you rephrase it a bit?

src/crawlee/crawlers/_adaptive_playwright/_result_comparator.py
```diff
@@ -0,0 +1,38 @@
...
+    # Fallback default comparator.
+    return push_data_only_comparator
+
+
+def full_result_comparator(result_1: RequestHandlerRunResult, result_2: RequestHandlerRunResult) -> bool:
+    """Compare results by comparing all their parts."""
+    # PlayWright can produce links with extra arguments compared to pure BS. Default comparator ignores this.
+    # Maybe full comparator should have flag about taking into account only urls without parameters.
+    # https://sdk.apify.com/docs/guides/getting-started
+    # https://sdk.apify.com/docs/guides/getting-started?__hsfp=1136113150&__hssc=7591405.1.1735494277124&__hstc=7591405.e2b9302ed00c5bfaee3a870166792181.1735494277124.1735494277124.1735494277124.1
```
vdusek, line 27: This seems more a note from a code review discussion ("maybe") rather than documentation meant to help someone understand the code, could you rephrase it a bit?

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,425 @@
...
+class AdaptivePlaywrightCrawler(
+    Generic[TStaticCrawlingContext, TStaticParseResult],
+    BasicCrawler[AdaptivePlaywrightCrawlingContext, AdaptivePlaywrightCrawlerStatisticState],
+):
+    """Adaptive crawler that uses both specific implementation of `AbstractHttpCrawler` and `PlaywrightCrawler`.
+
+    It tries to detect whether it is sufficient to crawl without browser (which is faster) or if
+    `PlaywrightCrawler` should be used (in case previous method did not work as expected for specific url.).
+
+    # TODO: Add example
```
vdusek, line 96: Forgotton TODO? If it is intented and it is going to happen in the future please link as issue to it.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,425 @@
...
+                non-default configuration.
+            kwargs: Additional keyword arguments to pass to the underlying `BasicCrawler`.
+        """
+        # Some sub crawler kwargs are internally modified. Prepare copies.
+        basic_crawler_kwargs_for_static_crawler = deepcopy(kwargs)
+        basic_crawler_kwargs_for_pw_crawler = deepcopy(kwargs)
+
+        # Adaptive crawling related.
+        self.rendering_type_predictor = rendering_type_predictor or DefaultRenderingTypePredictor()
+        self.result_checker = result_checker or (lambda result: True)  #  noqa: ARG005  # Intentionally unused argument.
```
vdusek, line 131:
```suggestion
        self.result_checker = result_checker or (lambda _: True)
```

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
```diff
@@ -87,6 +97,31 @@ def __init__(
         kwargs.setdefault('_logger', logging.getLogger(__name__))
         super().__init__(**kwargs)
 
+    @staticmethod
```
vdusek, line 100: Isn't this more a @classmethod?

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,425 @@
...
+
+        self._additional_context_managers = [*self._additional_context_managers, playwright_crawler._browser_pool]  # noqa: SLF001 # Intentional access to private member.
+
+        # Sub crawler pipeline related
+        self._pw_context_pipeline = playwright_crawler._context_pipeline  # noqa:SLF001  # Intentional access to private member.
+        self._static_context_pipeline = static_crawler._context_pipeline  # noqa:SLF001  # Intentional access to private member.
+        self._static_parser = static_parser
+
+    @staticmethod
+    def with_beautifulsoup_static_parser(
```
vdusek, line 189: No @classmethod?

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,425 @@
...
+            result_comparator=result_comparator,
+            static_parser=BeautifulSoupParser(parser=parser_type),
+            static_crawler_specific_kwargs=static_crawler_specific_kwargs,
+            playwright_crawler_specific_kwargs=playwright_crawler_specific_kwargs,
+            statistics=adaptive_statistics,
+            **kwargs,
+        )
+
+    @staticmethod
+    def with_parsel_static_parser(
```
vdusek, line 216: No @classmethod?

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,425 @@
...
+        static_parser: AbstractHttpParser[TStaticParseResult],
+        rendering_type_predictor: RenderingTypePredictor | None = None,
+        result_checker: Callable[[RequestHandlerRunResult], bool] | None = None,
+        result_comparator: Callable[[RequestHandlerRunResult, RequestHandlerRunResult], bool] | None = None,
+        static_crawler_specific_kwargs: _HttpCrawlerAdditionalOptions | None = None,
+        playwright_crawler_specific_kwargs: _PlaywrightCrawlerAdditionalOptions | None = None,
+        statistics: Statistics[AdaptivePlaywrightCrawlerStatisticState] | None = None,
+        **kwargs: Unpack[_BasicCrawlerOptions],
+    ) -> None:
+        """A default constructor. Recommended way to create instance is to call factory methods `with_*_static_parser`.
```
vdusek, line 111: We should probably use full names of alternative constructors rather than `with_*_static_parser` (because of the future symbol API docs rendering - https://github.com/apify/crawlee-python/issues/625).

docs/examples/code/adaptive_playwright_crawler.py
```diff
@@ -0,0 +1,57 @@
...
+        max_requests_per_crawl=5, playwright_crawler_specific_kwargs={'headless': False}
+    )
+
+    @crawler.router.handler(label='label')
+    async def request_handler_for_label(context: AdaptivePlaywrightCrawlingContext) -> None:
+        # Do some processing using `page`
+        some_locator = context.page.locator('div').first
+        await some_locator.wait_for()
+        # Do stuff with locator...
+        context.log.info(f'Playwright processing of: {context.request.url} ...')
```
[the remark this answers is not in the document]
janbuchar: Static crawls will throw an exception (no page for static crawling), so the crawler will retry it with playwright. Gradually, it will learn that `label='label'` means that it should crawl with playwright in all cases.

src/crawlee/crawlers/_adaptive_playwright/_result_comparator.py
Pijukatel: It is used only in adaptive crawling, so adding it to RequestHandlerRunResult might not make much sense in any other crawler. Also there can be many different comparators, these are just our default implementations, so it is kind of different from `__eq__`. But I see your point and I have no strong opinion about this. So if you are sure about this point, I will move it there.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
Pijukatel: (same as above)

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
Pijukatel: (same as above)

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
Pijukatel: What is the benefit? 

When I tried using it as class method, the only use of `cls` argument was to call the class itself for `__init__` and then I was unable to make mypy happy with generics `cls[genric_arg1, generic_arg2](...)`

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
Pijukatel: Ok

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
Pijukatel: Done

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
Pijukatel: Done

src/crawlee/crawlers/_adaptive_playwright/_result_comparator.py
Pijukatel: Rephrased

src/crawlee/crawlers/__init__.py
```diff
@@ -18,10 +18,31 @@
...
+    'RenderingTypePrediction',
+    'RenderingTypePredictor',
+):
+    from ._adaptive_playwright import (
+        AdaptivePlaywrightCrawler,
+        AdaptivePlaywrightCrawlingContext,
+        AdaptivePlaywrightPreNavCrawlingContext,
+        RenderingType,
+        RenderingTypePrediction,
+        RenderingTypePredictor,
```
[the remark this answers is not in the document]
Pijukatel: Exactly

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler_statistics.py
Pijukatel: Added

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
Pijukatel: Added

src/crawlee/crawlers/_adaptive_playwright/_rendering_type_predictor.py
Pijukatel: Added

src/crawlee/crawlers/_adaptive_playwright/_rendering_type_predictor.py
Pijukatel: Added

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
Pijukatel: Rephrased

docs/examples/code/adaptive_playwright_crawler.py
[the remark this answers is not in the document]
vdusek: thanks

src/crawlee/crawlers/__init__.py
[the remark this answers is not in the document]
vdusek: thanks

src/crawlee/crawlers/_adaptive_playwright/_result_comparator.py
vdusek: It does not have to be just the eq dunder method. Maybe just normal methods. But I will leave it up to you.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
vdusek: there is still:

```
from crawlee.crawlers._adaptive_playwright._result_comparator import (
    create_default_comparator,
)
```

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler_statistics.py
vdusek: and the fields? :slightly_smiling_face:

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
vdusek: I believed the `@classmethod` decorator is primarily used to indicate an alternative constructor. We can also leverage it for documentation rendering (if we aren't already doing so).

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
Pijukatel: Added

src/crawlee/crawlers/_adaptive_playwright/_result_comparator.py
Pijukatel: I looked at it again and when I move it inside `RequestHandlerRunResult`, then it looks weird for the `RequestHandlerRunResult` to contain static methods like `push_data_only_comparator(result_1: RequestHandlerRunResult, result_2: RequestHandlerRunResult)` so next natural step would be making them bound methods `push_data_only_comparator(self, other_result: RequestHandlerRunResult)` which  goes against the way it is supposed to be used as standalone unbound comparator function that can be passed by user without any connection to any `RequestHandlerRunResult` instance.

So all that considered I think it is better in separate file and isolated to only adpative crawler.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler_statistics.py
Pijukatel: Now they have it as well :-)

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
Pijukatel: Right, done.

src/crawlee/crawlers/_adaptive_playwright/_result_comparator.py
vdusek: Agreed. Thanks for the explanation.

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
vdusek: I am ok with the current state.

PR #876, opened 2025-01-06
fix: Make result of RequestList.is_empty independent of fetch_next_request calls

janbuchar:
The old version of `RequestList` only updated the `_is_empty` flag on `fetch_next_request` calls, which was not enough. This PR updates it so that the `is_empty` method also tries to dequeue a request from the iterator before returning a result.

### DIFF
--- a/src/crawlee/request_loaders/_request_list.py
+++ b/src/crawlee/request_loaders/_request_list.py
@@ -39,13 +39,12 @@ class RequestList(RequestLoader):
         self._assumed_total_count = 0
 
         self._in_progress = set[str]()
-        self._is_empty = False
+        self._next: Request | None = None
 
         if isinstance(requests, AsyncIterable):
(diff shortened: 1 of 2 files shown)


src/crawlee/request_loaders/_request_list.py
```diff
@@ -95,6 +91,17 @@ async def mark_request_as_handled(self, request: Request) -> None:
     async def get_handled_count(self) -> int:
         return self._handled_count
 
+    async def _ensure_next_request(self) -> None:
+        if self._requests_lock is None:
```
Pijukatel, line 95: I see it was like this before the change, but what is the reason for not setting the lock in the __init__ and then avoiding this None check?

vdusek (commented):
Would it be possible to add a test case that covers the issue from the SDK?

vdusek (approved):
LGTM

PR #883, opened 2025-01-08
docs: fix multiline `Args` documentation

barjin:
|before|after|
|---|---|
| obrazek | obrazek |

Bumps the `@apify/docusaurus-plugin-typedoc-api` package to the version with patched comment parsing (see https://github.com/apify/docusaurus-plugin-typedoc-api/pull/37)

The second commit turns naked links to Markdown links with the `[]()` syntax. This is purely for aesthetics (I didn't like seeing the same long URL twice under each other). If this is undesired - e.g. breaks something or goes against our style guide - I'm happy to revert.


vdusek (approved):
Thanks

PR #899, opened 2025-01-13
fix: Fix the usage of Configuration

vdusek:
### Description

- Utilized all the `Configuration` fields, except for the `chrome_executable_path` and `xvfb`.
- Enhanced the `Configuration` fields documentation.

### Issues

- Closes: #670

### DIFF
--- a/src/crawlee/_autoscaling/snapshotter.py
+++ b/src/crawlee/_autoscaling/snapshotter.py
@@ -41,7 +41,7 @@ class Snapshotter:
         *,
         event_loop_snapshot_interval: timedelta = timedelta(milliseconds=500),
         client_snapshot_interval: timedelta = timedelta(milliseconds=1000),
-        max_used_cpu_ratio: float = 0.95,
+        max_used_cpu_ratio: None | float = None,
         max_memory_size: ByteSize | None = None,
         max_used_memory_ratio: float = 0.9,
         max_event_loop_delay: timedelta = timedelta(milliseconds=50),
@@ -58,10 +58,10 @@ class Snapshotter:
             event_loop_snapshot_interval: The interval at which the event loop is sampled.
             client_snapshot_interval: The interval at which the client is sampled.
             max_used_cpu_ratio: Sets the ratio, defining the maximum CPU usage. When the CPU usage is higher than
-                the provided ratio, the CPU is considered overloaded.
+                the provided ratio, the CPU is considered overloaded. The default value is taken from
+                the `Configuration`.
             max_memory_size: Sets the maximum amount of system memory to be used by the `AutoscaledPool`. If `None`
-                is provided, the max amount of memory to be used is set to one quarter of total system memory.
-                I.e. on a system with 8192 MB, the `AutoscaledPool` will only use up to 2048 MB of memory.
+                is provided, the max amount of memory to be used is determined by the `available_memory_ratio`.
             max_used_memory_ratio: Sets the ratio, defining the maximum ratio of memory usage. When the memory usage
                 is higher than the provided ratio of `max_memory_size`, the memory is considered overloaded.
             max_event_loop_delay: Sets the maximum delay of the event loop. When the delay is higher than the provided
@@ -69,7 +69,9 @@ class Snapshotter:
             max_client_errors: Sets the maximum number of client errors (HTTP 429). When the number of client errors
                 is higher than the provided number, the client is considered overloaded.
             snapshot_history: Sets the time interval for which the snapshots are kept.
-            available_memory_ratio: How big part of the system memory should be used if `max_memory_size` is not given.
+            available_memory_ratio: The maximum proportion of system memory to use. If `max_memory_size` is not
+                provided, this ratio is used to calculate the maximum memory. The default value is taken from
+                the `Configuration`.
             reserve_memory_ratio: Fraction of memory kept in reserve. Used to calculate critical memory overload
                 threshold.
             memory_warning_cooldown_period: Minimum time interval between logging successive critical memory overload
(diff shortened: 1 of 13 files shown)


src/crawlee/storage_clients/_memory/_memory_storage_client.py
```diff
@@ -68,6 +68,9 @@ def __init__(
     ) -> None:
         """A default constructor.
 
+        In the most cases you should use the `from_config` constructor to create an new instance based on
```
janbuchar, line 71:
```suggestion
        In most cases you should use the `from_config` constructor to create an new instance based on
```

src/crawlee/events/_local_event_manager.py
```diff
@@ -32,8 +33,12 @@ def __init__(
     ) -> None:
         """A default constructor.
 
+        In the most cases you should use the `from_config` constructor to create an new instance based on
+        the provided configuration.
+
         Args:
-            system_info_interval: Interval at which `SystemInfo` events are emitted.
+            system_info_interval: Interval at which `SystemInfo` events are emitted. The default value is taken
+                from the `Configuration`.
```
janbuchar, line 41: Weird thing to say here - the constructor does not receive the `Configuration` object even if `from_config` is used.

src/crawlee/events/_event_manager.py
```diff
@@ -68,6 +68,7 @@ def __init__(
 
         Args:
             persist_state_interval: Interval between emitted `PersistState` events to maintain state persistence.
+                The default value is taken from the `Configuration`.
```
janbuchar, line 71: Same as in `LocalEventManager` - this does not know anything about the Configuration.

src/crawlee/_autoscaling/snapshotter.py
janbuchar, line 1: Weird thing that the `max_used_cpu_ratio` and `max_used_memory_ratio` arguments don't seem to be passed anywhere in the codebase - that makes them effectively equal to whatever is in `Configuration`.

src/crawlee/_autoscaling/snapshotter.py
```diff
@@ -112,8 +113,11 @@ def _get_sorted_list_by_created_at(input_list: list[T]) -> SortedList[T]:
         return SortedList(input_list, key=attrgetter('created_at'))
 
     @staticmethod
-    def _get_default_max_memory_size(available_memory_ratio: float) -> ByteSize:
-        """Default `memory_max_size` is 1/4 of the total system memory."""
+    def _get_default_max_memory_size(memory_mbytes: int | None, available_memory_ratio: float) -> ByteSize:
```
janbuchar, line 116: The name doesn't fit well after the changes :slightly_smiling_face:

src/crawlee/configuration.py
```diff
@@ -207,7 +217,7 @@ class Configuration(BaseSettings):
             )
         ),
     ] = None
-    """This setting is currently unused. For more details, see https://github.com/apify/crawlee-python/issues/670."""
+    """This setting is currently unused."""
```
janbuchar, line 220: I think it's safe to just drop this and never look back.

src/crawlee/configuration.py
```diff
@@ -229,7 +243,7 @@ class Configuration(BaseSettings):
             )
         ),
     ] = False
-    """This setting is currently unused. For more details, see https://github.com/apify/crawlee-python/issues/670."""
+    """This setting is currently unused."""
```
janbuchar, line 246: I think it's safe to just drop this and never look back.

janbuchar (commented):
Nice, good stuff!

src/crawlee/configuration.py
vdusek: same as the other one

src/crawlee/events/_local_event_manager.py
vdusek: This is a forgotten docstring update from the previous version. Thanks.

src/crawlee/events/_event_manager.py
vdusek: This is a forgotten docstring update from the previous version. Thanks.

src/crawlee/_autoscaling/snapshotter.py
vdusek: Yeah, that is a good point. Do you suggest removing them?

Or maybe, in the future, we could consider exposing `Snapshotter` and `AutoscaledPool` to the public, allowing (advanced) users to configure and pass their own instances to the Crawler. What do you think?

src/crawlee/configuration.py
vdusek: I wrote it down to this issue: https://github.com/apify/crawlee-python/issues/906

src/crawlee/_autoscaling/snapshotter.py
janbuchar: We already have `ConcurrencySettings`, and it's hard for me to imagine a use case where someone would want to swap out the `AutoscaledPool` completely. Same goes for `Snapshotter`, which is super internal.

janbuchar: So yeah, I wouldn't mind nuking them :grin:

src/crawlee/configuration.py
janbuchar: I mean, I don't think there's any harm in removing an option that doesn't do anything, but suit yourself :slightly_smiling_face:

src/crawlee/configuration.py
janbuchar, line 1: I like how the configuration options refer to the places where they're supposed to be used. Hope we can keep that in sync :slightly_smiling_face:

src/crawlee/_autoscaling/snapshotter.py
B4nan: Makes sense to me, these things were initially designed before the configuration object became a thing, I think it all originated there.

vdusek: Alright, but why just `max_used_cpu_ratio` and `max_used_memory_ratio`.

In the `BasicCrawler`, we do this:

```python
self._snapshotter = Snapshotter()
```

And it is the only place where it gets instantiated.

Following this approach, we can eliminate all input arguments, convert them to constants, and configure the remaining parameters through `Configuration`.

janbuchar: Why bot do the same as we do in `EventManager` and `MemoryStorageClient`? A constructor that accepts stuff and a `from_config` factory method?

vdusek: Yeah, that makes sense.

src/crawlee/_autoscaling/snapshotter.py
```diff
@@ -36,84 +38,100 @@ class Snapshotter:
...
+        )
+
     @staticmethod
     def _get_sorted_list_by_created_at(input_list: list[T]) -> SortedList[T]:
         return SortedList(input_list, key=attrgetter('created_at'))
 
     @staticmethod
-    def _get_default_max_memory_size(available_memory_ratio: float) -> ByteSize:
-        """Default `memory_max_size` is 1/4 of the total system memory."""
+    def _determine_max_memory_size(memory_mbytes: int | None, available_memory_ratio: float) -> ByteSize:
```
janbuchar, line 126: I wouldn't mind inlining this into `from_config` :slightly_smiling_face: But I'll leave that up to you

janbuchar (approved):

PR #904, opened 2025-01-14
docs: update request loaders & storages guide

vdusek:


docs/guides/request_loaders.mdx
```diff
@@ -39,20 +39,31 @@ classDiagram
...
+    + is_empty()
+    + is_finished()
+    + get_handled_count()
+    + get_total_count()
+    + to_tandem()
 }
 
 class RequestManager {
     <<abstract>>
     _attributes_
```
janbuchar, line 62: Are the `_attributes_` lines necessary?

vdusek: Otherwise there is an empty row. But they are not necessary. What do you think?

image

Pijukatel (approved):

docs/guides/request_loaders.mdx
vdusek: Nice, but unfortunately, it seems to work only for entirely empty classes.

image

Additionally, there doesn't seem to be any further configuration options available at this moment - [Mermaid Class Diagram Config Properties](https://mermaid.js.org/config/schema-docs/config-defs-class-diagram-config.html#classdiagramconfig-properties).

So I would probably leave it as it is for now.

janbuchar: It may be worth it to raise that with the maintainers... In my opinion, the empty attributes section is marginally better than the "attributes" label.

vdusek: Okay, updated.

janbuchar (approved):

Issue #895, opened 2025-01-10
High RAM consumption by any HttpCrawlers when working with large proxy pools


Mantisus:
When working with large proxy pools (e.g., Apify RESIDENTIAL), I observe significant RAM usage growth. Memory consumption increases by more than a gigabyte within a few minutes during active scraping.

My version:
The issue appears to be related to creating an HTTP session for each proxy https://github.com/apify/crawlee-python/blob/master/src/crawlee/http_clients/_httpx.py#L132 combined with the high default maximum SessionPool size of 1000 sessions. This results in the crawler creating a new HTTP session for almost every new request during its initial run.

The absence of cleanup logic for created HTTP sessions will likely worsen the situation when the proxy pool contains a large number of "bad" proxies.

B4nan:
AFAIK this works just fine in the JS version, it feels like a memory leak in the python version somewhere to me. We should investigate why this happens, I don't see why 1000 sessions should be a problem.

Either way, a complete reproduction for this is needed unless you plan to deal with this on your own.

Mantisus:
I think there is a difference between Python and JS at the level of HTTP clients being used, as memory testing for 1000 httpx.AsyncClient instances shows consumption of about 1.7 gigabytes.

```python
import asyncio
import resource
from httpx import AsyncClient, AsyncHTTPTransport


def log_memory_usage():
    usage = resource.getrusage(resource.RUSAGE_SELF)
    memory_mb = usage.ru_maxrss / 1024  
    print(f"Memory usage: {memory_mb:.2f} MB")

async def create_httpx_sessions():
    sessions = {
        i: AsyncClient(
            http1=True,
            http2=True,
            proxy=f"http://{i}:{i}",
            transport=(AsyncHTTPTransport(
                                    proxy=f"http://{i}:{i}", 
                                    http1=True, 
                                    http2=True))
        ) for i in range(1000)
    }
    return sessions

async def run():
    print("Starting test...")
    log_memory_usage()
    
    sessions = await create_httpx_sessions()

    print("Sessions created")
    log_memory_usage()
    
def main() -> None:
    asyncio.run(run())
```

```bash
Starting test...
Memory usage: 26.57 MB
Sessions created
Memory usage: 1710.93 MB
```

The same thing happens in the `get_client` method - https://github.com/apify/crawlee-python/blob/master/src/crawlee/http_clients/_httpx.py#L217

janbuchar:
[...]

PR #905, opened 2025-01-15
fix: Optimize memory consumption for `HttpxHttpClient`, fix proxy handling

Mantisus:
### Description

- Use a shared transport and SSL context for all `httpx.AsyncClient` instances
- Fix proxy handling by implementing proper binding of `session_id` to `proxy_url`

### Issues

- Closes: #895

### Testing

- Updated test `tests/unit/proxy_configuration/test_tiers.py::test_successful_request_makes_tier_go_down`. In the previous version, when `session_id` was passed, we should not obtain a new proxy even if the tier changes. The updated test now aligns with both the documentation and the `test_retrying_request_makes_tier_go_up` test.

### DIFF
--- a/src/crawlee/http_clients/_httpx.py
+++ b/src/crawlee/http_clients/_httpx.py
@@ -125,10 +125,16 @@ class HttpxHttpClient(BaseHttpClient):
         )
         self._http1 = http1
         self._http2 = http2
-        self._verify = verify
+
         self._async_client_kwargs = async_client_kwargs
         self._header_generator = header_generator
 
+        self._ssl_context = httpx.create_ssl_context(verify=verify)
+        self._transport = _HttpxTransport(
+            http2=http2,
+            verify=self._ssl_context,
+            limits=httpx.Limits(max_connections=1000, max_keepalive_connections=200),
+        )
         self._client_by_proxy_url = dict[Optional[str], httpx.AsyncClient]()
 
     @override
@@ -222,13 +228,11 @@ class HttpxHttpClient(BaseHttpClient):
         if proxy_url not in self._client_by_proxy_url:
             # Prepare a default kwargs for the new client.
             kwargs: dict[str, Any] = {
-                'transport': _HttpxTransport(
-                    proxy=proxy_url, http1=self._http1, http2=self._http2, verify=self._verify
-                ),
+                'transport': self._transport,
                 'proxy': proxy_url,
                 'http1': self._http1,
                 'http2': self._http2,
-                'verify': self._verify,
+                'verify': self._ssl_context,
             }
 
             # Update the default kwargs with any additional user-provided kwargs.
--- a/src/crawlee/proxy_configuration.py
+++ b/src/crawlee/proxy_configuration.py
@@ -197,7 +197,7 @@ class ProxyConfiguration:
         else:
             raise RuntimeError('Invalid state')
 
-        if session_id is None or request is not None:
+        if session_id is None:
             url = urls[self._next_custom_url_index % len(urls)]
             self._next_custom_url_index += 1
             return url, proxy_tier
--- a/tests/unit/proxy_configuration/test_tiers.py
+++ b/tests/unit/proxy_configuration/test_tiers.py
@@ -84,13 +84,13 @@ async def test_successful_request_makes_tier_go_down() -> None:
 
     info = None
     for tier in tiered_proxy_urls:
-        info = await config.new_proxy_info('session_id', request_1, None)
+        info = await config.new_proxy_info(None, request_1, None)
         assert info is not None
         assert info.url == tier[0]
 
     for i in range(100):
         new_request = Request(url=f'http://some.domain/{i}', unique_key=str(i), id=str(i))
-        info = await config.new_proxy_info('session_id', new_request, None)
+        info = await config.new_proxy_info(None, new_request, None)
 
     assert info is not None
     assert info.url == tiered_proxy_urls[0][0]


Mantisus:
Something very strange is going on inside `httpx`. Since the changes made reduce memory consumption from 1.7 Gigabytes to 120 Megabytes, for 1000 httpx.AsyncClient

This is so, weird that someone might want to double-check it and tell me where I went wrong.

In order to test to exclude delays and errors that can cause real proxies, I tested using a local environment. Here is the configuration used

```docker
# compose.yaml

version: '3'
services:
  target-server:
    image: kennethreitz/httpbin
    ports:
      - "8000:80"
    networks:
      - test-net

  toxiproxy:
    image: ghcr.io/shopify/toxiproxy
    ports:
      - "8474:8474"
      - "8001-10000:8001-10000"
    networks:
      - test-net

networks:
  test-net:
    driver: bridge
```

```python
import asyncio
import os
from contextlib import suppress

import psutil
from toxiproxy import Toxiproxy
from crawlee import ConcurrencySettings, Request
from crawlee.http_clients import HttpxHttpClient
from crawlee.sessions import SessionPool
from crawlee.crawlers import ParselCrawler, ParsedHttpCrawlingContext

current_process = psutil.Process(os.getpid())

def log_memory_usage():
    current_size_bytes = int(current_process.memory_info().rss)
    for child in current_process.children(recursive=True):
        with suppress(psutil.NoSuchProcess):
            current_size_bytes += int(child.memory_info().rss)
    
    memory_mb = current_size_bytes / 1024 /1024
    return memory_mb

async def setup_toxiproxy(proxy_count: int = 1000):
    toxiproxy = Toxiproxy()
    
    toxiproxy.destroy_all()
    
    proxies = []
    i = 0
    while proxy_count > 0:
        try:
            port = 8001 + i
            toxiproxy.create(
                name=f"proxy_{i}",
                upstream="target-server:80",
                enabled=True,
                listen=f"0.0.0.0:{port}"
            )
            proxies.append(f'http://localhost:{port}')
            proxy_count -= 1
        except:
            pass
        i += 1
    
    print(f"Created {len(proxies)} proxies")
    return proxies

async def run():
    proxies = await setup_toxiproxy(1500)

    session_pool = SessionPool()
    http_client = HttpxHttpClient(
        headers={'accept-encoding': 'gzip, deflate, br, zstd',
                 'user-agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36', 'accept-language': 'en',
                 },
                 )
    crawler = ParselCrawler(
        http_client=http_client,
        concurrency_settings=ConcurrencySettings(min_concurrency=20, desired_concurrency=50),
        proxy_configuration=ProxyConfiguration(proxy_urls=proxies),
        max_requests_per_crawl=4000,
        session_pool=session_pool
    )

    async def request_handler(context: ParsedHttpCrawlingContext) -> None:
        ip = context.parsed_content.jmespath('origin').get()
        memory = log_memory_usage()
        context.log.info(f'Processing {context.request.url} with response ip {ip}, {context.session.id} {context.proxy_info.url} clients {len(http_client._client_by_proxy_url)} {memory:.2f} MB"')

    requests = [Request.from_url(url='http://httpbin.org/get', always_enqueue=True) for _ in range(5000)]
    await crawler.run(requests)


if __name__ == "__main__":
    asyncio.run(run())
```

Mantisus:
Also @janbucharwas right. Binding of session_id and proxy_url worked incorrectly. As a result, we were creating `httpx.AsyncClient` for each proxy in the pool. This didn't show up while I was using the Apify platform with Apify proxy for testing, but it did in the local test environment.

src/crawlee/http_clients/_httpx.py
```diff
@@ -125,10 +125,16 @@ def __init__(
...
         self._http1 = http1
         self._http2 = http2
-        self._verify = verify
+
         self._async_client_kwargs = async_client_kwargs
         self._header_generator = header_generator
 
+        self._ssl_context = httpx.create_ssl_context(verify=verify)
+        self._transport = _HttpxTransport(
+            http2=http2,
```
Pijukatel, line 134: Question: Previously this kwarg was passed `http1=self._http1`, why is it not passed now ?

src/crawlee/http_clients/_httpx.py
```diff
@@ -125,10 +125,16 @@ def __init__(
         )
         self._http1 = http1
         self._http2 = http2
-        self._verify = verify
+
         self._async_client_kwargs = async_client_kwargs
         self._header_generator = header_generator
 
+        self._ssl_context = httpx.create_ssl_context(verify=verify)
```
Pijukatel, line 132: Is there any potential drawback in sharing this context for all clients? Or are there any scenarios where different contexts for different clients would be desired?

src/crawlee/http_clients/_httpx.py
Mantisus: I missed that, thank you. 🙂

src/crawlee/http_clients/_httpx.py
Mantisus: In the context of our use case, I don't see what the possible problems are.

Using a separate context for each client might make sense in scenarios where individual clients use specific configurations (certificates, ciphers, TLS versions). But we still don't give users that level of control.

Maybe I'm missing something. But I think it should work for us.

B4nan: As long as it works with different domains, it should be fine.

src/crawlee/http_clients/_httpx.py
```diff
@@ -222,13 +229,11 @@ def _get_client(self, proxy_url: str | None) -> httpx.AsyncClient:
         if proxy_url not in self._client_by_proxy_url:
             # Prepare a default kwargs for the new client.
             kwargs: dict[str, Any] = {
-                'transport': _HttpxTransport(
-                    proxy=proxy_url, http1=self._http1, http2=self._http2, verify=self._verify
-                ),
```
janbuchar, line 227: I checked now and it looks like the `transport` doesn't need to know the proxy URL since proxying is handled by the `httpx.AsyncClient` (it creates a separate transport for that).

src/crawlee/http_clients/_httpx.py
```diff
@@ -125,10 +125,17 @@ def __init__(
...
+
         self._async_client_kwargs = async_client_kwargs
         self._header_generator = header_generator
 
+        self._ssl_context = httpx.create_ssl_context(verify=verify)
+        self._transport = _HttpxTransport(
+            http1=http1,
+            http2=http2,
+            verify=self._ssl_context,
+            limits=httpx.Limits(max_connections=1000, max_keepalive_connections=200),
```
janbuchar, line 137:
```suggestion
            limits=async_client_kwargs.get('limits', httpx.Limits(max_connections=1000, max_keepalive_connections=200)),
```

src/crawlee/http_clients/_httpx.py
```diff
@@ -222,13 +229,11 @@ def _get_client(self, proxy_url: str | None) -> httpx.AsyncClient:
...
             kwargs: dict[str, Any] = {
-                'transport': _HttpxTransport(
-                    proxy=proxy_url, http1=self._http1, http2=self._http2, verify=self._verify
-                ),
+                'transport': self._transport,
                 'proxy': proxy_url,
                 'http1': self._http1,
                 'http2': self._http2,
-                'verify': self._verify,
+                'verify': self._ssl_context,
```
janbuchar, line 236: We should probably set the `transport` and `verify` fields *after* the `kwargs.update(self._async_client_kwargs)` line. I believe we don't want to let users override `transport` ever, and `verify` is not part of `self._async_client_kwargs`, but it'll be more obvious how we handle it that way.

src/crawlee/proxy_configuration.py
```diff
@@ -197,7 +197,7 @@ async def _pick_url(
         else:
             raise RuntimeError('Invalid state')
 
-        if session_id is None or request is not None:
+        if session_id is None:
```
janbuchar, line 200: Did the `request is None` condition cause any trouble here?

tests/unit/proxy_configuration/test_new_proxy_info.py
```diff
@@ -106,12 +104,20 @@ async def test_rotates_proxies_with_sessions() -> None:
...
+    assert info is not None
+    assert info.url == proxy_urls[0]
+
     # Different sessions should get rotated proxies
     info = await config.new_proxy_info(sessions[1], None, None)
     assert info is not None
     assert info.url == proxy_urls[1]
 
-    info = await config.new_proxy_info(sessions[2], None, None)
+    info = await config.new_proxy_info(sessions[2], request, None)
```
janbuchar, line 120: But that way, we won't test the `request=None` case anywhere, right?

janbuchar (commented):
Just a couple of notes, looks great as a whole

tests/unit/proxy_configuration/test_new_proxy_info.py
janbuchar: True. We should probably test requests with no session though, assuming that there is a use case for this. Maybe @barjin would know better? He wrote the original tiered proxies functionality for JS crawlee.

src/crawlee/proxy_configuration.py
Mantisus: The condition `session_id is None or request is None`, will cause the proxy rotation for 
``python
info = await config.new_proxy_info(`session_id`, None, None)
info = await config.new_proxy_info(`session_id`, None, None)
```

Comment (janbuchar) [reply src/crawlee/proxy_configuration.py:200]: Sorry, I mean `request is not None` - the same as it was in the original code.

Comment (Mantisus) [reply src/crawlee/proxy_configuration.py:200]: Yes, it caused the proxy to rotate on each request (
```
info = await config.new_proxy_info(`session_id`, request, None)
info = await config.new_proxy_info(`session_id`, request, None)
```
Different proxies for the same session

tests/unit/proxy_configuration/test_new_proxy_info.py
Mantisus: added a similar test, without the session.

vdusek (commented):
Nice! And thanks for adding the tests. Could you please also update the docstring of `ProxyConfiguration.new_proxy_info` and `ProxyConfiguration.new_url` to explain the behavior when `None` values are passed in?

vdusek (approved):
Amazing, thank you

Issue #907, opened 2025-01-15
Timeout in user defined handler should cause retry


Pijukatel:
Timeouts in user defined handlers should be treated the same way as other exceptions raised from within the user handler.
This should cause retry.

Currently the request that times out will end up in strange state in RequestQueue and it will block crawler from closing for 300 secs until following triggers:
"The request queue seems to be stuck for 300.0s, resetting internal state."

(Regarding blocking sync code timeouts from user defined handlers: https://github.com/apify/crawlee-python/issues/908)

Issue #908, opened 2025-01-15
Request handler should be able to timeout blocking sync code in user defined handler


Pijukatel:
Current purely async implementation of request handler is not capable of triggering timeout for blocking sync code. Such code is created by users and we have no control over it. So we can't expect only async blocking code.

Following simple user defined handler can't currently trigger timeout even if it should.
```
@crawler.router.default_handler
async def handler(context: BasicCrawlingContext) -> None:
    time.sleep(some_big_time)
```

PR #909, opened 2025-01-15
fix: Fix crawler not retrying user handler if there was timeout in the handler

Pijukatel:
### Description

Fix crawler not retrying user handler if there was timeout in the handler.
Add parametrized test with one parameter set skipped with reason https://github.com/apify/crawlee-python/issues/908

### Issues

- Closes: #907


vdusek (approved):
Nice, thanks

janbuchar (approved):
Cool!

Issue #912, opened 2025-01-15
Code sample on home page does not work


gdvallance:
The documentation on [https://crawlee.dev/python/](https://crawlee.dev/python/), is incorrect, specifically, the code in "Now we are ready to execute our first Crawlee project:"

This is on a fresh install as of today (v0.5.1): via: `pip install crawlee[all] && playwright install` as per [https://crawlee.dev/python/docs/introduction/setting-up](https://crawlee.dev/python/docs/introduction/setting-up)

The error is here:

```python
from crawlee.playwright_crawler import PlaywrightCrawler, PlaywrightCrawlingContext
```
The import doesn't resolve.

Correct (taken from other documention) [https://crawlee.dev/python/docs/quick-start](https://crawlee.dev/python/docs/quick-start) is:

```python
from crawlee.crawlers import PlaywrightCrawler, PlaywrightCrawlingContext
```

vdusek:
Hi @gdvallance, thank you for reporting this! I will fix it ASAP.

PR #913, opened 2025-01-16
docs: Fix code sample on home page

vdusek:
### Description

- Fix code sample on home page and move it to a separate Python file.

### Issues

- Closes: #912

### Testing

- Docs were rendered locally.

### Checklist

- [ ] CI passed


docs/home_page_example.py
```diff
@@ -0,0 +1,43 @@
+import asyncio
```
B4nan, line 1: I would put this elsewhere, the `docs` folder will be versioned at some point (it will be copied to https://github.com/apify/crawlee/tree/master/website/versioned_docs for each feature release), but the homepage won't. I actually wanted to look at this rather sooner and at least have "latest stable and latest dev" so we don't confuse users with unreleased things in docs.

vdusek: I am wondering where to place it. The `website/` dir is currently excluded from the Python static analysis. And I am not sure whether we want to include it because of `node_modules`. Any suggestions?

B4nan: You could include just `*.py` files maybe? Where is that exclusion defined?

vdusek: In the `pyproject.toml` and `makefile`. 

I moved it to the `website/`, tested static analysis then, and it seems that there is no noticeable slowdown.

docs/introduction/07_saving_data.mdx
```diff
@@ -7,31 +7,21 @@ import ApiLink from '@site/src/components/ApiLink';
...
 
 import FinalCodeExample from '!!raw-loader!./code/07_final_code.py';
+import FirstCodeExample from '!!raw-loader!./code/07_first_code.py';
 
 A data extraction job would not be complete without saving the data for later use and processing. You've come to the final and most difficult part of this tutorial so make sure to pay attention very carefully!
 
 ## Save data to the dataset
 
-Crawlee provides a <ApiLink to="class/Dataset">`Dataset`</ApiLink> class, which acts as an abstraction over tabular storage, making it useful for storing scraping results. First, add the following import to the top of your file:
+Crawlee provides a <ApiLink to="class/Dataset">`Dataset`</ApiLink> class, which acts as an abstraction over tabular storage, making it useful for storing scraping results. To get started:
```
B4nan, line 16:
```suggestion
Crawlee provides a <ApiLink to="class/Dataset">`Dataset`</ApiLink> class, which acts as an abstraction over tabular storage, making it useful for storing scraping results. To get started:

```

B4nan (approved):
just one nit

Issue #304, opened 2024-07-15
Improve API docs of public components


vdusek:
Improve API docs of the public components, mainly:
- [x] `BasicCrawler`
- [x] `HttpCrawler`
- [x] `BeautifulSoupCrawler`
- [x] `ParselCrawler`
- [x] `PlaywrightCrawler`
- [x] `Dataset`
- [x] `KeyValueStore`
- [x] `RequestQueue`
- [x] `MemoryStorageClient`
- [x] `HttpxHttpClient`
- [x] `CurlImpersonateHttpClient`
- [x] `Configuration`
- [x] `EventManager`
- [x] `LocalEventManager`
- [x] `Request`
- [x] `Session`
- [x] `SessionPool`
- [x] `BrowserPool`
- [x] `PlaywrightBrowserController`
- [x] `PlaywrightBrowserPlugin`
- [x] `Statistics`

And functions:
- [x] `AddRequestsFunction`
- [x] `EnqueueLinksFunction`
- [x] `ExportToFunction`
- [x] `GetDataFunction`
- [x] `GetKeyValueStoreFunction`
- [x] `PushDataFunction`
- [x] `SendRequestFunction`
- [x] `UseStateFunction`

belloibrahv:
Hi @vdusek,

I’d like to work on this issue to improve the API docs for the listed public components. Before starting, I have a few quick questions:

1. Are there any specific guidelines or formats I should follow for the documentation?
2. Should the focus be on adding descriptions and examples, or should I also cover parameter types, return values, etc.?

Looking forward to your guidance!

vdusek:
Hi @belloibrahv, thanks for your interest in Crawlee. 

> Are there any specific guidelines or formats I should follow for the documentation?

- We follow the [Google style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).
- Maybe you can also check out the #324.

> Should the focus be on adding descriptions and examples, or should I also cover parameter types, return values, etc.?

- This should be only about the class docstrings. 
- You can check out the previous PRs (#465, #534) and maybe [Crawlee JS API docs](https://crawlee.dev/api) for inspiration.

belloibrahv:
Hi @vdusek,

I'd like to contribute to improving the API docs for the public components as described in this issue. After reviewing the provided references, guidelines, and the list of components, here's how I plan to approach this task:

1. Focus on improving class docstrings for the unchecked components, starting with:
   - `BasicCrawler`
[...]

PR #914, opened 2025-01-16
docs: Update API docs of public components

vdusek:
### Description

- Update API docs of public components.
- Add functions (context/crawler helpers) to a "Functions" API docs section.

### Issues

- Closes: #304

### Testing

- The docs were rendered locally.

### Checklist

- [x] CI passed

### DIFF
--- a/src/crawlee/events/_local_event_manager.py
+++ b/src/crawlee/events/_local_event_manager.py
@@ -24,7 +24,12 @@ logger = getLogger(__name__)
 
 @docs_group('Classes')
 class LocalEventManager(EventManager):
-    """Local event manager for emitting system info events."""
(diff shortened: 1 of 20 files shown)


src/crawlee/events/_local_event_manager.py
```diff
@@ -24,7 +24,12 @@
 
 @docs_group('Classes')
 class LocalEventManager(EventManager):
-    """Local event manager for emitting system info events."""
+    """Local event manager for emitting system info events.
```
janbuchar, line 27: The fact that it emits `SystemInfo` probably doesn't need to be on the first line, does it?

janbuchar (approved):

Issue #891, opened 2025-01-09
Add a `keep_alive` flag to `BasicCrawler`


janbuchar:
See https://github.com/apify/crawlee-python/discussions/890 and https://github.com/apify/crawlee/pull/1452

PR #921, opened 2025-01-20
feat: Add keep_alive flag to `crawler.__init__`

Pijukatel:
### Description
Add keep_alive flag to `crawler.__init__`

If True, this flag will keep crawler alive even when there are no more requests in queue. Crawler is then waiting for more requests to be added or to be explicitly stopped by `crawler.stop()`.

Add test, add code example in docs.

### Issues
- Closes: #891

### DIFF
--- a/docs/examples/code/beautifulsoup_crawler_keep_alive.py
+++ b/docs/examples/code/beautifulsoup_crawler_keep_alive.py
@@ -1,0 +1,47 @@
+import asyncio
+
+from crawlee._types import BasicCrawlingContext
+from crawlee.crawlers import BeautifulSoupCrawler
+
+
+async def main() -> None:
+    crawler = BeautifulSoupCrawler(
+        # Keep the crawler alive even when there are no requests to be processed at the moment.
+        keep_alive=True,
+    )
+
+    def stop_crawler_if_ulr_visited(context: BasicCrawlingContext) -> None:
+        """Stop crawler once specific url is visited. Just an example of guard condition to stop the crawler."""
+        if context.request.url == 'https://crawlee.dev/docs/examples':
+            crawler.stop('Stop crawler that was in keep_alive state after specific url was visited.')
+        else:
+            context.log.info('keep_alive=True, waiting for more requests to come.')
+
+    async def add_request_later(url: str, after_s: int) -> None:
+        """Add requests to the queue after some time. This can be done by external code."""
+        # Just an example of request being added to the crawler later, when it is waiting due to `keep_alive=True`.
+        await asyncio.sleep(after_s)
+        await crawler.add_requests([url])
+
+    # Define the default request handler, which will be called for every request.
+    @crawler.router.default_handler
+    async def request_handler(context: BasicCrawlingContext) -> None:
+        context.log.info(f'Processing {context.request.url} ...')
+
+        # Stop crawler if some guard condition has been met.
+        stop_crawler_if_ulr_visited(context)
+
+    # Start some tasks that will add some requests later to simulate real situation,
+    # where requests are added later by external code.
+    add_request_later_task1 = asyncio.create_task(add_request_later(url='https://crawlee.dev', after_s=1))
+    add_request_later_task2 = asyncio.create_task(add_request_later(url='https://crawlee.dev/docs/examples', after_s=5))
+
+    # Run the crawler without the initial list of requests.
+    # It will wait for more requests to be added to the queue later due to `keep_alive=True`.
+    await crawler.run()
+
+    await asyncio.gather(add_request_later_task1, add_request_later_task2)
+
+
+if __name__ == '__main__':
+    asyncio.run(main())
--- a/docs/examples/code/beautifulsoup_crawler_keep_alive.py
+++ b/docs/examples/code/beautifulsoup_crawler_keep_alive.py
@@ -10,10 +10,10 @@ async def main() -> None:
         keep_alive=True,
     )
 
-    def stop_crawler_if_ulr_visited(context: BasicCrawlingContext) -> None:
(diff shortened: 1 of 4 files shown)


tests/unit/crawlers/_basic/test_basic_crawler.py
```diff
@@ -1124,3 +1124,37 @@ async def handler(context: BasicCrawlingContext) -> None:
...
+
+    @crawler.router.default_handler
+    async def handler(context: BasicCrawlingContext) -> None:
+        mocked_handler(context.request.url)
+        crawler.stop()
+
+    crawler_run_task = asyncio.create_task(crawler.run())
+
+    # Give some time to crawler to finish(or be in keep_alive state) and add new request.
+    await asyncio.sleep(1)
```
janbuchar, line 1152: We should check here that the crawler did not stop by itself before the additional request was added.

tests/unit/crawlers/_basic/test_basic_crawler.py
```diff
@@ -1124,3 +1124,37 @@ async def handler(context: BasicCrawlingContext) -> None:
...
+
+
+@pytest.mark.parametrize(
+    ('keep_alive', 'max_requests_per_crawl', 'should_process_added_request'),
+    [
+        pytest.param(True, 1, True, id='keep_alive'),
+        pytest.param(True, 0, False, id='keep_alive, but max_requests_per_crawl achieved'),
+        pytest.param(False, 1, False, id='Crawler without keep_alive (default)'),
+    ],
+)
```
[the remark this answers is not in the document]
Pijukatel: Done

tests/unit/crawlers/_basic/test_basic_crawler.py
Pijukatel: Done

tests/unit/crawlers/_basic/test_basic_crawler.py
```diff
@@ -1124,3 +1124,46 @@ async def handler(context: BasicCrawlingContext) -> None:
...
+    @crawler.router.default_handler
+    async def handler(context: BasicCrawlingContext) -> None:
+        mocked_handler(context.request.url)
+        if context.request == additional_urls[-1]:
+            crawler.stop()
+
+    crawler_run_task = asyncio.create_task(crawler.run())
+
+    # Give some time to crawler to finish(or be in keep_alive state) and add new request.
+    await asyncio.sleep(1)
```
janbuchar, line 1163: Isn't 1 second like... a lot?

janbuchar (approved):

tests/unit/crawlers/_basic/test_basic_crawler.py
Pijukatel: Well any time related test is tricky if there is no event to wait for. How do I make sure that the crawler is alive because keep_alive = True and not just because it is randomly slow and takes time to shut down?
I could wrap basic_crawler.__is_finished_function in some mock and wait until it is called at least once, instead of wait time. It will be faster test, but it will leak implementation. Do you prefer that or some other option?

Pijukatel: Crawler already has bunch of internal private state flags. This method could inspect those state flags and maybe also _autoscaled_pool state and report back some sort of public state assessment?
I don't mind the idea, but that looks very much like standalone PR. Then I could modify this test with newly added "state" method.

Just quick guess of possible states:
"Starting"
"Processing requests"
"Waiting in keep alive"
"Shutting down due to unexpected stop"
"Aborted"
"Finished"
"Stopped"
....

janbuchar: Yeah, with the introduction of the `keep_alive` flag, being able to inspect the crawler state makes a lot of sense IMO. Feel free to make an issue and add a TODO to the test that references it.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -216,6 +220,8 @@ def __init__(
                 this limit. Setting this value can help avoid infinite loops in misconfigured crawlers. `None` means
                 no limit. Due to concurrency settings, the actual number of pages visited may slightly exceed
                 this value.
+                If used together with `keep_alive`, then the crawler will be kept alive only until
+                `max_requests_per_crawl` is achieved.
```
vdusek, line 224: Could we align this as a single paragraph? The line wrapping do not have impact on the rendered docs anyways.

vdusek (changes requested):
Could we mention it in the documentation please? Either guide or an example. It might fit well alongside the `crawler.`stop method and maybe more.

src/crawlee/crawlers/_basic/_basic_crawler.py
Pijukatel: Done

docs/examples/code/beautifulsoup_crawler_keep_alive.py
```diff
@@ -0,0 +1,47 @@
...
+from crawlee.crawlers import BeautifulSoupCrawler
+
+
+async def main() -> None:
+    crawler = BeautifulSoupCrawler(
+        # Keep the crawler alive even when there are no requests to be processed at the moment.
+        keep_alive=True,
+    )
+
+    def stop_crawler_if_ulr_visited(context: BasicCrawlingContext) -> None:
```
janbuchar, line 13: typo

Pijukatel: Done

vdusek (approved):
Nice, thanks.

PR #922, opened 2025-01-20
chore: rm old templates directory

vdusek:
It was left there for backward compatibility. We can probably remove it now.


janbuchar (approved):
Yup, thanks for your service, bye bye

Issue #894, opened 2025-01-10
Add Support for Specifying Headers in `context.enqueue_links` Method


zhouya2010:
When using the `context.enqueue_links` method, there is currently no way to specify custom headers (e.g., cookies or other required headers) for the links being enqueued. Some websites require specific headers or cookies to properly retrieve data from detail pages. It would be helpful to add support for passing headers or request options to context.enqueue_links to handle such cases.

Example use case:

```
headers = {
    "Cookie": "session_id=abc123",
    "User-Agent": "Custom User Agent"
}
context.enqueue_links(selector="a.detail-page", headers=headers)
```

This feature would greatly enhance flexibility when crawling websites that require authentication or specific headers for access.

Thank you!

janbuchar:
I believe that adding a `transform_request_function` parameter to `enqueue_links` would help, am I correct? The JS version of Crawlee has it - https://crawlee.dev/api/core/interface/EnqueueLinksOptions#transformRequestFunction

PR #923, opened 2025-01-20
feat: add `transform_request_function` for `enqueue_links`

Mantisus:
### Description

- Add `transform_request_function` for `enqueue_links`
- Add example in introduction section
- Add tests

### Issues

- Closes: #894


docs/introduction/code/03_transform_request.py
```diff
@@ -0,0 +1,36 @@
...
+
+        # Add custom headers to requests to specific sections
+        if '/docs' in request.url:
+            request.headers = HttpHeaders({'Custom-Header': 'value'})
+
+        # Add label for certain URLs
+        if '/blog' in request.url:
+            request.user_data['label'] = 'BLOG'
+
+        return request
```
janbuchar, line 23: Could we move the function to the top level?

janbuchar (commented):
Very opinion based, but we should consider returning some other special value for skipping a request instead of None. Seeing e.g. `return Request.SKIP` would be more readable than `return None`.

Moreover, an empty transform function would mean "skip all requests", which I'm not too fond of - I'd expect that to just let any request go through without any modification.

Mantisus:
> Very opinion based, but we should consider returning some other special value for skipping a request instead of None. Seeing e.g. `return Request.SKIP` would be more readable than `return None`.
> 
> Moreover, an empty transform function would mean "skip all requests", which I'm not too fond of - I'd expect that to just let any request go through without any modification.

I like the use of some special flag, in terms of readability. But I'm not sure how convenient it would be in regular use.

Also the current behavior is closer to TS, if I'm not mistaken about the `filter` operation

```TypeScript
if (transformRequestFunction) {
    requestOptions = requestOptions
        .map((request) => transformRequestFunction(request))
        .filter((r) => !!r) as RequestOptions[];
}
```

B4nan:
What about returning `false` explicitly?

janbuchar:
> What about returning `false` explicitly?

I think that you can't really tell what `false` means in this context either, but at least it's not the implicit return value...

> I would rather not force users to import anything to achieve this behaviour.

I get that, that's why I suggest to put it on `Request` which you're pretty likely to import anyways. We could also use string literals, but that's not commonplace in Python.

B4nan:
> I get that, that's why I suggest to put it on Request which you're pretty likely to import anyways.

I don't think so, at least not in the JS version.

But we can have both, I agree its more readable with some kind of symbol. Docs could use that with a comment saying `false` also works.

src/crawlee/_request.py
```diff
@@ -108,6 +108,23 @@ def __eq__(self, other: object) -> bool:
 user_data_adapter = TypeAdapter(UserData)
 
 
+class RequestOptions(TypedDict, total=False):
```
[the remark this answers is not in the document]
vdusek: Yeah, please no more `Request`-whatever.

vdusek:
I have no issue with using string literals in combination with the "new" `Literal` type. The type checker will handle typos and other problems, and users won't need to import anything more.

```python
def transform(request_options: RequestOptions) -> RequestOptions | Literal['skip', 'unchanged']:
    ...
```

src/crawlee/_types.py
```diff
@@ -353,6 +354,8 @@ def __call__(
                 - `BeautifulSoupCrawler` supports CSS selectors.
             label: Label for the newly created `Request` objects, used for request routing.
             user_data: User data to be provided to the newly created `Request` objects.
+            transform_request_function: A function that processes request options before creating a request.
```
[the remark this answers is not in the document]
janbuchar: Dunno, callback kinda sounds like something that just performs some random side effect. This function actually directly changes the result of `enqueue_links`

src/crawlee/_request.py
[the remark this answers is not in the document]
janbuchar: I'm sorta in favor of `RequestOptions` here - it is simpler to handle than `Request` and the dict structure invites people to mutate it, if you know what I mean. On the other hand, shoving the full `Request` in would let us use validation on assignment, which may be more user friendly than getting an error only after the options are used to construct a `Request`.

docs/introduction/code/03_transform_request.py
```diff
@@ -0,0 +1,37 @@
+from __future__ import annotations
+
+from crawlee import HttpHeaders, RequestOptions
+from crawlee.crawlers import BeautifulSoupCrawler, BeautifulSoupCrawlingContext
+
+
+def transform_request(request: RequestOptions) -> RequestOptions | None:
```
Pijukatel, line 7: Maybe do not call the variable request. I know there is a type hint next to it. But my first though was that I should actually pass the request object to it.

Pijukatel:
> > > Very opinion based, but we should consider returning some other special value for skipping a request instead of None. Seeing e.g. `return Request.SKIP` would be more readable than `return None`.
> > 
> > 
> > I like the use of some special flag, in terms of readability. But I'm not sure how convenient it would be in regular use.
> 
> I imagine something like this - it would be cleaner to put the special flags in a separate enum and not `Request`, but you usually import that anyway. Thoughts?
> 
> ```python
> def transform(request_options: RequestOptions) -> RequestOptions | Literal[Request.SKIP, Request.UNCHANGED]:
>     if 'page' in request_options["url"]:
>         return {*request_options, "label": "page"}
>     if 'campaign' in request_options["url"]:
>         return Request.SKIP
>     return Request.UNCHANGED
> ```
> 
> > Also the current behavior is closer to TS, if I'm not mistaken about the `filter` operation
> > ```ts
> > if (transformRequestFunction) {
> >     requestOptions = requestOptions
> >         .map((request) => transformRequestFunction(request))
> >         .filter((r) => !!r) as RequestOptions[];
> > }
> > ```
> 
> Yeah, it does filter out results that convert to `false`. I'd argue that this behavior is counter-intuitive, even though changing it in the JS version would be breaking in subtle ways. I would be happy to sacrifice parity between versions here, but I'm open to the opinions of others - @Pijukatel @vdusek @B4nan

From point of view of a person that sees this function for the first time and is not aware of the JS implementation I would expect this **transform_request** function to accept Request object, mutate it in place and return nothing. I think we should be more explicit with the naming of functions. This name implies side effects on request.
What we really do is more close to some name like "get_transformed_request_options" ...

I would suggest to make a choice of either doing side effects and not returning anything or creating copy of object and returning it.

janbuchar:
> From point of view of a person that sees this function for the first time and is not aware of the JS implementation I would expect this **transform_request** function to accept Request object, mutate it in place and return nothing. I think we should be more explicit with the naming of functions. This name implies side effects on request. What we really do is more close to some name like "get_transformed_request_options" ...

That's why I wouldn't accept functions that return `None` - it forces the caller to pass a function that returns either the modified options or one of the two other possible conclusions. Empty functions won't pass the type check.

B4nan:
> What we really do is more close to some name like "get_transformed_request_options" ...

Let me disagree, anything that is called `get*` doesn't imply it can be used to mutate things. I would personally keep the JS name.

src/crawlee/_request.py
```diff
@@ -108,27 +108,57 @@ def __eq__(self, other: object) -> bool:
 user_data_adapter = TypeAdapter(UserData)
 
 
-class BaseRequestData(BaseModel):
-    """Data needed to create a new crawling request."""
+class RequestOptions(TypedDict):
+    """Options that can be used to customize request creation.
 
-    model_config = ConfigDict(populate_by_name=True)
+    This type exactly matches the parameters of Request.from_url() method.
```
vdusek, line 114:
```suggestion
    This type exactly matches the parameters of `Request.from_url` method.
```

src/crawlee/_types.py
```diff
@@ -353,6 +355,10 @@ def __call__(
                 - `BeautifulSoupCrawler` supports CSS selectors.
             label: Label for the newly created `Request` objects, used for request routing.
             user_data: User data to be provided to the newly created `Request` objects.
+            transform_request_function: A function that takes RequestOptions and returns either:
+                - Modified RequestOptions to update the request configuration
+                - 'skip' to exclude the request from being enqueued
+                - 'unchanged' to use the original request options without modification
```
vdusek, line 361:
```suggestion
            transform_request_function: A function that takes `RequestOptions` and returns either:
                - Modified `RequestOptions` to update the request configuration,
                - `'skip'` to exclude the request from being enqueued,
                - `'unchanged'` to use the original request options without modification.
```

vdusek (commented):
Nice!

src/crawlee/_request.py
```diff
@@ -108,27 +108,57 @@ def __eq__(self, other: object) -> bool:
 user_data_adapter = TypeAdapter(UserData)
 
 
-class BaseRequestData(BaseModel):
-    """Data needed to create a new crawling request."""
+class RequestOptions(TypedDict):
```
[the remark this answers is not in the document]
Mantisus: We can't, because these variables from `RequestOptions` are used in the `from_url` method.

src/crawlee/_types.py
```diff
@@ -44,6 +44,8 @@
 
 HttpPayload: TypeAlias = bytes
 
+RequestTransformAction: TypeAlias = Literal['skip', 'unchanged']
```
vdusek, line 47: not sure whether this is necessary :slightly_smiling_face:

vdusek (approved):
Good job!

src/crawlee/_types.py
janbuchar: You mean the `TypeAlias` annotation? No, not necessary, but it's nice for readability I suppose.

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
```diff
@@ -140,20 +140,31 @@ async def enqueue_links(
...
-                user_data.setdefault('label', label)
+            requests = list[Request]()
+            base_user_data = user_data or {}
+
             for link in self._parser.find_links(parsed_content, selector=selector):
                 url = link
                 if not is_url_absolute(url):
                     url = convert_to_absolute_url(context.request.url, url)
+
+                request_option = RequestOptions(url=url, user_data={**base_user_data}, label=label)
```
janbuchar, line 157: Please rename to `request_options`

janbuchar (commented):
Just a couple of nitpicks :slightly_smiling_face:

src/crawlee/_types.py
vdusek: Yeah, why not

PR #926, opened 2025-01-21
docs: bump `docusaurus-plugin-typedoc-api`

barjin:
Bumps the `@apify/docusaurus-plugin-typedoc-api` version to include the latest fixes (`optional` / `keyword-only` parameter tags and info about symbol inheritance). Unhides the `Page options (Hide inherited)` button, as it now actually works.

obrazek


B4nan (approved):

barjin:
The new versions of the package seem to still have some cosmetic problems (e.g. links for overwritten symbols are still pointing to the parent implementation). While not crucial, I'd like to address those before merging this.

barjin:
The latest commit adds all the changes from above and unblocks bumping the plugin in the SDK docs (there were some compatibility issues with the reexports).

PR #930, opened 2025-01-23
feat: Add LogisticalRegressionPredictor - rendering type predictor for adaptive crawling

Pijukatel:
### Description
Add `LogisticalRegressionPredictor` which is going to be default `RenderingTypePredictor` for `AdaptivePlaywrightCrawler`.

### Issues

Split from: https://github.com/apify/crawlee-python/issues/249


B4nan:
> LogisticalRegressionPredictor

Isn't that a bit... weird name? :]

pyproject.toml
```diff
@@ -52,6 +52,7 @@ eval-type-backport = ">=0.2.0"
 html5lib = { version = ">=1.0", optional = true }
 httpx = {version = ">=0.27.0", extras = ["brotli", "http2", "zstd"]}
 inquirer = ">=3.3.0"
+jaro-winkler = ">=2.0.3"
```
janbuchar, line 55:
```suggestion
jaro-winkler = { version = ">=2.0.3", optional = true }
```

pyproject.toml
```diff
@@ -62,6 +63,7 @@ pydantic = ">=2.8.1, !=2.10.0, !=2.10.1, !=2.10.2"
 # https://github.com/apify/crawlee-python/issues/814
 pydantic-settings = ">=2.2.0 <2.7.0"
 pyee = ">=9.0.0"
+scikit-learn = ">=1.6.0"
```
janbuchar, line 66:
```suggestion
scikit-learn = { version = ">=1.6.0", optional = true }
```

pyproject.toml
```diff
@@ -92,7 +94,10 @@ types-psutil = "~5.9.5.20240205"
 types-python-dateutil = "~2.9.0.20240316"
 
 [tool.poetry.extras]
-all = ["beautifulsoup4", "curl-cffi", "lxml", "html5lib", "parsel", "playwright"]
+# Support for re-using groups in other groups https://peps.python.org/pep-0735/ in poetry:
+# https://github.com/python-poetry/poetry/issues/9751
+adaptive_playwright = ["jaro-winkler", "playwright", "scikit-learn"]
```
janbuchar, line 99: Maybe it should be `adaptive-playwright` to match `curl-impersonate`?

tests/unit/crawlers/_adaptive_playwright/test_predictor.py
```diff
@@ -0,0 +1,139 @@
...
+    detection_ratio = 0.01
+    label = 'some label'
+    predictor = LogisticalRegressionPredictor(detection_ratio=detection_ratio)
+
+    # Learn first prediction of this label
+    predictor.store_result(Request.from_url(url='http://www.aaa.com/some/stuff', label=label), rendering_type='static')
+    # Increased detection_probability_recommendation
+    prediction = predictor.predict(Request.from_url(url='http://www.aaa.com/some/stuffa', label=label))
+    assert prediction.rendering_type == 'static'
+    assert prediction.detection_probability_recommendation == detection_ratio * 4
```
janbuchar, line 58: These exact checks are probably too implementation-dependent. Could we just check that the recommendation is, like, higher than `detection_ratio * 2` for the first couple of results so that we don't have to update the tests each time we fine-tune some constant by a tiny bit?

janbuchar (commented):
This seems sound to me, I have just some minor comments. Also please fix typos in the PR title :slightly_smiling_face:

tests/unit/crawlers/_adaptive_playwright/test_predictor.py
Pijukatel: This test is just describing how it actually behaves.  Is it requirement or implementation detail? Impossible to say without actually having the requirements. I think it acts nicely as documentation of current behavior and is very easy to change if needed.
If we fine-tune some constant and this test fails, it will be very nice feedback in seeing what those constants actually influence and we can then do informed decision about whether we want to change the test or not.

pyproject.toml
Pijukatel: ok

Pijukatel:
> LogisticalRegressionPredictor

I thought about naming my future kid like this. But maybe it is not so great name after all. It is some implementation of abstract `RenderingTypePredictor` and this one will stay internal and will be used on in AdaptivePwCrawler as default version of predictor. Maybe I should call it `DefaultRenderingTypePredictor` ?

vdusek (commented):
Nice! And thanks for taking this out from `AdaptiveCrawler` PR.

pyproject.toml
Pijukatel, line 1: I have to either make scikit-learn for version span Pythop3.9-3.13 or find another logistical regression implementation

tests/unit/crawlers/_adaptive_playwright/test_predictor.py
janbuchar: Not a hill I want to die on, but this particular example seems like an overspecified test. Almost to the point where if it breaks, you just adjust it without a second thought, which undermines the value of such test.

src/crawlee/crawlers/_adaptive_playwright/_rendering_type_predictor.py
```diff
@@ -0,0 +1,170 @@
+from abc import ABC, abstractmethod
+from collections import defaultdict
+from dataclasses import dataclass
+from itertools import zip_longest
+from statistics import mean
+from typing import Literal
+from urllib.parse import urlparse
+
+from jaro import jaro_winkler_metric  # type: ignore[import-untyped]  # Untyped and stubs not available
+from sklearn.linear_model import LogisticRegression  # type: ignore[import-untyped]  # Untyped and stubs not available
```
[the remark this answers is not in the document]
Pijukatel: Ok

src/crawlee/crawlers/_adaptive_playwright/_rendering_type_predictor.py
```diff
@@ -0,0 +1,170 @@
...
+        self._labels_coefficients: dict[str, float] = defaultdict(lambda: n + 2)
+
+    @override
+    def predict(self, request: Request) -> RenderingTypePrediction:
+        """Get `RenderingTypePrediction` based on the input request.
+
+        Args:
+            request: `Request` instance for which the prediction is made.
+        """
+        similarity_threshold = 0.1  #  Prediction probability difference threshold to consider prediction unreliable.
```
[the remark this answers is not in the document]
Pijukatel: In JS implementation it is unnamed inline constant. My intention in making it named is just to make it more readable.
It is used only in this single method and it is not intended to be configurable, so I would rather keep it here where it is used at the moment.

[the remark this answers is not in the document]
vdusek: Okay, but constants should be upper-cased.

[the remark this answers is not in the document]
Pijukatel: I tried, but then ruff tells me it is not a "proper" constant https://github.com/astral-sh/ruff/issues/15721#issuecomment-2613168454.

So I would rather keep it lower case if don't mind.

tests/unit/crawlers/_adaptive_playwright/test_predictor.py
```diff
@@ -0,0 +1,139 @@
...
+        ('http://www.aab.com/some/stuff', 'static'),
+        ('http://www.aac.com/some/stuff', 'static'),
+        ('http://www.ddd.com/some/stuff', 'client only'),
+        ('http://www.dde.com/some/stuff', 'client only'),
+        ('http://www.ddf.com/some/stuff', 'client only'),
+    )
+
+    # Learn from small set
+    for learning_input in learning_inputs:
+        predictor.store_result(Request.from_url(url=learning_input[0], label=label), rendering_type=learning_input[1])
```
janbuchar, line 40:
```suggestion
    for url, rendering_type in learning_inputs:
        predictor.store_result(Request.from_url(url=url, label=label), rendering_type=rendering_type)
```

src/crawlee/crawlers/_adaptive_playwright/_rendering_type_predictor.py
```diff
@@ -0,0 +1,172 @@
+from abc import ABC, abstractmethod
+from collections import defaultdict
+from dataclasses import dataclass
+from itertools import zip_longest
+from statistics import mean
+from typing import Literal
+from urllib.parse import urlparse
+
+from jaro import jaro_winkler_metric
+from sklearn.linear_model import LogisticRegression
```
janbuchar, line 10: We should wrap this in a way that the exception tells the user to install the `adaptive-playwright` extra. But since this is a private subpackage, maybe it can wait until the adaptive playwirght functionality is made public.

pyproject.toml
```diff
@@ -92,7 +97,10 @@ types-psutil = "~5.9.5.20240205"
 types-python-dateutil = "~2.9.0.20240316"
 
 [tool.poetry.extras]
-all = ["beautifulsoup4", "curl-cffi", "lxml", "html5lib", "parsel", "playwright"]
+# Support for re-using groups in other groups https://peps.python.org/pep-0735/ in poetry:
+# https://github.com/python-poetry/poetry/issues/9751
+adaptive-playwright = ["jaro-winkler", "playwright", "scikit-learn"]
+all = ["beautifulsoup4","curl-cffi", "html5lib", "jaro-winkler", "lxml","parsel", "playwright","scikit-learn"]
```
vdusek, line 103:
```suggestion
all = ["beautifulsoup4", "curl-cffi", "html5lib", "jaro-winkler", "lxml", "parsel", "playwright", "scikit-learn"]
```

src/crawlee/crawlers/_adaptive_playwright/_rendering_type_predictor.py
```diff
@@ -0,0 +1,172 @@
...
+            request: `Request` instance for which the prediction is made.
+        """
+
+    @abstractmethod
+    def store_result(self, request: Request, rendering_type: RenderingType) -> None:
+        """Store prediction results and retrain the model.
+
+        Args:
+            request: Used request.
+            rendering_type: Known suitable RenderingType.
```
vdusek, line 41:
```suggestion
            rendering_type: Known suitable `RenderingType`.
```

vdusek (approved):
Just a few nits, LGTM otherwise.

PR #934, opened 2025-01-24
feat: Return handler upon decorator invocation

mylank:
### Description
Currently, wrapping a handler with a decorator nullifies the function because it returns `None` [in the code](https://github.com/apify/crawlee-python/blob/28579996e454e19dd69ed6ee09e4293d4e056a6e/src/crawlee/router.py#L44-L45). See the "None" popup over here:
image

By having the router.handler function return the original input handler, we can now reuse the handler function again in other places as it's not set to `None` anymore. This allows for invoking the original function even after applying the decorator and for adding multiple decorators to the same handler function. The latter for example allows for using the same handler function for multiple labels (a use case I ran into).

### Issues
Not applicable.

### Testing

- [x] Added a test to ensure handler is not nullified
- [x] Added a test for multiple labels assigned to a single handler

### Checklist

- [x] CI passed

Please let me know your thoughts! :raised_hands:

### DIFF
--- a/src/crawlee/router.py
+++ b/src/crawlee/router.py
@@ -33,7 +33,9 @@ class Router(Generic[TCrawlingContext]):
 
         return handler
 
-    def handler(self, label: str) -> Callable[[RequestHandler[TCrawlingContext]], None]:
+    def handler(
+        self, label: str
+    ) -> Callable[[RequestHandler[TCrawlingContext]], Callable[[TCrawlingContext], Awaitable]]:
         """A decorator used to register a label-based handler.
 
         The registered will be invoked only for requests with the exact same label.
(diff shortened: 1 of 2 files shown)


janbuchar (approved):
Thanks for your contribution @mylank! Would you mind sharing the use case for this? Is it just routing multiple labels to the same handler, or something more sophisticated?

mylank:
> Thanks for your contribution @mylank! Would you mind sharing the use case for this? Is it just routing multiple labels to the same handler, or something more sophisticated?

Thanks for the quick response! 🙌 Of course; so I have two pages that are almost identical but one requires just a couple of actions more. I tried a couple of different options but in the end I just thought the cleanest would be to tag the same handler with 2 labels and just use an if statement [like in the docs](https://crawlee.dev/python/docs/introduction/crawling#the-label-of-enqueue_links). 

So yes, my specific use case is indeed just for routing multiple labels to the same handler that require only a small deviation from the general logic!

Thanks for merging this in!

PR #936, opened 2025-01-24
docs: update setting up and running in cloud

vdusek:
- I updated the "Setting up" and "Running in cloud" pages to make them more beginner-friendly (hopefully) and also removed mentions of Poetry. Including Poetry was unnecessary and could be confusing. Not to mention in Python there are X other package managers that we could hypotetically mention as well. So I decided to stay only with Pip and Pipx for templates. People using a package manager other than Pip will likely know how to handle the installation on their own.
- This was reported on Slack by @honzajavorek.

### DIFF
--- a/docs/introduction/01_setting_up.mdx
+++ b/docs/introduction/01_setting_up.mdx
@@ -4,13 +4,22 @@ title: Setting up
 ---
 
 import ApiLink from '@site/src/components/ApiLink';
+import CodeBlock from '@theme/CodeBlock';
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
 
-To run Crawlee on your computer, ensure you meet the following requirements:
+This guide will help you get started with Crawlee by setting it up on your computer. Follow the steps below to ensure a smooth installation process.
 
-1. [Python](https://www.python.org/) 3.9 or higher installed,
-2. [Pip](https://pip.pypa.io/en/stable/) installed.
+## Prerequisites
 
-You can verify these by running the following commands:
+Before installing Crawlee itself, make sure that your system meets the following requirements:
+
+- **Python 3.9 or higher**: Crawlee requires Python 3.9 or a newer version. You can download Python from the [official website](https://www.python.org/downloads/).
+- **Python package manager**: While this guide uses Pip (the most common package manager), you can also use any package manager you want. You can download Pip from the [official website](https://pip.pypa.io/en/stable/installation/).
+
+### Verifying prerequisites
+
+To check if Python and the Pip package manager are installed, run the following commands:
 
 ```sh
 python --version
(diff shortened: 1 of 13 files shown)


honzajavorek (commented):
🔴 I'd like to see the comment about `apify init` addressed, as that's my current roadblock when working on the Python course.

🔴 I don't know if you aim to support Windows, but if so, these documents need more love and verification that all works there as intended. AFAIK the venv creation and activation won't work for sure.

🟡 Otherwise I didn't find anything serious, just various unimportant nitpicks about lower/uppercase letters (I'm sorry!) and I offered several improvements to the flow of the text. I dig removing poetry from the mix.

vdusek:
Thanks @honzajavorek for your review.

> I'd like to see the comment about apify init addressed, as that's my current roadblock when working on the Python course.

I'm not sure how much detail we should provide about the Apify CLI and `apify init` specifically. Crawlee projects should be Actorizable like any other Python projects. If not, only the Crawlee-specific aspects should be mentioned here, while everything related to `apify init` should be covered in the [Apify CLI docs](https://docs.apify.com/cli/docs). Mostly to ensure future-proof relevance. At least I've extended the section and added more reference to the CLI's docs.

> I don't know if you aim to support Windows, but if so, these documents need more love and verification that all works there as intended. AFAIK the venv creation and activation won't work for sure.

@Pijukatel If it's not a problem, could you please address this? It is [this](https://github.com/apify/crawlee-python/pull/936#discussion_r1930235275) comment.

docs/introduction/01_setting_up.mdx
```diff
@@ -94,26 +99,54 @@ You can install multiple extras at once by using a comma as a separator:
...
+### Running your project
+
+To run your newly created project, navigate to the project directory, activate the virtual environment, and execute the Python interpreter with the project module:
+
+```sh
+cd -m my_crawler
 ```
 
+```sh
+source .venv/bin/activate
```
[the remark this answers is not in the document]
Pijukatel:
````
<Tabs>
  <TabItem value="Linux" label="Linux" default>
```sh
source .venv/bin/activate
``` 
  </TabItem>
<TabItem value="Windows" label="Windows" default>
```sh
venv\Scripts\activate
```
  </TabItem>
</Tabs>


````

File: docs/introduction/01_setting_up.mdx [hunk 2]
```diff
@@ -20,9 +29,15 @@ python --version
 pip --version
```
Pijukatel, line 29: All those direct **pip** calls might be better replaced by **python -m pip calls**. I think that will work in more setups - Windows included.

docs/introduction/01_setting_up.mdx
```diff
@@ -94,26 +99,54 @@ You can install multiple extras at once by using a comma as a separator:
...
+```
+
+Follow the interactive prompts in the CLI to choose a crawler type and set up your new project.
+
+### Running your project
+
+To run your newly created project, navigate to the project directory, activate the virtual environment, and execute the Python interpreter with the project module:
+
+```sh
+cd -m my_crawler
```
Pijukatel, line 137: -m ?

honzajavorek:
@vdusek That's a good point that while addressing the `apify init` part is important, it may be a task for a different tutorial than this one. Even in such case though, I think if you provide any kind of tutorial, the reader must be able to follow it seamlessly step-by-step, even if one of the steps is "setup your project according to this link", perhaps with "run this and if it works, you verified you can continue to another step" to make it more bulletproof. If you've done that, then I'm 👍 (I don't have time to re-read it right now, just replying to your comment.)

docs/introduction/01_setting_up.mdx
vdusek: Thanks

PR #940, opened 2025-01-29
feat: add `time_remaining_secs` property to `MIGRATING` event data

fnesveda:
The Apify platform now sends a `timeRemainingSecs` property in the `MIGRATING` event data, this adds it to the event data model.

### DIFF
--- a/src/crawlee/events/_types.py
+++ b/src/crawlee/events/_types.py
@@ -61,7 +61,7 @@ class EventMigratingData(BaseModel):
 
     # The remaining time in seconds before the migration is forced and the process is killed
     # Optional because it's not present when the event handler is called manually
-    time_remaining_secs: Annotated[float | None, Field(alias='timeRemainingSecs')]
+    time_remaining_secs: Annotated[float | None, Field(alias='timeRemainingSecs')] = None
 
 
 @docs_group('Event payloads')
--- a/src/crawlee/events/_types.py
+++ b/src/crawlee/events/_types.py
@@ -7,8 +7,8 @@ from typing import Annotated, Any, TypeVar, Union
 from pydantic import BaseModel, ConfigDict, Field
 
 from crawlee._utils.docs import docs_group
-from crawlee._utils.system import CpuInfo, MemoryUsageInfo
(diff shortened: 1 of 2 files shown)


vdusek (approved):

src/crawlee/events/_types.py
```diff
@@ -59,6 +59,10 @@ class EventMigratingData(BaseModel):
 
     model_config = ConfigDict(populate_by_name=True)
 
+    # The remaining time in seconds before the migration is forced and the process is killed
+    # Optional because it's not present when the event handler is called manually
+    time_remaining_secs: Annotated[float | None, Field(alias='timeRemainingSecs')]
```
vdusek, line 64: Could we use `datetime`/`time` types rather than float + secs suffix identifier?

janbuchar, line 64: Shouldn't it default to `None` as well?

fnesveda: But if you want it as `timedelta` in Python, I can add it, it shouldn't be hard. Only the property names will be slightly different between JS and Python, but I guess it's fine.

vdusek: Yes please. We use `timedelta` for all time-related fields. For Pydantic models, we have a custom type called [timedelta_ms](https://github.com/apify/crawlee-python/blob/master/src/crawlee/_utils/models.py#L38), which includes a custom validator and serializer. You can see it in use e.g. [here](https://github.com/apify/crawlee-python/blob/ae0d328c85cc98a962b320ad876342269157acca/src/crawlee/statistics/_models.py#L68).

And the property will only differ in Python’s internal code, the env var name can remain unchanged.

fnesveda: It's not an env var, it's data in an Actor event, and it's not internal, since it's something that users will be working with.

I'll rename the field exposed to users to `time_remaining`, which will be a `timedelta`, and internally it will be parsed from the `timeRemainingSecs` property of the received event data.

vdusek (approved):
LGTM

janbuchar (approved):

Issue #933, opened 2025-01-24
Standardize cookie handling


Mantisus:
Currently we have 3 main cookie handling mechanisms depending on the HTTP client or browser, and none work correctly.

1. `HttpxHttpClient`.
This solution is closest to expected. Cookies are stored in `Session`. However, we use `dict` which loses the cookie-domain relationship. This can cause issues during cross-domain crawling.

2. `CurlImpersonateHttpClient`.
`Session` knows nothing about cookies and all cookies are stored at the `AsyncSession` level. As a result, if we don't use proxies, all sessions have identical cookies. If we work with proxies, cookies become tied to the proxy.

3. `Playwright`.
`Session` knows nothing about cookies and all cookies are stored at the `PlaywrightContext` level, meaning all sessions working from one context will operate with the same cookies.

B4nan:
@Mantisus thanks for bringing this up. Let's split it into 3 separate PRs please.

B4nan:
I would focus on 3 first, that feels like the biggest issue to me. Scraping multiple domains in a single crawler is not a very common use case.

janbuchar:
> I would focus on 3 first, that feels like the biggest issue to me. Scraping multiple domains in a single crawler is not a very common use case.

Agreed, even though 2. is similar in terms of severity (but yeah, playwright is a bit more popular)

Mantisus:
2 of 3. But for multi-domain cookie support, we'd really have to go to something like `http.cookiejar.CookieJar` instead of dict.

Issue #722, opened 2024-11-21
Implement option for incognito pages to PlaywrightCrawler


vdusek:
- Implement an option for incognito pages to PlaywrightCrawler in a similar way as it is in the Crawlee JS.
- https://crawlee.dev/api/browser-pool/interface/BrowserPluginOptions#useIncognitoPages
- Before implementation sync with @barjin, as he can provide further context and also suggest potential improvements.

vdusek:
closing as it was resolved in https://github.com/apify/crawlee-python/pull/941

PR #941, opened 2025-01-29
feat: add support `use_incognito_pages` for `browser_launch_options` in `PlaywrightCrawler`

Mantisus:
### Description

- Improve cookie handling for `PlaywrightCrawler`. Cookies are now stored in the `Session` and set in Playwright Context from the `Session`.
- Add `use_incognito_pages` option for `browser_launch_options` allowing each new page to be launched in a separate context. 

### Issues

- #722 
- #933


janbuchar:
Does the incognito pages option have any relationship with the cookie handling? Also, a test would be nice :slightly_smiling_face:

Mantisus:
> Does the incognito pages option have any relationship with the cookie handling?

Yes. When we work in basic mode we are working with one common context. In this case, cookies will be strayed between sessions.

However, by using `incognito pages` one page per context, we get a more controlled situation. When a session works only with the cookies intended for it.

janbuchar:
> > Does the incognito pages option have any relationship with the cookie handling?
> 
> Yes. When we work in basic mode we are working with one common context. In this case, cookies will be strayed between sessions.

Can this be a desirable state for anyone?

> However, by using `incognito pages` one page per context, we get a more controlled situation. When a session works only with the cookies intended for it.

I sort of think that this should be the default. Do we really need to make it configurable? What does the JS version do?

B4nan:
> What does the JS version do?

It's not the default, because it means no browser cache, so a huge perf cost, when we were testing this, things took literally twice the time to finish because of that.

janbuchar:
> > What does the JS version do?
> 
> It's not the default, because it means no browser cache, so a huge perf cost, when we were testing this, things took literally twice the time to finish because of that.

So does the JS `PlaywrightCrawler` also not store cookies in the `Session`?

Mantisus:
> So does the JS `PlaywrightCrawler` also not store cookies in the `Session`?

Stores, but when using the basic setup, they are just as flowing between sessions from the Playwright context.

B4nan:
This is the thing we really need to redesign in next major. IDK how cookies behave, but the more important part there is that because of this, we keep the same proxy in one browser instance with the defaults (so persistent contexts). In other words, sessions rotate per request, but proxies only per browser (we have some limits on how many times a browser instance can be used).

src/crawlee/browsers/_playwright_browser_plugin.py
```diff
@@ -70,6 +70,7 @@ def __init__(
         self._browser_launch_options = default_launch_browser_options | (browser_launch_options or {})
         self._browser_new_context_options = browser_new_context_options or {}
         self._max_open_pages_per_browser = max_open_pages_per_browser
+        self._use_incognito_pages: bool = self._browser_launch_options.pop('use_incognito_pages', False)
```
vdusek, line 73: The `use_incognito_pages` option should not be included in the `browser_launch_options`. This parameter is intended for browser launch options supported by Playwright ([docs](https://playwright.dev/python/docs/api/class-browsertype#browser-type-launch)).

The `use_incognito_pages` is our custom flag used for internal logic in our code, not by Playwright. Mixing it with `browser_launch_options` does not seem correct. Instead, it should be a dedicated argument.

And it should be probably exposed in the same places as `browser_launch_options`, specifically: `PlaywrightCrawler.__init__`, `BrowserPool.with_default_plugin`, `PlaywrightBrowserPlugin.__init__`.

tests/unit/crawlers/_playwright/test_playwright_crawler.py
```diff
@@ -242,3 +245,69 @@ async def some_hook(context: PlaywrightPreNavCrawlingContext) -> None:
     await crawler.run(['https://test.com'])
 
     assert handler_data.get('proxy') == proxy_value
+
+
+@pytest.mark.parametrize(
+    'use_incognito_pages',
+    [False, True],
+    ids=['without use_incognito_pages', 'with use_incognito_pages'],
+)
```
vdusek, line 254: Could you please use `pytest.param`?

vdusek (approved):
Last nit, otherwise LGTM, but let's wait for @janbuchar's approve as well.

tests/unit/crawlers/_playwright/test_playwright_crawler.py
```diff
@@ -242,3 +245,71 @@ async def some_hook(context: PlaywrightPreNavCrawlingContext) -> None:
...
+    @crawler.router.default_handler
+    async def handler(context: PlaywrightCrawlingContext) -> None:
+        if context.session:
+            sessions_ids.append(context.session.id)
+            if context.request.unique_key in {'1', '2'}:
+                sessions_cookies[context.session.id] = context.session.cookies
+                response_data = json.loads(await context.response.text())
+                response_cookies[context.session.id] = response_data.get('cookies')
+                if context.request.unique_key == '1':
+                    context.session.retire()
```
janbuchar, line 277: The nesting could be reduced here using early returns.

janbuchar (approved):

PR #947, opened 2025-01-30
fix: fix session managment with retire

Mantisus:
### Description

- Fix cases where a session was not excluded from the pool after calling `session.retire()` inside a successful `request_handler`.


janbuchar (approved):

vdusek (approved):
LGTM, thanks

PR #948, opened 2025-01-31
ci: Do not wait for code checks on release if the head commit skips CI

janbuchar:


vdusek (approved):
LGTM

B4nan (approved):
looks good, but my brain is not good at processing bash commands on its own so I hope you've tested this 🙃

janbuchar:
> looks good, but my brain is not good at processing bash commands on its own so I hope you've tested this 🙃

It's the same fix that we applied to the CLI, so it should be... fine

Issue #955, opened 2025-02-03
SAME_HOSTNAME not working on non www URLs


ROYOSTI:
When using the `EnqueueStrategy.SAME_HOSTNAME` I noticed it does not work properly on non www urls.

In the debugger I noticed it passes `origin` to the `_check_enqueue_strategy` but it uses the `context.request.loaded_url` if available.
So every URL that is checked will mismatch because of the difference in hostname

![Image](https://github.com/user-attachments/assets/d9c20dfb-96e8-47ac-b87f-6632182b3812)

I tested this with multiple urls with & without www prefix and got the same behaviour.

![Image](https://github.com/user-attachments/assets/f8a10c77-988a-4228-bddf-36b4780afcc8)

Changing the line to `origin = context.request.url` fix this issue, but I have no idea what implications this would have on the other code.

I use the `PlaywrightCrawler` in my code with `context.enqueue_links`

vdusek:
Hi @ROYOSTI, thanks for reporting this. It does seem like a bug and we'll try to look into it soon.

PR #956, opened 2025-02-03
fix: fix convert relative link to absolute in `enqueue_links` for response with redirect

Mantisus:
### Description

-  fix `enqueue_links` for response with redirect. 

### Issues

- Closes: #955


vdusek (approved):
Any idea for a test? Couldn't we utilize httpbin? ... if there is no easy way, it is not necessary I would say. Otherwise LGTM :slightly_smiling_face:.

Pijukatel:
> Any idea for a test? Couldn't we utilize httpbin? ... if there is no easy way, it is not necessary I would say. Otherwise LGTM 🙂.

I agree it would be nice to have a test for it to properly describe the scenario in which this is a problem. Maybe something from https://httpbin.org/#/Redirects  can be used to create such scenario?

Mantisus:
Updated the current test for `enqueue_links` to work as expected behavior when a redirect exists

vdusek (approved):

Pijukatel (approved):

Issue #60, opened 2024-03-01
Implement `Snapshotter._snapshot_client()`


vdusek:
- Currently, there is only a dummy version of [`Snapshotter._snapshot_client()`](https://github.com/apify/crawlee-python/blob/v0.4.0/src/crawlee/_autoscaling/snapshotter.py#L258) without a real measurement.
- Once [`ApifyStorageClient`](https://github.com/apify/apify-sdk-python/tree/master/src/apify/apify_storage_client) is implemented, use it there to measure the real values.
- Check TypeScript implementation for an inspiration - [Snapshotter._snapshotClient()]( https://github.com/apify/crawlee/blob/v3.7.3/packages/core/src/autoscaling/snapshotter.ts#L338:L368).

vdusek:
@Mantisus

PR #957, opened 2025-02-04
feat: implement `_snapshot_client` for `Snapshotter`

Mantisus:
### Description

- implement `_snapshot_client` for `Snapshotter`

### Issues

- Closes: #60

### DIFF
--- a/src/crawlee/_autoscaling/snapshotter.py
+++ b/src/crawlee/_autoscaling/snapshotter.py
@@ -307,7 +307,7 @@ class Snapshotter:
         """
         client = service_locator.get_storage_client()
 
-        rate_limit_errors: dict[int, int] = getattr(client, 'rate_limit_errors', {})
+        rate_limit_errors: dict[int, int] = client.get_rate_limit_errors()
 
         error_count = rate_limit_errors.get(self._CLIENT_RATE_LIMIT_ERROR_RETRY_COUNT, 0)
         snapshot = ClientSnapshot(error_count=error_count, max_error_count=self._max_client_errors)
--- a/src/crawlee/_autoscaling/snapshotter.py
+++ b/src/crawlee/_autoscaling/snapshotter.py
@@ -305,11 +305,11 @@ class Snapshotter:
         Only errors produced by a 2nd retry of the API call are considered for snapshotting since earlier errors may
         just be caused by a random spike in the number of requests and do not necessarily signify API overloading.
         """
-        # TODO: This is just a dummy placeholder. It can be implemented once `StorageClient` is ready.
-        # Attribute `self._client_rate_limit_error_retry_count` will be used here.
-        # https://github.com/apify/crawlee-python/issues/60
+        client = service_locator.get_storage_client()
 
-        error_count = 0
+        rate_limit_errors: dict[int, int] = getattr(client, 'rate_limit_errors', {})
+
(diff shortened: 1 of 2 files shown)


vdusek (approved):
LGTM

src/crawlee/_autoscaling/snapshotter.py
```diff
@@ -305,11 +305,11 @@ def _snapshot_client(self) -> None:
         Only errors produced by a 2nd retry of the API call are considered for snapshotting since earlier errors may
         just be caused by a random spike in the number of requests and do not necessarily signify API overloading.
         """
-        # TODO: This is just a dummy placeholder. It can be implemented once `StorageClient` is ready.
-        # Attribute `self._client_rate_limit_error_retry_count` will be used here.
-        # https://github.com/apify/crawlee-python/issues/60
+        client = service_locator.get_storage_client()
 
-        error_count = 0
+        rate_limit_errors: dict[int, int] = getattr(client, 'rate_limit_errors', {})
```
vdusek, line 310: I need some explanation here... What values can be in `rate_limit_errors`? What value is going to be there for the `MemoryStorageClient` (nothing?) and for the `ApifyStorageClient`? Thanks.

src/crawlee/_autoscaling/snapshotter.py
```diff
@@ -305,11 +305,11 @@ def _snapshot_client(self) -> None:
...
+
+        error_count = rate_limit_errors.get(self._CLIENT_RATE_LIMIT_ERROR_RETRY_COUNT, 0)
```
vdusek, line 312: And this does not seem to be correct.

vdusek (changes requested):
I have some concerns/questions.

src/crawlee/_autoscaling/snapshotter.py
Mantisus: For `MemoryStorageClient` it is empty `dict`. Since getattr will return the default value if the specified attribute does not exist
For `ApifyStorageClient` it is `rate_limit_errors` from https://github.com/apify/apify-client-python/blob/master/src/apify_client/_statistics.py#L15. Relate PR - https://github.com/apify/apify-sdk-python/pull/387

src/crawlee/_autoscaling/snapshotter.py
Mantisus: Maybe I'm missing something.

But since we use `dict` for `rate_limit_errors`, but I kept the TS logic to make the list index match the key in the dictionary. I think it should work as expected.
https://github.com/apify/apify-client-python/blob/master/src/apify_client/_statistics.py#L27

If the dictionary is empty or the key is missing the value is always 0. Otherwise we get the corresponding value

vdusek (approved):
Thanks for the explanation.

Mantisus:
> Apart from that one comment, I'm also not sure that this passes as a `chore: ` 🙂

I'm not sure about the `chore` type either.

On the one side, it has no effect on clients outside of the Apify platform. On the other side, we now add a new method to the public interface....

janbuchar:
> > Apart from that one comment, I'm also not sure that this passes as a `chore: ` 🙂
> 
> I'm not sure about the `chore` type either.
> 
> On the one side, it has no effect on clients outside of the Apify platform. On the other side, we now add a new method to the public interface....

I mean, we're still on 0.x, so I wouldn't think twice about making it a `feat` :slightly_smiling_face: Let's edit that and merge!

janbuchar (approved):

Issue #953, opened 2025-02-03
Add ignore_http_error_status_codes and additional_http_error_status_codes arguments to PlaywrightCrawler


Pijukatel:
Currently arguments that allow to change how different return codes are handled are [available only to static http-based crawlers](https://github.com/apify/crawlee-python/blob/master/src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py#L39). Those arguments can be used in crawler `__init__`, but are not available in `PlaywrightCrawler`. If someone wants to for example ignore 403 error:

```
crawler = ParselCrawler(..., ignore_http_error_status_codes = {403})
```

but in PlaywrightCrawler they have to do something like this:

```
crawler = PlaywrightCrawler(...)
crawler._http_client._ignore_http_error_status_codes = {403}
```

That is very confusing and users will hardly even know about it. The `PlaywrightCrawler` behavior should be aligned with other crawlers and these should be possible to set in `__init__`

Pijukatel:
> In `PlaywrightCrawler`, they could do
> 
> crawler = PlaywrightCrawler(
>   ...,
>   http_client=HttpxHttpClient(
>     ignore_http_error_status_codes = {403}
>   )
> )
> 
> ...which is less bad than touching protected attributes.
> 
> But still, this does something different than the `ParselCrawler` example. In `ParselCrawler`, the http client is used for fetching the website itself, while in `PlaywrightCrawler`, you only use it for `context.send_request`. So even if you ignore the 403 status code, it will have no effect on what the crawler does during regular crawling, which is also confusing.
> 
> To handle http status codes received during navigation, we'd have to implement this separately for `PlaywrightCrawler`.

PlaywrightCrawler already handles requests recieved during navigation, but in somewhat non-obious way
https://github.com/apify/crawlee-python/blob/master/src/crawlee/crawlers/_playwright/_playwright_crawler.py#L255

Where it inherits [_is_session_blocked_status_code](https://github.com/apify/crawlee-python/blob/master/src/crawlee/crawlers/_basic/_basic_crawler.py#L1111) from BasicCrawler, that looks into `self._http_client.additional_blocked_status_codes`. So that is why even navigation return status codes can be handled through `crawler._http_client`

janbuchar:
> PlaywrightCrawler already handles status codes received during navigation, but in somewhat non-obvious way https://github.com/apify/crawlee-python/blob/master/src/crawlee/crawlers/_playwright/_playwright_crawler.py#L255
[...]

Issue #830, opened 2024-12-18
Implementing correct handling of 403 and the other codes that should trigger a SessionError


Mantisus:
After #812, we discussed with @janbuchar that currently the trigger for SessionError based on status codes is not working correctly. To be more precise, when using the `http client`, it doesn't trigger at all.

There are several points that need to be clarified in this case:
- Should blocked_status_codes in Session always take priority over any error status codes in `BaseHttpClient` and trigger session rotation instead of request retry?
- Should `additional_http_error_status_codes` and `ignore_http_error_status_codes` be considered in the `is_blocked_status_code` method in `Session`, or should we only consider status codes defined in `blocked_status_codes` in `Session`?

PR #959, opened 2025-02-05
feat!: Enable additional status codes arguments to PlaywrightCrawler

Pijukatel:
### Description

Add `additional_http_error_status_codes` and `ignore_http_error_status_codes` to PlaywrightCrawler.
Since they exist now on all crawlers, move them to `BasicCrawler` level.
Do not use `_http_client` attributes for getting additional status codes related variables.

**Breaking:** Remove `HttpCrawlerOptions` -> No unique options compared to `BasicCrawlerOptions` anymore.

### Issues

- Closes: #953


src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -256,7 +268,29 @@ def __init__(
...
+        )
+
+        self._http_client = http_client or HttpxHttpClient(
+            additional_http_error_status_codes=self._additional_http_error_status_codes,
+            ignore_http_error_status_codes=self._ignore_http_error_status_codes,
+        )
+
+        if self._http_client.additional_blocked_status_codes != self._additional_http_error_status_codes:
+            raise ValueError(
+                'Used `additional_blocked_status_codes` argument does not match with  with '
```
Mantisus, line 286: double `with`

Sorry, can't commit with the quick fix due to limited permissions.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -276,7 +288,29 @@ def __init__(
...
+        if self._http_client.additional_blocked_status_codes != self._additional_http_error_status_codes:
+            raise ValueError(
+                'Used `additional_blocked_status_codes` argument does not match with '
+                f'{self._http_client.additional_blocked_status_codes=}. They have to be the same.'
+            )
+        if self._http_client.ignore_http_error_status_codes != self._ignore_http_error_status_codes:
+            raise ValueError(
+                'Used `ignore_http_error_status_codes` argument does not match with '
+                f'{self._http_client.ignore_http_error_status_codes=}. They have to be the same.'
+            )
```
vdusek, line 313: Couldn't we just keep them only in the `http_client` instance? (PW Crawler has HTTP client as well)

Pijukatel: I was considering that option, but it felt like misuse to me, especially when it comes to PlaywrightCrawler. PlaywrightCrawler is not using HTTP client for `page.navigate` so it would be really strange if it would use some attribute of this unrelated component to decide whether response status code of `page.navigate` is ok or not.

But I see it looks like unnecessary code duplication, so I am not 100% happy with this either.

vdusek: Yeah, I got it... However, in my opinion, having it duplicated seems like a worse option to me.

Mantisus: I think we can think about taking this logic out of the http client. And put it in the `BasicCrawler`. Then it will work uniformly for any crawler and we will avoid code duplication

Mantisus: In that case, I don't see a problem if we keep the duplication of code at this point. It will be solved during refactoring.

Mantisus (approved):
LGTM

src/crawlee/crawlers/_basic/_basic_crawler.py
Pijukatel: Yes, Maybe it will be solved at the same time as https://github.com/apify/crawlee-python/issues/830, but if not, here is the issue: https://github.com/apify/crawlee-python/issues/998

vdusek (approved):
LGTM

PR #962, opened 2025-02-06
chore: improve http crawler tests and cookies tests

Mantisus:
### Description

- Improve cookies tests for more stability
- Update `parametrize` in http crawler tests for use `pytest.param`

### DIFF
--- a/tests/unit/crawlers/_http/test_http_crawler.py
+++ b/tests/unit/crawlers/_http/test_http_crawler.py
@@ -133,16 +133,14 @@ async def test_handles_redirects(
 @pytest.mark.parametrize(
     ('additional_http_error_status_codes', 'ignore_http_error_status_codes', 'expected_number_error'),
     [
-        ([], [], 1),
-        ([403], [], 3),
-        ([], [403], 0),
-        ([403], [403], 3),
-    ],
-    ids=[
-        'default_behavior',  # error without retry for all 4xx statuses
-        'additional_status_codes',  # make retry for codes in `additional_http_error_status_codes` list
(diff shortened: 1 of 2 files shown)


tests/unit/crawlers/_http/test_http_crawler.py
```diff
@@ -133,16 +133,14 @@ async def test_handles_redirects(
...
-        'ignore_error_status_codes',  # take as successful status codes from the `ignore_http_error_status_codes` list
-        'additional_and_ignore',  # check precedence for `additional_http_error_status_codes`
+        # error without retry for all 4xx statuses
+        pytest.param([], [], 1, id='default_behavior'),
+        # make retry for codes in `additional_http_error_status_codes` list
+        pytest.param([403], [], 3, id='additional_status_codes'),
+        # take as successful status codes from the `ignore_http_error_status_codes` list
+        pytest.param([], [403], 0, id='ignore_error_status_codes'),
+        # check precedence for `additional_http_error_status_codes`
+        pytest.param([403], [403], 3, id='additional_and_ignore'),
```
janbuchar, line 143: I think this would warrant using a single dataclass param so that it's obvious which value is which (additional vs ignored)

Pijukatel (approved):

tests/unit/crawlers/_http/test_http_crawler.py
Mantisus: I think it's a bit of an overhead to add a `dataclass` that will only be used in one test. 🙂

vdusek (approved):
LGTM

janbuchar (approved):

PR #964, opened 2025-02-07
feat: Add adaptive context helpers

Pijukatel:
### Description

Add adaptive context helpers and documentation for AdaptivePlaywrightCrawler.

### Issues

- Closes: #249


docs/examples/playwright_crawler_adaptive.mdx
```diff
@@ -0,0 +1,20 @@
...
+title: AdaptivePlaywrightCrawler
+---
+
+import ApiLink from '@site/src/components/ApiLink';
+import CodeBlock from '@theme/CodeBlock';
+
+import AdaptivePlaywrightCrawlerExample from '!!raw-loader!./code/adaptive_playwright_crawler.py';
+
+This example demonstrates how to use <ApiLink to="class/AdaptivePlaywrightCrawler">`AdaptivePlaywrightCrawler`</ApiLink>. An <ApiLink to="class/AdaptivePlaywrightCrawler">`AdaptivePlaywrightCrawler`</ApiLink> is a combination of <ApiLink to="class/PlaywrightCrawler">`PlaywrightCrawler`</ApiLink> and some implementation of HTTP-based crawler like <ApiLink to="class/ParselCrawler">`ParselCrawler`</ApiLink> or <ApiLink to="class/BeautifulSoupCrawler">`BeautifulSoupCrawler`</ApiLink>.
+It uses more limited request handler interface so that it is able to switch to HTTP-only crawling when it detects it may be possible.
```
vdusek, line 12: There is missing explanation what `AdaptivePlaywrightCrawler` is doing. It seems like you assume the readers already know what it is.

docs/examples/playwright_crawler_adaptive.mdx
```diff
@@ -0,0 +1,20 @@
+---
```
vdusek, line 1: Could you wrap (all) the examples more aggresively so they fit on the page without the horizontal slider?

image

docs/examples/playwright_crawler_adaptive.mdx
```diff
@@ -0,0 +1,20 @@
...
+import CodeBlock from '@theme/CodeBlock';
+
+import AdaptivePlaywrightCrawlerExample from '!!raw-loader!./code/adaptive_playwright_crawler.py';
+
+This example demonstrates how to use <ApiLink to="class/AdaptivePlaywrightCrawler">`AdaptivePlaywrightCrawler`</ApiLink>. An <ApiLink to="class/AdaptivePlaywrightCrawler">`AdaptivePlaywrightCrawler`</ApiLink> is a combination of <ApiLink to="class/PlaywrightCrawler">`PlaywrightCrawler`</ApiLink> and some implementation of HTTP-based crawler like <ApiLink to="class/ParselCrawler">`ParselCrawler`</ApiLink> or <ApiLink to="class/BeautifulSoupCrawler">`BeautifulSoupCrawler`</ApiLink>.
+It uses more limited request handler interface so that it is able to switch to HTTP-only crawling when it detects it may be possible.
+
+A **pre-navigation hook** can be used to perform actions before navigating to the URL. This hook provides further flexibility in controlling environment and preparing for navigation. Hooks will be executed both for the pages crawled by HTTP-bases sub crawler and playwright based sub crawler. Use `playwright_only`=`True` to mark hooks that should be executed only for playwright sub crawler.
```
vdusek, line 14: pre-navigation hook - maybe link to API docs?

vdusek, line 14:
```suggestion
A **pre-navigation hook** can be used to perform actions before navigating to the URL. This hook provides further flexibility in controlling environment and preparing for navigation. Hooks will be executed both for the pages crawled by HTTP-bases sub crawler and playwright based sub crawler. Use `playwright_only=True` to mark hooks that should be executed only for playwright sub crawler.
```

src/crawlee/crawlers/_abstract_http/_abstract_http_parser.py
```diff
@@ -30,6 +30,29 @@ async def parse(self, response: HttpResponse) -> TParseResult:
...
+            Parsed text.
+        """
+
+    @abstractmethod
+    async def select(self, parsed_content: TParseResult, selector: str) -> TSelectResult | None:
+        """Use css selector to select page element and return it.
+
+        Args:
+            parsed_content: content where the page element will be located
+            selector: css selector
```
vdusek, line 50: Use sentences please.

src/crawlee/crawlers/_abstract_http/_abstract_http_parser.py
```diff
@@ -30,6 +30,29 @@ async def parse(self, response: HttpResponse) -> TParseResult:
             Parsed HTTP response.
         """
 
+    @abstractmethod
+    async def parse_text(self, text: str) -> TParseResult:
+        """Parse text.
+
```
vdusek, line 36: could you add something meaningful? like what the text is supposed to be for example

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -86,19 +87,45 @@ async def __aexit__(
...
-    Generic[TStaticCrawlingContext, TStaticParseResult],
+    Generic[TStaticCrawlingContext, TStaticParseResult, TStaticSelectResult],
     BasicCrawler[AdaptivePlaywrightCrawlingContext, AdaptivePlaywrightCrawlerStatisticState],
 ):
     """Adaptive crawler that uses both specific implementation of `AbstractHttpCrawler` and `PlaywrightCrawler`.
 
-    It tries to detect whether it is sufficient to crawl without browser (which is faster) or if
-    `PlaywrightCrawler` should be used (in case previous method did not work as expected for specific url.).
+    It uses more limited request handler interface so that it is able to switch to HTTP-only crawling when it detects it
+    may be possible.
```
vdusek, line 96: This is important docstring and this is really not much helpful if I do not know already what it shoudl be.

image

I would imagine something like:
- The first line - one sentence description of the class without referencing other things.
- Multiline paragraph description, with other references.

Check the others for inspiration.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
```diff
@@ -73,17 +76,76 @@ def response(self) -> Response:
...
         return self._response
 
+    async def wait_for_selector(self, selector: str, timeout: timedelta = timedelta(seconds=5)) -> None:
+        """Locate element by css selector a return once it is found.
+
+        If element is not found within timeout, `TimeoutError` is raised.
+
+        Args:
+            selector: css selector to be used to locate specific element on page.
+            timeout: timeout that defines how long the function wait for the selector to appear.
```
vdusek, line 86: Sentences

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
```diff
@@ -73,17 +76,76 @@ def response(self) -> Response:
...
+        await self.page.locator(selector).wait_for(timeout=timeout.total_seconds() * 1000)
+
+    async def query_selector(self, selector: str, timeout: timedelta = timedelta(seconds=5)) -> TStaticSelectResult:
+        """Locate element by css selector a return it once it is found.
+
+        If element is not found within timeout, `TimeoutError` is raised.
+
+        Args:
+            selector: css selector to be used to locate specific element on page.
+            timeout: timeout that defines how long the function wait for the selector to appear.
```
vdusek, line 99: Sentences

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
```diff
@@ -73,17 +76,76 @@ def response(self) -> Response:
...
+
+        Returns:
+            `TStaticSelectResult` which is result of used static parser `select` method.
```
vdusek, line 102: Don't be rendundant, the type is already in tyhe annotation and will be rendered anyways.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
```diff
@@ -73,17 +76,76 @@ def response(self) -> Response:
...
+        """Parse whole page with static parser. If `selector` argument is used, wait for selector first.
+
+        If element is not found within timeout, TimeoutError is raised.
+
+        Args:
+            selector: css selector to be used to locate specific element on page.
+            timeout: timeout that defines how long the function wait for the selector to appear.
+
+        Returns:
+            `TStaticParseResult` which is result of used static parser `parse_text` method.
```
vdusek, line 130: same as above

docs/examples/playwright_crawler_adaptive.mdx
janbuchar: Any chance we could let `ruff` do this?

docs/examples/playwright_crawler_adaptive.mdx
janbuchar: This is just an example though... A link to the appropriate guide would be nice though.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
```diff
@@ -73,17 +76,76 @@ def response(self) -> Response:
             raise AdaptiveContextError('Page was not crawled with PlaywrightCrawler.')
         return self._response
 
+    async def wait_for_selector(self, selector: str, timeout: timedelta = timedelta(seconds=5)) -> None:
+        """Locate element by css selector a return once it is found.
```
janbuchar, line 80: I think this does not return anythin... right?

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
```diff
@@ -73,17 +76,76 @@ def response(self) -> Response:
...
+        Args:
+            selector: css selector to be used to locate specific element on page.
+            timeout: timeout that defines how long the function wait for the selector to appear.
+
+        Returns:
+            `TStaticSelectResult` which is result of used static parser `select` method.
+        """
+        static_content = await self._static_parser.select(self.parsed_content, selector)
+        if static_content is not None:
+            return static_content
```
janbuchar, line 106: I'm not sure about this. If the element changes dynamically, this method will always return the version from the initial HTTP response. That might surprise users.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
```diff
@@ -73,17 +76,76 @@ def response(self) -> Response:
...
+
+        locator = self.page.locator(selector)
+        await locator.wait_for(timeout=timeout.total_seconds() * 1000)
+
+        parsed_selector = await self._static_parser.select(
+            await self._static_parser.parse_text(await locator.evaluate('el => el.outerHTML')), selector
+        )
+        if parsed_selector is not None:
+            return parsed_selector
+        raise AdaptiveContextError('Used selector is not a valid static selector')
```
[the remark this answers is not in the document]
Pijukatel: If selector did not match, then you will get `playwright._impl._errors.TimeoutError ` even before, when waiting for it.

This `AdaptiveContextError` happens in very rare case:
Waiting for the selector is done through `Playwright`.
Parsing is done through `static parser` which then uses the selector as well.

So in theory it can happen that user will pick selector supported by `Playwright`, but not the `static parser`. So Playwright will find the selector and wait for it,  `static parser` will manage to parse the page,but it will fail to do `select` with such unsupported selector. 

Playwright supported selectors and static parser supported selectors are two sets with some overlap.

So you will get this error only if you picked invalid `static parser` selector, but valid `Playwright` selector.

docs/guides/playwright_crawler_adaptive.mdx
```diff
@@ -0,0 +1,92 @@
...
+It uses more limited request handler interface so that it is able to switch to HTTP-only crawling when it detects it may be possible.
+
+Detection is done based on the <ApiLink to="class/RenderingTypePredictor">`RenderingTypePredictor`</ApiLink> with default implementation <ApiLink to="class/DefaultRenderingTypePredictor">`DefaultRenderingTypePredictor`</ApiLink>. It predicts which crawling method should be used and learns from already crawled pages.
+
+## When to use AdaptivePlaywrightCrawler
+
+Use <ApiLink to="class/AdaptivePlaywrightCrawler">`AdaptivePlaywrightCrawler`</ApiLink> in scenarios where some target pages have to be crawled with <ApiLink to="class/PlaywrightCrawler">`PlaywrightCrawler`</ApiLink>, but for others faster HTTP-based crawler is sufficient. This way, you can achieve lower costs when crawling multiple different websites.
```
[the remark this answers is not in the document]
Pijukatel: Added.

docs/guides/playwright_crawler_adaptive.mdx
```diff
@@ -0,0 +1,92 @@
...
+import AdaptivePlaywrightCrawlerInitParsel from '!!raw-loader!./code/adaptive_playwright_crawler/adaptive_playwright_crawler_init_parsel.py';
+import AdaptivePlaywrightCrawlerInitPrediction from '!!raw-loader!./code/adaptive_playwright_crawler/adaptive_playwright_crawler_init_prediction.py';
+import AdaptivePlaywrightCrawlerHandler from '!!raw-loader!./code/adaptive_playwright_crawler/adaptive_playwright_crawler_handler.py';
+import AdaptivePlaywrightCrawlerPreNavHooks from '!!raw-loader!./code/adaptive_playwright_crawler/adaptive_playwright_crawler_pre_nav_hooks.py';
+
+
+
+An <ApiLink to="class/AdaptivePlaywrightCrawler">`AdaptivePlaywrightCrawler`</ApiLink> is a combination of <ApiLink to="class/PlaywrightCrawler">`PlaywrightCrawler`</ApiLink> and some implementation of HTTP-based crawler like <ApiLink to="class/ParselCrawler">`ParselCrawler`</ApiLink> or <ApiLink to="class/BeautifulSoupCrawler">`BeautifulSoupCrawler`</ApiLink>.
+It uses more limited request handler interface so that it is able to switch to HTTP-only crawling when it detects it may be possible.
```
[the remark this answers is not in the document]
Pijukatel: Updated.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
Pijukatel: Fixed

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
Pijukatel: Fixed

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
Pijukatel: Fixed

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
Pijukatel: Updated

src/crawlee/crawlers/_abstract_http/_abstract_http_parser.py
Pijukatel: Fixed

docs/examples/playwright_crawler_adaptive.mdx
Pijukatel: That is already `ruff`ed. If I try to wrap more, `ruff` will fix my changes.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
Pijukatel: In scenario where element exists in static page, but will change dynamically in the future, the adaptive crawler has no chance of seeing this future change. So it will pick the static value from the presence.
I am having hard time imagining the scenario for this. 

The only solution that comes to my mind is keeping optional attribute "latest_parsed_content" that would default to `parsed_content`, but would be updated each time someone already explicitly re-parsed the content or you would have to re-parse the page in the beginning of each call to make sure it is doing the select on the most up-to-date version. So there could be quite a performance penalty for this. 

I was under the impression that this is the reason for having the `parse_with_static_parser` context helper.

docs/examples/playwright_crawler_adaptive.mdx
janbuchar: Could we set a different line length for the examples?

vdusek: > Could we set a different line length for the examples?

If it is possible, that would be great.

janbuchar: In the worst case, we could run ruff on the examples folder separately, with a different config. But I hope there's a built-in way.

Pijukatel: There is per-file ignore, but I do not think there is per-file override. Given the fact, that this is current state of the documentation and exactly the same situation is present in many current examples I would not like to fix it in this change.
Let's do it in dedicated PR: https://github.com/apify/crawlee-python/issues/970

docs/examples/playwright_crawler_adaptive.mdx
Pijukatel: In this rare case I think it would not be suitable to add the link to API docs. AdaptivePlaywrightCrawler#pre_navigation_hook is showing the parametrized decorator doc page and even though it is correct, I think it would be confusing to regular users. 
For example return type is `Callable[[Callable[[AdaptivePlaywrightPreNavCrawlingContext], Awaitable[None]]], None]` :-)

So I would suggest to show link to the guide page that shows how to use hooks and describes them in more detail:
...python/docs/guides/adaptive-playwright-crawler#page-configuration-with-pre-navigation-hooks

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
Pijukatel: Updated with automatic re-parsing

vdusek (approved):
LGTM

janbuchar:
@TC-MO could you please check the guide for style issues and general readability? I'm heavily biased since I already understand the feature :slightly_smiling_face:

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
```diff
@@ -73,17 +76,82 @@ def response(self) -> Response:
             raise AdaptiveContextError('Page was not crawled with PlaywrightCrawler.')
         return self._response
 
+    async def _get_latest_parsed_content(self) -> TStaticParseResult:
+        """Return parsed page. No need to parse again in static context. Reparse in Playwright context."""
+        if self._page:
+            return await self.parse_with_static_parser()
```
janbuchar, line 82: I would prefer if `parse_with_static_parser` delegated to this method rather than vice versa, but no need to bend over backwards if it's not easily doable.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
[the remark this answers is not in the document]
janbuchar: I think we're not completely aligned on how the method should work. The JS version waits for the element when it's running in the browser and delegates to the static parser otherwise. If the static parser doesn't find anything, that's a valid result. It should be perfectly possible to make two branches here based on `self._page is None`

I'm sorry if I caused a misunderstanding how this should work.

janbuchar (commented):
Nice, I have a couple of nitpicks and one more serious comment.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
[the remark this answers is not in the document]
Pijukatel: There are following possible scenarios:

1. static_parser does not find the selector and browser does not find the selector -> will raise `playwright._impl._errors.TimeoutError`
2. static_parser finds the selector immediately (whether browser finds it as well is irrelevant then, so two scenarios collapse to one) -> will return parsed selector
3. static_parser not does find the selector immediately and browser finds it after a while. static_parser finds it after the wait time as well -> will return parsed selector
4. static_parser not does find the selector immediately and browser finds it after a while. static_parser does not find it even after wait time-> will raise `AdaptiveContextError('Used selector is not a valid static selector')` . Because the element actually exists in the page, it was found by the browser by used selector, but it was not found using static_parser, because it does not support such selector.

Branch where `self._page is None` and the static parser does not find anything is "hidden" to the user as this will raise `AdaptiveContextError`, this will force the AdaptiveCrawler to use Playwright instead and this will lead to scenario 1, 3 or 4.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
```diff
@@ -73,17 +76,82 @@ def response(self) -> Response:
...
+
+        Args:
+            selector: Css selector to be used to locate specific element on page.
+            timeout: Timeout that defines how long the function wait for the selector to appear.
+        """
+        if await self._static_parser.select(await self._get_latest_parsed_content(), selector):
+            return
+        await self.page.locator(selector).wait_for(timeout=timeout.total_seconds() * 1000)
+
+    async def query_selector(self, selector: str, timeout: timedelta = timedelta(seconds=5)) -> TStaticSelectResult:
```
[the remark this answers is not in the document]
Pijukatel: Yes, that is a good point. I am wondering whether it is even needed to have `query_selector_one` which is basically `query_selector_all(...)[0]`, but it is not uncommon is such search query context to have such helper method.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -86,19 +87,46 @@ async def __aexit__(
...
+    It uses a more limited crawling context interface so that it is able to switch to HTTP-only crawling when it detects
+    that it may bring a performance benefit.
+    It uses specific implementation of `AbstractHttpCrawler` and `PlaywrightCrawler`.
+
+    ### Usage
+    ```python
+    from crawlee.crawlers import AdaptivePlaywrightCrawler, AdaptivePlaywrightCrawlingContext
+
+    crawler = AdaptivePlaywrightCrawler.with_beautifulsoup_static_parser(
+        max_requests_per_crawl=5, playwright_crawler_specific_kwargs={'headless': False}
```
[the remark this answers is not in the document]
Pijukatel: This was just an example argument to make it clear that how to pass Playwright related arguments.

(Also I though that this argument is actually good for users that will try to run example locally and it would be useful for their understanding to see Playwright in headless=False mode so that they could see, that only some pages are crawled by browser. But this is just a guess, I can replace it by other argument like `browser_type`)

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
Pijukatel: Got rid of the method. It can be in the `parse_with_static_parser`.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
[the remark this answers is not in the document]
Pijukatel: Changed to use `browser_type` argument as example

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawling_context.py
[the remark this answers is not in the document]
Pijukatel: Added `query_selector_all`

docs/examples/code/adaptive_playwright_crawler.py
```diff
@@ -10,40 +11,43 @@
...
 
     @crawler.pre_navigation_hook
     async def hook(context: AdaptivePlaywrightPreNavCrawlingContext) -> None:
-        """Hook executed both in static sub crawler and playwright sub crawler."""
-        # Trying to access context.page in this hook would raise `AdaptiveContextError`
-        # for pages crawled without playwright.
+        """Hook executed both in static sub crawler and playwright sub crawler.
+
+        Trying to access `context.page` in this hook would raise `AdaptiveContextError`
+        for pages crawled without  playwright."""
```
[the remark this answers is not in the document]
Pijukatel: Fixed

src/crawlee/crawlers/_beautifulsoup/_beautifulsoup_parser.py
```diff
@@ -32,8 +32,8 @@ def is_matching_selector(self, parsed_content: Tag, selector: str) -> bool:
         return parsed_content.select_one(selector) is not None
 
     @override
-    async def select(self, parsed_content: Tag, selector: str) -> Tag | None:
-        return parsed_content.select_one(selector)
+    async def select(self, parsed_content: Tag, selector: str) -> tuple[Tag, ...]:
```
janbuchar: this is breaking - are you sure we want to go through with it? why not expose the one/all variants while we're at it?

Pijukatel: It is not breaking, this method is added in this change. I am not sure about benefit of adding one/all variants to `Parser`. `BeautifulSoup` has it, `Parsel` does not have it on `Selector` level. It returns `SelectorList` and one/all choice is done on next level, in data extraction. `SelectorList.get()` / `SelectorList.getAll()`.

In the end it is just a convenience method, so I think it is fine to have only on the final user facing context.

janbuchar: Right, sorry! In the longer run, I'd like to see a unified interface for picking an element by selector on all crawling context variants. I'm not sure if this is the right PR to do this, but we should be careful not to shoot ourselves in the foot here... so that we can design something uniform later.

PR #974, opened 2025-02-11
chore: Fix flakiness in test and use explicit test id

Pijukatel:
### Description
Replace unreliable `asyncio.sleep` in flaky `test_crawler_multiple_stops_in_parallel` by deterministic `asyncio.Barrier` and `asyncio.Semaphore`
Use explicit test name through `id=` in parametrized `test_html_to_text`  where parameter is several line long string. (To prevent strange test case name in logs.)


janbuchar (approved):
Nice!

Issue #906, opened 2025-01-15
Breaking changes for v0.6


vdusek:
### Remove unused fields in the `Configuration`
 
- Fields to remove: `chrome_executable_path`, `xvfb`.
- Additionally, consider removing the `verbose_log` field:
    - The `verbose_log` field is only used to derive the log level, and a `log_level` option already exists. So I consider it as redundant.

### Refactor abstract class naming

- Remove "Base" from the names of abstract classes (e.g., `BaseStorage`, `BaseStorageClient`, ...).
- Together with @janbuchar we came to a conclusion not to use the Hungarian notation (where the name of a variable indicates its intention or kind). @Pijukatel, feel free to share your thoughts.

### ~Rename `enqueue_links` for clarity~

- ~The naming of `enqueue_links` causes confusion due to its similarity to `add_requests`.~
- ~Proposed solution:~
  - ~Rename `enqueue_links` to something more descriptive, such as `extract_links` (indicating what it does more precisely).~
- ~I am open to name suggestions here.~

### Update of `enqueue_links`

- `enqueue_links` will have the same interface as its JavaScript counterpart, [enqueueLinks](https://crawlee.dev/api/core/function/enqueueLinks).
- `add_requests` will remain dedicated to adding requests to the RQ only.
- A new `extract_links` function can be introduced for link extraction only.
- Internally, `enqueue_links` can utilize both `add_requests` and `extract_links`.

Pijukatel:
Regarding abstract class naming, just one more thing to think about. It will involve both **Base**SomeClass and **Abstract**SomeClass. What will we do when we have default implementation of such class? For example HttpCrawler is one implementation of AbstractHttpCrawler. (Where HttpCrawler was reserved for backwards compatibility)

vdusek:
Maybe for `AbstractHttpCrawler` we would have to make an exception. Or rename the `HttpCrawler` to something else.

Mantisus:
I'm thinking of moving from a `dict` to a `CookieJar`-based approach for storing cookies in `Session` for #933, would be appropriate for 0.6

vdusek:
One more thing from Ruff 0.9:

```
src/crawlee/statistics/__init__.py:1:1: A005 Module `statistics` shadows a Python standard-library module
```

Since `statistics` is a public module, we should consider renaming it for v0.6 release. Maybe just `stats`? Or do you have any suggestions?

@Pijukatel @janbuchar

janbuchar:
[...]

PR #981, opened 2025-02-12
refactor!: rename `enqueue_links` to `extract_links`

vdusek:
Relates: https://github.com/apify/crawlee-python/issues/906


Pijukatel (commented):
Change is OK, but I am still not 100% sure about the rename. 
Previously the `enqueue_links` function could cause confusion, because some people didn't realize, that it is extracting the links first. 

I fear we will be in similar situation, where some people could be confused with the new name and not realize it is enqueuing links.

This function does two separate actions `extract` and `enqueue`, so if we choose just one of those actions to be in the name, we will always risk confusing users.

janbuchar:
It's quite a significant change that we want to propagate to the JS version as well, so it's better to raise concerns now than to wait until we get flak from users :slightly_smiling_face: 

It is true that `context.extract_links` and `context.add_requests` are not exactly orthogonal with the current implementation, but to achieve that, we would have to make users call `context.add_requests(context.extract_links())`, and that kinda kills the magic (in a bad way). The flip side is that new users might expect `context.extract_links` to return a bunch of URLs or whatever, but they should get the hint when they see that the return type is `None`. But feel free to challenge this.

In any way, the confusion raised by `enqueue_links` and `add_requests` feels much larger than any of this...

Pijukatel:
> It's quite a significant change that we want to propagate to the JS version as well, so it's better to raise concerns now than to wait until we get flak from users 🙂
> 
> It is true that `context.extract_links` and `context.add_requests` are not exactly orthogonal with the current implementation, but to achieve that, we would have to make users call `context.add_requests(context.extract_links())`, and that kinda kills the magic (in a bad way). The flip side is that new users might expect `context.extract_links` to return a bunch of URLs or whatever, but they should get the hint when they see that the return type is `None`. But feel free to challenge this.
> 
> In any way, the confusion raised by `enqueue_links` and `add_requests` feels much larger than any of this...
> 
> EDIT: Maybe we could make it more vague and call it `process_links`, that would push the meaning a bit closer to "extract AND enqueue", but still further from "accept a bunch of links and put them in the queue" (which is what people seem to expect from `enqueue_links`. Not sure, just a thought.

I would prefer to be more explicit at the cost of more ugly name and since the function does two things, I would not mind having "and" in the name. 

Common argument against function names with "and" in the name is that it is a red flag telling that the function does more than one thing and should be split. Hiding that fact by abstract name is not going to change that, so I would rather be explicit and open about it.

vdusek:
After internal discussions, we have decided not to proceed with this change. For more details and next steps, see #906.

PR #982, opened 2025-02-12
chore: rm Apify as optional dependecy

vdusek:


Pijukatel (commented):
I see that `from apify import Actor` is used in several code examples. I would suggest to add comment to each such line, explicitly mentioning why those examples will not work out of the box. Similar to comment about external package in Camoufox example: 
https://github.com/apify/crawlee-python/blob/master/docs/examples/code/playwright_crawler_with_camoufox.py#L3

vdusek:
I'm unsure if it's necessary to mention this directly in the source code since there's already a "Requirements" section (or a similar one) explaining the need to install the Apify SDK. This was not the case for one of the related page, so I added it there.

Issue #710, opened 2024-11-18
Add session cookies to crawling context


Mantisus:
Add to the context, the cookie of the session from which the request was made, both for HTTP crawlers and Playwright

janbuchar:
Could you elaborate please? For plain HTTP crawlers, you can use `context.http_response.headers`. Is this about accessing the request in Playwright?

Mantisus:
I missed that, for HTTP crawlers we have `context.http_response.headers` for Playwright this is not as relevant.

janbuchar:
It is true that it would be hard to reach the headers via `PlaywrightCrawlingContext` though.

Mantisus:
It can be useful for Playwright to have access to the cookie of the session from which the request was made.

But not directly to the headers

janbuchar:
Feel free to rephrase the issue title and description then :slightly_smiling_face:

oldsiks:
When using PlaywrightCrawler in Crawlee for web scraping, how can I add cookies? Could you provide an example?

Mantisus:
Hey @oldsiks

Thank you for your interest in crawlee.

Here is an example of using a cookie at the request header level

```python
import asyncio

from crawlee import Request
from crawlee.playwright_crawler import PlaywrightCrawler, PlaywrightCrawlingContext


async def main() -> None:
    crawler = PlaywrightCrawler()

    @crawler.router.default_handler
    async def request_handler(context: PlaywrightCrawlingContext) -> None:
        print(await context.page.content())

    await crawler.run([
        Request.from_url(url='https://httpbin.org/get', headers={'cookie': 'my_cookies'})
        ])

asyncio.run(main())
```

Also after release 0.5 it will be possible to set cookie in `PlaywrightBrowserPlugin` using `browser_new_context_options`

Example for 0.5:
```python
import asyncio

from crawlee.crawlers import PlaywrightCrawler, PlaywrightCrawlingContext
from crawlee.browsers import BrowserPool, PlaywrightBrowserPlugin


async def main() -> None:
    user_plugin = PlaywrightBrowserPlugin(
        browser_new_context_options={"extra_http_headers": {'cookie': 'my_cookies'}}
        )

    browser_pool = BrowserPool(plugins=[user_plugin])

    crawler = PlaywrightCrawler(browser_pool=browser_pool)

    @crawler.router.default_handler
    async def request_handler(context: PlaywrightCrawlingContext) -> None:
        print(await context.page.content())

[...]

Issue #963, opened 2025-02-06
Add support for launching Playwright with a persistent browser context


vdusek:
### Description

It looks like our current Python implementation always uses an incognito context when launching Playwright ([launch](https://playwright.dev/python/docs/api/class-browsertype#browser-type-launch), which is the default behavior. However, in JavaScript, the default is a persistent context unless explicitly set otherwise ([launch_persistent_context](https://playwright.dev/python/docs/api/class-browsertype#browser-type-launch-persistent-context)).

This will probably be a breaking change (it's changing the default behaviour).

### Relevant code

- https://github.com/apify/crawlee-python/blob/v0.5.4/src/crawlee/browsers/_playwright_browser_controller.py#L142:L155
- https://github.com/apify/crawlee-python/blob/v0.5.4/src/crawlee/browsers/_playwright_browser_plugin.py#L156:L163

B4nan:
PoC implementation #410

Pijukatel:
There should be one unit test that will test that when using **persistent browser context** cookies are retained even when **context.session** changes, as discussed here https://github.com/apify/crawlee-python/pull/984#discussion_r1955999007

PR #984, opened 2025-02-13
refactor!: change `Session` cookies from `dict` to `SessionCookies` with `CookieJar`

Mantisus:
### Description

- Change the cookie storage in the `Session` from `dict` to `SessionCookies` which uses `CookieJar`. It supports basic cookie parameters and multiple domains.

### Issues

- Closes: #933
- Closes: #710

### Testing

- Update old tests
- Add tests for new class
- Add tests for multidomain support


src/crawlee/crawlers/_playwright/_playwright_crawler.py
```diff
@@ -202,7 +201,8 @@ async def _navigate(
         """
         async with context.page:
             if context.session:
-                await self._set_cookies(context.page, context.request.url, context.session.cookies)
+                cookies = context.session.cookies.get_cookies_as_browser_format()
```
Pijukatel, line 204: I am little confused about what we want in the end regarding the scope of cookies. This is more of a discussion point to make sure we are all on the same page.

Here in this change we make cookies connected to the session. So cookies are "session scoped",meaning that cookies are shared in the same session.

Then there is [this issue ](https://github.com/apify/crawlee-python/issues/963). That will probably be implemented by using [launch](https://playwright.dev/python/docs/api/class-browsertype#browser-type-launch) and [launch_persistent_context](https://playwright.dev/python/docs/api/class-browsertype#browser-type-launch-persistent-context). From perspective of that issue I would expect the cookies to be "browser context scoped", meaning that cookies are shared in the same browser context.

Those two approaches seems mutually exclusive to me. Maybe you have clear vision of how it should be done. Could you please describe it clear any confusion.

B4nan: > Those two approaches seems mutually exclusive to me, as you can use same browser context with different (crawling)context.session and vice versa same (crawling)context.session with different browser contexts. Maybe you have clear vision of how it should be done. Could you please describe it, to clear any confusion.

I wouldn't call that mutually exclusive, persistent context should keep cookies, just like a real browser would. That is a feature of the browser itself. Sessions are our abstraction, they are not supposed to isolate cookies in one browser context, for that you need incognito contexts (which is our current default). I can imagine the limitations will be the same in python - you cannot change a proxy in an existing browser context, and a session is not just about cookies, it's about the proxy too.

Pijukatel: Ok, I think I am starting to understand. One more question about our current implementation, especially change that introduced "use_incognito_pages" https://github.com/apify/crawlee-python/pull/941/files

Currently we always use just `launch` and we never use `launch_persistent_context`.

When looking into [Add support for launching Playwright with a persistent browser](https://github.com/apify/crawlee-python/issues/963) there is a specific wording `our current Python implementation always uses an incognito context when launching PlaywrightI`, so I got under impression, that "use_incognito_pages" argument would map to `launch` method and normal (non-incognito) would map to `launch_persistent_context`.

But in current implementation we by default use incognito context, but still share cookies. Only when we use `use_incognito_pages` we use incognito context without sharing the cookies.

So `use_incognito_pages` argument and `incognito context` are two similarly sounding, but different things after all?

Pijukatel: > I wouldn't call that mutually exclusive, persistent context should keep cookies, just like a real browser would. That is a feature of the browser itself. Sessions are our abstraction, they are not supposed to isolate cookies in one browser context, for that you need incognito contexts (which is our current default). I can imagine the limitations will be the same in python - you cannot change a proxy in an existing browser context, and a session is not just about cookies, it's about the proxy too.

So those different use cases are ok as described?:

1. We use same persistent context, but session changes. -> We will loose cookies previously used in the same persistent context.
2. We use same session, but browser context changes. -> We will loose cookies previously used in the same session.

B4nan: > So use_incognito_pages argument and incognito context are two similarly sounding, but different things after all?

Nope, it's the very same thing. Our current implementation was just reusing one incognito context, which is not much different from using a persistent context.

B4nan: > We use same persistent context, but session changes. -> We will loose cookies previously used in the same persistent context.

I don't think cookies should be lost by that in a persistent context. I am not completely sure how things behave right now in the JS version, but I believe we keep the cookies, again, its the browser that keeps them, not our code.

> We use same session, but browser context changes. -> We will loose cookies previously used in the same session.

I don't think this should be possible. In crawlee v4 we want to make sure the session is tied to a browser context basically, in other words, we want the session to actually represent a user, with the same proxy, cookies and fingerprint.

Pijukatel: >  again, its the browser that keeps them, not our code.

Ok, I think that is the main source of my confusion, because in current Python implementation it is our code that keeps them and not the browser.

Mantisus: > We use same persistent context, but session changes. -> We will loose cookies previously used in the same persistent context.

When we use a single context, there will be cookie flows between sessions. So in fact all sessions will have the same cookies.

> We use same session, but browser context changes. -> We will loose cookies previously used in the same session.

Since we store the cookie in the session. We will write the cookie to a new context before calling the page transition. The cookie will not be lost in the session.

Mantisus: > Ok, I think that is the main source of my confusion, because in current Python implementation it is our code that keeps them and not the browser.

They are stored by both the browser and our code. This is similar to the current TS implementation.

https://github.com/apify/crawlee/blob/master/packages/browser-crawler/src/internals/browser-crawler.ts#L560
https://github.com/apify/crawlee/blob/master/packages/browser-crawler/src/internals/browser-crawler.ts#L672

Cookies in a session, allow the user to access the cookie. Read them, set new ones. And also restore cookies from the session if the context has been closed or changed.

Pijukatel: > Cookies in a session, allow the user to access the cookie. Read them, set new ones. And also restore cookies from the session if the context has been closed or changed.

So it can happen that cookies on the session are different than cookies on the browser context?

Mantisus: Yes, if for example the user writes new cookies to the session in the handler with `pre_navigation_hook`.
However, before navigating to the page, we will write these cookies to the browser context.

vdusek (changes requested):
Since it contains a breaking change, could you please describe the breaking changes in the PR's description and also summarize it in the [Upgrading guide](https://github.com/apify/crawlee-python/blob/master/docs/upgrading/upgrading_to_v0x.md)?

src/crawlee/crawlers/_playwright/_playwright_crawler.py
```diff
@@ -329,17 +329,14 @@ def pre_navigation_hook(self, hook: Callable[[PlaywrightPreNavCrawlingContext],
...
 
-    async def _get_cookies(self, page: Page) -> dict[str, str]:
+    async def _get_cookies(self, page: Page) -> list[dict[str, Any]]:
         """Get the cookies from the page."""
         cookies = await page.context.cookies()
-        return {cookie['name']: cookie['value'] for cookie in cookies if cookie.get('name') and cookie.get('value')}
+        return [{**cookie} for cookie in cookies]
 
-    async def _set_cookies(self, page: Page, url: str, cookies: dict[str, str]) -> None:
+    async def _set_cookies(self, page: Page, cookies: list[dict[str, Any]]) -> None:
```
Pijukatel, line 337: This name was here already before, but I would consider changing it to `_add_cookies` / `_update_cookies` or similar

`_set_cookies` creates an impression that we will set context cookies to whatever is the list of input cookies and throw away any previous values. But in fact we are just adding to existing cookies and not throwing them away.

Or maybe this alias is no longer needed and we can just call `page.context.add_cookies(cookies)` directly

src/crawlee/sessions/_cookies.py
```diff
@@ -0,0 +1,161 @@
...
+            rest={'HttpOnly': ''} if http_only else {},
+            rfc2109=False,
+        )
+
+        if same_site:
+            cookie.set_nonstandard_attr('SameSite', same_site)
+
+        self.jar.set_cookie(cookie)
+
+    def _convert_cookie_to_dict(self, cookie: Cookie) -> dict[str, Any]:
```
Pijukatel, line 70: Isn't this type `dict[str, Any]` too permissive? (Same comment for all places where this cookie dict is used)

Maybe it would be worth creating TypedDict for cookies. It is basically this one, right?

https://playwright.dev/python/docs/api/class-browsercontext#browser-context-add-cookies-option-cookies
`from playwright._impl._api_structures import SetCookieParam`

Mantisus: Yes, I think TypedDict will indeed be better in this situation.

Playwright uses different TypedDict for `add_cookies` and `cookies`. And they have slight differences with `cookie.set`. So our TypedDict is slightly different.

vdusek (commented):
There are failing tests for Windows Python 3.12.

Mantisus:
> There are failing tests for Windows Python 3.12.

The problem is not related to this PR.

It's solved in the PR #1007

Pijukatel (dismissed):

docs/upgrading/upgrading_to_v0x.md
```diff
@@ -13,6 +13,10 @@ This section summarizes the breaking changes between v0.5.x and v0.6.0.
 
 - Removed `HttpCrawlerOptions` - which contained options from `BasicCrawlerOptions` and unique options `additional_http_error_status_codes` and `ignore_http_error_status_codes`. Both of the unique options were added to `BasicCrawlerOptions` instead.
 
+### SessionCookies
+
+- Replaces the `dict` used for cookie storage in `Session` with a new `SessionCookies` class. `SessionCookies` uses `CookieJar`, which enables support for multiple domains.
```
vdusek, line 18: Could you please mention also the field name - `Session.cookies`?

src/crawlee/sessions/_cookies.py
```diff
@@ -0,0 +1,193 @@
...
+    secure: NotRequired[bool]
+
+
+class CookieParam(BaseCookieParam, total=False):
+    http_only: NotRequired[bool]
+    expires: NotRequired[int]
+    same_site: NotRequired[Literal['Lax', 'None', 'Strict']]
+
+
+class PWCookieParam(BaseCookieParam, total=False):
```
vdusek, line 25: We haven't used the "PW" anywhere so far. Could you please use the full name - `PlaywrightCookieParam`, instead?

src/crawlee/crawlers/_playwright/_playwright_crawler.py
```diff
@@ -329,17 +330,14 @@ def pre_navigation_hook(self, hook: Callable[[PlaywrightPreNavCrawlingContext],
         """
         self._pre_navigation_hooks.append(hook)
 
-    async def _get_cookies(self, page: Page) -> dict[str, str]:
+    async def _get_cookies(self, page: Page) -> list[PWCookieParam]:
         """Get the cookies from the page."""
         cookies = await page.context.cookies()
-        return {cookie['name']: cookie['value'] for cookie in cookies if cookie.get('name') and cookie.get('value')}
+        return [cast(PWCookieParam, cookie) for cookie in cookies]
```
vdusek, line 336: Couldn't we get rid of the caste here please?

src/crawlee/sessions/_cookies.py
```diff
@@ -0,0 +1,193 @@
...
+    value: Required[str]
+    domain: NotRequired[str]
+    path: NotRequired[str]
+    secure: NotRequired[bool]
+
+
+class CookieParam(BaseCookieParam, total=False):
+    http_only: NotRequired[bool]
+    expires: NotRequired[int]
+    same_site: NotRequired[Literal['Lax', 'None', 'Strict']]
```
vdusek, line 22: This is being exposed to public, correct? In that case, please add docstrings and docs group decorator.

src/crawlee/sessions/_cookies.py
```diff
@@ -0,0 +1,193 @@
+from __future__ import annotations
+
+from copy import deepcopy
+from http.cookiejar import Cookie
+from typing import Any, ClassVar, Literal, cast
+
+from httpx import Cookies
```
vdusek, line 7: I would probably use `HttpxCookies` alias here for better readability across the module (our cookie types vs http's vs httpx's :exploding_head:)

src/crawlee/sessions/_cookies.py
```diff
@@ -0,0 +1,193 @@
...
+    same_site: NotRequired[Literal['Lax', 'None', 'Strict']]
+
+
+class PWCookieParam(BaseCookieParam, total=False):
+    httpOnly: NotRequired[bool]
+    expires: NotRequired[float]
+    sameSite: NotRequired[Literal['Lax', 'None', 'Strict']]
+
+
+class SessionCookies(Cookies):
```
vdusek, line 31: docs group decorator please so it can be rendered in API docs

src/crawlee/sessions/_cookies.py
vdusek: To be honest, I'm not sure it's a good idea to depend on HTTPX it this case anyway. How difficult would it be to build our own Cookies class?

Primarily, it's a library for HTTP communication, not for providing types for working with cookies. Its types could be coupled with its own client usage, whereas we also use Playwright, curl-impersonate, and our [impit](https://github.com/apify/impit) in the future.

For example, we previously relied on the `URL` class from HTTPX, but it didn't work out well, leading us to remove it and switch to `yarl` instead.

vdusek (changes requested):
Looks good, just a few minor comments and one concern regarding the usage of HTTPX's cookies.

src/crawlee/sessions/_cookies.py
Mantisus: I agree.

Updated without using `httpx.Cookies` as a parent class. Since our main task is to store cookies and convert them into appropriate formats. We don't need the same functionality that http clients use.

src/crawlee/sessions/_cookies.py
```diff
@@ -177,17 +195,59 @@ def set_cookies(self, cookie_dicts: list[CookieParam]) -> None:
...
-        # This is necessary because cookijars use `RLock`, which prevents `deepcopy`.
+        # This is necessary because `CookieJar` use `RLock`, which prevents `deepcopy`.
         cookie_dicts = self.get_cookies_as_dicts()
-        return self.__class__.from_dict_list(deepcopy(cookie_dicts, memo))
+        return self.__class__(deepcopy(cookie_dicts, memo))
+
+    def __len__(self) -> int:
+        return len(self._jar)
+
+    def __set__(self, name: str, value: str) -> None:
```
Pijukatel: I am not sure I understand this line as I haven't used descriptors much, but isn't the second argument supposed to be the instance of the object containing `SessionCookies` as attribute?

https://docs.python.org/3/reference/datamodel.html#object.__set__

src/crawlee/sessions/_cookies.py
```diff
@@ -177,17 +195,59 @@ def set_cookies(self, cookie_dicts: list[CookieParam]) -> None:
...
+
+        if len(self) != len(other):
+            return False
+
+        self_cookies = {
+            (cookie.name, cookie.value, cookie.domain, cookie.path): self._convert_cookie_to_dict(cookie)
+            for cookie in self._jar
+        }
+
+        for cookie in other.jar:
```
Pijukatel: If I guess it correctly this is some optimization? Is it really necessary, why not just compare two sets? 
`return self_cookies==other_cookies`

src/crawlee/sessions/_cookies.py
```diff
@@ -0,0 +1,268 @@
...
+    def store_cookie(self, cookie: Cookie) -> None:
+        """Store a Cookie object in the session cookie jar.
+
+        Args:
+            cookie: The Cookie object to store in the jar.
+        """
+        self.jar.set_cookie(cookie)
+
+    def store_cookies(self, cookies: list[Cookie]) -> None:
+        """Store multiple Cookie objects in the session cookie jar.
```
vdusek, line 197:
```suggestion
        """Store multiple cookie objects in the session cookie jar.
```

src/crawlee/sessions/_cookies.py
```diff
@@ -0,0 +1,268 @@
...
+
+        Args:
+            cookies: A list of Cookie objects to store in the jar.
```
vdusek, line 200:
```suggestion
            cookies: A list of cookie objects to store in the jar.
```

src/crawlee/sessions/_cookies.py
```diff
@@ -0,0 +1,268 @@
...
+    """Domain for which the cookie is set."""
+
+    path: NotRequired[str]
+    """Path on the specified domain for which the cookie is set."""
+
+    secure: NotRequired[bool]
+    """Set the `Secure` flag for the cookie."""
+
+    http_only: NotRequired[bool]
+    """Set the `HttpOnly` flag for the cookie."""
```
vdusek, line 35: Or does `HttpOnly` reference anything specific?

```suggestion
    """Set the HTTP only flag for the cookie."""
```

src/crawlee/sessions/_cookies.py
```diff
@@ -0,0 +1,268 @@
...
+    """Cookie value."""
+
+    domain: NotRequired[str]
+    """Domain for which the cookie is set."""
+
+    path: NotRequired[str]
+    """Path on the specified domain for which the cookie is set."""
+
+    secure: NotRequired[bool]
+    """Set the `Secure` flag for the cookie."""
```
vdusek, line 32: Or does `Secure` reference anything specific?

```suggestion
    """Set the secure flag for the cookie."""
```

src/crawlee/sessions/_cookies.py
```diff
@@ -0,0 +1,268 @@
...
+
+    http_only: NotRequired[bool]
+    """Set the `HttpOnly` flag for the cookie."""
+
+    expires: NotRequired[int]
+    """Expiration date for the cookie, None for a session cookie."""
+
+    same_site: NotRequired[Literal['Lax', 'None', 'Strict']]
+    """Set the `SameSite` attribute for the cookie."""
```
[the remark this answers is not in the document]
Mantisus: Yep, ['SameSite'](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#samesitesamesite-value) is the exact spelling of the flag in the cookie string.

src/crawlee/sessions/_cookies.py
Mantisus: [Secure](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#secure) - is the exact spelling of the flag in the cookie string.

vdusek: thanks

src/crawlee/sessions/_cookies.py
vdusek: thanks

src/crawlee/sessions/_cookies.py
```diff
@@ -0,0 +1,268 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
vdusek: After a second thought and a Slack discussion, I am not sure whether this is 100% good idea. 

@janbuchar Could we ask you for an opinion?

Otherwise, this PR is LGTM for me.

vdusek (approved):
LGTM and great job, just wait for the resolution of the last unresolved conversation

src/crawlee/sessions/_cookies.py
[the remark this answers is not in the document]
janbuchar: What I like about the current solution is how it makes it clear that `PlaywrightCookieParam` is, well, for Playwright. If you put the names used by Playwright into the field aliases, this information will be hidden.

On the other hand, I don't like that the `TypedDict` has camelCased keys :grin: Are there any other problems with the current implementation?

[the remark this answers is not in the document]
vdusek: resolving after internal discussion

Issue #721, opened 2024-11-21
Implement option for persistent context to PlaywrightCrawler


vdusek:
- Implement an option for persistent context (`user_data_dir`) to PlaywrightCrawler in a similar way as it is in the Crawlee JS.
- https://crawlee.dev/api/browser-pool/interface/BrowserPluginOptions#userDataDir
- Before implementation sync with @barjin, as he can provide further context and also suggest potential improvements.

PR #985, opened 2025-02-15
refactor!: сhange default `incognito context` to `persistent context` for `Playwright`

Mantisus:
### Description

- Changes the `PlaywrightCrawler` from using the standard [browser context](https://playwright.dev/python/docs/api/class-browser#browser-new-context) to using a [persistent browser context](https://playwright.dev/python/docs/api/class-browsertype#browser-type-launch-persistent-context).
- Allows passing a `user_data_dir` with the path to the directory for the context. If `user_data_dir` is not provided, a temporary directory will be created.

### Issues

- Closes: #963
- Closes: #721


vdusek (changes requested):
Since it contains a breaking change, could you please describe the breaking changes in the PR's description and also summarize it in the [Upgrading guide](https://github.com/apify/crawlee-python/blob/master/docs/upgrading/upgrading_to_v0x.md)?

src/crawlee/browsers/_playwright_browser.py
```diff
@@ -0,0 +1,68 @@
...
+        launch_options = self._browser_launch_options | context_options
+
+        if self._user_data_dir:
+            user_data_dir = self._user_data_dir
+            temp_dir = False
+        else:
+            user_data_dir = tempfile.mkdtemp(prefix='apify-playwright-firefox-taac-')
+            temp_dir = True
+
+        self._context = await self._browser_type.launch_persistent_context(
```
Pijukatel, line 54: I am not really sure how this is going to work with [Fingerprints](https://github.com/apify/crawlee-python/blob/master/src/crawlee/browsers/_playwright_browser_controller.py#L211C17-L211C39)

Since in order to have fingerprints we inject them through `browserforge` and that is already wrapping [browser.new_context](https://github.com/daijro/browserforge/blob/main/browserforge/injectors/playwright/injector.py#L29) call.

Seems to me that this wrapper replaces that wrapper and no fingerprints will be injected.

Maybe I am wrong, but I would like to see at least one test that checks that both `fingerprints` and `persistent context` works together

Pijukatel (changes requested):

src/crawlee/browsers/_playwright_browser.py
Pijukatel: Or to be more specific, using Fingerprints will disable the use of persistent context. I believe that the implementation should be done in a way that these two options are not mutually exclusive.

src/crawlee/browsers/_browser_pool.py
```diff
@@ -101,6 +102,7 @@ def with_default_plugin(
         cls,
         *,
         browser_type: BrowserType | None = None,
+        user_data_dir: str | Path | None = None,
```
Pijukatel, line 105: I think there should be one test that will test usage of the `user_data_dir`. Such test should demonstrate use case for using this argument.

src/crawlee/browsers/_playwright_browser.py
Mantisus: Great point, I'll add the tests.

But it works. It calls `new_context` of the browser that was passed into it, this provides support for different browser types.  We pass it our browser wraper for which `new_context` will be called.

Pijukatel: My mistake, you are correct. I just run it in debug and indeed it seems to work.

Mantisus: But adding tests, that's always a great idea)

vdusek (commented):
There are failing tests for Windows Python 3.12.

Mantisus:
> There are failing tests for Windows Python 3.12.

The problem is not related to this PR.

It's solved in the PR #1007

tests/unit/crawlers/_playwright/test_playwright_crawler.py
```diff
@@ -30,12 +33,23 @@
...
+    from collections.abc import Generator
+
     from yarl import URL
 
     from crawlee._request import RequestOptions
     from crawlee.crawlers import PlaywrightCrawlingContext, PlaywrightPreNavCrawlingContext
 
 
+@pytest.fixture
+def user_folder() -> Generator[Path]:
```
Pijukatel, line 45: Maybe you could use existing [tmp_path fixture](https://docs.pytest.org/en/4.6.x/tmpdir.html#the-tmp-path-fixture) ?

tests/unit/crawlers/_playwright/test_playwright_crawler.py
```diff
@@ -438,3 +452,47 @@ async def some_hook(context: PlaywrightPreNavCrawlingContext) -> None:
...
+        fingerprints['window.navigator.userAgent'] = await context.page.evaluate('()=>window.navigator.userAgent')
+        await context.page.route('**/*', lambda route: route.fulfill(status=200))
+
+    assert not check_path.exists()
+
+    await crawler.run(['https://test.io'])
+
+    assert check_path.exists()
+
+    assert 'headless' not in fingerprints['window.navigator.userAgent'].lower()
```
Pijukatel, line 498: Super tiny detail. I would also suggest to `assert fingerprints['window.navigator.userAgent']`  (because `'headless' not in "" == True`)

Pijukatel (commented):
Thanks for the tests, just two minor comments.

tests/unit/crawlers/_playwright/test_playwright_crawler.py
Mantisus: I didn't know there was such a thing. Thanks!

Pijukatel (approved):

src/crawlee/browsers/_playwright_browser.py
```diff
@@ -0,0 +1,68 @@
...
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+    from playwright.async_api import BrowserContext, BrowserType
+
+logger = getLogger(__name__)
+
+
+class PlaywrightPersistentBrowser:
```
vdusek, line 17: Taking into account we have only a single browser class, I would rather stay with the name `PlaywrightBrowser`. Or do we plan to support a non-persistent context browser in the future as well?

src/crawlee/browsers/_playwright_browser_controller.py
```diff
@@ -210,7 +209,9 @@ async def _create_browser_context(
 
         if self._fingerprint_generator:
             return await AsyncNewContext(
-                browser=self._browser, fingerprint=self._fingerprint_generator.generate(), **browser_new_context_options
+                browser=cast(Browser, self._browser),
```
vdusek, line 212: I don't like this cast, isn't there a better way?

vdusek (changes requested):
Looks good, just a few comments.

src/crawlee/browsers/_playwright_browser.py
```diff
@@ -0,0 +1,68 @@
...
+        if self._context:
+            raise RuntimeError('Persistent browser can have only one context')
+
+        launch_options = self._browser_launch_options | context_options
+
+        if self._user_data_dir:
+            user_data_dir = self._user_data_dir
+            temp_dir = False
+        else:
+            user_data_dir = tempfile.mkdtemp(prefix='apify-playwright-firefox-taac-')
```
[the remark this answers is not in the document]
Mantisus: I used the prefix from [TS Crawlee](https://github.com/apify/crawlee/blob/master/packages/browser-pool/src/playwright/playwright-plugin.ts#L113)

src/crawlee/browsers/_playwright_browser.py
Mantisus: I suppose that's a possibility in the future.

src/crawlee/browsers/_playwright_browser.py
```diff
@@ -0,0 +1,68 @@
...
+
+        self._context: BrowserContext | None = None
+        self._is_connected = True
+
+    @property
+    def browser_type(self) -> BrowserType:
+        return self._browser_type
+
+    def is_connected(self) -> bool:
+        return self._is_connected
```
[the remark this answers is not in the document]
vdusek: In this case, could you please mention in the class docstring, that this should have the same interface as Playwright's [Browser](https://playwright.dev/python/docs/api/class-browser)?

src/crawlee/browsers/_playwright_browser.py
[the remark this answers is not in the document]
vdusek: Ah, okay. Could you please make it a module-private constant then?

vdusek (approved):
LGTM

PR #986, opened 2025-02-15
chore(deps): update curl-impersonate to 0.9

Mantisus:
### Description

- update curl-cffi version requirement to >=0.9.0.
- update default `impersonate` from `chrome124` to `chrome131`


Mantisus:
This is the first stable version after chrome131 impersonation support was added. It has a significantly different fingerprint compared to chrome124.
Because of this, we should use chrome131 as the default impersonation option, since it matches the current version of Chrome.

vdusek (approved):
Thanks.

cc @janbuchar Will `chore(deps)!` commit type be rendered in the changelog? Just to be sure.

Issue #86, opened 2024-04-03
Better approach of making a cache


vdusek:
- This is a follow-up issue to the discussion in https://github.com/apify/crawlee-py/pull/82#discussion_r1548009445.
- Currently, we have our own implementation of LRU cache in `crawlee/_utils/lru_cache.py`.
- Let's do it in a more Pythonic way, maybe utilizing the built-in caching from `functools` std module (`lru_cache` decorator)?

Azathoth-X:
Was reading on this and wanted to clear up somethings.
The lru_cache from functools has no delete function for a given key, it just clears whole cache.
Can you define what do you mean by pythonic?
The cache implementation looks good as it.

29deepanshutyagi:
i want to work on this issue ,kindly assign me @B4nan , if it's still opened

janbuchar:
> i want to work on this issue ,kindly assign me @B4nan , if it's still opened

We don't assign issues for hacktoberfest. If you want to work on this, open a PR. First mergeable one gets merged.

Azathoth-X:
Hi @vdusek ,
Just wanted to ask do we need the delete individual type saved in cache function, in this class. I could only trace references of get, clear and creating key value pair for LRUcache class. 
Only place where the delete function is used is unit tests.
```python
def test_del(lru_cache: LRUCache[int]) -> None:
    # Key error on non-existent key
    with pytest.raises(KeyError):
        del lru_cache['non-existent-key']
    # No error with existing key
    len_before_del = len(lru_cache)
    del lru_cache['a']
    assert len(lru_cache) == len_before_del - 1
    assert 'a' not in lru_cache
```

vdusek:
Hi @Azathoth-X, if the delete is used only in the tests, I believe we do not need it.

PR #988, opened 2025-02-16
chore(deps): change custom `LRUCache` to `cachetools.LRUCache`

Mantisus:
### Description

- change custom `LRUCache` to `cachetools.LRUCache`.  In my opinion, `functools.lru_cache's` logic isn't well-suited for this use case. Therefore, if we want to modify our caching approach, using `cachetools` appears to be a better option.

### Issues

- Closes: #86


pyproject.toml
```diff
@@ -72,6 +72,7 @@ tldextract = ">=5.1.0"
 typer = ">=0.12.0"
 typing-extensions = ">=4.1.0"
 yarl = ">=1.18.0"
+cachetools = "^5.5.1"
```
vdusek, line 75: Could you please use ">=" version specifier? And please put it in alphabetical order 🙂.

pyproject.toml
```diff
@@ -95,6 +96,7 @@ types-beautifulsoup4 = "~4.12.0.20240229"
 types-colorama = "~0.4.15.20240106"
 types-psutil = "~5.9.5.20240205"
 types-python-dateutil = "~2.9.0.20240316"
+types-cachetools = "^5.5.0.20240820"
```
vdusek, line 99: Could you please use "~=" version specifier in dev deps? And please put it in alphabetical order 🙂.

(We use `~=` for dev deps because we have a Renovate bot, which updates them to the next major/minor.)

vdusek (commented):
Nice! Do we have please dome mechanism how to test/ensure that the cache is being used?

Mantisus:
> Do we have please dome mechanism how to test/ensure that the cache is being used?

Now, we have 🙂.

vdusek (approved):
Thanks

PR #993, opened 2025-02-18
refactor: set storage_object on open of storages

danpoletaev:
### Description

Currently, when opening a KV store, dataset or request queue we only store the id and name from the get() response.

This PR updates the logic to store the entire storage_object, allowing access to additional attributes without requiring extra requests.

More context in [slack](https://apify.slack.com/archives/CD0SF6KD4/p1738936334553189)

P.S. Python Crawlee storage constructors are a bit different than JS storage consturctors, that's why I've added storage_object to all of the classes - KV store, dataset & request queue

[Similar PR in Crawlee JS](https://github.com/apify/crawlee/pull/2837)

### Testing

I've added one test, which checks that KV store class has storage_object after open.

### DIFF
--- a/src/crawlee/storages/_base.py
+++ b/src/crawlee/storages/_base.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING
 if TYPE_CHECKING:
     from crawlee.configuration import Configuration
     from crawlee.storage_clients._base import StorageClient
+    from crawlee.storage_clients.models import _StorageMetadata
 
 
 class Storage(ABC):
@@ -20,6 +21,11 @@ class Storage(ABC):
     @abstractmethod
     def name(self) -> str | None:
         """Get the storage name."""
+
+    @property
+    @abstractmethod
+    def storage_object(self) -> _StorageMetadata:
+        """Get the full storage info object."""
 
     @classmethod
     @abstractmethod
--- a/src/crawlee/storages/_creation_management.py
+++ b/src/crawlee/storages/_creation_management.py
@@ -169,7 +169,9 @@ async def open_storage(
             resource_collection_client = _get_resource_collection_client(storage_class, storage_client)
             storage_info = await resource_collection_client.get_or_create(name=name)
 
-        storage = storage_class(id=storage_info.id, name=storage_info.name, storage_client=storage_client)
+        storage = storage_class(
+            id=storage_info.id, name=storage_info.name, storage_client=storage_client, storage_object=storage_info
+        )
 
         # Cache the storage by ID and name
         _add_to_cache_by_id(storage.id, storage)
--- a/src/crawlee/storages/_dataset.py
+++ b/src/crawlee/storages/_dataset.py
@@ -12,7 +12,7 @@ from crawlee import service_locator
 from crawlee._utils.byte_size import ByteSize
 from crawlee._utils.docs import docs_group
 from crawlee._utils.file import json_dumps
-from crawlee.storage_clients.models import DatasetMetadata
+from crawlee.storage_clients.models import DatasetMetadata, _StorageMetadata
 
 from ._base import Storage
 from ._key_value_store import KeyValueStore
@@ -194,9 +194,12 @@ class Dataset(Storage):
     _EFFECTIVE_LIMIT_SIZE = _MAX_PAYLOAD_SIZE - (_MAX_PAYLOAD_SIZE * _SAFETY_BUFFER_PERCENT)
     """Calculated payload limit considering safety buffer."""
 
-    def __init__(self, id: str, name: str | None, storage_client: StorageClient) -> None:
+    def __init__(
+        self, id: str, name: str | None, storage_client: StorageClient, storage_object: _StorageMetadata
+    ) -> None:
         self._id = id
         self._name = name
+        self._storage_object = storage_object
 
         # Get resource clients from the storage client.
         self._resource_client = storage_client.dataset(self._id)
@@ -211,6 +214,11 @@ class Dataset(Storage):
     @override
     def name(self) -> str | None:
         return self._name
+
+    @property
+    @override
+    def storage_object(self) -> _StorageMetadata:
+        return self._storage_object
 
     @override
     @classmethod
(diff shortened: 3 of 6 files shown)


src/crawlee/storages/_base.py
```diff
@@ -21,6 +22,11 @@ def id(self) -> str:
     def name(self) -> str | None:
         """Get the storage name."""
 
+    @property
+    @abstractmethod
+    def storage_object(self) -> _StorageMetadata:
```
Pijukatel, line 27: This is public property, that returns private type. I guess we should expose `_StorageMetadata` if we use it as type hint for public property now.

src/crawlee/storages/_creation_management.py
```diff
@@ -169,7 +169,9 @@ async def open_storage(
             resource_collection_client = _get_resource_collection_client(storage_class, storage_client)
             storage_info = await resource_collection_client.get_or_create(name=name)
 
-        storage = storage_class(id=storage_info.id, name=storage_info.name, storage_client=storage_client)
+        storage = storage_class(
+            id=storage_info.id, name=storage_info.name, storage_client=storage_client, storage_object=storage_info
```
Pijukatel, line 173: It looks kind of strange as the arguments for the `__init__`  are now redundant as the storage_info contains first two args as well. Maybe it would be more suitable to create an alternate constructor. For example:

`storage=storage_class.from_storage_info(storage_client=storage_client, storage_object=storage_info)`

Pijukatel (commented):
One more comment to the whole change. Seems to me that same thing is refereed under different names in different places. It would be nice to unify it and use only one name variant. For example one of following `storage_object` or `storage_info` or maybe `storage_metadata` since it is instance of `_StorageMetadata`.

src/crawlee/storages/_dataset.py
```diff
@@ -194,17 +195,30 @@ class Dataset(Storage):
...
+    @classmethod
+    def from_storage_object(cls, storage_client: StorageClient, storage_object: StorageMetadata) -> Dataset:
+        """Create a new instance of Dataset from a storage metadata object."""
+        dataset = Dataset(
+            id=storage_object.id,
+            name=storage_object.name,
+            storage_client=storage_client,
+        )
+
+        dataset.set_storage_object(storage_object)
```
Pijukatel, line 219: When you add the setter through the decorator,then you can assign value like this:

`dataset.storage_object  = storage_object`

src/crawlee/storages/_dataset.py
```diff
@@ -217,9 +231,14 @@ def name(self) -> str | None:
...
     @property
     @override
-    def storage_object(self) -> _StorageMetadata:
+    def storage_object(self) -> StorageMetadata:
         return self._storage_object
 
+    @classmethod
+    @override
+    def set_storage_object(cls, storage_object: StorageMetadata) -> None:
+        cls._storage_object = storage_object
```
Pijukatel, line 240: As mentioned above, this should be instance method

src/crawlee/storages/_key_value_store.py
```diff
@@ -85,9 +99,14 @@ def name(self) -> str | None:
 
     @property
     @override
-    def storage_object(self) -> _StorageMetadata:
+    def storage_object(self) -> StorageMetadata:
         return self._storage_object
 
+    @classmethod
+    @override
+    def set_storage_object(cls, storage_object: StorageMetadata) -> None:
```
Pijukatel, line 107: As mentioned above, this should be instance method

src/crawlee/storages/_request_queue.py
```diff
@@ -152,9 +166,14 @@ def name(self) -> str | None:
 
     @property
     @override
-    def storage_object(self) -> _StorageMetadata:
+    def storage_object(self) -> StorageMetadata:
         return self._storage_object
 
+    @classmethod
+    @override
+    def set_storage_object(cls, storage_object: StorageMetadata) -> None:
```
Pijukatel, line 174: As mentioned above, this should be instance method

Pijukatel (commented):
Please add 3 simple test cases(one for each storage class) that will just call `from_storage_object` constructor and assert that the created object is as expected.

src/crawlee/storages/_base.py
```diff
@@ -21,6 +22,16 @@ def id(self) -> str:
     def name(self) -> str | None:
         """Get the storage name."""
 
+    @property
+    @abstractmethod
+    def storage_object(self) -> StorageMetadata:
```
vdusek, line 27: Wouldn't this be more precise?

```suggestion
    def storage_object(self) -> DatasetMetadata | KeyValueStoreMetadata | RequestQueueMetadata:
```

and `_StorageMetadata` could remain private.

I am also ok with having `StorageMetadata` as a type alias for the union mentioned above. And the `_StorageMetadata` could be renamed to `_CommonStorageMetadata`, `_BaseStorageMetadata`, ...

src/crawlee/storages/_dataset.py
```diff
@@ -197,11 +198,27 @@ class Dataset(Storage):
     def __init__(self, id: str, name: str | None, storage_client: StorageClient) -> None:
         self._id = id
         self._name = name
+        datetime_now = datetime.now(timezone.utc)
+        self._storage_object: StorageMetadata = StorageMetadata(
+            id=id, name=name, accessed_at=datetime_now, created_at=datetime_now, modified_at=datetime_now
+        )
```
vdusek, line 204: I believe type inference is enough here, isn't it?

```suggestion
        self._storage_object = StorageMetadata(
            id=id, name=name, accessed_at=datetime_now, created_at=datetime_now, modified_at=datetime_now
        )
```

vdusek (changes requested):

src/crawlee/storages/_base.py
danpoletaev: That's what I actually did in the beginning, but I was getting type error in _creation_managment.py

Screenshot 2025-03-03 at 14 52 34

Pijukatel (approved):

janbuchar (approved):
No issues here, just make sure that the new SDK works with this.

vdusek (approved):
On second thought, I'm OK with the common `StorageMetadata` model for every storage. LGTM.

PR #994, opened 2025-02-18
chore: migrate to uv package manager

vdusek:
- Migrate from `poetry` to `uv`.
- Relates: #628
- The update of templates to use `uv` will be implemented separately.


pyproject.toml
```diff
@@ -31,86 +36,81 @@ keywords = [
...
+
+[project.urls]
+homepage = "https://crawlee.dev/python"
+apify_homepage = "https://apify.com"
+changelog = "https://crawlee.dev/python/docs/changelog"
+documentation = "https://crawlee.dev/python/docs/"
+issue_tracker = "https://github.com/apify/crawlee-python/issues"
+repository = "https://github.com/apify/crawlee-python"
+
+#######################
```
Pijukatel, line 104: Same comment:
https://github.com/apify/apify-sdk-python/pull/406#discussion_r1961109252

Pijukatel (approved):

pyproject.toml
vdusek: yep, but the tools can have multiple such sections (see Ruff), however, if we don't have a consensus here, I'll remove it

vdusek:
I tested everything I could from the feature branch. Gonna merge it with `workflows@switch-to-uv` and see how it behaves from master. The releases (PyPI & docs) will be failing because of limit permissions only to master branch.

Issue #929, opened 2025-01-22
Prepare universal http interceptor for both static and browser crawlers tests


Pijukatel:
Currently in our tests sometimes respx is used to mock http traffic for static crawlers and for PlaywrightCrawler mostly real requests are done.
It would be convenient, faster and more robust to create a fixture that can mock both.
For Playwright related browser requests it can be done using custom browser **BrowserPool**,  **page.route** for example:

```python
class _StaticRedirectBrowserPool(BrowserPool):
    """BrowserPool for redirecting browser requests to static content."""

    async def new_page(
        self,
        *,
        page_id: str | None = None,
        browser_plugin: BaseBrowserPlugin | None = None,
        proxy_info: ProxyInfo | None = None,
    ) -> CrawleePage:
        crawlee_page = await super().new_page(page_id=page_id, browser_plugin=browser_plugin, proxy_info=proxy_info)
        await crawlee_page.page.route(
            '**/*',
            lambda route: route.fulfill(
                status=200, content_type='text/plain', body='<!DOCTYPE html><html><body>What a body!</body></html>'
            ),
        )
        return crawlee_page
```

Mantisus:
It also solves this issue #197

PR #1007, opened 2025-02-21
chore: fix tests and improve tests stability

Mantisus:
### Description

- Fix `kvs` isolation in the `use_state` test for `AdaptivePlaywrightCrawler` 
- Separate isolation cookie tests for `httpx` and `curl`. Having `httpx` use `respx` instead of `httpbin` will increase the stability of the test.
- Update some `PlaywrightCrawler` tests to use `route` instead of `httpbin`. This will speed up execution time and improve stability. It will also probably be useful for #929 in future

### DIFF
--- a/tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
+++ b/tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
@@ -32,15 +32,17 @@ from crawlee.crawlers._adaptive_playwright._adaptive_playwright_crawling_context
     AdaptiveContextError,
 )
 from crawlee.statistics import Statistics
+from crawlee.storages import KeyValueStore
 
 if TYPE_CHECKING:
-    from collections.abc import Iterator
+    from collections.abc import AsyncGenerator, Iterator
 
     import respx
 
     from crawlee.browsers._browser_plugin import BrowserPlugin
     from crawlee.browsers._types import CrawleePage
     from crawlee.proxy_configuration import ProxyInfo
+
 
 _H1_TEXT = 'Static'
 _H2_TEXT = 'Only in browser'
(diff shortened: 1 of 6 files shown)


tests/unit/crawlers/_playwright/test_playwright_crawler.py
```diff
@@ -146,56 +148,70 @@ async def request_handler(context: PlaywrightCrawlingContext) -> None:
...
-async def test_chromium_headless_headers(httpbin: URL) -> None:
+async def test_chromium_headless_headers() -> None:
     crawler = PlaywrightCrawler(headless=True, browser_type='chromium')
     headers = dict[str, str]()
 
+    @crawler.pre_navigation_hook
+    async def some_hook(context: PlaywrightPreNavCrawlingContext) -> None:
+        # Emulate server response to prevent Playwright from making real requests
+        await context.page.route(
+            '**/*', lambda route: route.fulfill(status=200, body=json.dumps(route.request.headers))
```
Pijukatel, line 159: Nice. It was really confusing to have those capitalized Sec headers in asserts.

Pijukatel (approved):

tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
```diff
@@ -74,6 +76,13 @@ def test_urls(respx_mock: respx.MockRouter) -> list[str]:
     return urls
 
 
+@pytest.fixture
+async def key_value_store() -> AsyncGenerator[KeyValueStore, None]:
+    kvs = await KeyValueStore.open()
+    yield kvs
+    await kvs.drop()
```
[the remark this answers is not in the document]
Mantisus: That's right.

I still don't understand why it failed only on python 3.12 and for windows. Perhaps the tests were run in a different order.

tests/unit/crawlers/_playwright/test_playwright_crawler.py
Mantisus: Yes, `httpbin` transforms headers into Title format for some reason. 

Maybe for better readability or to make them look like in browser devtools....

tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
vdusek, line 83: Just wondering whether to put this into `conftest` and utilize this fixture across different test modules?

vdusek (approved):
Nice, thanks!

Issue #1010, opened 2025-02-21
Running CLI via `pipx` or `uvx` is not possible anymore


janbuchar:
```
$ pipx run 'crawlee[all]==0.6.0b10' create
'crawlee' executable script not found in package 'crawlee'.
```

```
$ uvx 'crawlee==0.6.0b10' create
The executable `crawlee` was not found.
warning: Package `crawlee` does not provide any executables.
```

I can make an educated guess that this is after the `uv` migration - https://github.com/apify/crawlee-python/pull/994/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L112-L113.

Ideally, the CLI should work with both `pipx` and `uvx`.

vdusek:
Scripts configuration:

```
[tool.poetry.scripts]
crawlee = "crawlee._cli:cli"
```

from https://github.com/apify/crawlee-python/blob/v0.5.4/pyproject.toml#L109:L110

is missing in the new setup.

janbuchar:
@vdusek #1011 fixes that

Issue #703, opened 2024-11-15
Make CLI libraries optional dependencies


vdusek:
Since `cookiecutter`, along with other libraries like `inquirer` and `typer`, is used exclusively in the CLI for templating new projects, could we make these optional dependencies? This would reduce the installation footprint.

vdusek:
The mentioned packages & their (transitive) dependencies.

### Cookiecutter

```
cookiecutter 2.6.0 A command-line utility that creates projects from project templates, e.g. creating a Python package project from a Python package project template.
├── arrow *
│   ├── python-dateutil >=2.7.0 
│   │   └── six >=1.5 
│   └── types-python-dateutil >=2.8.10 
├── binaryornot >=0.4.4
│   └── chardet >=3.0.2 
├── click >=7.0,<9.0.0
│   └── colorama * 
├── jinja2 >=2.7,<4.0.0
│   └── markupsafe >=2.0 
├── python-slugify >=4.0.0
│   └── text-unidecode >=1.3 
├── pyyaml >=5.3.1
├── requests >=2.23.0
│   ├── certifi >=2017.4.17 
│   ├── charset-normalizer >=2,<4 
│   ├── idna >=2.5,<4 
│   └── urllib3 >=1.21.1,<3 
└── rich *
    ├── markdown-it-py >=2.2.0 
    │   └── mdurl >=0.1,<1.0 
    ├── pygments >=2.13.0,<3.0.0 
    └── typing-extensions >=4.0.0,<5.0
```

### Inquirer

```
inquirer 3.4.0 Collection of common interactive command line user interfaces, based on Inquirer.js
├── blessed >=1.19.0
│   ├── jinxed >=1.1.0 
│   │   └── ansicon * 
│   ├── six >=1.9.0 
│   └── wcwidth >=0.1.4 
├── editor >=1.6.0
│   ├── runs * 
│   │   └── xmod * 
│   └── xmod * (circular dependency aborted here)
└── readchar >=4.2.0
```

### Typer

```
typer 0.13.0 Typer, build great CLIs. Easy to code. Based on Python type hints.
├── click >=8.0.0
│   └── colorama * 
├── rich >=10.11.0
│   ├── markdown-it-py >=2.2.0 
│   │   └── mdurl >=0.1,<1.0 
│   ├── pygments >=2.13.0,<3.0.0 
│   └── typing-extensions >=4.0.0,<5.0 
├── shellingham >=1.3.0
└── typing-extensions >=3.7.4.3
[...]

PR #1011, opened 2025-02-23
refactor!: move `cli` dependencies to optional dependencies

Mantisus:
### Description

- Restore entrypoint `crawlee` for `cli`
- Move cli dependencies to optional dependencies

### Issues

- Closes: #1010
- Closes: #703

### Testing

- Tested on a locally created wheel and local index

for pipx
```bash
pipx run --pip-args="--index-url http://localhost:8080" crawlee[cli] create
```

For uvx.
```bash
uvx --index-url http://localhost:8080 crawlee[cli] create
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ pipx --help
 Then, run the CLI and choose from the available templates:
 
 ```sh
-pipx run crawlee create my-crawler
+pipx run crawlee[cli] create my-crawler
 ```
 
 If you already have `crawlee` installed, you can spin it up by running:
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ pipx --help
 Then, run the CLI and choose from the available templates:
 
 ```sh
-pipx run crawlee[cli] create my-crawler
+pipx run 'crawlee[cli]' create my-crawler
 ```
 
 If you already have `crawlee` installed, you can spin it up by running:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,6 +78,9 @@ curl-impersonate = ["curl-cffi>=0.9.0"]
 parsel = ["parsel>=1.10.0"]
 playwright = ["browserforge>=1.2.3", "playwright>=1.27.0"]
 
+[project.scripts]
+crawlee = "crawlee._cli:cli"
+
 [project.urls]
 "Homepage" = "https://crawlee.dev/python"
 "Apify homepage" = "https://apify.com"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,11 +35,9 @@ keywords = [
 dependencies = [
     "cachetools>=5.5.0",
     "colorama>=0.4.0",
-    "cookiecutter>=2.6.0",
     "docutils>=0.21.0",
     "eval-type-backport>=0.2.0",
     "httpx[brotli,http2,zstd]>=0.27.0",
-    "inquirer>=3.3.0",
     "more-itertools>=10.2.0",
     "psutil>=6.0.0",
     "pydantic-settings>=2.2.0,<2.7.0",
@@ -48,7 +46,6 @@ dependencies = [
     "rich>=13.9.0",
     "sortedcollections>=2.1.0",
     "tldextract>=5.1.0",
-    "typer>=0.12.0",
     "typing-extensions>=4.1.0",
     "yarl>=1.18.0",
 ]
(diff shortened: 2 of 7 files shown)


Mantisus:
After https://github.com/astral-sh/uv/pull/11337, `uvx` correctly supports optional dependencies without using `--with`.

janbuchar:
Nice! Couple of points for discussion:

1. square brackets are special characters for the `zsh` shell, which is quite a popular one - we should either change the commands to `uvx 'crawlee[cli]' create` (add single quotes) or add a warning that you need to do this if you use `zsh`
2. if you call just plain `uvx crawlee create`, it will crash because of missing dependencies, right? we should print a user-friendly warning in that case, something like "looks like you're running just `crawlee`, try using `crawlee[cli]`"
3. this should be mentioned more visibly in the changelog - can we change the title? Or we could put it into the upgrading guide, but it seems kinda strange to mention changes to project bootstrapping there

pyproject.toml
```diff
@@ -77,6 +74,10 @@ beautifulsoup = ["beautifulsoup4[lxml]>=4.12.0", "html5lib>=1.0"]
 curl-impersonate = ["curl-cffi>=0.9.0"]
 parsel = ["parsel>=1.10.0"]
 playwright = ["browserforge>=1.2.3", "playwright>=1.27.0"]
+cli = ["cookiecutter>=2.6.0", "inquirer>=3.3.0", "typer>=0.12.0"]
```
vdusek, line 77: Could you please alphabetize the list and include these deps into `all` extra?

README.md
```diff
@@ -71,7 +71,7 @@ pipx --help
 Then, run the CLI and choose from the available templates:
 
 ```sh
-pipx run crawlee create my-crawler
+pipx run crawlee[cli] create my-crawler
```
vdusek, line 74: please add single quotes, because of zsh

```suggestion
pipx run 'crawlee[cli]' create my-crawler
```

website/src/pages/index.js
```diff
@@ -56,7 +56,7 @@ function Hero() {
                     </div>
                     <div className={styles.codeBlock}>
                         <CodeBlock className="language-bash">
-                            pipx run crawlee create my-crawler
+                            pipx run crawlee[cli] create my-crawler
```
vdusek, line 59: please add single quotes, because of zsh

```suggestion
                            pipx run 'crawlee[cli]' create my-crawler
```

website/src/pages/index.js
```diff
@@ -116,7 +116,7 @@ function ActorExample() {
                     The fastest way to try Crawlee out is to use the <b>Crawlee CLI</b> and choose one of the provided templates. The CLI will prepare a new project for you, and add boilerplate code for you to play with.
                 </p>
                 <CodeBlock className="language-bash">
-                    pipx run crawlee create my-crawler
+                    pipx run crawlee[cli] create my-crawler
```
vdusek, line 119: please add single quotes, because of zsh

```suggestion
                    pipx run 'crawlee[cli]' create my-crawler
```

vdusek (commented):
Merge this just before the release of v0.6, as we do not have the doc versioning yet.

src/crawlee/_cli.py
```diff
@@ -6,11 +6,14 @@
...
-from inquirer.render.console import ConsoleRender  # type: ignore[import-untyped]
-from rich.progress import Progress, SpinnerColumn, TextColumn
+try:
+    import inquirer  # type: ignore[import-untyped]
+    import typer
+    from cookiecutter.main import cookiecutter  # type: ignore[import-untyped]
+    from inquirer.render.console import ConsoleRender  # type: ignore[import-untyped]
+    from rich.progress import Progress, SpinnerColumn, TextColumn
+except ModuleNotFoundError as exc:
+    raise UserWarning("Looks like you're running just 'crawlee', try using 'crawlee[cli]'") from exc
```
janbuchar, line 16: Um, does this work differently than raising a `RuntimeError` for instance?

Mantisus: No, it's the same thing.

It just seems to me that using `UserWarning` + show that it's an error imports through with `from exc`.

Would be more transparent to the user.

janbuchar: Hm, I think that `UserWarning` is meant to be used with the `warnings` module, but at the same time, this is pretty innocent, so feel free to keep it this way.

Mantisus: I did some reflection on this and replaced it with `ImportError` :smile:.

src/crawlee/_cli.py
```diff
@@ -6,11 +6,14 @@
...
-from inquirer.render.console import ConsoleRender  # type: ignore[import-untyped]
-from rich.progress import Progress, SpinnerColumn, TextColumn
+try:
+    import inquirer  # type: ignore[import-untyped]
+    import typer
+    from cookiecutter.main import cookiecutter  # type: ignore[import-untyped]
+    from inquirer.render.console import ConsoleRender  # type: ignore[import-untyped]
+    from rich.progress import Progress, SpinnerColumn, TextColumn
+except ModuleNotFoundError as exc:
+    raise ImportError("Looks like you're running just 'crawlee', try using 'crawlee[cli]'") from exc
```
vdusek, line 16: Could we use a bit more user-friendly message? Smt like this:

> Missing required dependencies for the Crawlee CLI. It looks like you're running 'crawlee' without the CLI extras. Try using 'crawlee[cli]' instead.

Issue #1013, opened 2025-02-24
HttpxHttpClient always follows redirects


2tunnels:
I've encountered an issue with the redirection handling logic in `HttpxHttpClient`. Currently, there's no straightforward way to disable automatic redirect following when initializing `HttpxHttpClient`. Ideally, it should be possible to configure it like this:

```python
crawler = HttpCrawler(http_client=HttpxHttpClient(follow_redirects=False))
```

While the `follow_redirects` argument is correctly passed to the underlying `AsyncClient`, `HttpxHttpClient` overrides this behavior in the `crawl` method by explicitly setting `follow_redirects=True`:

https://github.com/apify/crawlee-python/blob/179ec93e8802fb230b1c5b635ab559d4ef7c6dd8/src/crawlee/http_clients/_httpx.py#L168

I assume this was done because `AsyncClient` defaults to `follow_redirects=False`, ensuring that HTTP-based crawlers follow redirects by default. However, would it be preferable to handle this differently? For example adding a default `follow_redirects` argument in `kwargs` in `_get_client` method?

PR #1015, opened 2025-02-24
fix: Remove follow_redirects override in HttpxHttpClient

2tunnels:
### Description

Allow users to disable redirect following for HttpxHttpClient.

### Issues

- Closes: #1013 

### Testing

After removing `follow_redirects=True` a couple of tests actually failed, updating default `kwargs` fixed them.

### Checklist

- [ ] CI passed


Pijukatel (commented):
Hi, thanks for the PR. Could you please add two tests that will set both possible values of `follow_redirects` in `async_session_kwargs` for `HttpxHttpClient`. For the test case maybe you can use some [redirects endpoint](https://httpbin.org/#/Redirects) of the httpbin that we use[ in the existing tests](https://github.com/apify/crawlee-python/blob/master/tests/unit/http_clients/test_httpx.py).

Response for the same endpoint should depend on the value of this option. So you can do assert on that.

2tunnels:
@Pijukatel added tests, but I'm not totally sure about asserts and if status + location header check is enough. `HttpCrawlingResult` and `HttpResponse` classes do not expose a lot of attributes and not the best candidates for introspection. Maybe you have better ideas?

Mantisus:
Hey @2tunnels, Thank you for this PR

> Maybe you have better ideas?

This will sound a bit unexpected, but you can use `request.loaded_url`.

if the redirect works:
```python
assert request.loaded_url == final_url
```

2tunnels:
@Mantisus yep, that works. Thank you 😄 Tests updated.

Pijukatel (approved):
LGTM. Thanks.

Issue #1016, opened 2025-02-24
CurlImpersonateHttpClient always follows redirects


2tunnels:
I've encountered an issue with the redirection handling logic in `CurlImpersonateHttpClient`. Currently, there's no straightforward way to disable automatic redirect following when initializing `CurlImpersonateHttpClient`. Ideally, it should be possible to configure it like this:

```python
crawler = HttpCrawler(http_client=CurlImpersonateHttpClient(allow_redirects=False))
```

While the `allow_redirects` argument is correctly passed to the underlying `AsyncSession`, `CurlImpersonateHttpClient` overrides this behavior in the `crawl` method by explicitly setting `allow_redirects=True`:

https://github.com/apify/crawlee-python/blob/179ec93e8802fb230b1c5b635ab559d4ef7c6dd8/src/crawlee/http_clients/_curl_impersonate.py#L151

Actaully, this is a default value for `AsyncSession`, so there is no need to force it. I guess this is somehow connected to `HttpxHttpClient` and it's default behaviour. Related issue: #1013

PR #1017, opened 2025-02-24
fix: Remove allow_redirects override in CurlImpersonateHttpClient

2tunnels:
### Description

Allow users to disable redirect following for CurlImpersonateHttpClient.

### Issues

- Closes: #1016 

### Testing

Not sure how to properly tests this. This is a default behaviour for underlying `AsyncSession`.

### Checklist

- [ ] CI passed


Pijukatel (commented):
Hi, thanks for the PR. Could you please add two tests that will set both possible values of `allow_redirects` in `async_session_kwargs` for `CurlImpersonateHttpClient`. For the test case maybe you can use some [redirects endpoint](https://httpbin.org/#/Redirects) of the httpbin that we use[ in the existing tests](https://github.com/apify/crawlee-python/blob/master/tests/unit/http_clients/test_curl_impersonate.py).

Response for the same endpoint should depend on the value of this option. So you can do assert on that.

Mantisus:
Hey @2tunnels.

Could you add `allow_redirects=True` to `kwargs` inside `_get_client`. To keep the default behavior unchanged.

Mantisus:
> headers.items() call is missing, so I've added it in this PR.

oh, thanks for catching that.

Missed it. :worried:

Pijukatel (approved):
LGTM. Thanks.

Issue #991, opened 2025-02-17
Default value of `persist_state_kvs_name` in `Statistics` and `SessionPool` causes the use of a named storage on Apify


janbuchar:
The default value of the argument is `"default"`, but it is used like this: `KeyValueStore.open(name=persist_state_kvs_name)`, which opens a named storage called "default" when running on Apify.

### Possible solutions
1. accept a pre-opened `KeyValueStore` instance instead
2. accept a type that distinguishes names/IDs/nothing - kinda verbose in Python

Note - after #809, this change will be easier to do consistently.

PR #1018, opened 2025-02-24
fix: Fix default migration storage

Pijukatel:
### Description
Fix the "default" storage used for the migration.
It should not be named "default", it should have `None` name to create default storage.

### Issues

- Closes: #991


vdusek (approved):
LGTM

janbuchar (approved):
That was easier than I thought, cool! Thanks!

PR #1019, opened 2025-02-24
refactor!: Replace enum with literal for EnqueueStrategy

vdusek:
So that there is no need for additional import when specifying the enqueue strategy.


src/crawlee/_types.py
```diff
@@ -46,6 +46,20 @@
...
+Strategies:
+    all: Enqueue every link encountered, regardless of the target domain. Use this option to ensure that all links,
+        including those leading to external websites, are followed.
+    same-domain: Enqueue links that share the same domain name as the current page, including any subdomains. This
+        strategy is ideal for crawling within the same top-level domain while still allowing for subdomain exploration.
+    same-hostname: Enqueue links only if they match the exact hostname of the current page. This is the default
+        behavior and restricts the crawl to the current hostname, excluding subdomains.
+    same-origin: Enqueue links that share the same origin as the current page. The origin is defined by the
+        combination of protocol, domain, and port, ensuring a strict scope for the crawl.
+"""
```
janbuchar, line 61: Inconsistent dashes and underscores :slightly_smiling_face:

src/crawlee/_types.py
```diff
@@ -46,6 +46,20 @@
 
 RequestTransformAction: TypeAlias = Literal['skip', 'unchanged']
 
+EnqueueStrategy: TypeAlias = Literal['all', 'same_domain', 'same_hostname', 'same_origin']
+"""Specify the strategy for determining which links to extract and enqueue.
+
+Strategies:
```
janbuchar, line 52: I discussed this with @barjin and it seems that we won't be able to display this docs anywhere in the API reference, which is a disadvantage of enums (those get their own generated pages). It's probably still worth it, I just want to make sure we make an informed decision.

docs/upgrading/upgrading_to_v0x.md
```diff
@@ -24,6 +24,10 @@ We decided to move away from [Hungarian notation](https://en.wikipedia.org/wiki/
 - `BaseBrowserController` -> `BrowserController`
 - `BaseBrowserPlugin` -> `BrowserPlugin`
 
+### EnqueueStrategy
+
+The `EnqueueStrategy` has been changed from an enum to a string literal type. All its values and their meaning remain unchanged.
```
janbuchar, line 29: Did you consider making this change backwards compatible and accepting both the enum and the plain string?

src/crawlee/_types.py
```diff
@@ -46,6 +46,20 @@
 
 RequestTransformAction: TypeAlias = Literal['skip', 'unchanged']
 
+EnqueueStrategy: TypeAlias = Literal['all', 'same_domain', 'same_hostname', 'same_origin']
```
janbuchar, line 49: This could break if you run the new version with storage populated by the previous version - the `enqueue_strategy` is stored in the request queue and the old version has dashes instead of underscores.

janbuchar (commented):
The code seems well written - just a couple of comments to consider.

docs/upgrading/upgrading_to_v0x.md
vdusek: Is it really necessary, considering we're still 0.X?

src/crawlee/_types.py
vdusek: Ok, we can explain it in the enqueue links function docstring then.

src/crawlee/_types.py
janbuchar: I don't have a strong opinion here, but it's probably safer

src/crawlee/_types.py
janbuchar: That works for me. The only drawback I see is that if we added a new `EnqueueStrategy` (unlikely), we would have to document it in the function docstring too, which is easy to miss.

docs/upgrading/upgrading_to_v0x.md
vdusek: unpleasant but tolerable, especially when they are well documented ...

so you would rather prefer to keep both enum & literal types?

src/crawlee/_types.py
vdusek: updated

src/crawlee/_types.py
vdusek: getting back to dashes then

docs/upgrading/upgrading_to_v0x.md
janbuchar: I don't know. The advantage is avoiding a BC break, disadvantages are a longer type (more confusing documentation) and more code to handle it. Am I missing anything?

vdusek: You're right. The disadvantage is a BC break, that's clear from the beginning, and I accepted it going in. Having both options is IMO the worst case. I'd rather stick with enum than have both. This PR should be about correcting past mistakes, but not by creating a new mess.

janbuchar: I'm not sure if I'd call using an enum a mistake :slightly_smiling_face: but this is more consistent and probably more friendly to the end user, so let's do it and never look back.

janbuchar (approved):

Issue #1021, opened 2025-02-24
Wrong type of crawler shown when using crawler/context logging


vdusek:
### Reproduction

```python
import asyncio

from crawlee.crawlers import ParselCrawler, ParselCrawlingContext


async def main() -> None:
    crawler = ParselCrawler()
    crawler.log.info('Hello from crawler')

    @crawler.router.default_handler
    async def request_handler(context: ParselCrawlingContext) -> None:
        context.log.info('Hello from handler')

    await crawler.run(['https://crawlee.dev'])


if __name__ == '__main__':
    asyncio.run(main())
```

Output:
```
$ python crawl.py 
[crawlee.crawlers._abstract_http._abstract_http_crawler] INFO  Hello from crawler
[crawlee.crawlers._abstract_http._abstract_http_crawler] INFO  Current request statistics:
┌───────────────────────────────┬──────────┐
│ requests_finished             │ 0        │
│ requests_failed               │ 0        │
│ retry_histogram               │ [0]      │
│ request_avg_failed_duration   │ None     │
│ request_avg_finished_duration │ None     │
│ requests_finished_per_minute  │ 0        │
│ requests_failed_per_minute    │ 0        │
│ request_total_duration        │ 0.0      │
│ requests_total                │ 0        │
│ crawler_runtime               │ 0.082711 │
└───────────────────────────────┴──────────┘
[crawlee._autoscaling.autoscaled_pool] INFO  current_concurrency = 0; desired_concurrency = 2; cpu = 0; mem = 0; event_loop = 0.0; client_info = 0.0
[crawlee.crawlers._abstract_http._abstract_http_crawler] INFO  Hello from handler
[crawlee._autoscaling.autoscaled_pool] INFO  Waiting for remaining tasks to finish
[crawlee.crawlers._abstract_http._abstract_http_crawler] INFO  Final request statistics:
┌───────────────────────────────┬──────────┐
│ requests_finished             │ 1        │
│ requests_failed               │ 0        │
│ retry_histogram               │ [1]      │
│ request_avg_failed_duration   │ None     │
│ request_avg_finished_duration │ 0.598955 │
│ requests_finished_per_minute  │ 87       │
│ requests_failed_per_minute    │ 0        │
│ request_total_duration        │ 0.598955 │
│ requests_total                │ 1        │
│ crawler_runtime               │ 0.693418 │
└───────────────────────────────┴──────────┘
```

### Expected behaviour

[...]

PR #1023, opened 2025-02-25
fix: Fix logger name for http based loggers

Pijukatel:
### Description
Set logger name for http based loggers to class name.

### Issues

- Closes: #1021


vdusek (changes requested):
Could we use `"Crawler".__name__` in the tests rather than their names in the form as string literals? Thanks.

janbuchar:
> Could we use `"Crawler".__name__` in the tests rather than their names in the form as string literals? Thanks.

That borders on circular logic though :slightly_smiling_face: I slightly prefer copying the name in the tests so that they don't depend on the implementation, even though it's a super trivial dependency here.

vdusek (approved):

PR #1024, opened 2025-02-25
feat: Add `requests` argument to `EnqueueLinksFunction`

Pijukatel:
### Description
Add `requests` argument to `EnqueueLinksFunction`.
Split `EnqueueLinksFunction` implementations to `extract_links` and `add_requests`.
Add overload variants of `EnqueueLinksFunction`.
Raise error in `EnqueueLinksFunction` implementations if called with mutually exclusive arguments.

Relates to : https://github.com/apify/crawlee-python/issues/906


src/crawlee/_types.py
```diff
@@ -330,8 +330,9 @@ def __call__(
 
 @docs_group('Functions')
 class EnqueueLinksFunction(Protocol):
-    """A function for enqueueing new URLs to crawl based on elements selected by a given selector.
+    """A function for enqueueing new URLs to crawl based on elements selected by a given selector and explicit requests.
 
+    It adds explicitly passed requests to the `RequestManager`.
```
janbuchar, line 335: I can imagine at least two instances of confusing behavior:

1. you pass in `requests` and `selector`, the function enqueues the requests (even though their elements don't match the selector), and it extracts and enqueues some additional links from the current page
2. you pass in `requests` and `transform_request_function`, but the function won't be called

I know that the docblock makes it pretty clear that this is what will happen, but I also know that there will be bug reports about this :grin: 

In my opinion, we should make two overloads and throw a runtime error if someone passes both `requests` and one or more of the other arguments.

Pijukatel: > I can imagine at least two instances of confusing behavior:
> 
>     1. you pass in `requests` and `selector`, the function enqueues the requests (even though their elements don't match the selector), and it extracts and enqueues some additional links from the current page
> 
>     2. you pass in `requests` and `transform_request_function`, but the function won't be called
> 
> 
> I know that the docblock makes it pretty clear that this is what will happen, but I also know that there will be bug reports about this 😁
> 
> In my opinion, we should make two overloads and throw a runtime error if someone passes both `requests` and one or more of the other arguments.

Ok, so the function will basically become two in one alias for:

1. `add_requests(requests)`
2. `add_requests(extract_requests(...))`

src/crawlee/_types.py
```diff
@@ -324,20 +324,40 @@ def __call__(
...
+
     def __call__(
         self,
         *,
-        selector: str = 'a',
+        selector: str | None = None,
         label: str | None = None,
         user_data: dict[str, Any] | None = None,
         transform_request_function: Callable[[RequestOptions], RequestOptions | RequestTransformAction] | None = None,
+        requests: Sequence[str | Request] | None = None,
```
janbuchar, line 360: I now noticed that the JS counterpart accepts just `urls` as an array of strings. We should either restrict this, or extend the JS version :slightly_smiling_face: 

If we choose restricting this one, then most of the other parameters (barring `selector`) would actually start making sense.

vdusek: I would prefer to keep it as it is for consistency, since we use `request: str | Request` everywhere else.

vdusek (changes requested):
Shouldn't there be a new protocol `ExtractLinksFunction` in the `src/crawlee/_types.py`?

Pijukatel:
> Shouldn't there be a new protocol `ExtractLinksFunction` in the `src/crawlee/_types.py`?

I am not really sure that is necessary. `ExtractLinksFunction` is just inplace helper function. So even if I add the protocol, it will not be used anywhere.

Pijukatel:
> > I am not really sure that is necessary. ExtractLinksFunction is just inplace helper function. So even if I add the protocol, it will not be used anywhere.
> 
> However, `ExtractLinksFunction` should also be part of the public interface. Are we aligned on this? It could help address scenarios like the one mentioned here: #1084.

Ok, so adding it to contexts as well. I updated the change. Is that what you had in mind @vdusek ?

src/crawlee/_types.py
```diff
@@ -324,24 +324,76 @@ def __call__(
...
+                - `'skip'` to exclude the request from being enqueued,
+                - `'unchanged'` to use the original request options without modification.
+            requests: Requests to be added to the `RequestManager`.
+            **kwargs: Additional keyword arguments.
+        """
+
+
+@docs_group('Functions')
+class ExtractLinksFunction(Protocol):
+    """A function for extracting URLs to crawl based on elements selected by a given selector."""
```
vdusek, line 384: Could you make the docs better? something like enqueue links have?

extract links:
image

vs enqueue links:
image

vdusek (changes requested):
LGTM, just regarding the docs...

We have the following doc pages that exaplin `enqueue_links`:

- Crawl all links on website
- Crawl specific links on website
- Crawl website with relative links

We should extend them to explain the `extract_links` as well.

src/crawlee/_types.py
Pijukatel: Expanded.

Pijukatel:
> LGTM, just regarding the docs...
> 
> We have the following doc pages that exaplin `enqueue_links`:
> 
>     * Crawl all links on website
> 
>     * Crawl specific links on website
> 
>     * Crawl website with relative links
>  
>  We should extend them to explain the extract_links as well.

I added whole new section with two code examples to **Crawl specific links on website**. The other two pages seemed to be focused on something slightly different, so explaining  extract_links there might be little distracting from the focus of the example.

vdusek (approved):
Nice, good job!

PR #1026, opened 2025-02-25
chore: upgrade to Ruff v0.9

vdusek:
### Description

- Fix the incorrect usage of `Decimal` in tests.
- Rename internal modules because of `stdlib-module-shadowing`.
- Turn-off `stdlib-module-shadowing` for public modules.
- Codebase auto-format with a new version.


src/crawlee/storage_clients/_memory/_request_queue_client.py
```diff
@@ -514,12 +514,12 @@ def _calculate_order_no(self, request: Request, forefront: bool | None) -> Decim
             return None
 
         # Get the current timestamp in milliseconds
-        timestamp = Decimal(datetime.now(timezone.utc).timestamp()) * 1000
+        timestamp = Decimal(str(datetime.now(tz=timezone.utc).timestamp())) * Decimal('1000')
```
Pijukatel, line 517: Super tiny detail. Decimal('1000') == Decimal(1000). Not needed to use strings instead of ints.  (For floats we need strings)

In docs they do not use it for int either https://docs.python.org/3/library/decimal.html

Pijukatel (approved):

src/crawlee/storage_clients/_memory/_request_queue_client.py
vdusek: reverted

Issue #998, opened 2025-02-19
Refactor additional status codes arguments


Pijukatel:
Two arguments `additional_http_error_status_codes` and `ignore_http_error_status_codes` currently exists both on `BasicCrawler` and `HttpClient`. They are used to modify behavior of crawler based on the specific response status codes. 

Remove those arguments from `HttpClient`, but do retain Crawler functionality by relying only on the arguments in the Crawler.

Relates to:
https://github.com/apify/crawlee-python/pull/959
https://github.com/apify/crawlee-python/issues/830

PR #1028, opened 2025-02-26
refactor!: update status code handling

Mantisus:
### Description

* `additional_http_error_status_codes` and `ignore_http_error_status_codes` have been removed from the constructor parameters for HTTP clients
* The method `_raise_for_error_status_code` has been removed from `HttpClient` and its logic has been moved to `BasicCrawler`
* Prioritized checking of status codes that indicate `Session` blocking. Codes (401, 403, 429) trigger `retire` for the `Session` and retry, while other 4XX codes are handled as client errors (errors without retries).
* `additional_http_error_status_codes` is no longer used when checking status codes that indicate `Session` blocking. According to current documentation, these codes should trigger a retry, not `retire` for the `Session` and retry. Blocking status codes can be modified in `SessionPool` through `create_session_settings`.
* Standardized error handling for status codes in both `PlaywrightCrawler` and `HttpCrawler`

### Issues

- Closes: #998
- Closes: #830

### Testing

* Tests for client error status codes now use code 402.
* A separate test has been added for 403 since in this case the number of retries is affected by the 'max_session_rotations' parameter.

### DIFF
--- a/tests/unit/crawlers/_http/test_http_crawler.py
+++ b/tests/unit/crawlers/_http/test_http_crawler.py
@@ -82,6 +82,7 @@ async def server() -> AsyncGenerator[respx.MockRouter, None]:
             </html>""",
         )
 
+        # Endpoint with Session Blocked status code
         mock.get('/403', name='403_endpoint').return_value = Response(
             403,
             text="""<html>
@@ -91,6 +92,17 @@ async def server() -> AsyncGenerator[respx.MockRouter, None]:
             </html>""",
         )
 
+        # Endpoint with Client Error status code
+        mock.get('/402', name='402_endpoint').return_value = Response(
+            402,
+            text="""<html>
+                <head>
+                    <title>Not found</title>
+                </head>
+            </html>""",
+        )
+
+        # Endpoint with Server Error status code
         mock.get('/500', name='500_endpoint').return_value = Response(
             500,
             text="""<html>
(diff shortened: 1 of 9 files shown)


tests/unit/crawlers/_http/test_http_crawler.py
```diff
@@ -176,10 +188,54 @@ async def test_handles_client_errors(
...
+        pytest.param([], False, 0, 1, id='default_behavior_without_session_pool'),
+        # take as successful status codes from the `ignore_http_error_status_codes` list with Sessoion Pool
+        pytest.param([403], True, 0, 0, id='ignore_error_status_codes'),
+        # take as successful status codes from the `ignore_http_error_status_codes` list without Sessoion Pool
+        pytest.param([403], False, 0, 0, id='ignore_error_status_codes_without_session_pool'),
+    ],
+)
+async def test_handles_session_block_errors(
+    ignore_http_error_status_codes: list[int],
+    use_session_pool: bool,  # noqa: FBT001
```
Pijukatel, line 220: Super tiny detail. I prefer to put `*` into args to force named arguments and avoid using #noqa.

Pijukatel (approved):
Nice. One tiny optional comment. Approved anyway.

tests/unit/crawlers/_http/test_http_crawler.py
Mantisus: Good point!

janbuchar (approved):
Super nice! Thank you

PR #1044, opened 2025-03-03
fix: Add `browserforge` to mandatory dependencies

Pijukatel:
### Description

`Browserforge` is now used for header generation not only for `PlaywrightCrawler`, but also for http clients and so it needs to be mandatory dependency.


vdusek (approved):
LGTM

PR #1045, opened 2025-03-03
chore: Always run checks in release workflow

Pijukatel:
Previous optimization did not work as expected. Running checks always is safe and defensive approach.


vdusek (approved):
LGTM

Issue #954, opened 2025-02-03
Migrate project templates to poetry 2.x.x


Pijukatel:
Currently project templates lock the version of Poetry to <2
https://github.com/apify/crawlee-python/blob/master/src/crawlee/project_template/%7B%7Bcookiecutter.project_name%7D%7D/Dockerfile#L17

Remove the constraint to allow Poetry v2 and make sure that all templates works as expected.

janbuchar:
> or migrate to `uv` [#628](https://github.com/apify/crawlee-python/issues/628) 🙂

I believe that using newer poetry will be a much smaller undertaking.

vdusek:
Moving it to new issues and let's discuss it on Monday.

PR #1049, opened 2025-03-05
fix: update project templates for Poetry v2.x compatibility

Mantisus:
### Description

- Relax version poetry in template

### Issues

- Closes: #954

### DIFF
--- a/src/crawlee/project_template/hooks/pre_gen_project.py
+++ b/src/crawlee/project_template/hooks/pre_gen_project.py
@@ -7,6 +7,6 @@ try:
 except OSError as exc:
     raise RuntimeError('You chose to use the Poetry package manager, but it does not seem to be installed') from exc
 
-if not re.match(r'Poetry \(version 1\..*\)', version):
-    raise RuntimeError(f'Poetry below 3.x is required, but "{version}" is installed')
+if not re.match(r'Poetry \(version 2\..*\)', version):
+    raise RuntimeError(f'Poetry 2.x is required, but "{version}" is installed')
 # % endif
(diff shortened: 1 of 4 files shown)


Mantisus:
I tested the current templates with poetry 2.1.1 and found no problems.

If we want to add templates for `uv`, I think it should be done in a separate PR

janbuchar:
> I tested the current templates with poetry 2.1.1 and found no problems.

Awesome, does the dockerfile for deployment to apify still work?

> If we want to add templates for `uv`, I think it should be done in a separate PR

100% agree

vdusek (changes requested):
Could we please use the [PEP-621](https://peps.python.org/pep-0621/) format for the `pyproject.toml`? Poetry supports it from v2, so in that case, there should be a bottom version constraint as well.

Mantisus:
> Awesome, does the dockerfile for deployment to apify still work?

Yep. Yes, locally it runs correctly for `beautifulsoup`, `parsel` and `camoufox`. There is a problem for `Playwright`, but it is not `poetry` related

`Playwright` requires or modification of the docker template since I get the error 
`Running as root without --no-sandbox is not supported. See https://crbug.com/638180.`
Or running with the `browser_launch_options={'args': ['--no-sandbox']}` for `PlaywrightCrawler`.
Or using `firefox` as the browser in the template - I'd prefer this option

Mantisus:
> Could we please use the [PEP-621](https://peps.python.org/pep-0621/) format for the `pyproject.toml`? Poetry supports it from v2, so in that case, there should be a bottom version constraint as well.

Updated :)

vdusek (approved):
Awesome :rocket:, thanks

janbuchar (approved):

PR #1051, opened 2025-03-05
docs: website design uplift

webrdaniel:


baldasseva (approved):

website/docusaurus.config.js
```diff
@@ -340,13 +313,17 @@ module.exports = {
...
+                            href: 'https://www.youtube.com/apify',
+                        },
                     ],
                 },
                 {
                     title: 'More',
                     items: [
                         {
-                            label: 'Apify platform',
+                            label: 'Apify Platform',
```
B4nan, line 326: i believe we want to keep the `platform` lowercase, right @TC-MO ?

website/docusaurus.config.js
```diff
@@ -357,23 +334,24 @@ module.exports = {
                             label: 'GitHub',
                             href: 'https://github.com/apify/crawlee-python',
                         },
+
```
B4nan, line 337:
```suggestion
```

website/docusaurus.config.js
```diff
@@ -179,17 +216,13 @@ module.exports = {
...
         navbar: {
             hideOnScroll: true,
-            title: 'Crawlee for Python',
             logo: {
-                src: 'img/crawlee-light-new.svg',
-                srcDark: 'img/crawlee-dark-new.svg',
+                src: 'img/crawlee-python-light.svg',
+                srcDark: 'img/crawlee-python-dark.svg',
             },
+            title: 'Crawlee Python',
```
B4nan, line 225:
```suggestion
            title: 'Crawlee for Python',
```

website/docusaurus.config.js
TC-MO: Yup, but this also means that we have it wrong in JS Crawlee footer

webrdaniel: Updated in on both places 👍

PR #1056, opened 2025-03-06
refactor: allow extra attributes for storages

danpoletaev:
### Description

This PR follows up on https://github.com/apify/crawlee-python/pull/993. While aligning changes with the Apify SDK, I noticed that Pydantic models only select explicitly defined fields. However, we also need access to `urlSigningSecretKey`, so I’ve updated the `StorageMetadata` config **to allow extra fields.**

### Testing

Added tests for all storage types (datasets, key-value stores, and request queues) to ensure extra attributes are correctly handled.

### Checklist

- [x] CI passed

### DIFF
--- a/tests/unit/storages/test_dataset.py
+++ b/tests/unit/storages/test_dataset.py
@@ -145,6 +145,7 @@ async def test_from_storage_object() -> None:
         accessed_at=datetime.now(timezone.utc),
         created_at=datetime.now(timezone.utc),
         modified_at=datetime.now(timezone.utc),
+        extra_attribute='extra',
     )
 
     dataset = Dataset.from_storage_object(storage_client, storage_object)
@@ -152,3 +153,4 @@ async def test_from_storage_object() -> None:
     assert dataset.id == storage_object.id
     assert dataset.name == storage_object.name
     assert dataset.storage_object == storage_object
+    assert storage_object.model_extra.get('extra_attribute') == 'extra'  # type: ignore[union-attr]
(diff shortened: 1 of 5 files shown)


vdusek (commented):
Couldn't we add the `urlSigningSecretKey` field directly rather than allowing extra fields?

danpoletaev:
> Couldn't we add the `urlSigningSecretKey` field directly rather than allowing extra fields?

IMO: It doesn't make sense to provide `urlSigningSecretKey` as it is Apify specific field

Here is more [context](https://apify.slack.com/archives/CD0SF6KD4/p1738936334553189)

tests/unit/storages/test_dataset.py
```diff
@@ -145,10 +145,12 @@ async def test_from_storage_object() -> None:
...
         modified_at=datetime.now(timezone.utc),
+        extra_attribute='extra',
     )
 
     dataset = Dataset.from_storage_object(storage_client, storage_object)
 
     assert dataset.id == storage_object.id
     assert dataset.name == storage_object.name
     assert dataset.storage_object == storage_object
+    assert storage_object.model_extra.get('extra_attribute') == 'extra'  # type: ignore[union-attr]
```
vdusek, line 156: Could we anyhow get rid of the type ignores?

vdusek (approved):
> IMO: It doesn't make sense to provide urlSigningSecretKey as it is Apify specific field
> Here is more [context](https://apify.slack.com/archives/CD0SF6KD4/p1738936334553189)

Thanks, you're right, extra fields are a good solution then. LGTM.

Pijukatel (approved):

Issue #1053, opened 2025-03-05
Add an `uv` option to package manager selection in project bootstrapping


janbuchar:
The `Dockerfile` installation command can be the same as with poetry, but for exporting dependencies, `uv export --no-hashes --no-dev` should be used

PR #1057, opened 2025-03-06
feat: add project template with `uv` package manager

Mantisus:
### Description

- Add project template with `uv` package manager

### Issues

- Closes: #1053

### DIFF
--- a/src/crawlee/project_template/{{cookiecutter.project_name}}/Dockerfile
+++ b/src/crawlee/project_template/{{cookiecutter.project_name}}/Dockerfile
@@ -35,6 +35,28 @@ RUN echo "Python version:" \
     pip install -r /dev/stdin --no-dependencies \
  && echo "All installed Python packages:" \
  && pip freeze
+# % elif cookiecutter.package_manager == 'uv'
+RUN pip install -U pip setuptools \
+    && pip install 'uv<1'
+
+ENV UV_PROJECT_ENVIRONMENT="/usr/local"
+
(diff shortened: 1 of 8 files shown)


janbuchar (approved):
LGTM

src/crawlee/project_template/{{cookiecutter.project_name}}/Dockerfile
```diff
@@ -35,6 +35,28 @@ RUN echo "Python version:" \
...
+ENV UV_PROJECT_ENVIRONMENT="/usr/local"
+
+COPY pyproject.toml uv.lock ./
+
+RUN echo "Python version:" \
+    && python --version \
+    && echo "Installing dependencies:" \
+    # Check if playwright is already installed
+    && PLAYWRIGHT_INSTALLED=$(pip freeze | grep -q playwright && echo "true" || echo "false") \
+    && if [ "$PLAYWRIGHT_INSTALLED" = "true" ]; then \
```
Pijukatel, line 51: Not sure I follow, so I will think aloud:

This is here because we might already have some version of `playwright` installed in the base image in location "/usr/local". When we do, we exclude it from `uv sync` -> which btw syncs to where the python installation lives (so no isolated environment, but isolation is the image - so who cares).

If there is specific playwright version requirement(probably just theoretical use case) - we will ignore it and take whatever version already exists in the base image.  

Is my understanding correct?

Pijukatel (approved):

Issue #700, opened 2024-11-15
Add an option for JSON-compatible logs


vdusek:
### Description

Currently, Crawlee "statistics" logs are formatted as tables, which are human-readable but problematic when using JSON logs.

### Solution

Introduce a Crawler's flag that outputs logs in a JSON-compatible format. This would allow users to toggle between "table" and JSON-compatible logs.

janbuchar:
I remember successfully using https://github.com/madzak/python-json-logger before for the exact same purpose.

PR #1061, opened 2025-03-07
feat: Add `statistics_log_format` parameter to `BasicCrawler`

Mantisus:
### Description

- Add the `use_table_logs` parameter that allows disabling tables in logs. This makes log parsing easier when needed.

### Issues

- Closes: #700


Mantisus:
Thinking about this task, I believe we shouldn't add any third-party logger. A flag that disables tables in logs, which make data parsing difficult, is sufficient.

This will allow users to use any logger that's compatible with the standard one and enables customization of log output.

Example for loguru
```python
import inspect
import logging

from loguru import logger

class InterceptHandler(logging.Handler):
    def emit(self, record: logging.LogRecord) -> None:
        # Get corresponding Loguru level if it exists.
        try:
            level: str | int = logger.level(record.levelname).name
        except ValueError:
            level = record.levelno

        # Find caller from where originated the logged message.
        frame, depth = inspect.currentframe(), 0
        while frame:
            filename = frame.f_code.co_filename
            is_logging = filename == logging.__file__
            is_frozen = 'importlib' in filename and '_bootstrap' in filename
            if depth > 0 and not (is_logging | is_frozen):
                break
            frame = frame.f_back
            depth += 1

        logger.opt(depth=depth, exception=record.exc_info).log(level, record.getMessage())


logger.add('crawler.log', serialize=True, level='INFO')
logging.basicConfig(handlers=[InterceptHandler()], level=logging.INFO, force=True)

crawler = BeautifulSoupCrawler(configure_logging=False, use_table_logs=False)
```

Log record:
```json
{
    "text": "2025-03-07 16:51:09.947 | INFO     | crawlee.crawlers._basic._basic_crawler:run:580 - Final request statistics: requests_finished: 1; requests_failed: 0; retry_histogram: [1]; request_avg_failed_duration: None; request_avg_finished_duration: 0.795506; requests_finished_per_minute: 73; requests_failed_per_minute: 0; request_total_duration: 0.795506; requests_total: 1; crawler_runtime: 0.818803\n",
    "record": {
        "elapsed": { "repr": "0:00:01.921982", "seconds": 1.921982 },
        "exception": null,
        "extra": {},
        "file": {
            "name": "_basic_crawler.py",
            "path": "/home/mantisus/repos/apify/crawlee-python/src/crawlee/crawlers/_basic/_basic_crawler.py"
        },
        "function": "run",
        "level": { "icon": "ℹ️", "name": "INFO", "no": 20 },
        "line": 580,
        "message": "Final request statistics: requests_finished: 1; requests_failed: 0; retry_histogram: [1]; request_avg_failed_duration: None; request_avg_finished_duration: 0.795506; requests_finished_per_minute: 73; requests_failed_per_minute: 0; request_total_duration: 0.795506; requests_total: 1; crawler_runtime: 0.818803",
        "module": "_basic_crawler",
        "name": "crawlee.crawlers._basic._basic_crawler",
        "process": { "id": 32118, "name": "MainProcess" },
        "thread": { "id": 139760540858176, "name": "MainThread" },
        "time": {
            "repr": "2025-03-07 16:51:09.947345+00:00",
            "timestamp": 1741366269.947345
        }
    }
}
````

**Comment by janbuchar** [comment]
I like this take. Could you add this as an example to the docs about setting up JSON logs?

**Review by Pijukatel — APPROVED** [review]
Nice. Maybe add some tiny test that checks the non-default option(default option is already covered by existing tests.)

File: docs/examples/json_logging.mdx [hunk 1]
```diff
@@ -0,0 +1,46 @@
+---
+id: configure-jsonl-logging
```
vdusek, line 2:
```suggestion
id: configure-json-logging
```

or update the title

docs/examples/code_examples/configure_json_logging.py
```diff
@@ -0,0 +1,59 @@
+from __future__ import annotations
+
```
vdusek, line 2:
```suggestion
```

Do we need the __future__ annotations for an doc code example?

vdusek (commented):
Nice! Only two minor notes :slightly_smiling_face:...

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -134,6 +134,11 @@ class _BasicCrawlerOptions(TypedDict):
     configure_logging: NotRequired[bool]
     """If True, the crawler will set up logging infrastructure automatically."""
 
+    use_table_logs: NotRequired[bool]
```
janbuchar, line 137: I'm thinking, wouldn't it be better to have something like `statistics_log_format: Literal["table", "inline"]`? I think that most people won't know what a "table log" is...

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -564,8 +574,10 @@ def sigint_handler() -> None:
         await self._save_crawler_state()
 
         final_statistics = self._statistics.calculate()
-        self._logger.info(f'Final request statistics:\n{final_statistics.to_table()}')
-
+        if self._use_table_logs:
+            self._logger.info(f'Final request statistics:\n{final_statistics.to_table()}')
+        else:
+            self._logger.info(f'Final request statistics: {final_statistics.to_string()}')
```
janbuchar, line 580: Perhaps we could convert the statistics to a `dict` and then do this instead:

```py
self._logger.info('Final request statistics', extra=final_statistics.to_dict())
```

The preconfigured logger prints the extra in muted colors.

src/crawlee/crawlers/_basic/_basic_crawler.py
vdusek: I like this idea

vdusek (approved):
LGTM - resolve Honza's suggestions before merging

src/crawlee/crawlers/_basic/_basic_crawler.py
Mantisus: The `extra` parameters are not part of the standard logger message.

This will require the user to become more familiar with our log format in order to make a correct interceptor.

Current format, more transparent

janbuchar: Are they not? It seems pretty well documented [here](https://docs.python.org/3/library/logging.html)

image

Mantisus: I would call the `extra` parameter advanced log management.

It is documented and described for the standard logger, but it is not part of the standard message and requires additional processing. The standard `InterceptHandler` example I took from the `loguru` documentation for example will not handle it.

I would leave the `extra` parameter already directly for user logs. When the user knows where he uses it and why

janbuchar: Thing is, we already use it in crawlee and the SDK, so this would actually make the codebase more consistent.

Mantisus: Got it. Then it does make sense

docs/examples/json_logging.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+<CodeBlock className="language-python">
+    {JsonLoggingExample}
+</CodeBlock>
+
+Here's an example of what a crawler statistics log entry in JSONL format.
+
+```json
+{
+    "text": "2025-03-11 19:48:51.927 | INFO     | crawlee.crawlers._basic._basic_crawler:run:580 - Final request statistics:\n",
```
janbuchar, line 23: Shouldn't this say "http crawler" instead in the example?

janbuchar (approved):
One nit, feel free to resolve it at will. Otherwise LGTM

docs/examples/json_logging.mdx
janbuchar: I already see the "I set it up like the example say and now it's using the wrong crawler" issues coming :smile: any chance we could patch it up in the example, or would it be too hard?

Mantisus: I'll try to update )

PR #1067, opened 2025-03-10
refactor: Make `browserforge` dependency optional

Pijukatel:
### Description

Make `browserfore` dependency optional.
Add temporary fallback header generator for `HttpxHttpClient`.

This is temporary solution until browserforge is updated to avoid runtime downloads (https://github.com/daijro/browserforge/pull/29) and until `crawlee` is split into smaller more dedicated packages.

### Issues

- Relates to: [Logs from browserforge ](https://github.com/apify/apify-sdk-python/issues/423)


vdusek (changes requested):
I believe we do not need the PW-related stuff, correct?

Regarding the user-agents:
> Random 1000 user agents from Apify fingerprint dataset.

Since this is just a temporary fix, perhaps we could include e.g. only the 10 most common UAs from the list instead (now it is a subset of the dataset).

Pijukatel:
> I believe we do not need the PW-related stuff, correct?
> 
> Regarding the user-agents:
> 
> > Random 1000 user agents from Apify fingerprint dataset.
> 
> Since this is just a temporary fix, perhaps we could include e.g. only the 10 most common UAs from the list instead (now it is a subset of the dataset).

What is the benefit of that?
Including smaller set would downgrade the `HttpxClient` user agent header diversity compared to not only browserforge based headers, but also compared to the previous header generator. Maintenance cost is the same for 1000 or 10 samples. The only difference is in the size of the file - which is negligible compared to the size of the repo.

src/crawlee/http_clients/_httpx.py
```diff
@@ -10,7 +10,7 @@
 from crawlee._utils.blocked import ROTATE_PROXY_ERRORS
 from crawlee._utils.docs import docs_group
 from crawlee.errors import ProxyError
-from crawlee.fingerprint_suite import HeaderGenerator
+from crawlee.fingerprint_suite._fallback_header_generator import HeaderGenerator
```
janbuchar, line 13: Can't we instead import the real-deal browserforge-based generator when it's actually about to be used?

Pijukatel: I do not think that having runtime conditional imports is correct approach. As far as I see it the correct approach is fixing the `browserforge` to not execute code on import time and then split the repo into smaller pieces so that sdk does not need to import `BasicCrawler` -> `HttpxClient` -> `browserforge`, when it is not even using `BasicCrawler`.

When code is using `BasicCrawler`, then importing `browserforge` is legitimate.

vdusek (commented):
> What is the benefit of that?
> Including smaller set would downgrade the HttpxClient user agent header diversity compared to not only browserforge based headers, but also compared to the previous header generator. Maintenance cost is the same for 1000 or 10 samples. The only difference is in the size of the file - which is negligible compared to the size of the repo.

There are 1000 samples, but only ~200 are unique. We could use the X most common ones with even probability, achieving nearly the same result with 90% fewer lines of code.

Also, the Playwright headers aren't used at all.

However, considering it's a tmp fix, I won't discuss it further.

Also, it should be a fix, shouldn't it?

src/crawlee/http_clients/_httpx.py
janbuchar: Correct in what sense? If we can't control when `browserforge` stops fetching stuff on import time, then I believe importing conditionally (temporarily, until `browserforge` knows better) is slightly better than having a different implementation of HeaderGenerator in the httpx client.

Pijukatel: As per discussion on Monday I understood there are two separate issues:

1. browserforge doing stuff on import time
2. The fact that we need browserforge when `pip install crawlee` (without extras)

The open PR in browserforge repo is addressing the first point.
This PR is addressing the second point and is needed regardless of the first point (until we split the repo into smaller pieces).

Pijukatel:
Closed in favor of: https://github.com/apify/crawlee-python/pull/1075

PR #1068, opened 2025-03-10
fix: add a check thread before set `add_signal_handler`

Mantisus:
### Description

- Add a check thread before set `add_signal_handler`. Since `add_signal_handler` [can only be set in the main thread](https://docs.python.org/3/library/asyncio-eventloop.html#:~:text=Like%20signal.signal()%2C%20this%20function%20must%20be%20invoked%20in%20the%20main%20thread.).

### DIFF
--- a/src/crawlee/crawlers/_basic/_basic_crawler.py
+++ b/src/crawlee/crawlers/_basic/_basic_crawler.py
@@ -6,6 +6,7 @@ import logging
 import signal
 import sys
 import tempfile
+import threading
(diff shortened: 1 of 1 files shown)


Mantisus:
The problem is reproduced, for example in GCP Cloud Functions. Since under the hood they use Flask run via guvicorn

janbuchar (approved):
LGTM

vdusek (approved):
LGTM

PR #1080, opened 2025-03-13
chore(deps): mv eval-type-backport to dev deps

vdusek:


janbuchar:
Can you elaborate why this can be made a dev dependency? It is a dev dependency of `pydantic`, but then there's https://github.com/pydantic/pydantic/issues/7873 :thinking:

Issue #1076, opened 2025-03-12
Add support for request to use a specific session


dickermoshe:
I'm scraping a single site with 2 different logins. I don't see a way to specify which session/cookies a request should use?
The documentation for this is lacking.

Thanks

janbuchar:
Hello! You may specify additional cookies via `headers` when constructing a `Request` object. If you don't do this explicitly, you can use the `transform_request_function` argument of `context.enqueue_links`. In `PlaywrightCrawler`, this should take precedence over headers from session. In http-based crawlers, this is unfortunately implementation-dependent, so you'd need to experiment with your setup.

As @vdusek said, specifying which session should be used for a particular request is not supported as of now.

PR #1086, opened 2025-03-14
feat: add Session binding capability via `session_id` in `Request`

Mantisus:
### Description

- Add strict binding of a `Request` to a specific `Session`. If the `Session` is not available in the `SessionPool`, an error will be raised for the `Request` which can be handled in the `failed_request_handler`.

### Issues

- Closes: #1076 

### Testing

Added tests to verify functionality:
- Binding to a valid session
- Binding to a non-existent session
- Catching error in `failed_request_handler`


src/crawlee/_request.py
```diff
@@ -308,6 +326,11 @@ def label(self) -> str | None:
         """A string used to differentiate between arbitrary request types."""
         return cast(UserData, self.user_data).label
 
+    @property
+    def session_id(self) -> str | None:
+        """A string used to identify the bound session."""
+        return cast(UserData, self.user_data).session_id
```
Pijukatel, line 332: I see you are following the same pattern as already exists, but I think the cast is not necessary if we redefine model like this:

```
    user_data: Annotated[
        UserData,
        Field(alias='userData', default_factory=UserData),
        PlainValidator(user_data_adapter.validate_python),
        PlainSerializer(
            lambda instance: user_data_adapter.dump_python(
                instance,
                by_alias=True,
                exclude_none=True,
                exclude_unset=True,
                exclude_defaults=True,
            )
        ),
    ] = UserData()
```
I tried tests and type check and it seems to work fine like this.
@janbuchar was there any specific reason not covered by unit tests to have it defined as `dict[str, JsonSerializable] `and not as `UserData` ?

tests/unit/crawlers/_basic/test_basic_crawler.py
```diff
@@ -1194,3 +1195,52 @@ async def handler(context: BasicCrawlingContext) -> None:
...
+                used_sessions.append(context.session.id)
+
+        requests = [
+            Request.from_url('http://a.com/', session_id=check_session.id, always_enqueue=True) for _ in range(10)
+        ]
+
+        await crawler.run(requests)
+
+        assert len(used_sessions) == 10
+        assert all(session_id == check_session.id for session_id in used_sessions)
```
Pijukatel, line 1218: Maybe this more explicit `set(used_sessions) == {check_session.id}`

src/crawlee/_request.py
janbuchar: The reason for that is explained in the comment - it's for user convenience. With the `dict[str, JsonSerializable]`, you don't have to import `UserData` when constructing a `Request` - anything json-serializable will do.

src/crawlee/_request.py
```diff
@@ -296,6 +311,9 @@ def from_url(
         if label is not None:
             request.user_data['label'] = label
 
+        if session_id is not None:
+            request.user_data['session_id'] = session_id
```
janbuchar, line 315: I think it would be better to put this under `crawlee_data`.

src/crawlee/_request.py
Pijukatel: You do not have to import `UserData` even when you define your model with `user_data: Annotated[UserData`. It still allows you to initialize with json-serializable. Mypy is fine with that as well.

So with our current project setup, to me this seems like unnecessary missleading type + several casts for no benefit.

Pijukatel: So what is the real benefit of this in the end? (import argument not being applicable)

Pijukatel: We do not use this mypy settigns:
```
[tool.pydantic-mypy]
init_typed = true
```

therefore input to our models is can be whatever from mypy perspective

Pijukatel: So we should either use this settings and then there is some justification for this. Or if we do not use the settings, then this has no benefit at all.

janbuchar: You get precise type checking for `user_data` when creating a new `Request` object and the required type is understandable without any further investigation (dict of json-serializable values). 

I don't understand why the import argument shouldn't apply btw.

Pijukatel: > I don't understand why the import argument shouldn't apply btw.

Because there is no extra import needed in either case. Try to run it if you don't believe me.

Pijukatel (approved):

vdusek (changes requested):
Could we please cover this in the docs? :pray: Maybe [Session management](https://crawlee.dev/python/docs/guides/session-management)? Or find a better place. Thanks.

src/crawlee/_request.py
janbuchar: I tried :slightly_smiling_face: 

1. if you use `Request.from_url()`, it works OK because it has `**kwargs: Any`
2. if you construct it directly (`Request(url='https://crawlee.dev', user_data={'hello': 'darkness'}, unique_key="", id="")`), it's tricky
    a. mypy with `init_typed` is probably fine with it 
    b. pyright (what vscode uses) is not fine with it
    c. a different mypy configuration that our end users might use may or may not be fine with it

vdusek (approved):
That was easier than I thought, thanks, and good job! :slightly_smiling_face:

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1088,6 +1108,10 @@ async def __run_task_function(self) -> None:
 
             self._statistics.record_request_processing_finish(statistics_id)
 
+        except RequestCollisionError as request_error:
+            context.request.no_retry = True
+            await self._handle_request_error(context, request_error)
+
```
janbuchar, line 1114: So in case of a collision, the request will be aborted, right? Unless the user re-enqueues it in the `failed_request_handler`, which is kinda cumbersome. 

I imagine it can happen quite frequently that the session will get blocked and rotated out. I understand that silently using a new, different session might be confusing (but it might also be a viable option). Nevertheless, there should be an easier way to handle this in user code and it should be described in the documentation.

Mantisus: If we consider the case of the crawler working with authorization,

Each session is a separate authorized user.

Pages available for this user may not be available for another user. In this case, we cannot pass the request to another session. If a session becomes unavailable for some reason, then all we can do is give the user the opportunity to process these requests later.

Using this feature generally requires additional customizations to the `SessionPool`, so I don't think that having to use the `failed_request_handler` makes it any more cumbersome.

In general, I expect that users who will be using this know why they are doing it. 🙂

janbuchar: The tricky thing about `failed_request_handler` is that it is called after the request is marked as handled, so if you want to re-enqueue it with a different session, you need to make a new request. Are there any alternatives?

Mantisus: > Are there any alternatives?

Add a new handler for coliseums ?

> you need to make a new request.

That's right.

I would prefer it to be like this. A user created a `Request` with certain parameters, part of which is `session_id`. This `Request` has failed because one of the parameters is now invalid. I would expect that if the user wants to return it to the queue, a new `Request` with new parameters would be created

Mantisus:
> Do we support passing session ID to requests added by enqueue_links (or other Crawlee-native methods)?

No. Since using `session_id` requires additional `SessionPool` configuration (at least increasing the number of times a session is used, with extensive crawling). I would prefer not to give access to it via `enqueue_links`, this is for users who need more control. 🙂

> Requests are split between RQ and SessionPool

I understand your concerns. On the other hand, if we are talking about `Session`, then `Cookie` is a logical part of that entity.

If the user doesn't want to rely on `Session`, they can still pass `Cookie` as part of the `Request` headers

> Would one proxy hiccup (Apify proxies are afaik quite flaky) cause all the requests bound to the same session to fail?

If these are proxy errors related to timeouts and connections, the user should configure `SessionPool` so that the session does not die after a few errors 🙂

> but it still sounds like a very strict behavior

It is exactly what I expected, in the cases for which I think it can be used

src/crawlee/_request.py
```diff
@@ -119,6 +123,7 @@ class RequestOptions(TypedDict):
     headers: NotRequired[HttpHeaders | dict[str, str] | None]
     payload: NotRequired[HttpPayload | str | None]
     label: NotRequired[str | None]
+    session_id: NotRequired[str | None]
```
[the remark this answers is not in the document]
Mantisus: Yes, deduplication will affect this.

But I expect that users will use existing mechanisms to return a `Request` to the `Queue` avoiding deduplication. By passing either `unique_key` or `always_enqueue=True`.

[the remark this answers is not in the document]
vdusek: > Shouldn't session_id be used for unique_key computation? I expect that users might get hindered by deduplication if they try to re-enqueue a failed request with a different session.

Good point! Currently, it infers the `unique_key` from the URL, method, headers, and payload (in its extended form). You can, of course, use `session_id` together with `always_enqueue` and it will work, but that feels like a workaround to me. I believe we should include the `session_id` in the extended `unique_key` computation.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1226,3 +1250,18 @@ def _raise_for_session_blocked_status_code(self, session: Session | None, status
             ignore_http_error_status_codes=self._ignore_http_error_status_codes,
         ):
             raise SessionError(f'Assuming the session is blocked based on HTTP status code {status_code}')
+
+    def _raise_request_collision(self, request: Request, session: Session | None) -> None:
```
janbuchar, line 1254: I'd prefer to call this `check_request_collision` or something like that. Now the name sounds like it's always going to raise an error, which is not the case.

janbuchar (commented):
Some minor nits, but it's in good shape overall

src/crawlee/_utils/requests.py
```diff
@@ -114,9 +116,13 @@ def compute_unique_key(
     if use_extended_unique_key:
         payload_hash = _get_payload_hash(payload)
         headers_hash = _get_headers_hash(headers)
+        normilized_session = '' if session_id is None else session_id.lower()
```
janbuchar, line 119: One more typo, sorry :slightly_smiling_face: it should be `normalized_session`.

janbuchar (approved):

PR #1096, opened 2025-03-17
fix: add port for `same-origin` strategy check

Mantisus:
### Description

- add port for `same-origin` strategy check


Mantisus:
According to the documentation, we should also check the [`port'](https://crawlee.dev/python/docs/examples/crawl-website-with-relative-links#:~:text=The%20same%20origin%20refers%20to%20URLs%20that%20share%20the%20same%20protocol%2C%20domain%2C%20and%20port%2C%20ensuring%20a%20strict%20scope%20for%20the%20crawl.).

vdusek (approved):
Thanks

janbuchar (approved):
LGTM

PR #1100, opened 2025-03-18
docs: Use and enforce imperative mood

vdusek:
- Currently, our docstrings use a mix of imperative and indicative moods basically randomly.
- PEP 257 recommends using the imperative mood for the first line of docstrings, which is also enforced by the Ruff rule [D401](https://docs.astral.sh/ruff/rules/non-imperative-mood/#non-imperative-mood-d401). So I believe the imperative form is more suitable for the docs purposes.
- In most cases, I have simply changed the verb form, but a few cases required more rewording.
- I updated the pyproject configuration to enforce it.


src/crawlee/_autoscaling/autoscaled_pool.py
```diff
@@ -70,7 +70,7 @@ def __init__(
         is_task_ready_function: Callable[[], Awaitable[bool]],
         is_finished_function: Callable[[], Awaitable[bool]],
     ) -> None:
-        """A default constructor.
+        """Create a new instance.
```
janbuchar, line 73: `__init__` is not a constructor, it does not create the instance, it initializes it :slightly_smiling_face: Also, I'd prefer to just leave this blank, but I guess it's not possible to just have the `Args` section in a docblock.

vdusek: I also prefer leaving it blank, but as you mentioned, we use it for the args doc blocks.

What options do we have:
- Leave it blank and add a Ruff ignore (if our doc tooling supports it).
- Use a generic placeholder like "Create a new instance.", or maybe better as you pointed out: "Initialize a new instance.".

janbuchar: Yeah, b) with the updated wording is OK in my opinion.

vdusek: Resolved

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
```diff
@@ -71,7 +71,9 @@ def create_parsed_http_crawler_class(
         cls,
         static_parser: AbstractHttpParser[TParseResult, TSelectResult],
     ) -> type[AbstractHttpCrawler[ParsedHttpCrawlingContext[TParseResult], TParseResult, TSelectResult]]:
-        """Convenience class factory that creates specific version of `AbstractHttpCrawler` class.
+        """Create a specific version of `AbstractHttpCrawler` class.
+
+        Convenience class factory that.
```
Pijukatel, line 76: This is sentence is not complete

Pijukatel (approved):

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
vdusek: thanks

Issue #1093, opened 2025-03-17
Add integration tests for actor templates


Pijukatel:
**Motivation:**
Currently templates created through cli are not tested. The only test is that cli generating the templates is working, but the content of the templates is not tested. This situation leads to templates being broken from time to time and detection of such situation is rather random. 

**Proposed solution**
Create an apify-platform integration test for each template. Templates can run in different environments, but our apify-platform is the typical environment and is predictable and specific.
(This will mean adding dev dependency to apify-sahred for the sake of tests).

Each integration test should contain:

1. Creation of template through cli
2. running apify-init in template space
3. deploying to apify platform
4. running the actor
5. collecting the results and compare to expectations

PR #1109, opened 2025-03-20
chore: Add templates Apify E2E tests

Pijukatel:
### Description

Add `crawlee` templates integration tests with `Apify` platform. (only `Parsel` and `Beautifulsoup` templates)
Add workflow for manual triggering of integration tests (Will become automatic once it is running for a while to prove stability.)

### Issues

- Closes: #1093 

### Testing

Tested on forked repo:

- Example of no issue run:
https://github.com/Pijukatel/crawlee-python/actions/runs/13969844251
- Example of error in the template docker build (bad change in`uv` related part of the dockerfile):
https://github.com/Pijukatel/crawlee-python/actions/runs/13970390933
- Example of error in the actor execution expectations (bad change in `parsel` actor template):
https://github.com/Pijukatel/crawlee-python/actions/runs/13970267119

### Next steps

Add tests for `Playwright` based templates.
[Fix current problems with `Playwright`](https://github.com/apify/crawlee-python/issues/1110) template that has failing run in default settings.

Making the integration tests each run in separate job. Each test can take long time and can be unstable due to the network issues. This brings the necessity to have the possibility to re-trigger individual tests and not the whole test suite.

Since Playwright based crawlers are slower, this will be introduced together with their testing.
Example such run, where only one template test can be re-triggered:
image


pyproject.toml
```diff
@@ -211,8 +212,8 @@ timeout = 1200
 [tool.mypy]
 python_version = "3.9"
 plugins = ["pydantic.mypy"]
-exclude = ["project_template"]
-files = ["src", "tests", "docs", "website"]
+exclude = ["src/crawlee/project_template"]
+modules=["crawlee", "tests"]
```
vdusek, line 216: ?

Pijukatel: The exclude was done more specific to not exclude test/integration/project_template
The other line was not correct.

pyproject.toml
```diff
@@ -229,6 +230,7 @@ module = [
     "jaro",                         # Untyped and stubs not available
     "loguru",                       # Example code shows integration of loguru and crawlee for JSON logging.
     "sklearn.linear_model",         # Untyped and stubs not available
+    "cookiecutter.*",               # Untyped and stubs not available
```
vdusek, line 233: Could you add `inquirer` as well?

.github/workflows/templates_apify_integration_tests.yaml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Install dependencies
+        run: npm install -g apify-cli
+
+      - name: Set up Python ${{ env.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ env.python-version }}
+
+      - name: Install poetry
+        run: pipx install poetry
```
vdusek, line 39: Will it use the Python version we want and not the system one? Maybe we could use something from the GH marketplace, like for the `uv`?

src/crawlee/project_template/{{cookiecutter.project_name}}/Dockerfile
```diff
@@ -20,8 +20,7 @@ RUN pip install -U pip setuptools \
 # Second, copy just poetry.lock and pyproject.toml into the Actor image,
 # since those should be the only files that affects the dependency install in the next step,
 # in order to speed up the build
-COPY pyproject.toml ./
-COPY poetry.lock ./
+COPY pyproject.toml poetry.lock ./
```
vdusek, line 23: Is there any reason for this? If not, I would undo it.

tests/integration/conftest.py
```diff
@@ -0,0 +1,44 @@
...
+        # through an indicator file saying that the wheel was already built.
+        was_wheel_built_this_test_run_file = tmp_path_factory.getbasetemp() / f'wheel_was_built_in_run_{testrun_uid}'
+        if not was_wheel_built_this_test_run_file.exists():
+            subprocess.run(
+                args='python -m build',
+                cwd=_CRAWLEE_ROOT_PATH,
+                shell=True,
+                check=True,
+                capture_output=True,
+            )
```
vdusek, line 26: Shouldn't we add `build` as another dev dependency?

.github/workflows/templates_apify_integration_tests.yaml
```diff
@@ -0,0 +1,52 @@
+name: Templates Apify integration tests
```
vdusek, line 1: Could you explain how to execute the integration tests locally in the contributing guide?

vdusek (changes requested):
Looks good, just a few things...

Also, integration tests or end-to-end tests? :slightly_smiling_face: cc @janbuchar

src/crawlee/project_template/{{cookiecutter.project_name}}/Dockerfile
Pijukatel: Same pattern has already existed on line 43 for uv-related part of the template. So in order for the test to be generic, templates has to be as similar as possible. This one line notation is very convenient for me when detecting poetry vs uv in the same place infront of which I want to inject the wheel file related code.

https://github.com/apify/crawlee-python/pull/1109/files#diff-d27534684f00938b04217f5acb50047b794045bed996c79ddb50972edcca879aR22

It can be done also for the other notation, but unless there is good reason for it, I would prefer to keep it this way as it is very convenient.

tests/integration/conftest.py
Pijukatel: Good point

pyproject.toml
Pijukatel: Added.

.github/workflows/templates_apify_integration_tests.yaml
Pijukatel: I think that here it does not matter. This poetry is used only to inject the new wheel file to pyproject.toml and generate lock file. As far as I know, the lock file is generated not based on the current Python version, but based on the versions of the Python written in the pyproject.toml. 

Poetry used to install the project just before the test is different one and is the one that is in the docker container.

pyproject.toml
Pijukatel: Done

.github/workflows/templates_apify_integration_tests.yaml
Pijukatel: Added.

CONTRIBUTING.md
```diff
@@ -72,13 +72,19 @@ To run unit tests with HTML coverage report:
...
-TODO:
-
 ## Integration tests
 
-...
--->
+Pre-requisites for running integration tests:
+ - [apify-cli](https://docs.apify.com/cli/docs/installation) correctly installed
+ - apify-cli available in PATH environment variable
+ - Your [apify token](https://docs.apify.com/platform/integrations/api#api-token) is available in APIFY_TEST_USER_API_TOKEN environment variable
```
vdusek, line 80:
```suggestion
 - Your [apify token](https://docs.apify.com/platform/integrations/api#api-token) is available in `APIFY_TEST_USER_API_TOKEN` environment variable
```

CONTRIBUTING.md
```diff
@@ -72,13 +72,19 @@ To run unit tests with HTML coverage report:
...
-<!--
-TODO:
-
 ## Integration tests
 
-...
--->
+Pre-requisites for running integration tests:
+ - [apify-cli](https://docs.apify.com/cli/docs/installation) correctly installed
+ - apify-cli available in PATH environment variable
```
vdusek, line 79:
```suggestion
 - `apify-cli` available in `PATH` environment variable
```

src/crawlee/project_template/{{cookiecutter.project_name}}/Dockerfile
vdusek: :ok_hand:

.github/workflows/templates_apify_integration_tests.yaml
vdusek: I don't think it's a justification for doing it wrong.

Pijukatel: Not sure I understand your point. My point is, that the version here is not relevant, so in that case any attempt to enforce specific version is kind of misleading, but maybe I am wrong. Can you give me some example case where the current "version unspecific" solution causes an issue?

.github/workflows/templates_apify_integration_tests.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+on:
+  workflow_dispatch:
```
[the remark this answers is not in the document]
Pijukatel: Yes, but I would like to do it in two steps. First merge it with manual trigger and change it to automatic later when it is proven to work well in the repo.

Currently I am testing this workflow on forked repo and even though I had run it many times, it is still kind of a clumsy test setup, so it feels safer to merge it first with only manual trigger.

tests/integration/project_template/utils.py
```diff
@@ -0,0 +1,52 @@
...
+    # Copy prepared .whl file
+    shutil.copy(wheel_path, project_path)
+
+    # Get any extras
+    with open(project_path / 'pyproject.toml') as f:
+        pyproject = f.read()
+        crawlee_extras = re.findall(r'crawlee(\[.*\])', pyproject)[0] or ''
+
+    # Inject crawlee wheel file to the docker image and update project to depend on it."""
+    with open(project_path / 'Dockerfile') as f:
```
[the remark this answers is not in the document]
Pijukatel: The point of this somewhat complicated way of patching is to ensure it tests both the template package managers and the way they install the project and that it is as similar to user workflow as possible.

By using the `make actor` similar approach I think I would do slightly different installation steps, but more importantly I would have to make sure that `make actor` is maintained together with any changes to `apify-cli`. 

Imagine situation where there is some braking change to `apify push`. If I bypass the command by using `make actor` approach, then this test will be still passing, while out in the wilds it will start crashing for the people.

Pijukatel:
Example run on forked repo with latest changes: https://github.com/Pijukatel/crawlee-python/actions/runs/14165632652/job/39678213708

.github/workflows/templates_apify_integration_tests.yaml
Pijukatel: (Btw. Apart from no immediate benefit, another small additional reason why I am hesitant adding third party poetry github action is, that there is no "official" action like it is in case of uv. As far as I can see, poetry available actions are made by people unrelated to the poetry)

tests/integration/project_template/utils.py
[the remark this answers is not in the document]
janbuchar: Huh, testing `apify push` sounds way out of scope even for end-to-end tests. Also, non-backward-compatible changes to the API for building stuff are super unlikely. We already have a `make_actor` implementation in the SDK, so it would make a lot of sense to me to use the same approach here.

Makefile
```diff
@@ -35,6 +35,9 @@ unit-tests-cov:
 integration-tests:
 	uv run pytest --numprocesses=$(INTEGRATION_TESTS_CONCURRENCY) --verbose tests/integration
```
vdusek, line 36: I believe you can remove this and also all mentions of integration tests.

Makefile
```diff
@@ -35,6 +35,9 @@ unit-tests-cov:
...
 
+e2e-templates-tests:
+	uv run pytest --numprocesses=$(INTEGRATION_TESTS_CONCURRENCY) --verbose tests/e2e/project_template
```
vdusek, line 39: rename `INTEGRATION_TESTS_CONCURRENCY`?

pyproject.toml
```diff
@@ -92,6 +92,8 @@ crawlee = "crawlee._cli:cli"
 
 [dependency-groups]
 dev = [
+    "apify_client",  # For integration tests.
+    "build~=1.2.2",  # For integration tests.
```
vdusek, line 96: Update comment

tests/e2e/project_template/test_static_crawlers_templates.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+@pytest.mark.parametrize('http_client', ['httpx', 'curl-impersonate'])
+@pytest.mark.parametrize('crawler_type', ['parsel', 'beautifulsoup'])
+@pytest.mark.parametrize('package_manager', ['uv', 'poetry'])
+async def test_static_crawler_actor_at_apify(
+    tmp_path: Path, crawlee_wheel_path: Path, package_manager: str, crawler_type: str, http_client: str
+) -> None:
+    # Generate new actor name
+    actor_name = f'crawlee-python-template-integration-test-{crypto_random_object_id(8).lower()}'
```
vdusek, line 25: update

tests/e2e/project_template/test_static_crawlers_templates.py
```diff
@@ -0,0 +1,87 @@
...
+    subprocess.run(['apify', 'init', '-y', actor_name], capture_output=True, check=True, cwd=tmp_path / actor_name)  # noqa: ASYNC221, S603, S607
+
+    build_process = subprocess.run(['apify', 'push'], capture_output=True, check=False, cwd=tmp_path / actor_name)  # noqa: ASYNC221, S603, S607
+    # Get actor ID from build log
+    actor_id_regexp = re.compile(r'https:\/\/console\.apify\.com\/actors\/(.*)#\/builds\/\d*\.\d*\.\d*')
+    # Why is it in stderr and not in stdout???
+    actor_id = re.findall(actor_id_regexp, build_process.stderr.decode())[0]
+
+    client = ApifyClientAsync(
+        token=os.getenv('APIFY_TEST_USER_API_TOKEN'), api_url=os.getenv('APIFY_INTEGRATION_TESTS_API_URL')
```
vdusek, line 63: update `APIFY_INTEGRATION_TESTS_API_URL`?

tests/integration/README.md
```diff
@@ -1 +0,0 @@
-# Integration tests
```
vdusek, line 1: We can remove the whole `tests/integration` directory

tests/integration/project_template/utils.py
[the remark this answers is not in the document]
Pijukatel: `make_actor` is in different repo and is designed for code that is in the string format as the actors are generated on the fly by the test code. Here the actors are generated the way user generates them as a folder structure with files.

I would suggest to start with this approach and iteratively improve it later based on any potential issues we encounter. The tests are written in a way, that this discussed "patch function"  is well isolated from the rest of the test, so we can just change the patch function and the rest of the test remains the same.

Makefile
Pijukatel: Done

Makefile
Pijukatel: Done

tests/e2e/project_template/test_static_crawlers_templates.py
Pijukatel: Done

tests/integration/README.md
Pijukatel: Directly is probably just local leftover, but it is not stored in the repo.

vdusek (approved):
LGTM

PR #1115, opened 2025-03-24
fix(website): use correct image source

baldasseva:
The image sources as they are, are fetching the images from the root, i.e. https://crawlee.dev . Instead, we want them to fetch from crawlee python. There is namely one image that is different and has "Javascript" written on it ([see if you can find it](https://crawlee.dev/python)), but it should be "Python". The rest of the images are the same between js and py, so it doesn't matter as much.

### DIFF
--- a/website/src/pages/index.js
+++ b/website/src/pages/index.js
@@ -124,8 +124,8 @@ function BenefitsSection() {
                     <ThemedImage
                         alt="Work with your favorite tools"
                         sources={{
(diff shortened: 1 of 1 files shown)


B4nan (approved):

webrdaniel (approved):
Good catch 👍

B4nan:
dang, this should have been `docs` commit too 🙃

Issue #928, opened 2025-01-22
Implement a playwright-based `HttpClient` implementation


janbuchar:
There are two advantages to this:

1. `PlaywrightCrawler` could use it for `send_request` which would make those "AJAX calls" look more natural and probably less likely to be blocked
2. We could use it in e.g. `ParselCrawler` an another option to avoid blocking

PR #1128, opened 2025-04-01
feat: add playwright-based `HttpClient`

Mantisus:
### Description

- Add an asynchronous context for the `HttpClient` signature. This will allow correct opening and closing of resources, which is especially important for Playwright
- Add playwright-based `HttpClient` with [`APIRequestContext`](https://playwright.dev/python/docs/api/class-apirequestcontext)

### Issues

- Closes: #928


Mantisus:
Current Issues.

1. What advantages do we get using `APIRequestContext` Playwright, compared to a pure http client?

2. Cookie isolation. The `APIRequestContext` handles cookies and does not allow cookies to be set from the `Session` once created. Current solution, use one context per request. Cookies are set when the context is created and read into the `Session` after the request. What is the impact on resources and performance?

Mantisus:
I'm closing this PR in favor of #1134.

Issue #1055, opened 2025-03-06
bug: Request headers when using fingerprint generator are not saved in `context.request.headers`


Pijukatel:
Request headers are correctly present when making the request, but context is not aware of them.
This is due to the fact that the request headers are [set directly on Playwright in browserforge ](https://github.com/daijro/browserforge/blob/main/browserforge/injectors/playwright/injector.py#L31)and this bypasses our [common header setting approach](https://github.com/apify/crawlee-python/blob/master/src/crawlee/crawlers/_playwright/_playwright_crawler.py#L219)

Make sure that regardless of how the headers are set, they are present in the  `context.request.headers`

Pijukatel:
Either we do some internal refactoring to make it available or we wait for Playwright support.

Feature is tagged with 1.52 https://github.com/microsoft/playwright/issues/16239 (upcoming release)

We have to wait for such functionality to make it to the Python release as well, but maybe that would give as option how to handled this on Playwright level in straight forward way. Jeust get the headers from Playwright and save them to context.

PR #1134, opened 2025-04-03
feat: add an internal `HttpClient` to be used in `send_request` for `PlaywrightCrawler` using `APIRequestContext` bound to the browser context

Mantisus:
### Description

- This PR, adds an internal `HttpClient` implementation for `PlaywrightCrawler` using the `APIRequestContext` from `Playwright`. This ensures that HTTP requests use the same proxies as the browser context.

### Issues

- Closes: #928


Mantisus:
After discussions with @janbuchar in Slack, we came to the decision that this approach is best for `PlaywrightCrawler`. The requests will be closer to the browser context, and will use the same proxies that were set when the context was opened.

However, there is a problem. Playwright does not propagate the headers set with `set_extra_http_headers` to the request context. This results in the request context not receiving fingerprint headers.

This is already described in https://github.com/apify/crawlee-python/issues/1055

src/crawlee/crawlers/_playwright/_playwright_crawler.py
```diff
@@ -168,6 +168,8 @@ def __init__(
         kwargs.setdefault('_logger', logging.getLogger(__name__))
         self._pre_navigation_hooks: list[Callable[[PlaywrightPreNavCrawlingContext], Awaitable[None]]] = []
 
+        self._use_http_client = bool(kwargs.get('http_client'))
```
janbuchar, line 171: Is there any chance we could make an `HttpClient` implementation (it could be private to the `crawlee.crawlers._playwright` package) and use it instead of just overriding `context.send_request`? I admit I didn't read the code thoroughly, but it looks like you have most of the code in place already.

Mantisus: If we want to keep a tight coupling, between browser context and query context, I don't see that possibility (

janbuchar: can you elaborate what you mean by tight coupling?

janbuchar: Right... so the problem is that the http client would not know which browser instance should be used to create the request context?

Mantisus: Yep. That's right.

janbuchar: I'm still pondering a solution that would use contextvars to solve this, but I guess it's not worth it at the moment. The only upside is probably that it would make the playwright-request handling code easier to discover (same structure as other http clients) and that `http_client.send_request` called from a request handler would behave the same as `context.send_request`. Thoughts?

Mantisus: Updated to use an internal HTTP client using ‘contextvars’.

It seems cumbersome to me. But I agree that it is easier to navigate the code and more aligned with the overall structure.

I would ask you to check the use of `contextvars` as I have not used them before 😅

src/crawlee/crawlers/_playwright/_playwright_http_client.py
```diff
@@ -0,0 +1,74 @@
...
+    @override
+    async def send_request(
+        self,
+        url: str,
+        *,
+        method: HttpMethod = 'GET',
+        headers: HttpHeaders | dict[str, str] | None = None,
+        payload: HttpPayload | None = None,
+        session: Session | None = None,
+        proxy_info: ProxyInfo | None = None,
```
janbuchar, line 59: Can you also add a comment why these are ignored?

janbuchar (commented):
Nice, just a couple of nits :slightly_smiling_face:

janbuchar (approved):

vdusek (approved):
LGTM

PR #1147, opened 2025-04-10
fix: call `failed_request_handler` for `SessionError` when session rotation count exceeds maximum

Mantisus:
### Description

- Call `failed_request_handler` for `SessionError` when session rotation count exceeds maximum


vdusek (approved):
Nice! thanks

Issue #1079, opened 2025-03-13
Integrate `impit` as HTTP client


vdusek:
- Impit is already available on PyPI: [impit](https://pypi.org/project/impit/).
- Integrate it as the HTTP client, and make it a default one, while keeping `HTTPX` optional.
- Breaking change?
- The `apify_fingerprint_datapoints`, `browserforge` and `httpx[brotli,http2,zstd]` could be then removed from the mandatory dependencies.
- Relates to: https://github.com/apify/crawlee-python/issues/1077.

PR #1151, opened 2025-04-14
feat: add `ImpitHttpClient` http-client client using the `impit` library

Mantisus:
### Description

- add `ImpitHttpClient` http-client client using the `impit` library

### Issues

- Relates: #1079 

### Testing

Added tests for `ImpitHttpClient`. `ImpitHttpClient` is enabled for all tests using http-client


Mantisus:
For now, I suggest adding `impit` as an additional dependency, as it still needs some tweaking before it's ready to replace `httpx`.

Awaiting a decision - https://github.com/apify/impit/issues/123

Mantisus:
Python binding `Impit` has all the basic functionality to integrate into `Crawlee`.

The `_get_client` method is implemented based on [`ImpitHttpClient`](https://github.com/apify/crawlee/blob/master/packages/impit-client/src/index.ts#L38). However, this looks inefficient, especially when working without a proxy, but using a `SessionPool` of size greater than 1, because the client will be created anew for each request. I think we should improve this on the `impit` side. @barjin, maybe you'll have some ideas.

Replacing `httpx` with `impit` as the main client, I propose to do in a separate PR

tests/unit/http_clients/test_impit.py
```diff
@@ -0,0 +1,183 @@
...
+
+    crawling_result = await http_client.crawl(request)
+
+    assert crawling_result.http_response.status_code == 302
+    assert crawling_result.http_response.headers['Location'] == target_url
+    assert request.loaded_url == redirect_url
+
+
+async def test_stream(http_client: ImpitHttpClient, server_url: URL) -> None:
+    check_body = b"""\
```
Pijukatel: I would prefer to import HELLO_WORLD from tests.unit.server_endpoints

(same for the other occurrence)

tests/unit/http_clients/test_impit.py
Pijukatel, line 1: I would suggest to also add one test for multi-chunked compressed stream.
It seems to work correctly with impit, but it would be nice to have a test for it.

For example:
server.py:
```
async def get_compressed(_scope: dict[str, Any], _receive: Receive, send: Send) -> None:
    """Return large gzip compressed content."""

    await send(
        {
            'type': 'http.response.start',
            'status': 200,
            'headers': [[b'content-encoding', b'gzip']],
        }
    )
    await send({'type': 'http.response.body', 'body': gzip.compress(HELLO_WORLD*1000)})
``` 
and adding endpoint to server paths

test_impit.py:
```
async def test_compressed_chunked_stream(http_client: ImpitHttpClient, server_url: URL) -> None:
    content_body: bytes = b''

    async with http_client.stream(str(server_url/"get_compressed")) as response:
        assert response.status_code == 200
        async for chunk in response.read_stream():
            content_body += chunk

    assert content_body == HELLO_WORLD*1000
```

Pijukatel, line 1: I see you are following established pattern about adding new test file for the new client, but maybe now is the time to refactor the tests and have only one test file for all clients and parametrize all tests with client.

For example:
```
@pytest.mark.parametrize("http_client", [
    CurlImpersonateHttpClient(http_version=CurlHttpVersion.V1_1),
    ImpitHttpClient(),
    HttpxHttpClient(http2=False)])
async def test_http_1(http_client: HttpClient, server_url: URL) -> None:
    response = await http_client.send_request(str(server_url))
    assert response.http_version == 'HTTP/1.1'
```
Maybe we would need 3xclient factories instead and parametrize by that, but regardless of that I think it would be great to reduce code duplication and ensure that we have exactly the same tests for all clients and that they can all work in exactly the same way.

But this is just a suggestion. Maybe it can be done in separate PR as well to not mix new implementation with pure refactoring.

tests/unit/http_clients/test_impit.py
Mantisus: Great point. Updated for all clients

tests/unit/http_clients/test_impit.py
Mantisus: Added

Mantisus: That's a great idea. But yes, I think it should be done in a separate PR.

Also, we could take the same approach for `beautifulsoup` and `parsel` crawlers. Since the tests for them are also completely duplicated.

Pijukatel: https://github.com/apify/crawlee-python/issues/1292

Pijukatel (approved):

src/crawlee/http_clients/_impit.py
```diff
@@ -0,0 +1,252 @@
...
+        **async_client_kwargs: Any,
+    ) -> None:
+        """Initialize a new instance.
+
+        Args:
+            persist_cookies_per_session: Whether to persist cookies per HTTP session.
+            http3: Whether to enable HTTP/3 support.
+            verify: SSL certificates used to verify the identity of requested hosts.
+            header_generator: Header generator instance to use for generating common headers.
+            browser: Browser to impersonate ("chrome" or "firefox")
```
vdusek: The possible values it can acquire are already included in the type hints (`Browser` class).

```suggestion
            browser: Browser to impersonate.
```

vdusek (approved):
LGTM

Mantisus:
Not merging this PR until we resolve the test issue

tests/unit/http_clients/test_impit.py
Pijukatel: I will wait for this PR to be merged first, and then we can refactor: https://github.com/apify/crawlee-python/pull/1299

Makefile
```diff
@@ -34,8 +34,8 @@ unit-tests:
 	uv run pytest --numprocesses=auto -vv tests/unit -m "not run_alone"
 
 unit-tests-cov:
-	uv run pytest --numprocesses=1 -vv --cov=src/crawlee tests/unit -m "run_alone"
-	uv run pytest --numprocesses=auto -vv --cov=src/crawlee --cov-append --cov-report=html tests/unit -m "not run_alone"
+	uv run pytest --numprocesses=1 -vv --cov=src/crawlee tests/unit -m "run_alone" --timeout=120
+	uv run pytest --numprocesses=auto -vv --cov=src/crawlee --cov-append --cov-report=html tests/unit -m "not run_alone" --timeout=120
```
vdusek: We have "timeout = 1200" in pyproject toml, if you want to change, do it there please.

Mantisus: I apologize, this was leaked from the locale tests.

But, yes, I think this timeout should not exceed 5 minutes

tests/unit/server.py
```diff
@@ -454,6 +455,12 @@ async def watch_restarts(self) -> None:
             await self.shutdown()
             await self.startup()
 
+    def run(self, sockets: list[socket] | None = None) -> None:
+        """Run the server in a separate thread."""
+        if sys.version_info >= (3, 12) and sys.platform == 'win32':
```
Pijukatel, line 462: I think it would be good to put some explanatory comments here for future reference. Anyone not involved in chat messages will have no clue why is this done here.

tests/unit/server.py
```diff
@@ -454,6 +455,12 @@ async def watch_restarts(self) -> None:
...
+            asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
+        super().run(sockets=sockets)
```
vdusek: Could we please just add a comment explaining it?

vdusek (approved):
LGTM

PR #1159, opened 2025-04-17
feat: Handle unprocessed requests in `add_requests_batched`

Pijukatel:
### Description
Adds retry to unprocessed requests in call `add_requests_batched`.
Retry calls recursively `_process_batch`, which initially works on full request batch and then on batches of unprocessed requests until retry limit is reached or all requests are processed. Each retry is done after linearly increasing delay with each attempt.

Unprocessed requests are not counted in `request_queue.get_total_count`
Add test.

### Issues

- Closes: [Handle unprocessed requests in batch_add_requests](https://github.com/apify/apify-sdk-python/issues/456)

### DIFF
--- a/src/crawlee/storages/_request_queue.py
+++ b/src/crawlee/storages/_request_queue.py
@@ -240,12 +240,12 @@ class RequestQueue(Storage, RequestManager):
         # Wait for the first batch to be added
         first_batch = transformed_requests[:batch_size]
         if first_batch:
-            await self._process_batch(first_batch)
+            await self._process_batch(first_batch, base_retry_wait=wait_time_between_batches)
 
         async def _process_remaining_batches() -> None:
             for i in range(batch_size, len(transformed_requests), batch_size):
                 batch = transformed_requests[i : i + batch_size]
-                await self._process_batch(batch)
+                await self._process_batch(batch, base_retry_wait=wait_time_between_batches)
                 if i + batch_size < len(transformed_requests):
                     await asyncio.sleep(wait_time_secs)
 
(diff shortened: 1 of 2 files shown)


src/crawlee/storages/_request_queue.py
```diff
@@ -270,6 +264,31 @@ async def _process_remaining_batches() -> None:
                 timeout=wait_for_all_requests_to_be_added_timeout,
             )
 
+    async def _process_batch(self, batch: Sequence[Request], base_retry_wait: timedelta, attempt: int = 1) -> None:
+        max_attempts = 5
+        response = await self._resource_client.batch_add_requests(batch)
```
janbuchar, line 269: While you're at it, could you please make sure that `ApifyStorageClient` in the SDK does not do any retries of its own?

Of course, this is assuming we find that undesirable, but I firmly believe that the retry logic should be concentrated in a single place.

Pijukatel: I do not think it does. The retries that happens are currently in apify-client-python. First in Http clients and then in RequestQueueClients. Those retries are on different level and are both dealing with status codes of http responses and are not doing retries based on parsed content of successful API call - which is the case in this PR.

janbuchar: So `ApifyStorageClient` does its own retries via `apify-client-python`. Because the apify-client implementation also does its own kind of batching, the situation gets extra blurry. It doesn't make sense to change that in this PR, but I think we should make an issue to resolve this some day. If you agree, that is.

Pijukatel: Yes, I agree: https://github.com/apify/apify-client-python/issues/390

Issue #1144, opened 2025-04-08
Implement `respectRobotsTxtFile` crawler option


B4nan:
This option automatically fetches the robots.txt file based on the current request and adheres to the `disallow` directives.

JS version was implemented via the following PRs:
- https://github.com/apify/crawlee/pull/2910
- https://github.com/apify/crawlee/pull/2916
- https://github.com/apify/crawlee/pull/2913

We will first need to implement the `RobotsTxtFile` and `Sitemap` classes:
- https://github.com/apify/crawlee/blob/master/packages/utils/src/internals/robots.ts
- https://github.com/apify/crawlee/blob/master/packages/utils/src/internals/sitemap.ts

B4nan:
If this will be part of a major bump, we can enable the `respectRobotsTxtFile` by default, we want to do that in the JS version too once we get to the v4.

B4nan:
should be done now

PR #1162, opened 2025-04-17
feat: add  `respect_robots_txt_file` option

Mantisus:
### Description

- This PR implements automatic skipping of requests based on the robots.txt file. It works based on a new boolean flag in the crawler options called `respect_robots_txt_file`.

### Issues

- Related: #1144 

### Testing

- Add tests to check `respect_robots_txt_file` functioning in ‘EnqueueLinksFunction’ for crawlers
- Add tests for `RobotsTxtFile`


pyproject.toml
```diff
@@ -40,6 +40,7 @@ dependencies = [
     "eval-type-backport>=0.2.0",
     "httpx[brotli,http2,zstd]>=0.27.0",
     "more-itertools>=10.2.0",
+    "protego>=0.4.0",
```
janbuchar, line 43: It's fun to see another scrapy project here, but I guess that it guarantees some stability, so... all good.

src/crawlee/_utils/robots.py
```diff
@@ -0,0 +1,82 @@
...
+    from crawlee.http_clients import HttpClient
+    from crawlee.proxy_configuration import ProxyInfo
+
+
+class RobotsTxtFile:
+    def __init__(self, url: str, robots: Protego) -> None:
+        self._robots = robots
+        self._original_url = URL(url).origin()
+
+    @staticmethod
```
janbuchar, line 20: I'd prefer using `@classmethod` and the `Self` return type annotation

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -158,6 +161,10 @@ class _BasicCrawlerOptions(TypedDict):
     """A logger instance, typically provided by a subclass, for consistent logging labels. Intended for use by
     subclasses rather than direct instantiation of `BasicCrawler`."""
 
+    respect_robots_txt_file: NotRequired[bool]
+    """If set to `true`, the crawler will automatically try to fetch the robots.txt file for each domain,
```
janbuchar, line 165:
```suggestion
    """If set to `True`, the crawler will automatically try to fetch the robots.txt file for each domain,
```

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1265,3 +1307,30 @@ def _check_request_collision(self, request: Request, session: Session | None) ->
...
+
+        Args:
+            url: The URL to check.
+        """
+        if not self._respect_robots_txt_file:
+            return True
+        robots_txt_file = await self._get_robots_txt_file_for_url(url)
+        return not robots_txt_file or robots_txt_file.is_allowed(url)
+
+    async def _get_robots_txt_file_for_url(self, url: str) -> RobotsTxtFile | None:
```
janbuchar, line 1322: I believe we should use some synchronization mechanism so that we don't fetch the same robots.txt file multiple times in parallel.

pyproject.toml
Mantisus: Yes, I was planning to use [`RobotFileParser`](https://docs.python.org/3/library/urllib.robotparser.html), but it doesn't support Google's specification. 😞

src/crawlee/_utils/robots.py
```diff
@@ -0,0 +1,84 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from protego import Protego  # type: ignore[import-untyped]
```
vdusek, line 5: Could we please rather update the project toml rather than using a type ignore?

src/crawlee/_utils/robots.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class RobotsTxtFile:
+    def __init__(self, url: str, robots: Protego) -> None:
+        self._robots = robots
+        self._original_url = URL(url).origin()
+
+    @classmethod
+    async def from_content(cls, url: str, content: str) -> Self:
+        """Create a RobotsTxtFile instance from the given content.
```
vdusek, line 24:
```suggestion
        """Create a `RobotsTxtFile` instance from the given content.
```

src/crawlee/_utils/robots.py
```diff
@@ -0,0 +1,84 @@
...
+
+        Args:
+            url: the URL of the robots.txt file
```
vdusek, line 27: Could you please use sentences in arg description? (applies to all occurencies)

vdusek (changes requested):
Nice! I have a few details... And also, could you please write a new guide/example regarding this feature?

vdusek (approved):
Nice! LGTM

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -371,6 +383,8 @@ def __init__(
         self._additional_context_managers = _additional_context_managers or []
 
         # Internal, not explicitly configurable components
+        self._robots_txt_file_cache: LRUCache[str, RobotsTxtFile] = LRUCache(maxsize=1000)
+        self._robots_txt_locks_cache: LRUCache[str, asyncio.Lock] = LRUCache(maxsize=100)
```
janbuchar, line 387: I don't think these should be stored in a `LRUCache`. In my opinion, we can afford to keep it simple and have just one lock for all origins.

Mantisus: Fair enough. 🙂

Issue #1161, opened 2025-04-17
add Sitemap parser utility


Mantisus:
Sitemap parser is closely related to `RobotsTxtFile`. But it is not used for support `respect_robots_txt_file'

- Related: #1144

PR #1169, opened 2025-04-22
feat: add utility for load and parse Sitemap and `SitemapRequestLoader`

Mantisus:
### Description

- Add `SitemapRequestLoader` for comfortable working with `Sitemap` and easy integration into the framework
- Add utility for working with `Sitemap`, loads, and stream parsing

### Issues

- Closes: #1161

### Testing

- Add tests for `SitemapRequestLoader`
- Add new endpoints for the unicorn server for sitemaps tests


tests/unit/server.py
```diff
@@ -375,11 +369,32 @@ async def dynamic_content(scope: dict[str, Any], send: Send) -> None:
...
 
+async def get_sitemap_endpoint(scope: dict[str, Any], _receive: Receive, send: Send) -> None:
+    """Handle requests to serve XML sitemap content received in the request."""
+    query_params = get_query_params(scope.get('query_string', b''))
+
+    in_content = query_params.get('content', '')
+    in_base64 = query_params.get('base64', '')
+    c_type = query_params.get('c_type', 'application/xml; charset=utf-8')
+
+    out_content = base64.b64decode(in_base64) if in_base64 else in_content.encode()
```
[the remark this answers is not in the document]
Mantisus: It is not necessary for a server used only in tests. If Base64 contains an unplanned error, tests will report it

src/crawlee/_utils/robots.py
```diff
@@ -83,3 +85,14 @@ def get_crawl_delay(self, user_agent: str = '*') -> int | None:
...
         return int(crawl_delay) if crawl_delay is not None else None
+
+    async def parse_sitemaps(self) -> Sitemap:
+        """Parse the sitemaps from the robots.txt file and return a list of `Sitemap` instances."""
+        sitemaps = self.get_sitemaps()
+        proxy_url = self._proxy_info.url if self._proxy_info else None
+        return await Sitemap.load(sitemaps, proxy_url)
+
+    async def parse_urls_from_sitemaps(self) -> list[str]:
+        """Parse the URLs from the sitemaps in the robots.txt file and return a list of `Sitemap` instances."""
```
janbuchar: I think it should return URLs found in those sitemaps, right?

src/crawlee/_utils/sitemap.py
```diff
@@ -0,0 +1,462 @@
...
+class NestedSitemap:
+    loc: str
+    origin_sitemap_url: str | None = None
+
+
+class ParseSitemapOptions(TypedDict, total=False):
+    emit_nested_sitemaps: bool
+    max_depth: int
+    sitemap_retries: int
+    timeout: float | None
```
janbuchar: We usually do `timedelta` for timeouts.

tests/unit/server.py
```diff
@@ -375,11 +369,32 @@ async def dynamic_content(scope: dict[str, Any], send: Send) -> None:
     await send_html_response(send, html_content=content.encode())
 
 
-async def robots_txt(send: Send) -> None:
+async def robots_txt(_scope: dict[str, Any], _receive: Receive, send: Send) -> None:
     """Handle requests for the robots.txt file."""
     await send_html_response(send, ROBOTS_TXT)
 
 
+async def get_sitemap_endpoint(scope: dict[str, Any], _receive: Receive, send: Send) -> None:
```
janbuchar: If the endpoint just echoes whatever you send into it, maybe it doesn't need to be restricted to sitemaps? Wouldn't an all-purpose echo endpoint make more sense?

src/crawlee/_utils/sitemap.py
```diff
@@ -0,0 +1,462 @@
...
+            self._parser.feed(chunk)
+
+            # If we get here, the XML was valid and complete
+            for item in self._handler.items:
+                yield item
+
+            self._handler.items.clear()
+
+        except Exception as e:
+            logger.warning(f'Failed to parse XML data chunk: {e}')
```
janbuchar: Perhaps we could show the whole stack trace using `exc_info`?

src/crawlee/request_loaders/_sitemap_request_loader.py
```diff
@@ -0,0 +1,176 @@
...
+        self._total_count = 0
+
+        # URL queue and tracking
+        self._url_queue: asyncio.Queue[str] = asyncio.Queue(maxsize=max_buffer_size)
+        self._in_progress: set[str] = set()
+        self._processed_urls: set[str] = set()
+
+        # Loading state
+        self._loading_task = asyncio.create_task(self._load_sitemaps())
+        self._loading_finished = False
```
janbuchar: Can't you use `self._loading_task.done()` instead of making another property?

src/crawlee/request_loaders/_sitemap_request_loader.py
```diff
@@ -0,0 +1,176 @@
...
+    from collections.abc import Sequence
+
+    from crawlee.storage_clients.models import ProcessedRequest
+
+
+logger = getLogger(__name__)
+
+
+@docs_group('Classes')
+class SitemapRequestLoader(RequestLoader):
```
janbuchar, line 27: So the class doesn't have any automatically persisted state, right? Can you make a follow up issue to implement that?

src/crawlee/request_loaders/_sitemap_request_loader.py
```diff
@@ -0,0 +1,176 @@
...
+
+    The loader fetches and parses sitemaps in the background, allowing crawling to start
+    before all URLs are loaded. It supports filtering URLs using glob and regex patterns.
+    """
+
+    def __init__(
+        self,
+        sitemap_urls: list[str],
+        *,
+        proxy_url: str | None = None,
```
janbuchar: I'd prefer to accept an `HttpClient` instance here instead of just using `httpx`. I know that the Python `HttpClient` doesn't support streaming yet, but that's something we should fix anyway :grin:

janbuchar (commented):
Seems promising, thanks :slightly_smiling_face:

tests/unit/server.py
Mantisus: Great idea. thanks.

But I also saved the paths with `sitemap` in url as it is important to see that `.xml`, `.xml.gz` and `.txt` sitemaps are processed correctly.

src/crawlee/_utils/robots.py
Mantisus: Yep. Thanks for catching that.

src/crawlee/request_loaders/_sitemap_request_loader.py
Mantisus: Yes, I would also prefer for sitemap's utils to use `HttpClient`.

So it's time to add a `stream` method 😄

janbuchar: > So it's time to add a `stream` method 😄

It won't get easier later :slightly_smiling_face: If you can do it in a separate PR, please do.

Mantisus: Updated to use `HttpClient`

src/crawlee/_utils/sitemap.py
```diff
@@ -0,0 +1,474 @@
...
+                    # Process chunks as they arrive
+                    first_chunk = True
+                    async for raw_chunk in response.read_stream():
+                        # Check if the first chunk is a valid gzip header
+                        if first_chunk and raw_chunk.startswith(b'\x1f\x8b'):
+                            decompressor = zlib.decompressobj(zlib.MAX_WBITS | 16)
+                            first_chunk = False
+
+                        chunk = decompressor.decompress(raw_chunk) if decompressor else raw_chunk
+                        text_chunk = chunk.decode('utf-8', errors='replace')
```
janbuchar: Isn't there a chance that we'll try to decode a chunk that ends with a cut-off multibyte character? That would cause data corruption, right?

Mantisus: Yeah, there's a possibility. But, I'm not sure I can imagine a case where the sitemap would contain multibyte character

janbuchar: I mean, aren't all Chinese characters multibyte? And I assume that Chinese sitemaps will have those.

Mantisus: Yeah, I just believed in a better world where no one uses Cyrillic or hieroglyphics in the URL. But I found this case - https://www.hk01.com/sitemap.xml.

Apparently, my memory erased it from my head as a traumatic memory. :sweat_smile:

janbuchar: Yeah, then there's also emoji :grin: I guess that a streaming variant of `.decode()` must exist, somewhere...

Mantisus: Yep. Added [getincrementaldecoder](https://docs.python.org/3/library/codecs.html#codecs.getincrementaldecoder)

src/crawlee/request_loaders/_sitemap_request_loader.py
janbuchar: https://github.com/apify/crawlee-python/issues/1269

Pijukatel:
Looks good to me. I have a question though. This seems like a PR that deals with a problem that is already partially solved by existing packages. Probably not our specific request handling stuff, but at least the sitemap parsing part. For example this package seems pretty mature: https://github.com/GateNLP/ultimate-sitemap-parser/blob/main/usp/fetch_parse.py#L358 and it has word "ultimate" in name, so it must be really good :D
Could we maybe re-use some stuff from that package instead of having our own parsing in [sitemap.py](https://github.com/apify/crawlee-python/pull/1169/files#diff-d2ea030e308ff79dabe3d295461cb103f2febc8190cc9e5de6bd7cb7a669fc8b) ?

Mantisus:
> For example this package seems pretty mature: https://github.com/GateNLP/ultimate-sitemap-parser/blob/main/usp/fetch_parse.py#L358 and it has word "ultimate" in name, so it must be really good :D
Could we maybe re-use some stuff from that package instead of having our own parsing in [sitemap.py](https://github.com/apify/crawlee-python/pull/1169/files#diff-d2ea030e308ff79dabe3d295461cb103f2febc8190cc9e5de6bd7cb7a669fc8b) ?

Uh, thank you. I missed that when I was looking at available solutions before implementation. I'll see how promising it will be to use their solution. The only thing that confuses me is the presence of `requests` among the dependencies, not the lightest library out there

Mantisus:
> Could we maybe re-use some stuff from that package instead of having our own parsing in [sitemap.py](https://github.com/apify/crawlee-python/pull/1169/files#diff-d2ea030e308ff79dabe3d295461cb103f2febc8190cc9e5de6bd7cb7a669fc8b)?

I took a closer look at their code base. Unfortunately, despite the fact that in the documentation they write about stream processing, their code does not implement it.

In our case the possibility of streaming is quite important, because in some cases sitemaps can be..... take up a huge amount of memory.

Pijukatel:
> > Could we maybe re-use some stuff from that package instead of having our own parsing in [sitemap.py](https://github.com/apify/crawlee-python/pull/1169/files#diff-d2ea030e308ff79dabe3d295461cb103f2febc8190cc9e5de6bd7cb7a669fc8b)?
> 
> I took a closer look at their code base. Unfortunately, despite the fact that in the documentation they write about stream processing, their code does not implement it.
> 
> In our case the possibility of streaming is quite important, because in some cases sitemaps can be..... take up a huge amount of memory.

Thanks for checking it out and explanation.

Issue #1191, opened 2025-05-06
Update `push_data` annotations to use `JsonSerializable` type


vdusek:
Currently, we use on many places these annotations for `data` / `user_data`:

```python
data: list[dict[str, Any]] | dict[str, Any]
```

```python
data: dict[str, Any]
```

This works, but it isn't precise - we only accept JSON-serializable types.

We've got this recursive alias:

```python
J = TypeVar('J', bound='JsonSerializable')
JsonSerializable: TypeAlias = Union[
    list[J],
    dict[str, J],
    str,
    bool,
    int,
    float,
    None,
]
```

But if we use it for these variables:

```python
data: list[dict[str, JsonSerializable]] | dict[str, JsonSerializable]
```

```python
data: dict[str, JsonSerializable]
```

We run into variance-related errors, like this:

```
tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py:450: error: Argument 1 to "__call__" of "PushDataFunction" has incompatible type "dict[str, str]"; expected "Union[list[dict[str, Union[list[Any], dict[str, Any], str, bool, int, float, None]]], dict[str, Union[list[Any], dict[str, Any], str, bool, int, float, None]]]"  [arg-type]
tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py:450: note: "Dict" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py:450: note: Consider using "Mapping" instead, which is covariant in the value type
```

If we follow the suggestions, and use the `Mapping` and `Sequence`:

```python
data: Sequence[Mapping[str, JsonSerializable]] | Mapping[str, JsonSerializable]
```

We end up with even more errors on the usage side, e.g.

```python
item = {'key': 'value', 'number': 42}
await dataset_client.push_data(item)
```

Error (`dict[str, object]` vs. `Mapping[str, JsonSerializable]`)

```
Argument 1 to "push_data" of "MemoryDatasetClient" has incompatible type "dict[str, object]"; expected "Union[Sequence[Mapping[str, Union[list[Any], dict[str, Any], str, bool, int, float, None]]], Mapping[str, Union[list[Any], dict[str, Any], str, bool, int, float, None]]]" Mypy[arg-type](https://mypy.readthedocs.io/en/latest/_refs.html#code-arg-type)
```

[...]

PR #1193, opened 2025-05-07
Use `JsonSerializable` for `user_data`

Pijukatel:
### Description

Use `JsonSerializable` type for `user_data`

### Issues

- Closes: #1191

### DIFF
--- a/src/crawlee/_request.py
+++ b/src/crawlee/_request.py
@@ -129,7 +129,7 @@ class RequestOptions(TypedDict):
     keep_url_fragment: NotRequired[bool]
     use_extended_unique_key: NotRequired[bool]
     always_enqueue: NotRequired[bool]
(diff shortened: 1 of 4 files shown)


Pijukatel:
It seems ok according to mypy like this and unit tests are passing as well, but I am not really sure about this though. As discussed here: https://github.com/apify/crawlee-python/pull/1086/files/a67d3b3b42d2a1d59923810fe6f17dabeea3e98b#r1995086090

there are some casts down the road when it comes to user data, which makes me unsure about type checking result.

janbuchar:
The reason this is uncool is that crawlee puts its own stuff into `user_data` (more precisely `user_data['__crawlee']`), and this depends on it being a dictionary. Not sure if the platform would accept anything other than a dict, too...

Issue #783, opened 2024-12-05
Introduce new base & memory storage clients


vdusek:
### New interface of the `BaseStorageClient`

- Propose & implement a new `BaseStorageClient`.
- It should introduce a new interface, which does not have to be dependent on the `ApifyStorageClient`. Instead, we should implement an adapter for it.
- The `MemoryStorageClient` should implement this interface.

### The issues with the current `MemoryStorageClient`

- https://github.com/apify/crawlee-python/pull/691#discussion_r1871237112
  - it should accept just some smaller part of `Configuration` in the `open_storage`
- https://github.com/apify/crawlee-python/issues/816
  - there are problems with disk offloading
- https://github.com/apify/crawlee-python/issues/147
  - there are problems with initialization 
- There is probably no need to keep track of an `id` for each resource client - the memory storage deals with names only. However, we should consider that we will need to support some kind of an equivalent of Apify's non-default, unnamed storages.

janbuchar:
Regarding that comment on #691, I think that just a `Protocol` with select properties should do it.

vdusek:
After further discussions we're gonna:
- Remove collection clients.
- Get rid of the creation_management module as possible it would be.
- Get rid of client methods that do not make sense.
- Update the internals of FE storages.

Issue #147, opened 2024-05-10
Improve storage manager and merge it with `creation_management` module


vdusek:
The current [Crawlee / StorageClientManager](https://github.com/apify/crawlee-py/blob/master/src/crawlee/storage_client_manager.py) is more or less just copied from the [Python SDK / StorageClientManager](https://github.com/apify/apify-sdk-python/blob/master/src/apify/storages/storage_client_manager.py) and is extremely simple.  Its primary role is to maintain and provide access to storage client instances based on specific input parameters.

The [Crawlee TS / StorageManager](https://github.com/apify/crawlee/blob/master/packages/core/src/storages/storage_manager.ts) is more complex and it takes care of more things - creating instances of storages & their caching.

Currently, we have a helper module "creation_management" in `storages/` which helps with it.

Let's move logic from `storages/creation_management` to `StorageClientManager` and improve the creation & caching process.

Functions `get_or_create`, `find_or_create_client_by_id_or_name` a `create_*_from_directory` should be refactored.

Issue #92, opened 2024-04-05
Separate `MemoryStorageClient` and `FilesystemStorageClient`


vdusek:
### Description

Currently, we have a `MemoryStorageClient`, that can persist the data in the file system.

Let's separate them, `FilesystemStorageClient` could probably extend `MemoryStorageClient`

### Other related things

- There are memory storage-only data models in the `storage/models.py module`. Move them to the memory storage subpackage.

Issue #1175, opened 2025-04-25
Add support for non-default unnamed storages


vdusek:
## Problem

The Apify platform supports non-default unnamed storages. This functionality is also available in the Apify Python client, where you can do the following (example for dataset):

```py
await DatasetCollectionClientAsync.get_or_create()
```

Each call creates a new, unnamed dataset with a unique ID.

In contrast, Crawlee does not support this (in any storage client). For example, repeated calls to:

```py
await Dataset.open()
```

always return the same default unnamed storage.

## Goal state

Achieve feature parity between Crawlee storages (all storage clients, including the `ApifyStorageClient`) and the Apify platform (API client) by adding support for non-default unnamed storages.

## Possible solution

Introduce a new argument to the storage `open` constructor:

```py
async def open(
    cls,
    name: str,
    id: str,
    scope: Literal['run', 'global'] = 'global',
) -> Dataset | KeyValueStore | RequestQueue:
    ...
```

- `scope='run'` indicates a non-default unnamed storage.
- `scope='global'` refers to globally named storages.
- The `name` parameter cannot be entirely removed for run scope storages, as it's needed:
    - For the filesystem storage: to use as a directory name.
    - For Apify platform storage: to store the mapping of name -> ID in the default key-value store.

Behavior matrix...

### Open storage by ID and name

- Raise an exception.
- Scope argument is ignored.

### Open storage by ID

- Opens an existing storage by ID.
- Scope?

### Open storage by name

- Scope run:
    - Opens or creates a run-scope (non-default unnamed) storage.
        - `name` is used internally for reference-storage purposes but is not the actual storage's "name".
- Scope global:
    - Opens or creates a global named storage.

### Open storage without args

- Opens the default unnamed storage.
- Scope argument is ignored.

janbuchar:
When opening the storage by ID, the scope does not make sense. I think an exception would be appropriate.

B4nan:
We should ask for more feedback, e.g. on slack

metalwarrior665:
So to create a new persisted unnamed dataset, you would call `Dataset.open(name='debug', scope='run')` and then every time you call this (even after migration), it would return the same dataset, right?

Before releasing, I would have a short sync with the platform/output schema. There is e.g. this proposal so to make sure we don't use completely different terms https://github.com/apify/actor-whitepaper/pull/25

janbuchar:
[...]

PR #1194, opened 2025-05-10
refactor!: Introduce new storage client system

vdusek:
## Description

- I consolidated all commits from https://github.com/apify/crawlee-python/pull/1107 into this new PR.
- The previous storage-clients implementation was completely replaced with a redesigned clients, including:
    - new storage-client interface,
    - in-memory storage client,
    - file-system storage client,
    - Apify storage client (implemented in the SDK; see https://github.com/apify/apify-sdk-python/pull/470),
    - and various adjustments elsewhere in the codebase.
- The old "memory plus persist" client has been split into separate memory and file-system implementations.
    - The `Configuration.persist_storage` and `Configuration.persist_metadata` options were removed.
- All old collection clients have been removed, they're no longer needed.
- Each storage client now prints warnings if you pass method arguments it does not support.
- The creation management modules in the storage clients and storages were removed.
- Storage client parameters (e.g. `purge_on_start`, or `token` and `base_api_url` for the Apify client) are configured via the `Configuration`.
- Every storage, and its corresponding client, now provides both a `purge` method (which clears all items but preserves the storage and metadata) and a `drop` method (which removes the entire storage, metadata included).
- All unused types, models, and helper utilities have been removed.
- The detailed, per-storage/client changes are listed below...

### Dataset

- Properties:
    - `id`
    - `name`
    - `metadata`
- Methods:
    - `open`
    - `purge` (new method)
    - `drop`
    - `push_data`
    - `get_data`
    - `iterate_items`
    - `list_items` (new method)
    - `export_to`
- Breaking changes:
    - `from_storage_object` method has been removed - Use the `open` method with `name` or `id` instead.
    - `get_info` -> `metadata` property
    - `storage_object` -> `metadata` property
    - `set_metadata` method has been removed (it wasn't propage to clients)
        - Do we want to support it (e.g. for renaming)?
    - `write_to_json` -> method has been removed, use `export_to` instead
    - `write_to_csv` -> method has been removed, use `export_to` instead

```python
import asyncio

from crawlee.storage_clients import FileSystemStorageClient
from crawlee.storages import Dataset


async def main() -> None:
    dataset = await Dataset.open(storage_client=FileSystemStorageClient())
    print(f'default dataset - ID: {dataset.id}, name: {dataset.name}')

    await dataset.push_data({'name': 'John'})
    await dataset.push_data({'name': 'John', 'age': 20})
    await dataset.push_data({})

    dataset_with_name = await Dataset.open(
        name='my_dataset',
        storage_client=FileSystemStorageClient(),
    )
    print(f'named dataset - ID: {dataset_with_name.id}, name: {dataset_with_name.name}')

    await dataset_with_name.push_data([{'age': 30}, {'age': 25}])

    print('Default dataset items:')
    async for item in dataset.iterate_items(skip_empty=True):
        print(item)

    print('Named dataset items:')
    async for item in dataset_with_name.iterate_items():
        print(item)

    items = await dataset.get_data()
    print(items)

    dataset_by_id = await Dataset.open(id=dataset_with_name.id)
    print(f'dataset by ID - ID: {dataset_by_id.id}, name: {dataset_by_id.name}')


if __name__ == '__main__':
    asyncio.run(main())
```

### Key-value store

- Properties:
    - `id`
    - `name`
    - `metadata`
- Methods:
    - `open`
    - `purge` (new method)
    - `drop`
    - `get_value`
    - `set_value`
    - `delete_value` (new method, Apify platform's set_value support setting an empty value to a key, so having a separate method for deleting is useful)
    - `iterate_keys`
    - `list_keys` (new method)
    - `get_public_url`
    - `get_auto_saved_value`
    - `persist_autosaved_values`
- Breaking changes:
    - `from_storage_object` method has been removed - Use the `open` method with `name` or `id` instead.
    - `get_info` -> `metadata` property
    - `storage_object` -> `metadata` property
    - `set_metadata` method has been removed (it wasn't propage to clients)
        - Do we want to support it (e.g. for renaming)?

```python
import asyncio

import requests

from crawlee.storage_clients import FileSystemStorageClient
from crawlee.storages import KeyValueStore


async def main() -> None:
    print('Opening key-value store "my_kvs"...')
    storage_client = FileSystemStorageClient()
    kvs = await KeyValueStore.open(name='my_kvs', storage_client=storage_client)

    print('Setting value to "file.json"...')
    await kvs.set_value('file.json', {'key': 'value'})

    print('Setting value to "file.jpg"...')
    response = requests.get('https://avatars.githubusercontent.com/u/25082181')
    await kvs.set_value('file.jpg', response.content)

    print('Iterating over keys:')
    async for key in kvs.iterate_keys():
        print(f'Key: {key}')

    print('Listing keys:')
    keys = [key.key for key in await kvs.list_keys()]
    print(f'Keys: {keys}')

    for key in keys:
        print(f'Getting value of {key}...')
        value = await kvs.get_value(key)
        print(f'Value: {str(value)[:100]}')

    print('Deleting value of "file.json"...')
    await kvs.delete_value('file.json')

    kvs_default = await KeyValueStore.open(storage_client=storage_client)

    special_key = 'key with spaces/and/slashes!@#$%^&*()'
    test_value = 'Special key value'

    await kvs_default.set_value(key=special_key, value=test_value)

    record = await kvs_default.get_value(key=special_key)
    assert record is not None
    assert record == test_value

    result = await kvs_default.list_keys()
    print(f'kvs_default list keys = {result}')

    kvs_2 = await KeyValueStore.open()
    result = await kvs_2.list_keys()
    print(f'kvs_2 list keys = {result}')


if __name__ == '__main__':
    asyncio.run(main())
```

### Request queue

- Properties:
    - `id`
    - `name`
    - `metadata`
- Methods:
    - `open`
    - `purge` (new method)
    - `drop`
    - `add_request`
    - `add_requests_batched` -> `add_requests`
    - `fetch_next_request`
    - `get_request`
    - `mark_request_as_handled`
    - `reclaim_request`
    - `is_empty`
    - `is_finished`
- Breaking changes:
    - `from_storage_object` method has been removed - Use the `open` method with `name` or `id` instead.
    - `get_info` -> `metadata` property
    - `storage_object` -> `metadata` property
    - `set_metadata` method has been removed (it wasn't propage to clients)
        - Do we want to support it (e.g. for renaming)?
    - `get_handled_count` method had been removed - Use `metadata.handled_request_count` instead.
    - `get_total_count` method has been removed - Use `metadata.total_request_count` instead.
    - `resource_directory` from the `RequestQueueMetadata` was removed, use `path_to...` property instead.
    - `RequestQueueHead` model has been removed - Use `RequestQueueHeadWithLocks` instead.
- Notes:
    - New RQ `add_requests` contain `forefront` arg (Apify API supports it)

```python
import asyncio

from crawlee import Request
from crawlee.storage_clients import FileSystemStorageClient


async def main() -> None:
    rq = await RequestQueue.open(
        name='my-queue',
        storage_client=FileSystemStorageClient(),
        configuration=Configuration(purge_on_start=True),
    )

    print(f'RequestQueue: {rq}')
    print(f'RequestQueue client: {rq._client}')

    await rq.add_requests(
        requests=[
            Request.from_url('https://example.com', use_extended_unique_key=True),
            Request.from_url('https://crawlee.dev', use_extended_unique_key=True),
            Request.from_url('https://apify.com', use_extended_unique_key=True),
        ],
    )

    print('Requests were added to the queue')

    is_empty = await rq.is_empty()
    is_finished = await rq.is_finished()

    print(f'Is empty: {is_empty}')
    print(f'Is finished: {is_finished}')

    request = await rq.fetch_next_request()
    print(f'Fetched request: {request}')

    await rq.add_request('https://facebook.com', forefront=True)

    request = await rq.fetch_next_request()
    print(f'Fetched request: {request}')

    rq_default = await RequestQueue.open(
        storage_client=FileSystemStorageClient(),
        configuration=Configuration(purge_on_start=True),
    )

    await rq_default.add_request('https://example.com/1')
    await rq_default.add_requests(
        [
            'https://example.com/priority-1',
            'https://example.com/priority-2',
            'https://example.com/priority-3',
        ]
    )
    await rq_default.add_request('https://example.com/2')


if __name__ == '__main__':
    asyncio.run(main())
```

### BaseDatasetClient

- Properties:
    - `metadata`
- Methods:
    - `open`
    - `purge`
    - `drop`
    - `push_data`
    - `get_data`
    - `iterate_items`

### BaseKeyValueStoreClient

- Properties:
    - `metadata`
- Methods:
    - `open`
    - `purge`
    - `drop`
    - `get_value`
    - `set_value`
    - `delete_value`
    - `iterate_keys`
    - `get_public_url`

### BaseRequestQueueClient

- Properties:
    - `metadata`
- Methods:
    - `open`
    - `purge`
    - `drop`
    - `add_requests_batch` -> `add_batch_of_requests` (one backend method for 2 frontend methods)
    - `get_request`
    - `fetch_next_request`
    - `mark_request_as_handled`
    - `reclaim_request`
    - `is_empty`
- Models
    - `RequestQueueHeadWithLocks` -> `RequestQueueHead`
    - `BatchRequestsOperationResponse` -> `AddRequestsResponse`
- Notes:
    - Old file system (memory) version didn't persist the in-progress requests
    - Old file system (memory) version didn't persist the forefront values (now there is a FS-specific `_sequence` field in the FS Request)
    - The methods manipulating locks and listing heads are now only internal methods of Apify RQ client.

## Issues

- Closes: #92
- Closes: #147
- Closes: #783
- Closes: #1247
- Relates: #1175
- Relates: #1191

## Testing

- The original tests were mostly removed and replaced with a new ones.
- Each storage-client implementation now has its own dedicated tests at the client level (more targeted/edge-case coverage).
- On top of that, there are storage-level tests that use a parametrized fixture for each storage client (`file-system` and `memory`), ensuring every storage test runs against every client implementation.

## Checklist

- [x] CI passed


src/crawlee/storage_clients/_base/_key_value_store_client.py
```diff
@@ -6,126 +6,112 @@
...
 
-        Args:
-            key: The key of the record which to delete.
+        The backend method for the `KeyValueStore.iterate_keys` call.
         """
+        # This syntax is to make mypy properly work with abstract AsyncIterator.
+        # https://mypy.readthedocs.io/en/stable/more_types.html#asynchronous-iterators
+        raise NotImplementedError
+        if False:  # type: ignore[unreachable]
+            yield 0
```
Pijukatel, line 110: I am surprised that it is  Ok with `yield 0` even though it is `AsyncIterator[KeyValueStoreRecordMetadata]`...

src/crawlee/storage_clients/_file_system/_storage_client.py
```diff
@@ -0,0 +1,62 @@
...
+        self,
+        *,
+        id: str | None = None,
+        name: str | None = None,
+        configuration: Configuration | None = None,
+    ) -> FileSystemDatasetClient:
+        configuration = configuration or Configuration.get_global_configuration()
+        client = await FileSystemDatasetClient.open(id=id, name=name, configuration=configuration)
+
+        if configuration.purge_on_start and client.metadata.name is None:
```
Pijukatel, line 27: What is the scenario when `configuration.purge_on_start=True` but  `client.metadata.name is not None` ?
Maybe it would be worth adding some comment (or is it explained somewhere else?) about why `client.metadata.name` influences `purge_on_start` as I do not think it is self explanatory

src/crawlee/storage_clients/_memory/_storage_client.py
```diff
@@ -0,0 +1,62 @@
...
+        self,
+        *,
+        id: str | None = None,
+        name: str | None = None,
+        configuration: Configuration | None = None,
+    ) -> MemoryDatasetClient:
+        configuration = configuration or Configuration.get_global_configuration()
+        client = await MemoryDatasetClient.open(id=id, name=name, configuration=configuration)
+
+        if configuration.purge_on_start and client.metadata.name is None:
```
Pijukatel, line 27: Now I see this is repeated 6 time in the clients. Maybe it could be defined once in `StorageClient` as method with some nice explanatory docstring and just called in the clients?

tests/unit/crawlers/_basic/test_basic_crawler.py
```diff
@@ -544,6 +547,7 @@ async def test_crawler_get_storages() -> None:
     assert isinstance(kvs, KeyValueStore)
 
 
+# THIS
```
Pijukatel, line 550: Wha is THIS? :D

src/crawlee/storage_clients/_file_system/_storage_client.py
janbuchar: You are only supposed to purge unnamed storages to match platform behavior as closely as possible. Named storages are global and will typically outlive the run. 

Unnamed ones are most often "default" storages (almost 100% of cases). The platform creates a new one for each run, and we emulate this behavior with the `purge_on_start` behavior - if we're not resuming an interrupted run, we'll clear the contents of unnamed storages. This behavior is also desirable for non-default unnamed storages which are created ad-hoc by the Actor.

tests/unit/storage_clients/_memory/test_memory_kvs_client.py
```diff
@@ -0,0 +1,243 @@
...
+    # Verify keys are returned in sorted order
+    assert len(metadata_list) == 4
+    assert [m.key for m in metadata_list] == sorted(items.keys())
+    assert all(isinstance(m, KeyValueStoreRecordMetadata) for m in metadata_list)
+
+
+async def test_iterate_keys_with_exclusive_start_key(kvs_client: MemoryKeyValueStoreClient) -> None:
+    """Test that exclusive_start_key parameter returns only keys after it alphabetically."""
+    # Set some values
+    for key in ['a_key', 'b_key', 'c_key', 'd_key', 'e_key']:
```
Pijukatel, line 170: Tiny detail, add those test keys in non-alphabetical order to avoid some theoretical implementation detail sorting by insertion order and to really test alphabetical order.

tests/unit/storage_clients/_file_system/test_fs_kvs_client.py
```diff
@@ -0,0 +1,395 @@
...
+
+    # Iterate with limit
+    keys = [key.key async for key in kvs_client.iterate_keys(limit=2)]
+    assert len(keys) == 2
+
+
+async def test_iterate_keys_with_exclusive_start_key(kvs_client: FileSystemKeyValueStoreClient) -> None:
+    """Test that exclusive_start_key parameter returns only keys after it alphabetically."""
+    # Add some values with alphabetical keys
+    await kvs_client.set_value(key='a-key', value='value-a')
```
Pijukatel, line 314: Tiny detail, add those test keys in non-alphabetical order to avoid some theoretical implementation detail sorting by insertion order and to really test alphabetical order.

tests/unit/storage_clients/_memory/test_memory_kvs_client.py
```diff
@@ -0,0 +1,243 @@
...
+
+    # Get keys starting after 'b_key'
+    metadata_list = [metadata async for metadata in kvs_client.iterate_keys(exclusive_start_key='b_key')]
+
+    # Verify only keys after 'b_key' are returned
+    assert len(metadata_list) == 3
+    assert [m.key for m in metadata_list] == ['c_key', 'd_key', 'e_key']
+
+
+async def test_iterate_keys_with_limit(kvs_client: MemoryKeyValueStoreClient) -> None:
```
Pijukatel, line 181: It seems to me that this test would also benefit from non alphabetical insertion order in the setup as it would make it clear that data inside are sorted alphabetically. Or maybe add another test that inserts data in non alphabetical order and then assert that iterate_keys return it in alphabetical order?

Pijukatel (commented):
That was a  loooot of work. Very nice!

src/crawlee/_autoscaling/autoscaled_pool.py
```diff
@@ -142,7 +142,8 @@ async def run(self) -> None:
 
             logger.info('Waiting for remaining tasks to finish')
 
-            for task in run.worker_tasks:
+            tasks_to_wait = list(run.worker_tasks)
```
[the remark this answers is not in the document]
vdusek: Forgotten change from debugging.

src/crawlee/_types.py
```diff
@@ -656,3 +596,133 @@ async def get_snapshot(self) -> PageSnapshot:
...
+    """Keyword arguments for dataset's `export_data_json` method."""
+
+    skipkeys: NotRequired[bool]
+    """If True (default: False), dict keys that are not of a basic type (str, int, float, bool, None) will be skipped
+    instead of raising a `TypeError`."""
+
+    ensure_ascii: NotRequired[bool]
+    """Determines if non-ASCII characters should be escaped in the output JSON string."""
+
+    check_circular: NotRequired[bool]
```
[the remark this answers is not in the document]
vdusek: This is just moved from https://github.com/apify/crawlee-python/blob/v0.6.9/src/crawlee/storages/_dataset.py#L95, which is just copied from `json.dump`'s docstring.

src/crawlee/statistics/_error_snapshotter.py
```diff
@@ -23,41 +22,29 @@ class ErrorSnapshotter:
...
-            snapshot_save_tasks = []
-            if snapshot.html:
-                snapshot_save_tasks.append(
-                    asyncio.create_task(self._save_html(snapshot.html, base_name=snapshot_base_name))
-                )
-            if snapshot.screenshot:
-                snapshot_save_tasks.append(
-                    asyncio.create_task(self._save_screenshot(snapshot.screenshot, base_name=snapshot_base_name))
-                )
-            await asyncio.gather(*snapshot_save_tasks)
```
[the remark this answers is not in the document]
vdusek: The problem is in the parallel calls of

```python
kvs = await KeyValueStore.open(name=self._kvs_name)
```

which can result in flaky behaviour, see the `open` method for more context. If we want it to be more resistant to these errors, we can wrap the open method around another lock.

And since there are only two tasks, doing this in parallel offers only little benefit, that's the reason why I removed it completely. But if we want to keep it, I'll move the `open` call outside of the parallel block and keep only the `set_value` calls within it.

src/crawlee/storage_clients/_base/_key_value_store_client.py
vdusek: Old code, but if I remember it correctly, this is the recommended approach by Mypy.

src/crawlee/_utils/file.py
```diff
@@ -108,3 +52,89 @@ async def json_dumps(obj: Any) -> str:
         A string containing the JSON representation of the input object.
     """
     return await asyncio.to_thread(json.dumps, obj, ensure_ascii=False, indent=2, default=str)
+
+
+async def atomic_write_text(path: Path, data: str) -> None:
```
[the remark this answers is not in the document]
vdusek: Added, and I consolidated these 2 functions into a single one.

src/crawlee/_utils/file.py
```diff
@@ -108,3 +52,89 @@ async def json_dumps(obj: Any) -> str:
...
+    try:
+        await asyncio.to_thread(os.replace, tmp_path, str(path))
+    except (FileNotFoundError, PermissionError):
+        # fallback if tmp went missing
+        await asyncio.to_thread(path.write_text, data, encoding='utf-8')
+    finally:
+        await asyncio.to_thread(Path(tmp_path).unlink, missing_ok=True)
+
+
+async def atomic_write_bytes(path: Path, data: bytes) -> None:
```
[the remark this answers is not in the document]
vdusek: Added, and I consolidated these 2 functions into a single one.

src/crawlee/storage_clients/_memory/_storage_client.py
vdusek: Is it worth it to save 1 line of code? :slightly_smiling_face:

src/crawlee/_utils/file.py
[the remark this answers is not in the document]
Pijukatel: Nice.

src/crawlee/storage_clients/_file_system/_storage_client.py
Pijukatel: That makes perfect sense if you know context outside of this package :-)

src/crawlee/storage_clients/_memory/_storage_client.py
Pijukatel: I think it makes sense from readability point of view. Choosing good function name and summarizing @janbuchar [explanation](https://github.com/apify/crawlee-python/pull/1194#discussion_r2088415392) to the doc string of such method would make it very clear why it is there without the need to know some external context.

Pijukatel (commented):
I think it is good. Just please add section to the upgrading_to_v0x.md to summarize all the breaking changes in this.

src/crawlee/storage_clients/_file_system/_request_queue_client.py
```diff
@@ -0,0 +1,784 @@
...
+
+    This client persists requests to the file system as individual JSON files, making it suitable for scenarios
+    where data needs to survive process restarts. Each request is stored as a separate file in a directory
+    structure following the pattern:
+
+    ```
+    {STORAGE_DIR}/request_queues/{QUEUE_ID}/{REQUEST_ID}.json
+    ```
+
+    The implementation uses file timestamps for FIFO ordering of regular requests and maintains in-memory sets
```
janbuchar, line 40: So is this the same mechanism that was used in memory storage request queue previously, or is it new?

src/crawlee/storage_clients/_file_system/_request_queue_client.py
```diff
@@ -0,0 +1,784 @@
...
+
+        try:
+            return Request(**file_content)
+        except ValidationError as exc:
+            logger.warning(f'Failed to validate request file {request_path}: {exc!s}')
+
+        return None
+
+    @override
+    async def fetch_next_request(self) -> Request | None:
```
janbuchar, line 430: This is very long. Any chance you could split it into smaller chunks?

Mantisus:
That's excellent work!

janbuchar (commented):
I like this a lot. I do need to revisit the request queue related code though, it feels like we're throwing out the baby with the bathwater.

src/crawlee/_service_locator.py
```diff
@@ -77,13 +75,7 @@ def set_event_manager(self, event_manager: EventManager) -> None:
...
         """Get the storage client."""
         if self._storage_client is None:
-            from crawlee.storage_clients import MemoryStorageClient
-
-            self._storage_client = (
-                MemoryStorageClient.from_config(config=self._configuration)
-                if self._configuration
-                else MemoryStorageClient.from_config()
-            )
+            self._storage_client = FileSystemStorageClient()
```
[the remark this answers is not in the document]
vdusek: Storage client does not have from config method.

src/crawlee/storage_clients/_base/_key_value_store_client.py
```diff
@@ -6,126 +6,112 @@
...
+    async def get_value(self, *, key: str) -> KeyValueStoreRecord | None:
+        """Retrieve the given record from the key-value store.
 
-        Returns:
-            The requested record, or None, if the record does not exist
+        The backend method for the `KeyValueStore.get_value` call.
         """
 
     @abstractmethod
-    async def stream_record(self, key: str) -> AbstractAsyncContextManager[KeyValueStoreRecord[Response] | None]:
```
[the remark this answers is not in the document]
vdusek: Exactly

src/crawlee/configuration.py
```diff
@@ -118,21 +118,7 @@ class Configuration(BaseSettings):
             )
         ),
     ] = True
-    """Whether to purge the storage on the start. This option is utilized by the `MemoryStorageClient`."""
-
-    write_metadata: Annotated[bool, Field(alias='crawlee_write_metadata')] = True
-    """Whether to write the storage metadata. This option is utilized by the `MemoryStorageClient`."""
-
-    persist_storage: Annotated[
```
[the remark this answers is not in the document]
vdusek: If we were upgrading from v1 to v2, I'd be all doing it, but since we're moving from a zero‐based beta to v1, I'd rather keep things simple. Enabling this flag while also explicitly providing a storage client could create weird edge cases and trigger the storage client overriding-exceptions at the service-locator level.

src/crawlee/storage_clients/_file_system/_request_queue_client.py
vdusek: The whole FS RQ client was refactored, should be much better now.

src/crawlee/storages/_dataset.py
```diff
@@ -249,27 +109,55 @@ async def open(
         configuration: Configuration | None = None,
         storage_client: StorageClient | None = None,
     ) -> Dataset:
-        from crawlee.storages._creation_management import open_storage
+        if id and name:
```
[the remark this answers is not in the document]
vdusek: Added util opener

src/crawlee/request_loaders/_request_manager_tandem.py
```diff
@@ -32,9 +32,15 @@ def __init__(self, request_loader: RequestLoader, request_manager: RequestManage
...
     @override
-    async def get_total_count(self) -> int:
-        return (await self._read_only_loader.get_total_count()) + (await self._read_write_manager.get_total_count())
+    async def handled_count(self) -> int:
+        return await self._read_write_manager.handled_count
+
+    @property
+    @override
+    async def total_count(self) -> int:
+        return (await self._read_only_loader.total_count) + (await self._read_write_manager.total_count)
```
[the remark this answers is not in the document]
vdusek: Reverted to async getters.

vdusek:
All the feedback was addressed, including the upgrading guide. Could you guys please take a second look?

(FYI; this https://github.com/apify/crawlee-python/pull/1194/commits/9f10b955c6d8c0d82940ad1c6bec0be6d5274565 broke the docs build, @barjin will take a look at it later.)

src/crawlee/_utils/file.py
```diff
@@ -1,110 +1,156 @@
...
+            Path(tmp_path).unlink(missing_ok=True)
+            raise
+        return tmp_path
+
+    tmp_path = await asyncio.to_thread(_sync_write)
+
+    try:
+        await asyncio.to_thread(os.replace, tmp_path, str(path))
+    except (FileNotFoundError, PermissionError):
+        # fallback if tmp went missing
```
Pijukatel: Not sure I fully understand the use case for this exception. We use `_sync_write` instead of using directly `await asyncio.to_thread(path.write_bytes, data)` because of possibility of the code being interrupted at any moment. But in this exception handling we are taking the risk again? Shouldn't we try to call `atomic_write` again instead?

vdusek: > Not sure I fully understand the use case for this exception. We use _sync_write instead of using directly await asyncio.to_thread(path.write_bytes, data) because of possibility of the code being interrupted at any moment.

You're correct.

> But in this exception handling we are taking the risk again? Shouldn't we try to call atomic_write again instead?

Good catch, I'll add retries there.

src/crawlee/_utils/file.py
```diff
@@ -1,110 +1,178 @@
...
+                    tmp_file.write(data)  # type: ignore[arg-type]
+        except Exception:
+            Path(tmp_path).unlink(missing_ok=True)
+            raise
+        return tmp_path
+
+    try:
+        # We have to differentiate between Windows and Linux due to the permissions errors
+        # in Windows when working with temporary files.
+        if sys.platform == 'win32':
```
Pijukatel: Just a detail.Should it be maybe resolved only once during import time?  Having something like this in the top file that would be resolved on import time: 

```
if sys.platform == 'win32':
  def _write(...)
    #Win version
else:
  def _write(...)
    #Linux version
```

src/crawlee/_utils/file.py
```diff
@@ -95,41 +141,14 @@ async def atomic_write(
...
-        if sys.platform == 'win32':
-            await asyncio.to_thread(_write_windows)
-        else:
-            tmp_path = await asyncio.to_thread(_write_linux)
+        # Use the platform-specific write function resolved at import time.
+        tmp_path = await asyncio.to_thread(_write_file, path, data, is_binary=is_binary)
+
+        # On Linux/Unix, replace the destination file with tmp file.
+        if tmp_path is not None:
             await asyncio.to_thread(os.replace, tmp_path, str(path))
```
Pijukatel: Shouldn't this line be also part of the `_write_file` function? To me it seems, that this temp file,repalce,unlink are the implementation details of the Linux variant.

src/crawlee/_request.py
```diff
@@ -158,7 +158,23 @@ class Request(BaseModel):
     ```
     """
 
+    model_config = ConfigDict(populate_by_name=True, extra='allow')
```
janbuchar: Why the `extra='allow'`?

src/crawlee/_types.py
```diff
@@ -190,7 +195,7 @@ class PushDataKwargs(TypedDict):
 
 
 class PushDataFunctionCall(PushDataKwargs):
-    data: JsonSerializable
+    data: list[dict[str, Any]] | dict[str, Any]
```
janbuchar, line 198: Um, this goes against https://github.com/apify/crawlee-python/issues/1191, actually :grin:

src/crawlee/request_loaders/_request_loader.py
```diff
@@ -25,9 +25,13 @@ class RequestLoader(ABC):
         - Managing state information such as the total and handled request counts.
     """
 
+    @abstractmethod
+    async def get_handled_count(self) -> int:
+        """Get the number of requests in the loader that have been handled."""
+
     @abstractmethod
     async def get_total_count(self) -> int:
-        """Return an offline approximation of the total number of requests in the source (i.e. pending + handled)."""
```
janbuchar, line 30: The wording here is actually kinda important - in `RequestQueue`, this is not refetched on each call, so the number might be a bit off (but the method is more performant)

src/crawlee/storage_clients/_base/_storage_client.py
```diff
@@ -8,55 +6,76 @@
...
-    def key_value_stores(self) -> KeyValueStoreCollectionClient:
-        """Get a subclient for key-value store collection operations."""
+    Each storage client implementation is responsible for ensuring proper initialization, data persistence
+    (where applicable), and consistent access patterns across all storage types it supports.
+    """
 
     @abstractmethod
-    def request_queue(self, id: str) -> RequestQueueClient:
-        """Get a subclient for a specific request queue by its ID."""
+    async def open_dataset_client(
```
janbuchar: I'm not so sure about the naming of these methods - you don't really open the client,  you open the storage. Maybe we could drop the `_client` suffix?

src/crawlee/storage_clients/_base/_storage_client.py
```diff
@@ -8,55 +6,76 @@
...
+        self,
+        *,
+        id: str | None = None,
+        name: str | None = None,
+        configuration: Configuration | None = None,
```
janbuchar, line 37: Is there a use case for passing in a different `Configuration` than the one that the `StorageClient` may already hold in a private property?

src/crawlee/storages/_utils.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+async def open_storage_instance(
+    cls: type[T],
+    *,
+    id: str | None,
+    name: str | None,
+    configuration: Any,
+    cache_by_id: dict[str, T],
+    cache_by_name: dict[str, T],
```
janbuchar: I'm not sure about making these caches part of the storage "frontend" classes. I know we've moved them several times now, but could we try extracting the behavior into a separate class accessible via service locator that would hold instances of all storage types?

src/crawlee/storage_clients/_file_system/_dataset_client.py
```diff
@@ -0,0 +1,483 @@
...
+        return self._storage_dir / self._STORAGE_SUBDIR / self.metadata.name
+
+    @property
+    def path_to_metadata(self) -> Path:
+        """The full path to the dataset metadata file."""
+        return self.path_to_dataset / METADATA_FILENAME
+
+    @override
+    @classmethod
+    async def open(
```
janbuchar, line 92: It looks like there is a lot of duplication between `FileSystemDatasetClient.open` and its counterparts for kv-stores and RQs. Is it worth refactoring or not really?

src/crawlee/storage_clients/_file_system/_dataset_client.py
```diff
@@ -0,0 +1,483 @@
...
+                    name=name,
+                    created_at=now,
+                    accessed_at=now,
+                    modified_at=now,
+                    item_count=0,
+                    storage_dir=storage_dir,
+                )
+                await client._update_metadata()
+
+        return client
```
janbuchar, line 200: Pyright says that `client` is possibly unbound, is that so?

src/crawlee/storage_clients/_file_system/_storage_client.py
```diff
@@ -0,0 +1,66 @@
...
+from crawlee.storage_clients._base import StorageClient
+
+from ._dataset_client import FileSystemDatasetClient
+from ._key_value_store_client import FileSystemKeyValueStoreClient
+from ._request_queue_client import FileSystemRequestQueueClient
+
+
+@docs_group('Classes')
+class FileSystemStorageClient(StorageClient):
+    """File system implementation of the storage client.
```
janbuchar, line 16: Please mention that this is not safe for concurrent access (from multiple crawler processes)

janbuchar:
Awesome work @vdusek! Did you get a chance to test the SDK integration tests (with these changes https://github.com/apify/apify-sdk-python/pull/470) with updated Crawlee?

I'm asking because I'd love to avoid the situation where we need to make hotfix releases after we discover that we can't make SDK work with these changes.

docs/upgrading/upgrading_to_v1.md
```diff
@@ -0,0 +1,123 @@
...
+storage_client = MemoryStorageClient()
+
+# Either via the service locator:
+service_locator.set_storage_client(storage_client)
+
+# Or provide it directly to the crawler:
+crawler = ParselCrawler(storage_client=storage_client)
+```
+
+### Breaking changes
```
janbuchar, line 92: It'd be fair to mention that when you call for example `StorageClient.open_dataset_client()` or `DatasetClient.open()` twice with the same arguments, you may get different instances of `DatasetClient`. It's not a frequent use case by any means, but let's strive to maintain order.

src/crawlee/_request.py
vdusek: Because of the persistance of `__sequence` and `__forefront` flags.

src/crawlee/_types.py
vdusek: It's a follow-up issue, we discussed it more on Slack at the beginning of the May :slightly_smiling_face:.

src/crawlee/request_loaders/_request_loader.py
vdusek: Oh, will revert it.

src/crawlee/storage_clients/_base/_storage_client.py
vdusek: Yeah, but it returns the client (`DatasetClient`) instance, not the storage instance (`Dataset`).

src/crawlee/storage_clients/_base/_storage_client.py
vdusek: I don't think so, we can remove it.

vdusek: `StorageClient` holds a config in a private property? I don't see it.

... This is used when opening a new storage and providing config explicitly, e.g. `Dataset.open(configuration=...)` then it calls this method inside.

src/crawlee/storage_clients/_file_system/_dataset_client.py
vdusek: I tried it but always ended up with a feeling that we're going back where we were before implementing this :slightly_smiling_face:.

src/crawlee/storage_clients/_file_system/_dataset_client.py
vdusek: It can't happen. Either it is bound or an exception is raised.

src/crawlee/storage_clients/_base/_storage_client.py
janbuchar: Sorry for the confusion - I meant that if the `StorageClient` subclass wants to use something from `Configuration`, it can be made to accept it in the constructor and pass it to the individual resource client instances.

But since we already accept `Configuration` in `Dataset.open` and the like, I guess let's keep it this way. Even though, what is the use case for passing a config object to `open` methods?

src/crawlee/_request.py
janbuchar: Wouldn't a subclass of the model be more robust?

src/crawlee/storage_clients/_base/_storage_client.py
janbuchar: How about renaming them to `create_dataset_client` instead? It would have the added bonus of making it clear that there is no caching involved.

src/crawlee/storage_clients/_file_system/_dataset_client.py
janbuchar: That's all I needed to hear, let's keep it this way :slightly_smiling_face:

docs/upgrading/upgrading_to_v1.md
vdusek: Ok, I will address it.

src/crawlee/storage_clients/_base/_storage_client.py
vdusek: > But since we already accept Configuration in Dataset.open and the like, I guess let's keep it this way. Even though, what is the use case for passing a config object to open methods?

You can provide different config objects to different storage instances. If you don't provide any, it will use the global (default) one.

src/crawlee/storage_clients/_base/_storage_client.py
vdusek: Ok, that makes sense to me :slightly_smiling_face:.

src/crawlee/_request.py
vdusek: I'm afraid that would cause more problems than benefits :slightly_smiling_face:. We only work with these extra fields in the FS RQ client.

janbuchar: Right. Or you can have a separate model, run the deserialized JSON through both and return a tuple from `parse_request_file` :slightly_smiling_face: That way `Request` can stay the way it is and you get type-safe access to the extras.

src/crawlee/storage_clients/_base/_storage_client.py
janbuchar: I mean, yeah, but what do I actually want to change in the `Configuration`?

vdusek: It's currently the only way for a user to provide custom data to specific storage clients. For example, in the context of the FS client, you might want to open one storage with `purge_on_start` set to false, and another with it set to true. Similarly, when using the Apify client, you might want to open each storage with a different token - under different users (or with different scopes?), for instance.

I'm not saying this approach is ideal—it's just the current behavior, and this PR isn't changing it.

janbuchar: Well, `APIFY_TOKEN` is only present on the SDK subclass of `Configuration`, so you'd have to do some type shenanigans to use it here. So the options that make sense are:

- `purge_on_start` 
- `default_*_id` (but overriding those in an `open()` method call seems... unexpected)
- `storage_dir` - only makes sense for filesystem storage client

I just wrote this down to make sure I understand it right. I guess we don't need to change it, but it does seem to me that it creates more confusion than actual value.

vdusek: > I just wrote this down to make sure I understand it right. I guess we don't need to change it, but it does seem to me that it creates more confusion than actual value.

I've just kept the current state, and I don't have a strong preference, if we decide to remove it, I'm totally ok with that. cc @B4nan in case you would like to add something.

B4nan: If this is used for the static factory methods (e.g. `Dataset.open`) then we need to keep it, no? But this method is not a static one, right? If we already have a refrence to config that users can override in the `open` method, we dont need another one on instance level.

vdusek: It is a factory static method, let's keep it there then, thanks.

src/crawlee/storages/_utils.py
vdusek: Makes sense - storage instance manager is introduced.

src/crawlee/storages/_storage_instance_manager.py
```diff
@@ -0,0 +1,124 @@
...
+        # Create new instance
+        client = await client_opener(id=id, name=name, configuration=configuration)
+        instance = cls(client)  # type: ignore[call-arg]
+        instance_name = getattr(instance, 'name', None)
+
+        # Cache the instance
+        if cls not in self._cache_by_id:
+            self._cache_by_id[cls] = {}
+        if cls not in self._cache_by_name:
+            self._cache_by_name[cls] = {}
```
janbuchar: nit - `self._cache_by_id.setdefault(cls, {})`

src/crawlee/_request.py
vdusek: As we discussed on Slack, I removed the `__sequence` and `__forefront` extra fields from the `Request` and use `RecoverableState` instead.

vdusek:
# Benchmark

- 1000 requests to a local HTTP server.

## Crawlee Py - Old memory client

- Old memory client = memory with persitence false

All runtimes:
- Run 1: 4.280095s  
- Run 2: 4.388267s  
- Run 3: 4.322075s  
- Run 4: 4.520383s  
- Run 5: 4.222991s  
- Run 6: 4.359523s  
- Run 7: 4.182811s  
- Run 8: 4.322229s  
- Run 9: 4.072315s  
- Run 10: 4.032425s  

**Average crawler runtime: 4.270311s**

---

## Crawlee Py - Old file-system client

- Old file-system client = memory with persistence true

All runtimes:
- Run 1: 5.351646s  
- Run 2: 5.934761s  
- Run 3: 5.218038s  
- Run 4: 4.867312s  
- Run 5: 4.890084s  
- Run 6: 4.935311s  
- Run 7: 4.923271s  
- Run 8: 4.752518s  
- Run 9: 4.724725s  
- Run 10: 4.865203s  

**Average crawler runtime: 5.046287s**

---

## Crawlee Py - New memory client

All runtimes:
- Run 1: 1.582967s  
- Run 2: 1.723083s  
- Run 3: 1.539048s  
- Run 4: 1.622284s  
- Run 5: 1.802081s  
- Run 6: 1.556861s  
- Run 7: 1.436224s  
- Run 8: 1.635982s  
- Run 9: 1.633467s  
- Run 10: 1.727041s  

**Average crawler runtime: 1.625904s**

---

## Crawlee Py - New file-system client

All runtimes:
- Run 1: 4.299179s  
- Run 2: 4.576746s  
- Run 3: 4.359626s  
- Run 4: 4.305971s  
- Run 5: 4.480797s  
- Run 6: 4.511054s  
- Run 7: 4.316566s  
- Run 8: 4.503595s  
- Run 9: 4.378998s  
- Run 10: 4.427795s  

**Average crawler runtime: 4.416033s**

---

## Crawlee TS - Memory client

- Memory client = memory with persistence false

All runtimes:
- Run 1: 2.268000s  
- Run 2: 2.220000s  
- Run 3: 2.254000s  
- Run 4: 2.297000s  
- Run 5: 2.283000s  
- Run 6: 2.279000s  
- Run 7: 2.212000s  
- Run 8: 2.307000s  
- Run 9: 2.127000s  
- Run 10: 2.256000s  

**Average crawler runtime: 2.2503s**

---

## Crawlee TS - File-system client

- File-system client = memory with persistence true

All runtimes:
- Run 1: 3.446000s  
- Run 2: 3.186000s  
- Run 3: 3.390000s  
- Run 4: 3.145000s  
- Run 5: 3.102000s  
- Run 6: 3.260000s  
- Run 7: 3.178000s  
- Run 8: 3.160000s  
- Run 9: 3.369000s  
- Run 10: 3.247000s  

**Average crawler runtime: 3.2483s**

---

## Scrapy - memory*

- Scrapy provides only in-memory storage for requests.

All runtimes:
- Run 1: 1.476767s  
- Run 2: 1.451156s  
- Run 3: 1.463033s  
- Run 4: 1.521124s  
- Run 5: 1.498111s  
- Run 6: 1.494774s  
- Run 7: 1.487637s  
- Run 8: 1.479461s  
- Run 9: 1.459569s  
- Run 10: 1.447364s  

**Average crawler runtime: 1.477900s**

---

## Summary

| Configuration                       | Average Runtime (s) |
|------------------------------------|---------------------|
| **Crawlee Py - Old memory client** | 4.270311            |
| **Crawlee Py - Old file-system client** | 5.046287        |
| **Crawlee Py - New memory client** | 1.625904            |
| **Crawlee Py - New file-system client** | 4.416033        |
| **Crawlee TS - Memory client**     | 2.2503              |
| **Crawlee TS - File-system client**| 3.2483              |
| **Scrapy - memory***               | 1.477900            |

docs/guides/code_examples/storages/dataset_basic_example.py
```diff
@@ -6,7 +6,7 @@
 async def main() -> None:
     # Open the dataset, if it does not exist, it will be created.
     # Leave name empty to use the default dataset.
-    dataset = await Dataset.open()
+    dataset = await Dataset.open(name='my-dataset')
```
janbuchar, line 9: I'd argue that using the default (unnamed) dataset is the more frequent use case - why change this example?

docs/guides/storage_clients.mdx
```diff
@@ -0,0 +1,197 @@
...
+from crawlee.crawlers import ParselCrawler
+
+# Create memory storage client.
+storage_client = MemoryStorageClient()
+
+# Or pass it directly to the crawler.
+crawler = ParselCrawler(storage_client=storage_client)
+```
+
+The `MemoryStorageClient` is a good choice for testing, development, or short-lived operations where speed is more important than data persistence. It is not suitable for production use or long-running crawls, as all data will be lost when the program exits.
```
janbuchar: Also for HTTP APIs where each request should be handled with a fresh storage.

docs/guides/code_examples/storages/dataset_basic_example.py
vdusek: Unification - in the RQ basic example we used "my-request-queue", but in dataset, we didn't. 

Of course, I can unify it to the other way as well :slightly_smiling_face:.

So up to you.

janbuchar: Hm. It's an example, can we do both? :grin:

vdusek: I mean, in the comment above, the command of every one of these examples is: "Leave name empty to use the default {dataset|KVS|RQ}", so I would preferably leave it as it is :slightly_smiling_face:.

src/crawlee/storage_clients/_base/_key_value_store_client.py
janbuchar, line 1: Last minute thing :grin: could we add a `has_value` or `record_exists` method?

docs/guides/code_examples/storages/dataset_basic_example.py
vdusek: Resolving it... Feel free to open it again if you do not agree.

src/crawlee/storage_clients/_base/_key_value_store_client.py
vdusek: Added

B4nan:
end of an era 🎉

Issue #1201, opened 2025-05-19
PlaywrightCrawler silently switches POST requests to GET


phughesion-h3:
The POST request example doesn't work with PlaywrightCrawler. Instead, it silently gets switched to a GET request in the background, and the Crawlee Request object still has the original request data.

Result: `context.request` is the original Request object that you prepared (a POST request), but `context.response` is the playwright Response object that corresponds to the GET request's response, so the response you get is not actually for the request you intended to make.

Questions:
1. Is playwright capable of making POST requests?
2. This should not fail/switch silently and should scream loudly at the user that this happens under the hood.

If you switch the below example to use ParselCrawler instead, it works as expected.

```
import asyncio
from urllib.parse import urlencode

from crawlee import Request
from crawlee.crawlers import ParselCrawler, ParselCrawlingContext, PlaywrightCrawler, PlaywrightCrawlingContext
from crawlee.proxy_configuration import ProxyConfiguration

crawler = PlaywrightCrawler(
    proxy_configuration=ProxyConfiguration(
        proxy_urls=["http://localhost:8080"]
    )
)

@crawler.router.default_handler
async def default_handler(context: PlaywrightCrawlingContext) -> None:
    print("Default handler called")
    if isinstance(context, ParselCrawlingContext):
        print(f"Headers: {context.http_response.headers}")
    elif isinstance(context, PlaywrightCrawlingContext):
        print(f"Headers: {context.response.headers}")

async def main() -> None:
    # Prepare a POST request to the form endpoint.
    request = Request.from_url(
        url='http://localhost:80/login',
        method='POST',
        headers={'content-type': 'application/x-www-form-urlencoded'},
        payload=urlencode(
            {
                'username': 'admin',
                'password': 'password123'
            }
        ).encode('utf-8'),

    )

    await crawler.run([request])

if __name__ == '__main__':
    asyncio.run(main())
```

Mantisus:
> Is playwright capable of making POST requests?

No, as a browser. Only with `context.send_request`.

I checked how it works in the JS version of crawlee. The POST request is done via `route`, but this is not the recommended way.

phughesion-h3:
> No, as a browser. Only with `context.send_request`.

In that case, there is a bug because you can't send POST data (payload), because the payload arg isn't sent to the _http_client.

In _basic_crawler.py:

```
    def _prepare_send_request_function(
        self,
        session: Session | None,
        proxy_info: ProxyInfo | None,
[...]

PR #1208, opened 2025-05-21
fix: add support non-GET request for `PlaywrightCrawler`

Mantisus:
### Description

- Add support for non-GET requests for `PlaywrightCrawler` using [`Route`](https://playwright.dev/python/docs/api/class-route)

### Issues

- Closes: #1201

### DIFF
--- a/src/crawlee/crawlers/_playwright/_playwright_crawler.py
+++ b/src/crawlee/crawlers/_playwright/_playwright_crawler.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import warnings
 from functools import partial
 from typing import TYPE_CHECKING, Any, Callable, Generic, Literal, Union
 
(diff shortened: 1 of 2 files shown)


vdusek (approved):
Thanks, LGTM

src/crawlee/crawlers/_playwright/_playwright_crawler.py
```diff
@@ -210,6 +220,25 @@ async def _open_page(
                 await hook(pre_navigation_context)
         yield pre_navigation_context
 
+    def _prepare_request_handler(
```
janbuchar, line 223: `request_handler` is already taken, let's come up with a better name. How does `_prepare_non_get_request_interceptor` sound? Not great, huh... but it's a start, I guess.

src/crawlee/crawlers/_playwright/_playwright_crawler.py
```diff
@@ -210,6 +220,25 @@ async def _open_page(
...
+        self,
+        method: HttpMethod = 'GET',
+        headers: HttpHeaders | dict[str, str] | None = None,
+        payload: HttpPayload | None = None,
+    ) -> Callable:
+        """Create a route handler for Playwright to support non-GET methods with custom parameters.
```
janbuchar, line 229: I'd mention that the purpose of the interception is to add custom headers and payload to the request.

janbuchar (commented):
Cool, thanks! I onlyf found some naming and documentation issues.

janbuchar (approved):

Issue #1204, opened 2025-05-20
Default crawlee[cli] template project broken on Windows


solenum:
This could just be a misunderstanding on my part, first time using crawlee.  I followed the guide on the homepage and used pipx to create a project 

```pwsh
pipx run 'crawlee[cli]' create my-crawler
```

This produces a project that fails to run after I activate the venv in powershell and attempt to run the project

```pwsh
cd my-crawler
.\.venv\Scripts\activate
python -m my-crawler
```

```pwsh
[asyncio] ERROR Task exception was never retrieved
future: <Task finished name='Task-12' coro=<Connection.run() done, defined at C:\Users\redacted\Documents\Git\Work\redacted\redacted\.venv\lib\site-packages\playwright\_impl\_connection.py:272> exception=NotImplementedError()>
      Traceback (most recent call last):
        File "C:\Users\redacted\Documents\Git\Work\redacted\redacted\.venv\lib\site-packages\playwright\_impl\_connection.py", line 279, in run
          await self._transport.connect()
        File "C:\Users\redacted\Documents\Git\Work\redacted\redacted\.venv\lib\site-packages\playwright\_impl\_transport.py", line 133, in connect
          raise exc
        File "C:\Users\redacted\Documents\Git\Work\redacted\redacted\.venv\lib\site-packages\playwright\_impl\_transport.py", line 120, in connect
          self._proc = await asyncio.create_subprocess_exec(
        File "C:\Python310\lib\asyncio\subprocess.py", line 218, in create_subprocess_exec
          transport, protocol = await loop.subprocess_exec(
        File "C:\Python310\lib\asyncio\base_events.py", line 1667, in subprocess_exec
          transport = await self._make_subprocess_transport(
        File "C:\Python310\lib\asyncio\base_events.py", line 498, in _make_subprocess_transport
          raise NotImplementedError
      NotImplementedError
```

Annoyingly this took me ages to figure out.  It seems to be related to this line in the \__main__.py that gets generated, as commenting it out resolves the problem:

```python
    if platform.system() == 'Windows':
        # This mitigates a warning raised by curl-cffi. If you do not need to use curl-impersonate, you may remove this.
        asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
        # __comment out ^ to fix__
```

[...]

PR #1209, opened 2025-05-21
fix: set `WindowsSelectorEventLoopPolicy` only for curl-impersonate template without `playwright`

Mantisus:
### Description

- Sets `WindowsSelectorEventLoopPolicy` only for templates that use curl-impersonate, provided that playwright is not used. Since curl-impersonate does not support `ProactorEventLoop` used in Windows.

For all other cases, `ProactorEventLoop` is recommended for Windows.

### Issues

- Closes: #1204


janbuchar (approved):

janbuchar:
Did you have a chance to test this on Windows @Mantisus?

vdusek (approved):
LGTM

PR #1236, opened 2025-06-10
chore: fix `PytestCollectionWarning` by adding `__test__=False`

vdusek:
```
PytestCollectionWarning: cannot collect test class 'TestInput' because it has a __init__ constructor
  @dataclass
```


janbuchar:
Underscoring the name might be too harsh :slightly_smiling_face: We just need the class name to not start with `Test`. Or we can put `__test__ = False` inside it, apparently.

vdusek:
We already use `_TestInput` on some places, so I followed that, but I can use another approach, maybe just rephrasing to `InputTest`? :smile:

janbuchar:
Dunno. It's a test input. Or an input for tests :grin: Maybe using the `__test__ = False` flag is the least annoying option after all.

janbuchar (approved):
LGTM, if it does remove the warnings :grin:

PR #1239, opened 2025-06-10
chore: fix `UserWarning` `pkg_resources` is deprecated

vdusek:
```
UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  import pkg_resources
```

Issue: https://github.com/theY4Kman/pytest-only/pull/15

`pytest-only` is not maintained anymore, so I am replacing it with a fork.


janbuchar:
Can we just... not use `pytest-only`? If I understand it correctly, it's something that you only use if you're debugging a test and don't want to use `pytest -k` or a file filter or a button in your IDE.

vdusek:
> Can we just... not use pytest-only? If I understand it correctly, it's something that you only use if you're debugging a test and don't want to use pytest -k or a file filter or a button in your IDE.

I'm used to it - you can mark multiple tests and then run `make unit-tests` with predefined execution as usual. Also, the `-k` flag doesn't support the enumeration of multiple values (you can't do `-k test_asd -k test_qwer`). So if you want to run a subset of tests defined by enumeration, this is a good option.

With that in mind, I'll leave it there and take responsibility for any maintenance overhead it brings.

PR #1241, opened 2025-06-10
feat: Add `stream` method for `HttpClient`

Mantisus:
### Description

- Add `stream` method for `HttpClient`
- Add an async context manager for cleaning up resources when closing a `HttpClient`

Relates: #1169


Mantisus:
The addition of the `stream` method is necessary, to use the `HttpClient` instance utility in `Sitemap` instead of `httpx`.

https://github.com/apify/crawlee-python/pull/1169#discussion_r2134158694

src/crawlee/crawlers/_playwright/_types.py
```diff
@@ -42,6 +44,11 @@ class PlaywrightHttpResponse:
     def read(self) -> bytes:
         return self._content
 
+    async def iter_bytes(self, chunk_size: int | None = None) -> AsyncGenerator[bytes, None]:
```
janbuchar: Please add a comment that makes it clear that we don't actually stream anything here, intentionally.

src/crawlee/http_clients/_base.py
```diff
@@ -33,6 +38,9 @@ def headers(self) -> HttpHeaders:
     def read(self) -> bytes:
         """Read the content of the response body."""
 
+    def iter_bytes(self, chunk_size: int | None = None) -> AsyncIterator[bytes]:
```
janbuchar: What is the default (`chunk_size = None`) behavior? Can you document it please?

Also, will this fit in with the streaming sitemap loader? I'm honestly not sure and too lazy to check, just making sure :grin:

tests/unit/http_clients/test_curl_impersonate.py
```diff
@@ -116,3 +116,21 @@ async def test_send_request_allow_redirects_false(server_url: URL) -> None:
...
+    <title>Hello, world!</title>
+</head>
+<body>
+</body></html>"""
+
+    content_body: bytes = b''
+
+    async with http_client.stream(str(server_url)) as response:
+        assert response.status_code == 200
+        async for chunk in response.iter_bytes():
```
janbuchar: I'm not sure what the behavior is if you do `response.iter_bytes()` outside of a stream context (regular send_request/crawl). I'd prefer if it worked as usual, but in any way, there should be a test for both http clients that captures the behavior.

janbuchar (commented):
Nice, that seems pretty smooth! I have a couple of nits/questions.

src/crawlee/http_clients/_base.py
Mantisus: > What is the default (chunk_size = None) behavior? Can you document it please?

That's a good question 😁

I looked at the `httpx` and `curl_cffi` code a little deeper.

With `None` it depends on protocol, transport and server behavior.

To support `chunk_size`, `httpx` uses an internal buffer that either cuts or accumulates data to the right size. In `curl_cffi` this parameter is ignored as reported by `warning`

According to this, I think we can just remove this parameter and use the default behavior.

> Also, will this fit in with the streaming sitemap loader? I'm honestly not sure and too lazy to check, just making sure 😁

Well, all the tests from Sitemap PR pass for both clients.
And `SitemapRequestLoader` with `HttpCrawler` works on both clients for 'https://crawlee.dev/sitemap.xml'.

So I would say that, at the very least, it works

Mantisus:
After further testing, and because of possible issues related to the sharing of `read` and `iter_bytes` as noted by @janbuchar. I renamed `iter_bytes` to `read_stream`, which should be more transparent to users and clearer to use.

src/crawlee/http_clients/_base.py
```diff
@@ -119,3 +142,71 @@ async def send_request(
...
+        """Deinitialize the client and clean up resources when exiting the context manager.
+
+        Raises:
+            RuntimeError: If the context manager is already active.
+        """
+        if not self._active:
+            raise RuntimeError(f'The {self.__class__.__name__} is not active.')
+
+        await self.cleanup()
+        self._active = False
```
vdusek, line 221: So we use the `_active` flag only for the detection of multiple aenter/aexit calls?

If so, I remember we had some discussion a while ago with either @janbuchar or @Pijukatel on how the context managers should behave in that case. Do you guys remember?

Mantisus: Thank you for pointing that out. I added this flag to match the logic in `basic_crawler` - https://github.com/apify/crawlee-python/blob/master/src/crawlee/crawlers/_basic/_basic_crawler.py#L675

But I completely forgot about the public `property`.

Although you may have discussed some other details )

src/crawlee/crawlers/_playwright/_types.py
```diff
@@ -42,6 +44,11 @@ class PlaywrightHttpResponse:
     def read(self) -> bytes:
         return self._content
 
+    async def read_stream(self) -> AsyncGenerator[bytes, None]:
+        # Playwright not support `streaming` responses.
```
janbuchar:
```suggestion
        # Playwright does not support `streaming` responses.
```

tests/unit/http_clients/test_httpx.py
```diff
@@ -128,3 +128,65 @@ async def test_crawl_follow_redirects_false(server_url: URL) -> None:
...
+</body></html>"""
+
+    async with http_client.stream(str(server_url)) as response:
+        assert response.status_code == 200
+        content_body_first: bytes = b''
+        async for chunk in response.read_stream():
+            content_body_first += chunk
+
+        content_body_second: bytes = b''
+        async for chunk in response.read_stream():
```
janbuchar: This is a very common source of bugs - maybe we could raise an error here instead (fail fast)?

tests/unit/http_clients/test_curl_impersonate.py
```diff
@@ -116,3 +116,65 @@ async def test_send_request_allow_redirects_false(server_url: URL) -> None:
...
+
+        with pytest.raises(RuntimeError):
+            response.read()
+
+
+async def test_send_request_error_for_read_stream(http_client: CurlImpersonateHttpClient, server_url: URL) -> None:
+    response = await http_client.send_request(str(server_url))
+
+    assert response.status_code == 200
+    assert b''.join([item async for item in response.read_stream()]) == b''
```
janbuchar: Huh, I'm kinda confused here - the test is called `error_for_read_stream`, but there is no error involved. Also, `send_request` fetches the response eagerly, what's stopping us from returning it from `read_stream`?

src/crawlee/http_clients/_base.py
Pijukatel: We have 3 options. **Re-entrant , re-usable, single-use**:
https://docs.python.org/3/library/contextlib.html#single-use-reusable-and-reentrant-context-managers

Re-entrant does not make sense here in my opinion. 

So then second question. Does it make sense and is it possible to re-use context that was already left before? If yes, then it is re-usable. (Based on the flags, this is the setup in your change. But is it really possible to enter closed context again and everything will work as expected?)

If not then it is single use and we should probably throw some error if someone tries to enter it again. 
Based on the code in the cleanup, it might not be possible to re-use already closed client (unless `__aenter__` is implemented in a away to make the client ready again).

tests/unit/http_clients/test_curl_impersonate.py
Mantisus: > the test is called error_for_read_stream

I apologize, I had originally planned to call an error, but abandoned that idea without changing the name of the test.

> Also, send_request fetches the response eagerly, what's stopping us from returning it from read_stream?

A request made with `send_request` has the same state as `stream` after the response has been fully read. The `is_stream_consumed` and ‘is_closed’ flags are True.

Since this is a method designed to read from `stream`, I focused on what state the stream should have for this call. Since in `send_request` the stream is already subtracted at this point, this returns an empty chunk (or error after https://github.com/apify/crawlee-python/pull/1241#discussion_r2142888545).

tests/unit/http_clients/test_httpx.py
Mantisus: Makes sense and it would be closer to `httpx` behavior

tests/unit/http_clients/test_curl_impersonate.py
janbuchar: Let's look at this from a user perspective. Perhaps it would be best if we just raised an error along the lines of "If you want to stream, you have to use the stream method" in this case. Better be strict and stop the user immediately when we detect they're doing something that's likely wrong than try to guess what they wanted to do.

Mantisus: Updated with Error

src/crawlee/http_clients/_httpx.py
```diff
@@ -44,8 +47,17 @@ def headers(self) -> HttpHeaders:
         return HttpHeaders(dict(self._response.headers))
 
     def read(self) -> bytes:
+        if not self._response.is_closed:
+            raise RuntimeError('Use `read_stream` to read the body of the Response received from the `stream` method')
         return self._response.read()
 
+    async def read_stream(self) -> AsyncIterator[bytes]:
```
janbuchar, line 54: I'm tempted to ask what's going to happen if somebody tries to iterate over the same stream twice "in parallel". On the other hand, it's hard enough to do that on purpose, so maybe we can ignore that case.

janbuchar (approved):
LGTM

src/crawlee/http_clients/_base.py
Mantisus: Thank you for the detailed explanation.

Yes, `HttpClient` should be `reuse`. Fixed and added the tests

PR #1243, opened 2025-06-11
chore: Reduce flakiness of `test_timeout_in_handler` in CI

Pijukatel:
### Description

Increase the timeout in the `test_timeout_in_handler` to deal with overloaded CI executor that can slow down the test execution and make it fail.

### DIFF
--- a/tests/unit/crawlers/_basic/test_basic_crawler.py
+++ b/tests/unit/crawlers/_basic/test_basic_crawler.py
@@ -1180,8 +1180,10 @@ async def test_timeout_in_handler(sleep_type: str) -> None:
 
     handler_timeout = timedelta(seconds=1)
     max_request_retries = 3
-    double_handler_timeout_s = handler_timeout.total_seconds() * 2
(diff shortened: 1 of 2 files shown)


janbuchar (approved):
Honestly, this is one of the tests that could benefit from some event-loop mocking that would make sleeps instant (something like https://vitest.dev/guide/mocking.html#timers). But take my "Approve" and go in peace (for now :grin:)

vdusek (commented):
I tried it in https://github.com/apify/crawlee-python/pull/1194, and it didn't help. 

I would suggest either skipping this test for now and creating a follow-up issue, or - as we discussed on Slack - since the test passes when run in isolation (I ran it 100 times and it passed every time), let's look into running it in isolation somehow.

Pijukatel:
> Honestly, this is one of the tests that could benefit from some event-loop mocking that would make sleeps instant (something like https://vitest.dev/guide/mocking.html#timers). But take my "Approve" and go in peace (for now 😁)

Yes, that would be the best. Unfortunately I could not find any well maintained existing solution for this in Python :-(

PR #1245, opened 2025-06-12
chore: Fix incremental test coverage

Pijukatel:
Fix incremental test coverage


Pijukatel:
> I would suggest reverting the order of test groups -> first `run_alone`, which is much smaller, and then the big one.
> 
> And also, since you're updating it, could you please replace `--verbose` with `-vv` (everywhere) - to have more verbosity in CI - I found it quite useful recently.

Ok

PR #1251, opened 2025-06-16
fix: Do not raise an error to check 'same-domain' if there is no hostname in the url

Mantisus:
### Description

- The need for PR is due to the fact that now the enqueue strategy check, is used in `extract_links` and can be applied to invalid urls.

### DIFF
--- a/src/crawlee/crawlers/_basic/_basic_crawler.py
+++ b/src/crawlee/crawlers/_basic/_basic_crawler.py
@@ -941,12 +941,25 @@ class BasicCrawler(Generic[TCrawlingContext, TStatisticsState]):
         """Filter requests based on the enqueue strategy and URL patterns."""
         limit = kwargs.get('limit')
         parsed_origin_url = urlparse(origin_url)
+        strategy = kwargs.get('strategy', 'all')
+
+        if strategy == 'all' and not parsed_origin_url.hostname:
+            self.log.warning(f'Skipping enqueue: Missing hostname in origin_url = {origin_url}.')
+            return
+
+        # Emit a `warning` message to the log, only once per call
+        warning_flag = True
 
         for request in request_iterator:
             target_url = request.url if isinstance(request, Request) else request
+            parsed_target_url = urlparse(target_url)
+
+            if warning_flag and strategy != 'all' and not parsed_target_url.hostname:
+                self.log.warning(f'Skipping enqueue url: Missing hostname in target_url = {target_url}.')
+                warning_flag = False
 
             if self._check_enqueue_strategy(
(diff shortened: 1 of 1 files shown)


src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -967,7 +967,7 @@ def _check_enqueue_strategy(
 
         if strategy == 'same-domain':
             if origin_url.hostname is None or target_url.hostname is None:
-                raise ValueError('Both origin and target URLs must have a hostname')
+                return False
```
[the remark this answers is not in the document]
Mantisus: Make sense

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -962,13 +962,20 @@ def _check_enqueue_strategy(
         origin_url: ParseResult,
     ) -> bool:
         """Check if a URL matches the enqueue_strategy."""
+        if strategy == 'all':
+            return True
+
+        if origin_url.hostname is None or target_url.hostname is None:
+            self.log.debug(
```
janbuchar, line 982: Debug might be a bit low here, personally I'd go for warning.

janbuchar (approved):
Seems reasonable :+1:

src/crawlee/crawlers/_basic/_basic_crawler.py
Mantisus: I think that using `warning` will be quite annoying on some large sites.

janbuchar: Do you think this is a frequent enough problem?

Mantisus: Yes, this will probably include `href` such as `mailto`, `tel`.

Which, on some aggregator sites can be a real nightmare of `warning`

janbuchar: Ah, right. Maybe one warning per enqueue_links call then? Or we can add a new reason to `on_skipped_request` and use that?

Mantisus: Added a `warning` message that will only be called once per call and saved `debug` in case users need more details to analyze.

Mantisus: Using `on_skipped_request` gives me mixed feelings. On the one hand it will give the user more control over what happens and that's a good thing. On the other hand he could just use the strategy `all` and `extract_links`.

janbuchar: Fair enough!

Issue #1256, opened 2025-06-17
Investigate and fix flakiness of test_final_statistics


Pijukatel:
It seems that `FinalStatistics.request_avg_failed_duration` can be zero in some cases during tests even though there are some failed requests.

Example fail:
```>       assert final_statistics.request_avg_failed_duration > timedelta()
E       assert datetime.timedelta(0) > datetime.timedelta(0)
E        +  where datetime.timedelta(0) = FinalStatistics(requests_finished=45, requests_failed=5, retry_histogram=[25, 16, 9], request_avg_failed_duration=datetime.timedelta(0), request_avg_finished_duration=datetime.timedelta(microseconds=7503), requests_finished_per_minute=4428, requests_failed_per_minute=491, request_total_duration=datetime.timedelta(microseconds=337613), requests_total=50, crawler_runtime=datetime.timedelta(microseconds=609809)).request_avg_failed_duration
E        +  and   datetime.timedelta(0) = timedelta()```

**Comment by Pijukatel** [comment]
This can actually happen because the `RequestProcessingRecord._last_run_at` can be the same as datetime.now() called later. Which is pretty suspicions. It can happen even in the test where there is explicit 0.001s sleep, so at least in that test it should not be possible even in theory.

**Comment by Pijukatel** [comment]
Seems to be happening on Windows. It is possible that system clock on Windows has worse resolution and so this would yield same time in some cases (depending in the sampling luck)

```
  t1 = datetime.now(tz=timezone.utc)
  time.sleep(0.001)
  t2 = datetime.now(tz=timezone.utc)
```

Pijukatel:
Using `time.perf_counter()` confirms the assumption. While the `datetime.now()` can still yield the same time, `time.perf_counter()` with better resolution can detect the change.

PR #1260, opened 2025-06-19
fix: Use `perf_counter_ns` for request duration tracking

Pijukatel:
### Description

Replace `datetime.now` by `time.perf_counter_ns` in time difference calculations for request duration tracking to avoid possible issues  connected to measured time resolution.
This fixes the flakiness of `test_final_statistics` that was happening on Windows and Python versions < 3.13.

### Issues

- Closes: #1256

### Testing

Newly added test was stress tested 500x times on old implementation to see the test sensitivity. On the source code before this PR the new test failed for Windows and Python versions 3.9,3.10,3.11,3.12 all 500 times. It did not fail even once for Linux or Windows  + Python 3.13.

### DIFF
--- a/src/crawlee/statistics/_statistics.py
+++ b/src/crawlee/statistics/_statistics.py
@@ -43,7 +43,7 @@ class RequestProcessingRecord:
         if self._last_run_at_ns is None:
             raise RuntimeError('Invalid state')
 
-        self.duration = timedelta(microseconds=(time.perf_counter_ns() - self._last_run_at_ns) / 1000)
+        self.duration = timedelta(microseconds=math.ceil((time.perf_counter_ns() - self._last_run_at_ns) / 1000))
         return self.duration
 
     @property
--- a/src/crawlee/statistics/_statistics.py
+++ b/src/crawlee/statistics/_statistics.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import math
+import time
 from datetime import datetime, timedelta, timezone
 from logging import Logger, getLogger
 from typing import TYPE_CHECKING, Generic, Literal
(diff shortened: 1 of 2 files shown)


tests/unit/_statistics/test_request_processing_record.py
```diff
@@ -0,0 +1,17 @@
+from datetime import timedelta
+
+from crawlee.statistics._statistics import RequestProcessingRecord
+
+
+def test_tracking_time_resolution() -> None:
+    """Test that `RequestProcessingRecord` tracks time with sufficient resolution
```
vdusek:
```suggestion
    """Test that `RequestProcessingRecord` tracks time with sufficient resolution.
```

src/crawlee/statistics/_statistics.py
```diff
@@ -27,22 +28,22 @@ class RequestProcessingRecord:
     """Tracks information about the processing of a request."""
 
     def __init__(self) -> None:
-        self._last_run_at: datetime | None = None
```
vdusek, line 30: couldn't we stay with `datetime`?

Pijukatel: Unfortunately not. On Windows and Python versions 3.9,3.10,3.11 and 3.12 it uses somewhat low resolution time.
And converting `time.perf_counter_ns()` to `datetime` object is not really an option as `perf_counter_ns` is a relative value without any reference point and is suitable only to calculate difference - `timedelta` in our case.

vdusek: Ok, so we can use timedelta then, right?

Pijukatel: I do not get your point. `timedelta` is difference and requires two points in time and their subtraction to be constructed. First we record one point and as soon as we have the other point we create `timedelta` object from the difference and return it.

vdusek: You can use `timedelta` to represent a single time value (and we're doing it in many places).

The only issue is that it does not have nanoseconds precision:

```python 
from time import perf_counter_ns
from datetime import time

timedelta(microseconds=perf_counter_ns() / 1_000)
>>> datetime.timedelta(seconds=54450, microseconds=41707)
```

Pijukatel: I think it makes it more confusing. We also need to construct additional `timedelta` object and we loose the precision even before the subtraction. I do not see any advantage in this.
(It is also internal variable, so there is not even risk of someone using wrong units.)

vdusek: I have no issue with using integers, though I found timedelta to be the preferable option. But the loss of precision is, of course, bad. @janbuchar any thoughts?

janbuchar: As @Pijukatel wrote, it's an internal thing and it's completely justified (precision). So I'm fine with it.

janbuchar (approved):

Issue #96, opened 2024-04-09
BasicCrawler status logging


janbuchar:
- configurable interval
- configurable status message callback \(constructor parameter, property or decorator?\)
- we periodically set the crawler status via storage client
- in javascript crawlee, this does nothing when MemoryStorage is being used

vdusek:
> in javascript crawlee, this does nothing when MemoryStorage is being used

What's the reason behind this? I believe this could be useful for local execution as well.

B4nan:
We also log the message on the basic crawler level (to the debug level, since it's executed quite often, every 10 seconds iirc). What is ignored is the API call which this method represents, since there is no alternative to that when running locally.

https://github.com/apify/crawlee/blob/ad6fcd4a6c7ddd59afb48e763c0523992d263b54/packages/basic-crawler/src/internals/basic-crawler.ts#L797-L799

janbuchar:
I remember me and @tobice as well being confused by the "set status message" method being a part of the `StorageClient` - it has nothing to do with storage, except that it invokes an Apify API call - I suppose that is why it got slapped on top of that.

Could the crawler emit a "crawler status" event via `EventManager` instead? Then the SDK could hook into that and set the status message using its apify client instance.

B4nan:
Yeah that sounds good to me.

PR #1265, opened 2025-06-24
feat: add periodic status logging and `status_message_callback` parameter for customization

Mantisus:
### Description

- Add periodic logging of the current crawler status
- Add `status_message_logging_interval` parameter to configure the logging interval
- Add `status_message_callback` parameter to allow custom status message handling
- Add a new `CRAWLER_STATUS` `Event` that emits status messages through `EventManager`

### Issues

- Closes: #96

### DIFF
--- a/src/crawlee/_log_config.py
+++ b/src/crawlee/_log_config.py
@@ -4,7 +4,7 @@ import json
 import logging
 import sys
 import textwrap
-from typing import Any
+from typing import Any, Literal
 
 from colorama import Fore, Style, just_fix_windows_console
 from typing_extensions import assert_never
@@ -34,22 +34,27 @@ _LOG_LEVEL_SHORT_ALIAS = {
 _LOG_MESSAGE_INDENT = ' ' * 6
 
 
+def string_to_log_level(level: Literal['DEBUG', 'INFO', 'WARNING', 'ERROR', 'CRITICAL']) -> int:
+    """Convert a string representation of a log level to an integer log level."""
+    if level == 'DEBUG':
+        return logging.DEBUG
+    if level == 'INFO':
+        return logging.INFO
+    if level == 'WARNING':
+        return logging.WARNING
+    if level == 'ERROR':
+        return logging.ERROR
+    if level == 'CRITICAL':
+        return logging.CRITICAL
+
+    assert_never(level)
+
+
 def get_configured_log_level() -> int:
     config = service_locator.get_configuration()
 
     if 'log_level' in config.model_fields_set:
-        if config.log_level == 'DEBUG':
-            return logging.DEBUG
-        if config.log_level == 'INFO':
-            return logging.INFO
-        if config.log_level == 'WARNING':
-            return logging.WARNING
-        if config.log_level == 'ERROR':
-            return logging.ERROR
-        if config.log_level == 'CRITICAL':
-            return logging.CRITICAL
-
-        assert_never(config.log_level)
+        return string_to_log_level(config.log_level)
 
     if sys.flags.dev_mode:
         return logging.DEBUG
--- a/src/crawlee/crawlers/_basic/_basic_crawler.py
+++ b/src/crawlee/crawlers/_basic/_basic_crawler.py
@@ -25,7 +25,7 @@ from yarl import URL
 
 from crawlee import EnqueueStrategy, Glob, service_locator
 from crawlee._autoscaling import AutoscaledPool, Snapshotter, SystemStatus
-from crawlee._log_config import configure_logger, get_configured_log_level
+from crawlee._log_config import configure_logger, get_configured_log_level, string_to_log_level
 from crawlee._request import Request, RequestState
 from crawlee._types import (
     BasicCrawlingContext,
@@ -50,7 +50,7 @@ from crawlee.errors import (
     SessionError,
     UserDefinedErrorHandlerError,
 )
-from crawlee.events._types import Event
+from crawlee.events._types import Event, EventCrawlerStatusData
 from crawlee.http_clients import HttpxHttpClient
 from crawlee.router import Router
 from crawlee.sessions import SessionPool
@@ -172,6 +172,13 @@ class _BasicCrawlerOptions(TypedDict):
     """If set to `True`, the crawler will automatically try to fetch the robots.txt file for each domain,
     and skip those that are not allowed. This also prevents disallowed URLs to be added via `EnqueueLinksFunction`."""
 
+    status_message_logging_interval: NotRequired[timedelta]
+    """Interval for logging the crawler status messages."""
+
+    status_message_callback: NotRequired[Callable[[StatisticsState, StatisticsState | None, str], None]]
+    """Allows overriding the default status message. The callback needs to call `crawler.setStatusMessage()` explicitly.
+    The default status message is provided in the parameters."""
+
 
 class _BasicCrawlerOptionsGeneric(Generic[TCrawlingContext, TStatisticsState], TypedDict):
     """Generic options the `BasicCrawler` constructor."""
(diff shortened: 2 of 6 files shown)


src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -189,6 +191,15 @@ class _BasicCrawlerOptions(TypedDict):
...
     and skip those that are not allowed. This also prevents disallowed URLs to be added via `EnqueueLinksFunction`."""
 
+    status_message_logging_interval: NotRequired[timedelta]
+    """Interval for logging the crawler status messages."""
+
+    status_message_callback: NotRequired[
+        Callable[[BasicCrawler[TCrawlingContext, TStatisticsState], StatisticsState, StatisticsState | None, str], None]
+    ]
+    """Allows overriding the default status message. The callback needs to call `crawler.set_status_message()`
+    explicitly. The default status message is provided in the parameters."""
```
janbuchar: What was the reasoning behind this function's signature? Why does it have to call `crawler.set_status_message()` instead of just returning the new status message? Did you mean that you can delegate to the default logging functionality this way, if you want?

Also, I'm not sure why it should receive the crawler as the first parameter, especially when the parameter is typed as `BasicCrawler` (it might be missing some useful stuff from the child classes). The user can easily access the crawler in the outer scope or use `partial` if they need a reusable `status_message_callback`.

Last thing - maybe the callback could also be `async`?

src/crawlee/_log_config.py
```diff
@@ -34,22 +34,27 @@
 _LOG_MESSAGE_INDENT = ' ' * 6
 
 
+def string_to_log_level(level: Literal['DEBUG', 'INFO', 'WARNING', 'ERROR', 'CRITICAL']) -> int:
```
janbuchar: I guess you could make a type alias for that literal while you're at it :slightly_smiling_face:

src/crawlee/events/_types.py
```diff
@@ -79,7 +82,23 @@ class EventExitData(BaseModel):
...
 
 
-EventData = Union[EventPersistStateData, EventSystemInfoData, EventMigratingData, EventAbortingData, EventExitData]
+@docs_group('Event payloads')
+class EventCrawlerStatusData(BaseModel):
+    """Data for the crawler status event."""
+
+    model_config = ConfigDict(populate_by_name=True)
+
+    message: str
```
janbuchar, line 91: May I suggest adding something to distinguish crawlers for cases where there are multiple instances in a single run? If there's nothing more meaningful, `id(crawler)` should be fine.

src/crawlee/crawlers/_basic/_basic_crawler.py
Mantisus: > What was the reasoning behind this function's signature? Why does it have to call crawler.set_status_message() instead of just returning the new status message? Did you mean that you can delegate to the default logging functionality this way, if you want?

I've been following the Crawlee TS [signature](https://github.com/apify/crawlee/blob/ad6fcd4a6c7ddd59afb48e763c0523992d263b54/packages/basic-crawler/src/internals/basic-crawler.ts#L847) and [docs](https://github.com/apify/crawlee/blob/ad6fcd4a6c7ddd59afb48e763c0523992d263b54/packages/basic-crawler/src/internals/basic-crawler.ts#L318). However, returning a new message from the callback would indeed be cleaner. I like that idea

> Also, I'm not sure why it should receive the crawler as the first parameter, especially when the parameter is typed as BasicCrawler (it might be missing some useful stuff from the child classes). The user can easily access the crawler in the outer scope or use partial if they need a reusable status_message_callback.

Makes sense.

> Last thing - maybe the callback could also be async?

Makes sense. I was going from the idea that there is no need for an asynchronous call to call the logger. But if the user decides to do something else, the async function will give them more options.

janbuchar: > > What was the reasoning behind this function's signature? Why does it have to call crawler.set_status_message() instead of just returning the new status message? Did you mean that you can delegate to the default logging functionality this way, if you want?
> 
> I've been following the Crawlee TS [signature](https://github.com/apify/crawlee/blob/ad6fcd4a6c7ddd59afb48e763c0523992d263b54/packages/basic-crawler/src/internals/basic-crawler.ts#L847) and [docs](https://github.com/apify/crawlee/blob/ad6fcd4a6c7ddd59afb48e763c0523992d263b54/packages/basic-crawler/src/internals/basic-crawler.ts#L318). However, returning a new message from the callback would indeed be cleaner. I like that idea

Please, also update the docstring to clear out how `crawler.set_status_message` fits into this.

> > Last thing - maybe the callback could also be async?
> 
> Makes sense. I was going from the idea that there is no need for an asynchronous call to call the logger. But if the user decides to do something else, the async function will give them more options.

Setting the status message via Apify Client is probably async, for example :slightly_smiling_face: Even though you'll usually call that from an event handler...

Mantisus: > Please, also update the docstring to clear out how crawler.set_status_message fits into this.

Perhaps the idea of just returning a new message would be more user-friendly than using `crawler.set_status_message`.

janbuchar: And when you return the message, what happens? I assume `BasicCrawler` will emit an event with the message. Will it also call `crawler.set_status_message` for the user?

Mantisus: Yes, we will call `crawler.set_status_message`, which will log a result message with an `INFO` level, and use that message to `emit`. I think this approach would be cleaner for the user than suggesting that the user use `crawler.set_status_message` in the callback. But maybe there's some use case I'm missing

janbuchar: Well, this way, it will be hard to avoid logging the status message. I don't think that's an actual problem though.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1522,3 +1548,53 @@ async def _find_txt_file_for_url(self, url: str) -> RobotsTxtFile:
             url: The URL whose domain will be used to locate and fetch the corresponding robots.txt file.
         """
         return await RobotsTxtFile.find(url, self._http_client)
+
+    def set_status_message(self, message: str, level: LogLevel = 'DEBUG') -> None:
```
janbuchar: Now that the `status_message_callback` doesn't call this, shouldn't we make it protected/private? Also, maybe we could rename it to something like `log_status_message` - there is not "setting" going on.

Mantisus: Yes. That makes sense.

At first I thought about keeping the original ability to call ‘crawler.set_status_message’. But realized that in this case it's more intuitive for the user to use ‘crawler.log’.

janbuchar (approved):

src/crawlee/events/_types.py
```diff
@@ -79,7 +82,25 @@ class EventExitData(BaseModel):
...
+
+    crawler_id: int
```
vdusek: Field docstrings please.

tests/unit/crawlers/_basic/test_basic_crawler.py
```diff
@@ -1417,3 +1419,61 @@ async def handler(context: BasicCrawlingContext) -> None:
...
+
+    crawler = BasicCrawler(
+        status_message_callback=status_callback, status_message_logging_interval=timedelta(seconds=0.01)
+    )
+
+    @crawler.router.default_handler
+    async def handler(context: BasicCrawlingContext) -> None:
+        await asyncio.sleep(0.1)  # Simulate some processing time
+
+    await crawler.run(['http://a.com/'])
```
vdusek: why `'http://a.com/'` and not `server_url` fixture?

Mantisus: This is the `BasicCrawler` level, we don't make any real requests, so we don`t need `server_url`

vdusek (changes requested):
Could you please resolve conflicts?

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -322,6 +335,9 @@ def __init__(
             respect_robots_txt_file: If set to `True`, the crawler will automatically try to fetch the robots.txt file
                 for each domain, and skip those that are not allowed. This also prevents disallowed URLs to be added
                 via `EnqueueLinksFunction`
+            status_message_logging_interval:  Interval for logging the crawler status messages
```
```suggestion
            status_message_logging_interval: Interval for logging the crawler status messages.
```

tests/unit/crawlers/_basic/test_basic_crawler.py
vdusek: Still, I think `server_url` is the clearest option, alternatively, an explicit placeholder domain such as `'http://placeholder.com/'` could work.

vdusek (approved):
One last comment, otherwise LGTM

PR #1266, opened 2025-06-24
feat: Add `retire_browser_after_page_count` parameter for `BrowserPool`

Mantisus:
### Description

- Add `retire_browser_after_page_count` parameter for `BrowserPool`

### DIFF
--- a/src/crawlee/browsers/_browser_pool.py
+++ b/src/crawlee/browsers/_browser_pool.py
@@ -53,6 +53,7 @@ class BrowserPool:
         browser_inactive_threshold: timedelta = timedelta(seconds=10),
         identify_inactive_browsers_interval: timedelta = timedelta(seconds=20),
         close_inactive_browsers_interval: timedelta = timedelta(seconds=30),
+        retire_browser_after_page_count: int = 100,
     ) -> None:
         """Initialize a new instance.
 
@@ -67,7 +68,10 @@ class BrowserPool:
                 as retired.
             close_inactive_browsers_interval: The interval at which the pool checks for inactive browsers
                 and closes them. The browser is considered as inactive if it has no active pages and has been idle
-                for the specified period.
+                for the specified period. The browser is considered as retired if it has no active pages and has total
+                pages count greater than or equal to `retire_browser_after_page_count`.
+            retire_browser_after_page_count: The maximum number of processed pages after which the browser is considered
+                as retired.
         """
         self._plugins = plugins or [PlaywrightBrowserPlugin()]
         self._operation_timeout = operation_timeout
@@ -91,6 +95,7 @@ class BrowserPool:
         )
 
         self._total_pages_count = 0
+        self._retire_browser_after_page_count = retire_browser_after_page_count
         self._pages = WeakValueDictionary[str, CrawleePage]()  # Track the pages in the pool
         self._plugins_cycle = itertools.cycle(self._plugins)  # Cycle through the plugins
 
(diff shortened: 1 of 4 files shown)


Mantisus:
Parameter added following the [Crawlee TS](https://crawlee.dev/js/api/next/browser-pool/interface/BrowserPoolOptions#retireBrowserAfterPageCount).

src/crawlee/browsers/_browser_pool.py
```diff
@@ -305,6 +310,9 @@ async def _get_new_page(
         except RuntimeError as exc:
             raise RuntimeError('Browser pool is not initialized.') from exc
 
+        if browser_controller.total_pages >= self._retire_browser_after_page_count:
```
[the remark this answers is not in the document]
Mantisus: The browser will not be closed and deleted at this point; it will only be marked as `inactive`.

Actual deletion will only occur in the `_close_inactive_browsers` task, provided there are no more open pages.

src/crawlee/browsers/_playwright_browser_controller.py
```diff
@@ -74,11 +74,18 @@ def __init__(
...
+
     @property
     @override
     def pages(self) -> list[Page]:
         return self._pages
 
+    @property
+    @override
+    def total_pages(self) -> int:
+        return self._total_pages
```
janbuchar: Why can't we just use `len(self._pages)`?

tests/unit/browsers/test_browser_pool.py
```diff
@@ -160,3 +160,27 @@ async def test_with_plugin_contains_page_options(server_url: URL) -> None:
         await test_page.page.goto(str(server_url / 'user-agent'))
         assert 'My Best User-Agent' in await test_page.page.content()
         await test_page.page.close()
+
+
+@pytest.mark.parametrize(
+    ('page_count', 'equal_browsers'),
+    [
+        pytest.param(2, True, id='two pages create same browser'),
```
Pijukatel: I would suggest to rephrase the test id. Page creating browser sounds little strange. Maybe something like: "Two pages opened in the same browser" and "Each page opened in a new browser."

tests/unit/browsers/test_browser_pool.py
```diff
@@ -160,3 +160,27 @@ async def test_with_plugin_contains_page_options(server_url: URL) -> None:
         await test_page.page.goto(str(server_url / 'user-agent'))
         assert 'My Best User-Agent' in await test_page.page.content()
         await test_page.page.close()
+
+
+@pytest.mark.parametrize(
+    ('page_count', 'equal_browsers'),
```
janbuchar: I'd rename the first parameter to `retire_after_page_count` and the second parameter to something like `expect_equal_browsers`. Also maybe you could use `is` and `is not` instead of equality operator.

Also the `id` values in `pytest.param` are kinda misleading.

src/crawlee/browsers/_playwright_browser_controller.py
Mantisus: This will return the number of active pages. Currently, the `pages_count` property is responsible for this. It is expected that `total_pages` will return how many pages the browser has processed since launch. However, yes, this name is confusing; perhaps `total_pages_count` would be a better choice?

janbuchar: So inactive pages just vanish because there's no need to track them?

Mantisus: Correct. The page is deleted from `self._pages` after it is closed.

janbuchar: Cool. I'd consider renaming this to something like `total_pages_since_start` or `total_opened_pages`.

PR #1270, opened 2025-06-25
feat: Prevent overloading system memory when running locally

### LINKED ISSUE #1232: Autoscaling failsafe to prevent using up all memory when running locally
When running locally, Crawlee thinks "memory is overloaded" means "stuff that we own takes up more memory than some configured part of the total system memory" - if our calculation of the used memory fails to take something into account, we could easily smother the system.

We should have some safety mechanism that would say the system is overloaded if the system memory has something like >95% utilization (by any processes, not just current crawlee process), regardless of `CRAWLEE_AVAILABLE_MEMORY_RATIO`? shouldn't we?

_Originally posted by @janbuchar in https://github.com/apify/crawlee-python/issues/1210#issuecomment-2943875887_

janbuchar:
- closes #1232


src/crawlee/_autoscaling/_types.py
```diff
@@ -103,6 +109,11 @@ class MemorySnapshot:
     @property
     def is_overloaded(self) -> bool:
         """Indicate whether the memory is considered as overloaded."""
+        if self.system_wide_memory_size is not None and self.system_wide_used_size is not None:
+            system_wide_utilization = self.system_wide_used_size / self.system_wide_memory_size
+            if system_wide_utilization > 0.97:
```
janbuchar: It would be tricky to make this configurable. I guess I could make it a constant though.

src/crawlee/_autoscaling/snapshotter.py
```diff
@@ -273,8 +273,14 @@ def _snapshot_memory(self, event_data: EventSystemInfoData) -> None:
             max_memory_size=self._max_memory_size,
             max_used_memory_ratio=self._max_used_memory_ratio,
             created_at=event_data.memory_info.created_at,
+            system_wide_used_size=None,
+            system_wide_memory_size=None,
         )
 
+        if isinstance(memory_info := event_data.memory_info, MemoryInfo):
```
janbuchar, line 280: Some tests to cover this branch would be nice.

src/crawlee/_autoscaling/_types.py
```diff
@@ -91,9 +94,15 @@ class MemorySnapshot:
     current_size: ByteSize
     """Memory usage of the current Python process and its children."""
 
+    system_wide_used_size: ByteSize | None
```
Pijukatel, line 97: Maybe these two new values can have default `None`, I saw it is anyway initialized with `None` both in code and test

src/crawlee/_autoscaling/_types.py
```diff
@@ -103,6 +112,11 @@ class MemorySnapshot:
     @property
     def is_overloaded(self) -> bool:
         """Indicate whether the memory is considered as overloaded."""
+        if self.system_wide_memory_size is not None and self.system_wide_used_size is not None:
+            system_wide_utilization = self.system_wide_used_size / self.system_wide_memory_size
+            if system_wide_utilization > SYSTEM_WIDE_MEMORY_OVERLOAD_THRESHOLD:
+                return True
```
Pijukatel, line 118: Maybe add simple tests for this return value

vdusek (approved):
LGTM

src/crawlee/_autoscaling/_types.py
janbuchar: But I already updated all the call sites... and I'd need to reorder the fields in the dataclass :frowning_face: I'd prefer to leave it this way.

PR #1274, opened 2025-06-27
feat: Persist RequestList state

### LINKED ISSUE #99: Properly implement `RequestList`

janbuchar:
- closes #99

### DIFF
--- a/src/crawlee/request_loaders/_request_list.py
+++ b/src/crawlee/request_loaders/_request_list.py
@@ -159,6 +159,7 @@ class RequestList(RequestLoader):
 
     @override
     async def fetch_next_request(self) -> Request | None:
+        await self._get_state()
         await self._ensure_next_request()
 
         if self._next[0] is None:
@@ -185,6 +186,8 @@ class RequestList(RequestLoader):
         state.in_progress.remove(request.id)
 
     async def _ensure_next_request(self) -> None:
+        await self._get_state()
+
         if self._requests_lock is None:
             self._requests_lock = asyncio.Lock()
 


src/crawlee/_utils/recoverable_state.py
```diff
@@ -34,7 +34,7 @@ def __init__(
         *,
         default_state: TStateModel,
         persist_state_key: str,
-        persistence_enabled: bool = False,
+        persistence_enabled: Literal[True, False, 'explicit_only'] = False,
```
Pijukatel, line 37: Could you please describe how this is different from True? Probably this should also be explained in the docstring in the argument description.

src/crawlee/request_loaders/_request_list.py
```diff
@@ -1,46 +1,83 @@
...
-    from crawlee._request import Request
+logger = getLogger(__name__)
+
+
+class RequestListState(BaseModel):
+    model_config = ConfigDict(populate_by_name=True)
+
+    next_index: Annotated[int, Field(alias='nextIndex')] = 0
+    next_unique_key: Annotated[str | None, Field(alias='nextUniqueKey')] = None
+    in_progress: Annotated[set[str], Field(alias='inProgress')] = set()
```
vdusek, line 24: Are the `camelCase` aliases necessary? AFAIK I also did not use them in FS storage clients.

janbuchar: Probably not. Sessions and Statistics (other instances of recoverable state) use them too. I have no strong opinion here, if you do, say the word and I'll remove them.

janbuchar:
The request data snapshotting is pretty inefficient - it loads the whole thing into memory (same as the JS version) and stores it uncompressed in the key-value store (JS version uses gzip). 

After quite some trial and error, I believe we can use Ostrich algorithm now and optimize if it proves necessary.

src/crawlee/_utils/recoverable_state.py
```diff
@@ -101,12 +103,13 @@ async def teardown(self) -> None:
         if not self._persistence_enabled:
```
Pijukatel, line 103: These two lines are now redundant.

src/crawlee/_utils/recoverable_state.py
```diff
@@ -139,17 +157,21 @@ async def persist_state(self, event_data: EventPersistStateData | None = None) -
...
-        self._log.debug(f'Persisting state of the Statistics (event_data={event_data}).')
+        self._log.debug(
+            f'Persisting RecoverableState (model={self._default_state.__class__.__name__}, event_data={event_data}).'
+        )
 
         if self._key_value_store is None or self._state is None:
             raise RuntimeError('Recoverable state has not yet been initialized')
 
-        if self._persistence_enabled:
+        if self._persistence_enabled is True or self._persistence_enabled == 'explicit_only':
```
Pijukatel, line 167: Why this change? `if self._persistence_enabled:` seems to work the same

src/crawlee/request_loaders/_request_list.py
```diff
@@ -1,45 +1,83 @@
...
+            persist_state_key=persist_state_key or '',
+            logger=logger,
+        )
+
+        self._persist_request_data = bool(persist_requests_key)
+
+        self._requests_data = RecoverableState(
+            default_state=RequestListData(requests=[]),
+            # With request data persistence enabled, a snapshot of the requests will be done on initialization
+            persistence_enabled='explicit_only' if self._persist_request_data else False,
```
Pijukatel, line 77: So `RequestList` never has `RecoverableState` with `persistence_enabled=True`?

src/crawlee/request_loaders/_request_list.py
```diff
@@ -1,45 +1,83 @@
...
 
-if TYPE_CHECKING:
-    from crawlee._request import Request
+logger = getLogger(__name__)
+
+
+class RequestListState(BaseModel):
+    model_config = ConfigDict(populate_by_name=True)
+
+    next_index: Annotated[int, Field(alias='nextIndex')] = 0
```
Pijukatel, line 22: This deserves some description, pointing out that this is an index relative to current position and not an absolute index.

src/crawlee/request_loaders/_request_list.py
```diff
@@ -50,6 +88,44 @@ def __init__(
 
         self._requests_lock: asyncio.Lock | None = None
 
+    async def _get_state(self) -> RequestListState:
```
Pijukatel, line 91: This is a pretty complex function. After staring at it for several minutes, I am still not sure I understand it correctly. Having a descriptive docstring would be very nice, especially as understanding this requires some context knowledge.

src/crawlee/request_loaders/_request_list.py
```diff
@@ -50,6 +88,44 @@ def __init__(
...
+                        request if isinstance(request, Request) else Request.from_url(request)
+                        async for request in self._requests
+                    ]
+                    await self._requests_data.persist_state()
+
+                self._requests = self._iterate_in_threadpool(self._requests_data.current_value.requests)
+
+        for _ in range(self._state.current_value.next_index):
+            with contextlib.suppress(StopAsyncIteration):
+                await self._requests.__anext__()
```
Pijukatel: Didn't we empty self._requests in list comprehension on  line 104-107?

tests/unit/request_loaders/test_request_list.py
```diff
@@ -56,3 +57,132 @@ async def test_is_empty_does_not_depend_on_fetch_next_request() -> None:
...
+    first_request_again = await request_list_2.fetch_next_request()
+    assert first_request_again is not None
+    assert first_request_again.url == 'https://restore1.placeholder.com'
+    await request_list_2.mark_request_as_handled(first_request_again)
+
+    # Make sure that the second instance did not consume the input iterator
+    assert len(list(spy)) == 3
+
+
+async def test_requests_and_state_restoration() -> None:
```
Pijukatel, line 170: I tried extending this test by extra request and it failed.

So having these URLs

`urls = ['https://restore1.placeholder.com', 'https://restore2.placeholder.com', 'https://restore3.placeholder.com']`

and adding this

```
    third_request = await request_list_2.fetch_next_request()
    assert third_request is not None
    assert third_request.url == 'https://restore3.placeholder.com'
    await request_list_2.mark_request_as_handled(third_request)
```

Pijukatel (changes requested):
Some small comments and one more serious about the test.

src/crawlee/request_loaders/_request_list.py
janbuchar: In `self._state`, it does. In `self._requests_data`, it does not.

src/crawlee/_utils/recoverable_state.py
janbuchar: Just to make it more explicit and not require the reader to think about boolean conversions.

src/crawlee/request_loaders/_request_list.py
janbuchar: That... actually isn't what it should be - it should be an absolute index. Why do you think it's relative?

src/crawlee/request_loaders/_request_list.py
janbuchar: No, we replaced it with an iterator on line 110

Pijukatel: Ok, now I see and understand it is an absolute index. Is there any specific reason for emptying the self._requests in this way?

I think it would be more readable and probably even faster to do it on line 110 with just slicing

```
                self._requests = self._iterate_in_threadpool(
                    self._requests_data.current_value.requests[self._state.current_value.next_index:]
                )
```

src/crawlee/request_loaders/_request_list.py
Pijukatel: Ok, now I get it, considering the comment in the `_get_state` function

src/crawlee/request_loaders/_request_list.py
janbuchar: That's a good point, let me try that.

tests/unit/request_loaders/test_request_list.py
Pijukatel: I think that this can be fixed if in function `_ensure_next_request` you change the indexes:
from
`self._next[1].unique_key if self._next[1] is not None else None`
to
`self._next[0].unique_key if self._next[0] is not None else None`

But I am not sure if there is any side effect to it. All tests are passing with such change, even the modified one.

src/crawlee/request_loaders/_request_list.py
vdusek: Well, so currently somewhere we use them, and somewhere we don't - up to you then.

vdusek (approved):
LGTM

Pijukatel (approved):

Issue #1278, opened 2025-06-30
Integrate Stagehand into PlaywrightCrawler


vdusek:
- Stagehand - The AI Browser Automation Framework
- https://www.stagehand.dev/
- Python binding - https://github.com/browserbase/stagehand-python
- Explore it and integrate it - maybe in a similar way as we did for Camoufox (https://github.com/apify/crawlee-python/issues/684)

PR #1290, opened 2025-07-08
docs: Add guide about integrating Stagehand

Mantisus:
### Description

- Add guide about integrating `stagehand-python` v.0.4.0

### Issues

- Closes: #1278


Mantisus:
I had to use `cast` to avoid bloating the guide for the sake of typing.

docs/guides/code_examples/playwright_crawler_stagehand/support_classes.py
```diff
@@ -0,0 +1,60 @@
...
+        return await self._page._page.goto(  # noqa: SLF001
+            url,
+            referer=referer,
+            timeout=timeout,
+            wait_until=wait_until,
+        )
+
+    def __getattr__(self, name: str) -> Any:
+        """Delegate all other methods to the underlying StagehandPage."""
+        if name == 'goto':
```
Pijukatel: I am confused about this. Is this necessary? It will not be even called for `goto` as mentioned in the comment below.

docs/guides/code_examples/playwright_crawler_stagehand/browser_classes.py
```diff
@@ -0,0 +1,98 @@
...
+
+        self._stagehand = stagehand
+
+    @override
+    async def new_browser(self) -> StagehandBrowserController:
+        if not self._playwright:
+            raise RuntimeError('Playwright browser plugin is not initialized.')
+
+        browser = PlaywrightPersistentBrowser(
+            self._playwright.chromium, self._user_data_dir, self._browser_launch_options
```
Pijukatel: I read this https://docs.stagehand.dev/examples/customize_browser

Based on the docs, it seems that, for instance, Firefox won't work here. Maybe it would be worth adding a comment like "Stagehand can run only on a Chromium-based browser." to make it clear that selected `chromium` was not just a random pick for the sake of example code.

docs/guides/code_examples/playwright_crawler_stagehand/support_classes.py
```diff
@@ -0,0 +1,60 @@
...
+        """Enter the context manager."""
+        return self
+
+    async def __aexit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_value: BaseException | None,
+        exc_traceback: TracebackType | None,
+    ) -> None:
+        await self._page.close()
```
Pijukatel, line 48: self._page is an instance of `StagehandPage`, right? I don't think it has a close method.

I guess it is supposed to be self._page._page and this was supposed to rely on `__getattr__` to do the delegation, but since `_page` exists on `self`, the `__getattr__` will not be called.

There are different ways to handle this, but maybe you can use `self._stagehand_page` to avoid shadowing the `_page` that is supposed to be delegated and to be more explicit about which page is which.

```
class CrawleeStagehandPage:
    """StagehandPage wrapper for Crawlee."""

    def __init__(self, page: StagehandPage) -> None:
        self._stagehand_page = page
```

Pijukatel: Ok, ignore this comment. I see that `StagehandPage` has the delegation getattr as well :-)

docs/guides/code_examples/playwright_crawler_stagehand/support_classes.py
```diff
@@ -0,0 +1,60 @@
...
+        self,
+        url: str,
+        *,
+        referer: str | None = None,
+        timeout: int | None = None,
+        wait_until: str | None = None,
+    ) -> Any:
+        """Navigate to the specified URL."""
+        # Override goto to return navigation result that `PlaywrightCrawler` expects
+        return await self._page._page.goto(  # noqa: SLF001
```
Pijukatel, line 27: Isn't `StagehandPage.goto` relevant for the functionality?

Mantisus: [StagehandPage.goto](https://github.com/browserbase/stagehand-python/blob/main/stagehand/page.py#L87) returns `None` :neutral_face:

docs/guides/code_examples/playwright_crawler_stagehand/support_classes.py
Mantisus: Fair enough

docs/guides/code_examples/playwright_crawler_stagehand/support_classes.py
Pijukatel: But only `if not self._stagehand.use_api` , so that branch will not trigger by [default](https://github.com/browserbase/stagehand-python/blob/main/stagehand/config.py#L103)

I am not saying it is not correct, just that it is confusing that we are skipping  StagehandPage method and it still works :smile:

Mantisus: `use_api` is True, it signals that stagehand works with the ‘BROWSERBASE’ environment.

Since we are using local launch, this branch works for us - https://github.com/browserbase/stagehand-python/blob/main/stagehand/main.py#L174

Pijukatel: Ohhh, I see. Thanks for pointing it out. It makes more sense now.

docs/guides/playwright_crawler_stagehand.mdx
```diff
@@ -0,0 +1,66 @@
+---
```
vdusek, line 1: I suppose we can't use "Run on Apify" for these examples as it contains more than 1 file.

vdusek (commented):
Good job Max! 

The integration itself is not as easy as I expected. Maybe this could show us the direction in which we could improve/simplify the browsers/Playwright-related interface.

And/or we could introduce a dedicated crawler to this directly in Crawlee, something like `PlaywrightStagehandCrawler`.

Let's further discuss it with @B4nan and maybe @janbuchar once they're back from their vacations.

docs/guides/playwright_crawler_stagehand.mdx
Mantisus: That's right.

And in one file, it would look very cumbersome

Mantisus:
> The integration itself is not as easy as I expected. Maybe this could show us the direction in which we could improve/simplify the browsers/Playwright-related interface.

I think the integration comes out more complicated because of the current `Stagehand` API. Even though it's a wrapper around Playwright and the documentation says it's the same Playwright but with AI capabilities. The current code doesn't match that.

I hope that they will improve their API and then the guide can be simplified

docs/guides/code_examples/playwright_crawler_stagehand/browser_classes.py
```diff
@@ -0,0 +1,101 @@
...
+        # Handle page close event
+        pw_page.on(event='close', f=self._on_page_close)
+
+        # Update internal state
+        self._pages.append(pw_page)
+        self._last_page_opened_at = datetime.now(timezone.utc)
+
+        self._total_opened_pages += 1
+
+        # Wrap StagehandPage to provide Playwright Page interface
```
janbuchar, line 70: This comment seems inaccurate

docs/guides/code_examples/playwright_crawler_stagehand/browser_classes.py
```diff
@@ -0,0 +1,101 @@
...
+        pw_page = page._page  # noqa: SLF001
+
+        # Handle page close event
+        pw_page.on(event='close', f=self._on_page_close)
+
+        # Update internal state
+        self._pages.append(pw_page)
+        self._last_page_opened_at = datetime.now(timezone.utc)
+
+        self._total_opened_pages += 1
```
janbuchar, line 68: There is quite a bit of code copied over from `PlaywrightBrowserController`, isn't it? Any chance we could improve the `PlaywrightBrowserController` internal API so that integrating libraries that extend Playwright is easier?

docs/guides/code_examples/playwright_crawler_stagehand/stagehand_run.py
```diff
@@ -0,0 +1,64 @@
...
+
+        # Cast to StagehandPage for proper type hints in IDE
+        page = cast('StagehandPage', context.page)
+
+        # Use regular Playwright method
+        playwright_title = await page.title()
+        context.log.info(f'Playwright page title: {playwright_title}')
+
+        # Use AI-powered extraction with natural language
+        gemini_title = await page.extract('Extract title page')
```
janbuchar:
```suggestion
        gemini_title = await page.extract('Extract page title')
```

docs/guides/code_examples/playwright_crawler_stagehand/browser_classes.py
Mantisus: Perhaps context creation should be put into a separate public method. As well as updating states. That would make the same thing a bit cleaner.

But I would say that the main problem with this integration is that you have to do for example, this - `pw_page = page._page`.

docs/guides/code_examples/playwright_crawler_stagehand/stagehand_run.py
```diff
@@ -0,0 +1,64 @@
...
+        # Cast to StagehandPage for proper type hints in IDE
+        page = cast('StagehandPage', context.page)
+
+        # Use regular Playwright method
+        playwright_title = await page.title()
+        context.log.info(f'Playwright page title: {playwright_title}')
+
+        # Use AI-powered extraction with natural language
+        gemini_title = await page.extract('Extract page title')
+        context.log.info(f'Gemini page title: {gemini_title}')
```
vdusek, line 56: Could we highlight this?

```suggestion
        # highlight-start
        # Use AI-powered extraction with natural language
        gemini_title = await page.extract('Extract page title')
        context.log.info(f'Gemini page title: {gemini_title}')
        # highlight-end
```

src/crawlee/browsers/__init__.py
```diff
@@ -18,9 +18,7 @@
...
     from ._playwright_browser_plugin import PlaywrightBrowserPlugin
+with _try_import(__name__, 'PlaywrightPersistentBrowser'):
+    from ._playwright_browser import PlaywrightPersistentBrowser
 
-__all__ = [
-    'BrowserPool',
-    'PlaywrightBrowserController',
-    'PlaywrightBrowserPlugin',
-]
+__all__ = ['BrowserPool', 'PlaywrightBrowserController', 'PlaywrightBrowserPlugin', 'PlaywrightPersistentBrowser']
```
vdusek: It looks like this isn't just a docs‐only change - it also affects the public UI by exposing a new component. Could we split it into its own PR so that it's properly included in the changelog?

docs/guides/playwright_crawler_stagehand.mdx
```diff
@@ -0,0 +1,66 @@
...
+id: playwright-crawler-stagehand
+title: PlaywrightCrawler with Stagehand
```
vdusek: Could we please use a title that fits in the sidebar?

e.g.

```suggestion
title: Playwright with Stagehand
```

image

PR #1296, opened 2025-07-09
refactor!: Change method `HttpResponse.read` from sync to async

Mantisus:
### Description

- Makes the `read` method for `HttpResponse` asynchronous. This is more consistent in a fully asynchronous framework. More flexible than a synchronous method


vdusek (changes requested):
That makes sense, thanks, just please describe it in the upgrading guide :slightly_smiling_face:.

vdusek (approved):
LGTM

Issue #1292, opened 2025-07-08
Parametrize `HttpClient` implementations tests to remove code duplication


Pijukatel:
Currently, there are 3 different implementations of `HttpClient`. All of them have their own test file with duplicated test cases. Refactor the same tests to accept `HttpClient` (or some factory method) as test parameter to achieve something like this:

```
@pytest.mark.parametrize("http_client", [
  ImpitHttpClient(),
  CurlImpersonateHttpClient(),
  HttpxHttpClient(),
])
def some_http_client_test(http_client: HttpClient):
    ...

```

PR #1299, opened 2025-07-11
chore: Refactor `HttpClient` implementations tests

Pijukatel:
### Description

- Parametrize `HttpClient` tests.
- Add tests missing for one of the implementations.
- Delete non-unique, non-parametrized `HttpClient` tests.

### Issues

- Closes: #1292


Mantisus (approved):
Great job!

tests/unit/http_clients/test_http_clients.py
```diff
@@ -0,0 +1,375 @@
...
+@pytest.mark.parametrize(
+    'http_client',
+    [
+        pytest.param(CurlImpersonateHttpClient(), id='curl'),
+        pytest.param(HttpxHttpClient(), id='httpx'),
+        pytest.param(ImpitHttpClient(), id='impit'),
+    ],
+)
+async def test_reuse_context_manager(http_client: HttpClient, server_url: URL) -> None:
+    http_client = CurlImpersonateHttpClient()
```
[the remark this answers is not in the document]
Pijukatel: Good catch.

tests/unit/http_clients/test_http_clients.py
```diff
@@ -0,0 +1,375 @@
...
+
+    from _pytest.fixtures import SubRequest
+    from yarl import URL
+
+    from crawlee.http_clients import HttpClient
+    from crawlee.proxy_configuration import ProxyInfo
+
+
+@pytest.fixture
+async def http_client(request: SubRequest) -> AsyncGenerator[HttpClient]:
```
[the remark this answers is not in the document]
Pijukatel: Done

tests/unit/http_clients/test_http_clients.py
```diff
@@ -0,0 +1,333 @@
...
+
+_DEFAULT_CLIENTS = [
+    pytest.param(CurlImpersonateHttpClient, id='curl'),
+    pytest.param(HttpxHttpClient, id='httpx'),
+    pytest.param(ImpitHttpClient, id='impit'),
+]
+
+
+@pytest.fixture
+async def http_client(request: SubRequest) -> AsyncGenerator[HttpClient]:
```
janbuchar: Maybe you already tried and found a reason why it's not viable, but could this be made into a [parametrized fixture](https://docs.pytest.org/en/stable/how-to/fixtures.html#fixture-parametrize)? Then each test (or at least each test that uses `_DEFAULT_CLIENTS`) could just depend on `http_client` and it would get repeated for each possible value automatically.

Pijukatel: Sure, the only minor drawback is that I have to use a different variable name in tests that do not use the default Http clients. But I think it is still worth it.

tests/unit/http_clients/test_httpx.py
janbuchar, line 1: Please add a comment explaining why this one test needs to exist :slightly_smiling_face:

janbuchar (approved):
One nit, LGTM otherwise

tests/unit/http_clients/test_httpx.py
Pijukatel: I guess that each of the other clients would deserve their own header tests, but those would be different for each one of them? This refactoring exposes it, but I am not sure what the desired state is at the moment.

janbuchar: Me neither :shrug: And I don't think we need to be concerned about it right here, right now. After all, the other two HTTP clients are focused on impersonating "real" browsers, so the chances of them leaking stuff in headers this stupidly are pretty slim.

Pijukatel: Added some comment to the test

tests/unit/http_clients/test_httpx.py
```diff
@@ -24,6 +24,10 @@ async def http_client() -> AsyncGenerator[HttpClient]:
 
 
 async def test_common_headers_and_user_agent(server_url: URL, header_network: dict) -> None:
+    """Test that the relevant headers use header values from header generator instead of default Https headers.
```
janbuchar:
```suggestion
    """Test that the relevant headers use header values from header generator instead of default Httpx headers.
```

Issue #1293, opened 2025-07-08
Improve grouping in API docs


vdusek:
The current group names are not very helpful:

- Classes
- Abstract classes
- Data
- Errors
- Event
- Functions

It should be more descriptive/self-explanatory:

- Crawlers
- Storages
- Storage clients
- HTTP clients
- Error handling
- ...

PR #1309, opened 2025-07-16
docs: Improve API doc groups

vdusek:
### Description

Render of the whole page:

image

I am open to any suggestions for better grouping it. Maybe we can even hide something from "Others"?

### Issues

- Closes: #1293

### Testing

- Docs rendered locally.

### Checklist

- [x] CI passed

### DIFF
--- a/src/crawlee/_utils/docs.py
+++ b/src/crawlee/_utils/docs.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from collections.abc import Callable
 from typing import Any, Literal, TypeVar
 
+# The order of the rendered API groups is defined in the docusaurus-plugin-typedoc-api.
 GroupName = Literal[
     'Autoscaling',
     'Browser management',
@@ -15,13 +16,13 @@ GroupName = Literal[
     'Functions',
     'HTTP clients',
     'HTTP parsers',
-    'Other',
     'Request loaders',
     'Session management',
     'Statistics',
     'Storage clients',
     'Storage data',
     'Storages',
+    'Other',
 ]
 
 T = TypeVar('T', bound=Callable[..., Any])


src/crawlee/router.py
```diff
@@ -12,7 +12,7 @@
 RequestHandler = Callable[[TCrawlingContext], Awaitable[None]]
 
 
-@docs_group('Classes')
+@docs_group('Others')
 class Router(Generic[TCrawlingContext]):
```
Mantisus, line 16: I'm not sure about the `Others` group for `Router`.

But I'm not sure which group would be better

src/crawlee/_utils/docs.py
```diff
@@ -3,7 +3,26 @@
...
+    'Configuration',
+    'Crawlers',
+    'Crawling contexts',
+    'Errors',
+    'Event data',
+    'Event managers',
+    'Functions',
+    'HTTP clients',
+    'HTTP parsers',
+    'Others',
```
janbuchar: Could we use a singular (Other) here and also move it to the bottom of the class list?

Pijukatel:
Nice, this is way more readable. I was looking at `Others` trying to see some pattern and maybe two more categories could be extracted:

**Resource Management?**
`AutoscaledPool`, `ConcurencySettings`, `Snapshotter`, `SystemStatus`

**Request customization?**
`CookieParam`, `HeaderGenerator`, `SessionPool`, `HttpHeaders`, `ProxyConfiguration`???, `ProxyInfo`???, `Session`, `BrowserforgeFingerprintGenerator`, `SessionCookies`

src/crawlee/_utils/docs.py
vdusek: Updated to "Other". However, the sequence defined in `GroupName` doesn't affect the order in the rendered page.

vdusek:
Sorry, I added an outdated image to the PR description. This is the current state (so I'm not sure Pepa about your suggestions):

image

vdusek:
> Maybe we should also add a Middlewares group, currently, there is only CrawlerInstrumentor. However, I think there may be more in the future.

Hmm, maybe let's do it in the future when we have more "middlewares".

> I'm not sure about the Others group for Router. But I'm not sure which group would be better

Yeah, I'd also like to put it somewhere, but I have no idea where :D.

src/crawlee/_utils/docs.py
janbuchar: I know, but I don't know where to change this to actually change the order :grin: Could you please give it a try?

Mantisus (approved):

src/crawlee/_utils/docs.py
vdusek: Group sorting is implemented in the [docusaurus-plugin-typedoc-api](https://github.com/apify/docusaurus-plugin-typedoc-api/). @barjin will do the changes, and once that's done, we can bump the package version here.

PR #1327, opened 2025-07-28
chore: add `partitionKey` in `PlaywrightCookieParam`

Mantisus:
### Description

- add `partitionKey` in `PlaywrightCookieParam`


Mantisus:
In the latest release of Playwright, support for `partition_key` in cookies has been added.

However, Python CookieJar does not currently support `partition_key`. Therefore, this update is only necessary to avoid errors.

Pijukatel (approved):

Issue #1329, opened 2025-07-29
Memory estimation problem on MacOS


Pijukatel:
It seems that [fix: Use PSS instead of RSS to estimate children process memory usage on Linux](https://github.com/apify/crawlee-python/pull/1210) introduced bug for MacOS as the [memory_full_info](https://psutil.readthedocs.io/en/latest/index.html#psutil.Process.memory_full_info) might not be available there without some elevated privileges. 

Reported on discord: https://discord.com/channels/801163717915574323/1170038966817398894/threads/1398709287311638539

PR #1330, opened 2025-07-29
fix: Fix memory estimation not working on MacOS

Pijukatel:
### Description
- Use `pss` for memory estimation only on Linux.
- Use `rss` for memory estimation only on MacOs.

### Issues

- Closes: #1329

### DIFF
--- a/src/crawlee/_utils/system.py
+++ b/src/crawlee/_utils/system.py
@@ -1,10 +1,11 @@
 from __future__ import annotations
 
 import os
+import sys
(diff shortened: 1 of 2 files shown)


Pijukatel:
Manually tested on MacOs. Thanks @vladfrangu 

Two issues on MacOS in the original code before the PR:

- `memory_full_info` available only for self process. But privileges error on other processes.
- `pss` not available even in `memory_full_info` in MacOs

PR #1334, opened 2025-07-30
chore: Add `macos` executor for unit tests

Pijukatel:
### Description

- Run unit tests in CI also on `macOs` to prevent os OS-specific incompatible changes, for example: https://github.com/apify/crawlee-python/issues/1329


Pijukatel:
MacOS executors sometimes start with CPU usage of 100 % even before starting the test.

Pijukatel:
Looks like the [Spotlight](https://en.wikipedia.org/wiki/Spotlight_(Apple)) service on the executor can randomly start and use all the CPU. We should probably just turn it off on the executor, this service is completely useless in our context.

Pijukatel:
Executor related changes made `test_timeout_in_handler` test stable on MacOS, but `test_emit_system_info_event` i still very flaky

Issue #1333, opened 2025-07-30
Playwright cookie import fails due to unsupported 'partitionKey' and '_crHasCrossSiteAncestor' fields


enigmAsad:
When using PlaywrightCrawler, Crawlee throws TypeError when attempting to import cookies from Playwright. This happens because recent versions of Playwright (Chromium-based) include internal metadata fields in their cookie objects that are not supported by Crawlee's SessionCookies.set() method.

### **Error Messages**

**Error 1: Unsupported partitionKey**
`TypeError: SessionCookies.set() got an unexpected keyword argument 'partitionKey'`

**Stack trace:**
```
File "crawlee/sessions/_cookies.py", line 224, in set_cookies_from_playwright_format
  self.set(**cookie_param)
TypeError: SessionCookies.set() got an unexpected keyword argument 'partitionKey'
```

**Error 2: Unsupported _crHasCrossSiteAncestor**
`TypeError: SessionCookies.set() got an unexpected keyword argument '_crHasCrossSiteAncestor'`

**Stack trace:**
```
File "crawlee/sessions/_cookies.py", line 224, in set_cookies_from_playwright_format
  self.set(**cookie_param)
TypeError: SessionCookies.set() got an unexpected keyword argument '_crHasCrossSiteAncestor'
```

**### Root Cause**

Playwright sometimes includes Chromium-internal fields in cookies, such as:

1. partitionKey
2. _crHasCrossSiteAncestor

These are not part of standard web cookie specs and are not accepted by Crawlee’s cookie-setting API, causing crashes during cookie import.

**### Steps to Reproduce**

- Use PlaywrightCrawler with Chromium
- Extract cookies using Playwright (context.cookies())
- Try to import them using context.session.cookies.set_cookies_from_playwright_format(...)
- Crawlee throws TypeError on Chromium-specific fields

### ** Suggested Fix**

- Update set_cookies_from_playwright_format() to:
- Remove any key that starts with _
- Explicitly remove partitionKey (some versions don’t start this with _ )

```
for pw_cookie in pw_cookies:
    cookie_param = {k: v for k, v in self._from_playwright(pw_cookie).items() if not k.startswith('_')}
    cookie_param.pop('partitionKey', None)
    self.set(**cookie_param)
```

This prevents errors and future-proofs the code against upstream Playwright changes.

**### Environment:**

- Crawlee version: v0.6.11
- Python version: 3.12
- OS: Windows 11
- Browser: Chromium via Playwright
[...]

PR #1336, opened 2025-07-30
fix: Ignore unknown parameters passed in cookies

Mantisus:
### Description

- Ignore unknown parameters passed in cookies

### Issues

- Closes: #1333

### Testing

- Added a test that sets `Partitioned` cookies. To reproduce the behaviour that causes Playwright to throw an error


vdusek (approved):
Looks good! Thanks

Issue #1272, opened 2025-06-26
Persist the `DefaultRenderingTypePredictor` state


janbuchar:
- parity with https://github.com/apify/crawlee/pull/2987

PR #1340, opened 2025-08-04
feat: Persist `DefaultRenderingTypePredictor` state

Mantisus:
### Description

-  Persist `DefaultRenderingTypePredictor` state

### Issues

- Closes: #1272

### DIFF
--- a/src/crawlee/crawlers/_adaptive_playwright/_rendering_type_predictor.py
+++ b/src/crawlee/crawlers/_adaptive_playwright/_rendering_type_predictor.py
@@ -1,21 +1,47 @@
+from __future__ import annotations
+
 from abc import ABC, abstractmethod
 from collections import defaultdict
 from dataclasses import dataclass
 from itertools import zip_longest
+from logging import getLogger
 from statistics import mean
-from typing import Literal
+from typing import TYPE_CHECKING, Annotated, Literal
 from urllib.parse import urlparse
 
 from jaro import jaro_winkler_metric
+from pydantic import BaseModel, ConfigDict, Field, PlainSerializer, PlainValidator
 from sklearn.linear_model import LogisticRegression
 from typing_extensions import override
 
-from crawlee import Request
 from crawlee._utils.docs import docs_group
+from crawlee._utils.recoverable_state import RecoverableState
+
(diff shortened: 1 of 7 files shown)


src/crawlee/crawlers/_adaptive_playwright/_rendering_type_predictor.py
```diff
@@ -62,24 +119,55 @@ class DefaultRenderingTypePredictor(RenderingTypePredictor):
     https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html
     """
 
-    def __init__(self, detection_ratio: float = 0.1) -> None:
+    def __init__(
+        self, detection_ratio: float = 0.1, *, persistence_enabled: bool = False, persist_state_key: str | None = None
```
Pijukatel: Maybe directly set default, or is there a reason for `None`?
persist_state_key: str  = "rendering-type-predictor-state"

tests/unit/crawlers/_adaptive_playwright/test_predictor.py
```diff
@@ -23,99 +24,162 @@
...
+    ) as predictor:
+        # Learn some data
+        predictor.store_result(
+            Request.from_url(url='http://www.aaa.com/some/stuff', label='some label'), rendering_type='static'
+        )
+        before_recover_prediction = predictor.predict(
+            Request.from_url(url='http://www.aaa.com/some/stuff', label='some label')
+        )
+
+    # Recover predictor
```
Pijukatel, line 179: Could we also add to this test that another instance of `DefaultRenderingTypePredictor`  created without persistence gives a different prediction? (To avoid a situation where we would assert some default behavior of the predictor regardless of the persistence)

src/crawlee/crawlers/_adaptive_playwright/_rendering_type_predictor.py
Mantisus: Good point!

tests/unit/crawlers/_adaptive_playwright/test_predictor.py
Mantisus: It makes sense, thanks

Issue #1269, opened 2025-06-25
Persist the `SitemapRequestLoader` state


janbuchar:
- follow up to https://github.com/apify/crawlee-python/pull/1169#discussion_r2134152443

PR #1347, opened 2025-08-11
feat: Persist the `SitemapRequestLoader` state

Mantisus:
### Description

- Persist the `SitemapRequestLoader` state

### Issues

- Closes: #1269

### DIFF
--- a/docs/guides/code_examples/request_loaders/sitemap_example.py
+++ b/docs/guides/code_examples/request_loaders/sitemap_example.py
@@ -17,6 +17,7 @@ async def main() -> None:
             max_buffer_size=500,  # Buffer up to 500 URLs in memory
         )
 
+        await asyncio.sleep(0.1)  # Allow some time for the loader to start
         while request := await sitemap_loader.fetch_next_request():
             # Do something with it...
 
--- a/src/crawlee/request_loaders/_sitemap_request_loader.py
+++ b/src/crawlee/request_loaders/_sitemap_request_loader.py
@@ -1,19 +1,25 @@
 from __future__ import annotations
 
 import asyncio
+from collections import deque
 from contextlib import suppress
 from logging import getLogger
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Annotated, Any
+
+from pydantic import BaseModel, ConfigDict, Field
+from typing_extensions import override
 
 from crawlee import Request
 from crawlee._utils.docs import docs_group
 from crawlee._utils.globs import Glob
-from crawlee._utils.sitemap import ParseSitemapOptions, SitemapSource, SitemapUrl, parse_sitemap
+from crawlee._utils.recoverable_state import RecoverableState
+from crawlee._utils.sitemap import NestedSitemap, ParseSitemapOptions, SitemapSource, SitemapUrl, parse_sitemap
 from crawlee.request_loaders._request_loader import RequestLoader
 
 if TYPE_CHECKING:
     import re
     from collections.abc import Sequence
+    from types import TracebackType
 
     from crawlee.http_clients import HttpClient
     from crawlee.proxy_configuration import ProxyInfo
@@ -21,6 +27,28 @@ if TYPE_CHECKING:
 
 
 logger = getLogger(__name__)
+
+
+class SitemapRequestLoaderState(BaseModel):
+    """State for sitemap request loader persistence."""
+
+    model_config = ConfigDict(populate_by_name=True)
+
+    # URL queue and processing state
+    url_queue: Annotated[deque[str], Field(alias='urlQueue')]
+    in_progress: Annotated[set[str], Field(alias='inProgress')] = set()
+
+    # Sitemap state
+    pending_sitemap_urls: Annotated[deque[str], Field(alias='pendingSitemapUrls')]
+    in_progress_sitemap_url: Annotated[str | None, Field(alias='inProgressSitemapUrl')] = None
+    # URLs from the current sitemap that have been processed (for resuming within a sitemap)
+    current_sitemap_processed_urls: Annotated[set[str], Field(alias='currentSitemapProcessedUrls')] = set()
+    # Flag indicating if the sitemaps have been fully processed
+    completed: Annotated[bool, Field(alias='sitemapCompleted')] = False
+
+    # Counters
+    total_count: Annotated[int, Field(alias='totalCount')] = 0
+    handled_count: Annotated[int, Field(alias='handledCount')] = 0
 
 
 @docs_group('Request loaders')
@@ -40,7 +68,8 @@ class SitemapRequestLoader(RequestLoader):
         include: list[re.Pattern[Any] | Glob] | None = None,
         exclude: list[re.Pattern[Any] | Glob] | None = None,
         max_buffer_size: int = 200,
-        parse_sitemap_options: ParseSitemapOptions | None = None,
+        persist_state_key: str = 'SITEMAP_REQUEST_LOADER_STATE',
+        persist_enabled: bool = False,
     ) -> None:
         """Initialize the sitemap request loader.
 
@@ -50,27 +79,52 @@ class SitemapRequestLoader(RequestLoader):
             include: List of glob or regex patterns to include URLs.
             exclude: List of glob or regex patterns to exclude URLs.
             max_buffer_size: Maximum number of URLs to buffer in memory.
-            parse_sitemap_options: Options for parsing sitemaps, such as `SitemapSource` and `max_urls`.
             http_client: the instance of `HttpClient` to use for fetching sitemaps.
+            persist_state_key: Key in key-value store for persisting.
+            persist_enabled: Flag to enable/disable persistence.
         """
         self._http_client = http_client
-
         self._sitemap_urls = sitemap_urls
         self._include = include
         self._exclude = exclude
         self._proxy_info = proxy_info
-        self._parse_sitemap_options = parse_sitemap_options or ParseSitemapOptions()
+        self._max_buffer_size = max_buffer_size
 
-        self._handled_count = 0
-        self._total_count = 0
+        # Synchronization for queue operations
+        self._queue_has_capacity = asyncio.Event()
+        self._queue_has_capacity.set()
+        self._queue_lock = asyncio.Lock()
 
-        # URL queue and tracking
-        self._url_queue: asyncio.Queue[str] = asyncio.Queue(maxsize=max_buffer_size)
-        self._in_progress: set[str] = set()
-        self._processed_urls: set[str] = set()
+        # Initialize recoverable state
+        self._state = RecoverableState(
+            default_state=SitemapRequestLoaderState(
+                url_queue=deque(),
+                pending_sitemap_urls=deque(),
+            ),
+            persistence_enabled=persist_enabled,
+            persist_state_key=persist_state_key,
+            logger=logger,
+        )
 
-        # Loading state
+        # Start background loading
         self._loading_task = asyncio.create_task(self._load_sitemaps())
+
+    async def _get_state(self) -> SitemapRequestLoaderState:
+        """Initialize and return the current state."""
+        if self._state.is_initialized:
+            return self._state.current_value
+
+        await self._state.initialize()
+
+        # Initialize pending sitemaps on first run
+        if not self._state.current_value.pending_sitemap_urls and not self._state.current_value.completed:
+            self._state.current_value.pending_sitemap_urls.extend(self._sitemap_urls)
+
+        if self._state.current_value.url_queue and len(self._state.current_value.url_queue) >= self._max_buffer_size:
+            # Notify that the queue is full
+            self._queue_has_capacity.clear()
+
+        return self._state.current_value
 
     def _check_url_patterns(
         self,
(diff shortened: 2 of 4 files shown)


src/crawlee/request_loaders/_sitemap_request_loader.py
```diff
@@ -105,73 +159,136 @@ def _check_url_patterns(
...
-                options=self._parse_sitemap_options,
-            ):
-                # Only process URL items (not nested sitemaps)
-                if isinstance(item, SitemapUrl):
-                    url = item.loc
-
-                    # Skip if already processed
-                    if url in self._processed_urls:
+            # Get actual state
+            while (state := await self._get_state()) and (state.pending_sitemap_urls or state.in_progress_sitemap_url):
```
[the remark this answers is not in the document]
Mantisus: The condition uses `and` if the second half of the condition is False. Then the loop will be terminated.

src/crawlee/request_loaders/_sitemap_request_loader.py
```diff
@@ -50,27 +79,52 @@ def __init__(
...
+                url_queue=deque(),
+                pending_sitemap_urls=deque(),
+            ),
+            persistence_enabled=persist_enabled,
+            persist_state_key=persist_state_key,
+            logger=logger,
+        )
+
+        # Start background loading
+        self._loading_task = asyncio.create_task(self._load_sitemaps())
```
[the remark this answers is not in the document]
Mantisus: It is not expected that the instance will be created without full initialization. If it was stopped using `close`, then loading can be restarted using `start`.

src/crawlee/request_loaders/_sitemap_request_loader.py
```diff
@@ -40,7 +68,8 @@ def __init__(
         include: list[re.Pattern[Any] | Glob] | None = None,
         exclude: list[re.Pattern[Any] | Glob] | None = None,
         max_buffer_size: int = 200,
-        parse_sitemap_options: ParseSitemapOptions | None = None,
+        persist_state_key: str = 'SITEMAP_REQUEST_LOADER_STATE',
```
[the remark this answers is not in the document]
Mantisus: Documentation updated, standard `persist_state_key` for simple use cases.

src/crawlee/request_loaders/_sitemap_request_loader.py
```diff
@@ -23,6 +29,28 @@
...
+    pending_sitemap_urls: Annotated[deque[str], Field(alias='pendingSitemapUrls')]
+    in_progress_sitemap_url: Annotated[str | None, Field(alias='inProgressSitemapUrl')] = None
+    # URLs from the current sitemap that have been processed (for resuming within a sitemap)
+    current_sitemap_processed_urls: Annotated[set[str], Field(alias='currentSitemapProcessedUrls')] = set()
+    # Flag indicating if the sitemaps have been fully processed
+    completed: Annotated[bool, Field(alias='sitemapCompleted')] = False
+
+    # Counters
+    total_count: Annotated[int, Field(alias='totalCount')] = 0
+    handled_count: Annotated[int, Field(alias='handledCount')] = 0
```
vdusek: Can we rather use field docstrings, please?

src/crawlee/request_loaders/_sitemap_request_loader.py
```diff
@@ -40,7 +68,8 @@ def __init__(
         include: list[re.Pattern[Any] | Glob] | None = None,
         exclude: list[re.Pattern[Any] | Glob] | None = None,
         max_buffer_size: int = 200,
-        parse_sitemap_options: ParseSitemapOptions | None = None,
+        persist_state_key: str = 'SITEMAP_REQUEST_LOADER_STATE',
+        persist_enabled: bool = False,
```
vdusek: Can we do this consistently with `RequestList`? (no `persist_enabled`, `key=None` means state will not be persisted)

src/crawlee/request_loaders/_sitemap_request_loader.py
```diff
@@ -23,6 +29,36 @@
...
+    total_count: Annotated[int, Field(alias='totalCount')] = 0
+    """Total number of URLs found and added to the queue from all processed sitemaps."""
+
+    handled_count: Annotated[int, Field(alias='handledCount')] = 0
+    """Number of URLs that have been successfully handled."""
+
+
 @docs_group('Request loaders')
 class SitemapRequestLoader(RequestLoader):
     """A request loader that reads URLs from sitemap(s).
```
vdusek, line 91: Could you please mention the persistance feature in the class docstring?

src/crawlee/request_loaders/_sitemap_request_loader.py
```diff
@@ -105,73 +167,140 @@ def _check_url_patterns(
...
+    async def __aenter__(self) -> SitemapRequestLoader:
+        """Enter the context manager."""
+        await self.start()
+        return self
+
+    async def __aexit__(
+        self, exc_type: type[BaseException] | None, exc_value: BaseException | None, exc_traceback: TracebackType | None
+    ) -> None:
+        """Exit the context manager."""
+        await self.close()
```
vdusek, line 357: Is this here because of the persistence, or why? `RequestList` doesn't implement these.

Mantisus: Hmm. I don't see that `teardown` is called for `self._state` in `RequestList`.
That is, it only saves its state on the `PERSIST_STATE` event. If the work with the loader was completed between two events, the difference in states will not be saved. Or am I missing something?

For `SitemapRequestLoader`, `teardown` is called in `close`. This will stop the background task reading sitemaps and save the final state. The context manager is just a wrapper for the user so that they don't have to make this call manually.

docs/guides/code_examples/request_loaders/sitemap_example_with_persist.py
```diff
@@ -0,0 +1,48 @@
...
+        ImpitHttpClient() as http_client,
+        SitemapRequestLoader(
+            sitemap_urls=['https://crawlee.dev/sitemap.xml'],
+            http_client=http_client,
+            # Enable persistence
+            persist_state_key='my-persist-state',
+        ) as sitemap_loader,
+    ):
+        # Allow some time for the loader to fetch the sitemap and extract some URLs
+        await asyncio.sleep(1)
```
janbuchar: What happens if we omit this? Won't the `fetch_next_request` just take a bit longer?

src/crawlee/request_loaders/_sitemap_request_loader.py
```diff
@@ -105,73 +170,141 @@ def _check_url_patterns(
...
 
-    async def get_handled_count(self) -> int:
-        """Return the number of handled requests."""
-        return self._handled_count
-
     async def abort_loading(self) -> None:
         """Abort the sitemap loading process."""
         if self._loading_task and not self._loading_task.done():
             self._loading_task.cancel()
+            await asyncio.sleep(0.01)  # Allow time for cancellation
```
janbuchar: Huh, lots of random sleeps in this PR. Is there really no way around this?

src/crawlee/request_loaders/_sitemap_request_loader.py
```diff
@@ -23,12 +29,45 @@
 logger = getLogger(__name__)
 
 
+class SitemapRequestLoaderState(BaseModel):
```
janbuchar, line 32: Now that I see this from the outside, it would be great if you could write down how the persistence mechanism works in the docblock of this class.

Also, I don't see processed sitemap URLs being tracked in a any way, is that intentional?

src/crawlee/request_loaders/_sitemap_request_loader.py
Mantisus: If `Stream` closes too quickly, it causes errors from `impit`.

Perhaps when we optimize it, we will be able to remove this. I hope so.

docs/guides/code_examples/request_loaders/sitemap_example_with_persist.py
Mantisus: > What happens if we omit this? Won't the fetch_next_request just take a bit longer?

I decided not to wait inside `fetch_next_request` until there are links to process in the queue.
So it just returns `None`.

Neither the old nor the new behavior is a problem for Crawlee. But the old behavior with cyclic waiting may be unexpected for the user if they decide to use `SitemapRequestLoader` directly.

src/crawlee/request_loaders/_sitemap_request_loader.py
Mantisus: > I don't see processed sitemap URLs being tracked in a any way, is that intentional?

Yes, I may be wrong, but I think that cyclic links are not expected in sitemaps. Thanks to this, we don't need to store links to processed sitemaps.

JS uses similar behavior - https://github.com/apify/crawlee/blob/master/packages/core/src/storages/sitemap_request_list.ts#L108

src/crawlee/request_loaders/_sitemap_request_loader.py
Mantisus: Perhaps @janbuchar will provide more details about the absence of the `teardown` call for `self._state` in `RequestList`.

janbuchar: That's an oversight, but I'm not sure how to resolve it - `RequestList` is initialized lazily, so adding a manual teardown feels off.

docs/guides/code_examples/request_loaders/sitemap_example_with_persist.py
janbuchar: I think that is not the correct behavior. A 1-second wait does not guarantee that there will be something to process. Maybe it'll take less time, maybe more. How should a user know that?

If you want to check if there's something to process, you can always use `is_empty`.

src/crawlee/request_loaders/_sitemap_request_loader.py
janbuchar: We cannot design a class that uses an abstraction (HttpClient) around limitations of a library used in one of the implementations. Any chance this could be pushed down to `ImpitHttpClient`?

docs/guides/code_examples/request_loaders/sitemap_example_with_persist.py
Mantisus: Yes, I completely agree with that. This is poorly used in the example, I will update it.

src/crawlee/request_loaders/_sitemap_request_loader.py
Mantisus: Makes sense. I'll check it

Mantisus: [PR](https://github.com/apify/crawlee-python/pull/1355) solves this

src/crawlee/request_loaders/_sitemap_request_loader.py
janbuchar: I wouldn't be surprised to encounter a cyclic sitemap somewhere, but I don't have a real-world example :shrug:

docs/guides/code_examples/request_loaders/sitemap_example.py
```diff
@@ -17,11 +17,15 @@ async def main() -> None:
...
+        while not await sitemap_loader.is_finished():
+            if request := await sitemap_loader.fetch_next_request():
+                # Do something with it...
 
-            # And mark it as handled.
-            await sitemap_loader.mark_request_as_handled(request)
+                # And mark it as handled.
+                await sitemap_loader.mark_request_as_handled(request)
+            else:
+                await asyncio.sleep(0.01)
```
vdusek: Why this sleep? Maybe add comment?

vdusek (approved):
LGTM

docs/guides/code_examples/request_loaders/sitemap_example_with_persist.py
janbuchar: I didn't make myself clear the last time. I meant that this:

> I decided not to wait inside fetch_next_request until there are links to process in the queue.
So it just returns None.

is not the correct behavior. In the JS version, `fetchNextRequest` blocks until a link can be returned (or the sitemap is completely processed). What is the reason to do it differently here?

Mantisus: I understand your point, it is indeed different from JS. But in my opinion, this behavior is more consistent with the loader API

```python
    @abstractmethod
        """Return the next request to be processed, or `null` if there are no more pending requests."""
```

The loader will return `None` if there are no requests ready to be processed at the moment.

But I won't insist that my interpretation of the API description is correct. 🙂

janbuchar: Given that 1) I wrote the docblock and 2) the JS version waits for requests to appear, I think the "return None if and only if there will be no more requests, ever" is correct :grin: 

I'll expand the docblock in the meantime and fix that "`null`", too

src/crawlee/request_loaders/_sitemap_request_loader.py
janbuchar: Cyclic sitemaps aside, can you please briefly describe how is this state model used by the loader? Like "The crawler processes one sitemap at a time. The current one is kept in `in_progress_sitemap_url`..."

docs/guides/code_examples/request_loaders/sitemap_example_with_persist.py
Mantisus: Updated 🙂

src/crawlee/request_loaders/_sitemap_request_loader.py
Mantisus: Sure. 🙂

The crawler processes one sitemap at a time. The current sitemap is stored in `in_progress_sitemap_url`. The `parse_sitemap` function parses the sitemap and returns elements as an async iterator. Each element retrieved from the iterator is processed based on its type. If the element is a `NestedSitemap`, its URL is added to `pending_sitemap_urls`. If the element is a `SitemapUrl`, the system checks whether it already exists in `current_sitemap_processed_urls`. If it exists, the loader was restarted from a saved state and the URL is skipped.

If the URL is new, it is first added to `url_queue`, then to `current_sitemap_processed_urls`, and `total_count` is incremented by 1. When all elements from the current sitemap iterator have been processed, `in_progress_sitemap_url` is set to `None` and `current_sitemap_processed_urls` is cleared. The next sitemap is retrieved from `pending_sitemap_urls`. If `pending_sitemap_urls` is empty, `completed` is set to `True`.

When `fetch_next_request` is called, a URL is extracted from `url_queue` and placed in `in_progress`. When `mark_request_as_handled` is called for the extracted URL, it is removed from `in_progress` and `handled_count` is incremented by 1.

During initial startup or restart after persistence, state validation occurs in `_get_state`. If both `pending_sitemap_urls` and `in_progress_sitemap_url` are empty and `completed` is False, this indicates a fresh start. In this case, `self._sitemap_urls` are moved to `pending_sitemap_urls`. Otherwise, the system is restarting from a persisted state. If `in_progress` contains any URLs, they are moved back to `url_queue` and `in_progress` is cleared.

Mantisus: Regarding cyclic sitemaps. According to [sitemaps.org](https://sitemaps.org/protocol.html#:~:text=Using%20Sitemap%20index%20files%20(to%20group%20multiple%20sitemap%20files)), only a sitemapindex should contain links to sitemaps, and a sitemapindex should not contain links to other sitemapindex files.
This means the hierarchy should be: a sitemapindex contains links to one or more sitemaps, where each sitemap contains one or more links to website pages.
If a website follows the recommended protocol, this should prevent circular sitemaps. However, a website might have multiple sitemapindex files that reference the same sitemap (I don't think search engines would like this, but it's possible).

janbuchar: > Sure. 🙂 The crawler processes one sitemap at a time. The current sitemap is stored in...

Cool. Now put it inside the docblock please 😁 

> However, a website might have multiple sitemapindex files that reference the same sitemap (I don't think search engines would like this, but it's possible).

I could imagine a website implementing something like this as a scraper tarpit.

Mantisus: > Cool. Now put it inside the docblock please 😁

Added 🙂

> I could imagine a website implementing something like this as a scraper tarpit.

I added `processed_sitemap_urls` so that we can ensure we don't reprocess the same sitemap. 🙂

janbuchar (approved):
LGTM

PR #1350, opened 2025-08-12
docs: Polish "Upgrading to v1" guide

vdusek:
image

### DIFF
--- a/docs/upgrading/upgrading_to_v1.md
+++ b/docs/upgrading/upgrading_to_v1.md
@@ -5,15 +5,16 @@ title: Upgrading to v1
 
 This page summarizes the breaking changes between Crawlee for Python v0.6 and v1.0.
 
-## Distinct use of word `browser` in similar contexts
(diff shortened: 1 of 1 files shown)


docs/upgrading/upgrading_to_v1.md
```diff
@@ -31,169 +32,206 @@ HeaderGeneratorOptions(browsers=['chrome'])
...
+
+```python
+from crawlee.crawlers import ParselCrawler, ParselCrawlingContext
+
+async def main() -> None:
+    crawler = ParselCrawler()
+
+    @crawler.router.default_handler
+    async def request_handler(context: ParselCrawlingContext) -> None:
+        content = await context.http_response.read()
```
Pijukatel, line 235: I had to stare at the example really hard to see the difference. One extra `await`, it might be worth adding some comment to highlight the extra `await` as it is hidden in otherwise identical code.

Pijukatel (approved):

PR #1352, opened 2025-08-12
chore: fix `timeout` for `stream` method in `ImpitHttpClient`

Mantisus:
### Description

- fix `timeout` for `stream` method in `ImpitHttpClient`


vdusek (approved):
LGTM

PR #1355, opened 2025-08-13
chore: use sync close for impit stream response

Mantisus:
### Description

- When using `aclose` and canceling a task in which streaming reading is occurring, an error is generated from the generator in `impit`. When using `close`, the error does not occur.

### DIFF
--- a/src/crawlee/http_clients/_impit.py
+++ b/src/crawlee/http_clients/_impit.py
@@ -200,6 +200,7 @@ class ImpitHttpClient(HttpClient):
         try:
             yield _ImpitResponse(response)
         finally:
+            # TODO: https://github.com/apify/impit/issues/242
             # Quickly closing Response while reading the response body causes an error in the Rust generator in `impit`.
             # With a short sleep and sync closing, the error does not occur.
             # Replace with `response.aclose` when this is resolved in impit.


Mantisus:
So far, it looks like a solution, but we need to figure out what the problem is on the `impit` side.

src/crawlee/http_clients/_impit.py
```diff
@@ -199,7 +200,8 @@ async def stream(
         try:
             yield _ImpitResponse(response)
         finally:
-            await response.aclose()
+            await asyncio.sleep(0.01)
```
Pijukatel, line 207: Could you please add a comment to this mysterious sleep

Mantisus: Added.

For context. I encountered an error while testing `SitemapRequestLoader`. 
In some cases, if the generator closes in the middle of reading the response (for example, the read was canceled),  `response.aclose` causes a `RuntimeError` from the Rust code impit.

When switching to `response.close`, the error only reproduced once for me. It does not reproduce with `sleep`.

This is a temporary workaround. I am trying to create a minimal test case for reproduction to fix this in `impit`.

Pijukatel: Thanks for the explanation.

Pijukatel (approved):

src/crawlee/http_clients/_impit.py
janbuchar: It would be great if you could also add a link to the issue here so that we can quickly check if the problem has been resolved when coming back to this

Mantisus: Added

Issue #1077, opened 2025-03-12
Optimize package footprint by removing unnecessary deps


vdusek:
### Description

The current installation of the Crawlee package, when combined with all its direct and transitive dependencies, occupies ~75 MB. For context, Scrapy occupies ~88.4 MB.

```
$ (venv)  du -h venv/ --max-depth 0
6,6M	venv/
```

```
(venv) $ pip install crawlee
...
```

```
(venv) $ du -h venv/ --max-depth 0
82M	venv/
```

The large size is primarily due to several dependencies that may not be strictly necessary for the core functionality. Below is a detailed breakdown of the dependency tree, individual package sizes, and the summed sizes of Crawlee's direct dependencies.

### Dependency tree

```
crawlee v0.6.4
    ├── apify-fingerprint-datapoints v0.0.2
    ├── browserforge v1.2.3
    │   └── click v8.1.8
    ├── cachetools v5.5.2
    ├── colorama v0.4.6
    ├── docutils v0.21.2
    ├── eval-type-backport v0.2.2
    ├── httpx[brotli, http2, zstd] v0.28.1
    │   ├── anyio v4.8.0
    │   │   ├── idna v3.10
    │   │   └── sniffio v1.3.1
    │   ├── certifi v2025.1.31
    │   ├── httpcore v1.0.7
    │   │   ├── certifi v2025.1.31
    │   │   └── h11 v0.14.0
    │   ├── idna v3.10
    │   ├── brotli v1.1.0 (extra: brotli)
    │   ├── h2 v4.2.0 (extra: http2)
    │   │   ├── hpack v4.1.0
    │   │   └── hyperframe v6.1.0
    │   └── zstandard v0.23.0 (extra: zstd)
    ├── more-itertools v10.6.0
    ├── psutil v7.0.0
    ├── pydantic v2.10.6
    │   ├── annotated-types v0.7.0
    │   ├── pydantic-core v2.27.2
    │   │   └── typing-extensions v4.12.2
    │   └── typing-extensions v4.12.2
    ├── pydantic-settings v2.6.1
    │   ├── pydantic v2.10.6 (*)
    │   └── python-dotenv v1.0.1
    ├── pyee v12.1.1
    │   └── typing-extensions v4.12.2
    ├── rich v13.9.4
    │   ├── markdown-it-py v3.0.0
    │   │   └── mdurl v0.1.2
    │   └── pygments v2.19.1
    ├── sortedcollections v2.1.0
[...]

PR #1362, opened 2025-08-19
chore: Remove sortedcollections and sortedcontainers packages

vdusek:
### Description

- Remove `sortedcollections` and `sortedcontainers` packages from dependencies.
- Replace the sorted list in the Snapshotter with a custom class.

### Issues

- Closes: #1077

### Testing

- One new test for the `SortedSnapshotList` plus the current test set covers the changes.

### Checklist

- [x] CI passed

### DIFF
--- a/src/crawlee/_autoscaling/snapshotter.py
+++ b/src/crawlee/_autoscaling/snapshotter.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import bisect
 from datetime import datetime, timedelta, timezone
 from logging import getLogger
 from typing import TYPE_CHECKING, TypeVar, cast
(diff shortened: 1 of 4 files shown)


src/crawlee/_autoscaling/snapshotter.py
```diff
@@ -25,7 +22,25 @@
 
 logger = getLogger(__name__)
 
-T = TypeVar('T')
+T = TypeVar('T', bound=Snapshot)
+
+
+class _SortedSnapshotList(list[T]):
```
Pijukatel: Could you please add some very simple unit test for the add method.

src/crawlee/_autoscaling/snapshotter.py
```diff
@@ -25,7 +22,25 @@
...
 logger = getLogger(__name__)
 
-T = TypeVar('T')
+T = TypeVar('T', bound=Snapshot)
+
+
+class SortedSnapshotList(list[T]):
+    """A list that maintains sorted order by `created_at` attribute for snapshot objects."""
+
+    def add(self, item: T) -> None:
```
janbuchar, line 32: nit - you  could use the [bisect](https://docs.python.org/3/library/bisect.html) stdlib module instead of writing your own binary search

janbuchar (approved):
Nice, LGTM!

Issue #1354, opened 2025-08-13
All crawlee persistence names should be valid names for Apify platform


Pijukatel:
In [FileSystemRequestQueueClient](https://github.com/apify/crawlee-python/blob/master/src/crawlee/storage_clients/_file_system/_request_queue_client.py#L119) there is persistance to kvs with name `persist_state_kvs_name=f'__RQ_STATE_{self._metadata.id}',`

When used on Apify platform, it will throw an error:

> apify_client._errors.ApifyApiError: Invalid value provided in store: name can only contain letters 'a' through 'z', the digits '0' through '9', and the hyphen ('-') but only in the middle of the string (e.g. 'my-value-1')

which is caused by the underscores in the name. 

Fix the name to be compatible with the Apify constraints and check all similar names of `RecoverableState`s in code base

To reproduce, run this on Apify platform, which create file system base request queue, that persists to Apify kvs:
```
async def main() -> None:
    async with Actor:
        rq = await RequestQueue.open(
            storage_client=FileSystemStorageClient(),
        )
```

Pijukatel:
This is more of a problem that `RecoverableState` relies on a service locator to give it suitable storage, and this can lead to scenarios where `RecoverableState` for XYZ storage is being saved to ABC storage, which makes no sense. A similar problem is waiting to be triggered in `KeyValueStore.get_auto_saved_value`, which also uses `RecoverableState` that relies on the service locator.

Reasonable default should be that `RecoverableState` is using the same storage for which it is recovering...

PR #1339, opened 2025-08-01
feat:  Add `SqlStorageClient` based on `sqlalchemy` v2+

Mantisus:
### Description

- Add `SQLStorageClient` which can accept a database connection string or a pre-configured `AsyncEngine`, or creates a default `crawlee.db` database in `Configuration.storage_dir`.

### Issues

- Closes: #307

PR #1368, opened 2025-08-22
feat: Update `RecoverableState` and `StorageInstanceManager` to ensure proper persistence

Pijukatel:
### Description

This change enables scenarios with mixed storage clients. For example, running on the Apify platform, but also using some specialized storage based on a different storage client.

- `StorageInstanceManager` now also caches based on the used storage client, by using `client_opener.__qualname__` as cache key. This allows, for example, to keep two storages of the same name if they use different storage clients.
- `RecoverableState` now has an optional argument `key_value_store` to allow explicit setting of the desired `kvs` to be used for persistence. This is to prevent a situation where the storage client-specific state is persisted to a different storage client just because of the global service locator. For example, when using SQL based storage client on Apify platform, it is desired to persist the recoverable state to the SQL storage-based kvs and not to the Apify-based kvs.

### Issues

- Closes: #1354

### Testing

Added new unit tests.

### DIFF
--- a/src/crawlee/_utils/recoverable_state.py
+++ b/src/crawlee/_utils/recoverable_state.py
@@ -53,9 +53,14 @@ class RecoverableState(Generic[TStateModel]):
             persist_state_kvs_id: The identifier of the KeyValueStore to use for persistence.
                 If neither a name nor and id are supplied, the default store will be used.
             logger: A logger instance for logging operations related to state persistence
-            key_value_store: KeyValueStore to use for persistence. If not provided, the service locator is used to
-                provide suitable KeyValueStore.
+            key_value_store: KeyValueStore to use for persistence. If not provided, a system-wide KeyValueStore will be
+                used, based on service locator configuration.
         """
+        if key_value_store and (persist_state_kvs_name or persist_state_kvs_id):
+            raise ValueError(
+                'Cannot provide explicit key_value_store and persist_state_kvs_name or persist_state_kvs_id.'
+            )
+
         self._default_state = default_state
         self._state_type: type[TStateModel] = self._default_state.__class__
         self._state: TStateModel | None = None
@@ -63,8 +68,8 @@ class RecoverableState(Generic[TStateModel]):
         self._persist_state_key = persist_state_key
         self._persist_state_kvs_name = persist_state_kvs_name
         self._persist_state_kvs_id = persist_state_kvs_id
-        self._key_value_store = key_value_store
         self._log = logger
+        self._key_value_store = key_value_store
 
     async def initialize(self) -> TStateModel:
         """Initialize the recoverable state.
--- a/src/crawlee/storages/_storage_instance_manager.py
+++ b/src/crawlee/storages/_storage_instance_manager.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from collections import defaultdict
 from collections.abc import Awaitable, Callable
+from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, TypeVar, cast
 
 from crawlee.storage_clients._base import DatasetClient, KeyValueStoreClient, RequestQueueClient
@@ -20,16 +21,20 @@ ClientOpener = Callable[..., Awaitable[StorageClientType]]
 """Type alias for the client opener function."""
 
 
-class StorageClientCache:
+@dataclass
+class _StorageClientCache:
     """Cache for specific storage client."""
 
-    def __init__(self) -> None:
-        self.by_id: defaultdict[type[Storage], defaultdict[str, Storage]] = defaultdict(lambda: defaultdict())
-        """Cache for storage instances by ID, separated by storage type."""
-        self.by_name: defaultdict[type[Storage], defaultdict[str, Storage]] = defaultdict(lambda: defaultdict())
-        """Cache for storage instances by name, separated by storage type."""
-        self.default_instances: defaultdict[type[Storage], Storage] = defaultdict()
-        """Cache for default instances of each storage type."""
+    by_id: defaultdict[type[Storage], defaultdict[str, Storage]] = field(
+        default_factory=lambda: defaultdict(lambda: defaultdict())
+    )
+    """Cache for storage instances by ID, separated by storage type."""
+    by_name: defaultdict[type[Storage], defaultdict[str, Storage]] = field(
+        default_factory=lambda: defaultdict(lambda: defaultdict())
+    )
+    """Cache for storage instances by name, separated by storage type."""
+    default_instances: defaultdict[type[Storage], Storage] = field(default_factory=lambda: defaultdict())
+    """Cache for default instances of each storage type."""
 
 
 class StorageInstanceManager:
@@ -40,7 +45,7 @@ class StorageInstanceManager:
     """
 
     def __init__(self) -> None:
-        self._cache_by_storage_client: dict[str, StorageClientCache] = defaultdict(StorageClientCache)
+        self._cache_by_storage_client: dict[str, _StorageClientCache] = defaultdict(_StorageClientCache)
 
     async def open_storage_instance(
         self,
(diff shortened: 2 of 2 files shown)


src/crawlee/_utils/recoverable_state.py
```diff
@@ -52,6 +53,8 @@ def __init__(
             persist_state_kvs_id: The identifier of the KeyValueStore to use for persistence.
                 If neither a name nor and id are supplied, the default store will be used.
             logger: A logger instance for logging operations related to state persistence
+            key_value_store: KeyValueStore to use for persistence. If not provided, the service locator is used to
+                provide suitable KeyValueStore.
```
janbuchar: I'd reword this slightly to something like "a system-wide KeyValueStore will be used, based on service locator configuration"

src/crawlee/_utils/recoverable_state.py
```diff
@@ -38,6 +38,7 @@ def __init__(
         persist_state_kvs_name: str | None = None,
         persist_state_kvs_id: str | None = None,
         logger: logging.Logger,
+        key_value_store: None | KeyValueStore = None,
```
janbuchar, line 41: Can you raise an error if you get both `key_value_store` and `persist_state_kvs_id` or `persist_state_kvs_name`

src/crawlee/storages/_storage_instance_manager.py
```diff
@@ -19,6 +20,18 @@
 """Type alias for the client opener function."""
 
 
+class StorageClientCache:
```
janbuchar: This looks a lot like a dataclass. Also, could you make it private, or add `__all__` and exclude it from there? I'd like to make it clear that it's just an auxiliary class for `StorageInstanceManager`.

src/crawlee/storages/_storage_instance_manager.py
```diff
@@ -64,19 +70,23 @@ async def open_storage_instance(
             raise ValueError('Only one of "id" or "name" can be specified, not both.')
 
         # Check for default instance
-        if id is None and name is None and cls in self._default_instances:
-            return cast('T', self._default_instances[cls])
+        if (
+            id is None
+            and name is None
+            and cls in self._cache_by_storage_client[client_opener.__qualname__].default_instances
```
janbuchar, line 81: There's no guarantee that `client_opener` will be stable enough to use as a cache key. Something similar to `id(storage_client)` would be more appropriate IMO, but that would require a more sizable change to the code.

tests/unit/storages/test_storage_instance_manager.py
```diff
@@ -0,0 +1,96 @@
...
+        purge_on_start=True,
+    )
+
+    kvs1 = await KeyValueStore.open(storage_client=MemoryStorageClient(), configuration=config)
+
+    kvs2 = await KeyValueStore.open(
+        storage_client=MemoryStorageClient(),
+        configuration=config,
+    )
+    assert kvs1 is kvs2
```
janbuchar, line 60: Is this really correct? Those two clients could be configured differently (imagine e.g., two filesystem clients with different storage dirs)

src/crawlee/_utils/recoverable_state.py
Pijukatel: Reworded.

src/crawlee/_utils/recoverable_state.py
Pijukatel: Raised

src/crawlee/storages/_storage_instance_manager.py
Pijukatel: Updated

src/crawlee/storages/_storage_instance_manager.py
Pijukatel: Yes, with the current state, we do not have access to the storage_client before caching, so I was left with client_opener.

I was thinking about accessing `storage_client` through `client_opener.__self__`, but that would work only for bound methods of the storage clients, and client_opener can be any callable. That is why I opted for `client_opener.__qualname__`, which is generic enough and I think should be stable enough for the cache, as the most common case will be just one of our prepared factory methods, like `MemoryStorageClient.create_kvs_client`

tests/unit/storages/test_storage_instance_manager.py
Pijukatel: That is also current behavior, and I have not changed that. It is a fair point though, but I would question it based on the scope of the PR:

The current state is that it is allowed to have one storage of a specific type with the same ID or name.

Basically, caching is f(storage_type, id or name)

I am expanding it to f(storage_type, id or name, **client_opener**) to support use case in the issue I am dealing with.

To support the use case you are describing, I would have to expand it to f(storage_type, id or name, client_opener, **configuration**)

So I can further expand the caching to support this as well, if required, even though it is not necessary for this PR.

janbuchar: Okay, let's not make a behemoth PR here, but please create an issue to track this

src/crawlee/storages/_storage_instance_manager.py
janbuchar: I guess that that's another consideration related to the instance cache mechanism. We should resolve that before v1.

Mantisus (approved):
LGTM

janbuchar:
> For example, when using SQL based storage client on Apify platform, it is desired to persist the recoverable state to the SQL storage-based kvs and not to the Apify-based kvs.

After thinking about this for a while during the review of #1339, I'm not so sure. The Apify-based KVS has the advantage of having a default storage for each run. An SQL database server outside of the platform has no way of telling different Actor runs apart, so it can only have a single shared default KVS for all of them.

Point being... maybe `RecoverableState` should try hard to put its state in the Apify KVS if given the chance. I might be wrong though :shrug: I just wanted to write this idea down.

Pijukatel:
> > For example, when using SQL based storage client on Apify platform, it is desired to persist the recoverable state to the SQL storage-based kvs and not to the Apify-based kvs.
> 
> After thinking about this for a while during the review of #1339, I'm not so sure. The Apify-based KVS has the advantage of having a default storage for each run. An SQL database server outside of the platform has no way of telling different Actor runs apart, so it can only have a single shared default KVS for all of them.
> 
> Point being... maybe `RecoverableState` should try hard to put its state in the Apify KVS if given the chance. I might be wrong though 🤷 I just wanted to write this idea down.

I think that it is the storage's responsibility to ensure it is suitable for a recoverable state. This change is mainly about allowing (not enforcing) decoupling between `RecoverableState` and global kvs provided by the service locator.

tests/unit/storages/test_storage_instance_manager.py
Pijukatel: https://github.com/apify/crawlee-python/issues/1379

src/crawlee/storage_clients/_file_system/_request_queue_client.py
```diff
@@ -248,13 +230,40 @@ async def open(
...
+    @classmethod
+    async def _create_client(
+        cls, metadata: RequestQueueMetadata, storage_dir: Path, *, update_accessed_at: bool = False
+    ) -> FileSystemRequestQueueClient:
+        """Create client from metadata and storage directory."""
+        from crawlee.storage_clients import FileSystemStorageClient  # noqa: PLC0415 avoid circular imports
+        from crawlee.storages._key_value_store import KeyValueStore  # noqa: PLC0415 avoid circular imports
+
+        # Prepare kvs for recoverable state
+        kvs_client = await FileSystemStorageClient().create_kvs_client(name=f'__RQ_STATE_{metadata.id}')
```
janbuchar, line 246: Wait a minute. This is a subtle bug waiting to happen - the `FileSystemStorageClient` that creates this RQ client might have a different configuration than the one you create here.

Can you pass the actual storage client down to this method instead?

janbuchar (changes requested):

Pijukatel:
Close in favor of https://github.com/apify/crawlee-python/pull/1438

PR #1386, opened 2025-09-02
 feat: Rework storage creation and caching, configuration and services

Pijukatel:
Goals are:
- To enable proper caching based on all the storage properties.
- To make more predictable behavior of the service locator. No runtime crashes based on when we open the storage.
- To allow for combinations of different storage clients in Crawlers and Actor without causing `ServiceConflictError`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "crawlee"
-version = "1.0.0rc1"
+version = "0.6.13"
 description = "Crawlee for Python"
 authors = [{ name = "Apify Technologies s.r.o.", email = "support@apify.com" }]
 license = { file = "LICENSE" }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "crawlee"
-version = "0.6.13"
+version = "1.0.0rc1"
 description = "Crawlee for Python"
 authors = [{ name = "Apify Technologies s.r.o.", email = "support@apify.com" }]
 license = { file = "LICENSE" }
--- a/src/crawlee/configuration.py
+++ b/src/crawlee/configuration.py
@@ -28,7 +28,7 @@ class Configuration(BaseSettings):
     Settings can also be configured via environment variables, prefixed with `CRAWLEE_`.
     """
 
-    model_config = SettingsConfigDict(populate_by_name=True)
+    model_config = SettingsConfigDict(validate_by_name=True, validate_by_alias=True)
 
     internal_timeout: Annotated[timedelta | None, Field(alias='crawlee_internal_timeout')] = None
     """Timeout for the internal asynchronous operations."""
--- a/src/crawlee/crawlers/_basic/_basic_crawler.py
+++ b/src/crawlee/crawlers/_basic/_basic_crawler.py
@@ -347,10 +347,20 @@ class BasicCrawler(Generic[TCrawlingContext, TStatisticsState]):
             _logger: A logger instance, typically provided by a subclass, for consistent logging labels.
                 Intended for use by subclasses rather than direct instantiation of `BasicCrawler`.
         """
+        global_configuration: None | Configuration = None
+
         if not configuration:
-            configuration = service_locator.get_configuration()
+            global_configuration = service_locator.get_configuration()
+            configuration = global_configuration
+
         if not storage_client:
-            storage_client = service_locator.get_storage_client()
+            if global_configuration:
+                # If global configuration was used, reuse its storage client too
+                storage_client = service_locator.get_storage_client()
+            else:
+                # If unique configuration was used, create a unique storage client based on such configuration
+                storage_client = service_locator.get_storage_client().create_client(configuration)
+
         if not event_manager:
             event_manager = service_locator.get_event_manager()
 
--- a/src/crawlee/crawlers/_basic/_basic_crawler.py
+++ b/src/crawlee/crawlers/_basic/_basic_crawler.py
@@ -27,6 +27,7 @@ from crawlee import EnqueueStrategy, Glob, RequestTransformAction, service_locat
 from crawlee._autoscaling import AutoscaledPool, Snapshotter, SystemStatus
 from crawlee._log_config import configure_logger, get_configured_log_level, string_to_log_level
 from crawlee._request import Request, RequestOptions, RequestState
+from crawlee._service_locator import ServiceLocator
 from crawlee._types import (
     BasicCrawlingContext,
     EnqueueLinksKwargs,
@@ -346,14 +347,18 @@ class BasicCrawler(Generic[TCrawlingContext, TStatisticsState]):
             _logger: A logger instance, typically provided by a subclass, for consistent logging labels.
                 Intended for use by subclasses rather than direct instantiation of `BasicCrawler`.
         """
-        if configuration:
-            service_locator.set_configuration(configuration)
-        if storage_client:
-            service_locator.set_storage_client(storage_client)
-        if event_manager:
-            service_locator.set_event_manager(event_manager)
+        if not configuration:
+            configuration = service_locator.get_configuration()
+        if not storage_client:
+            storage_client = service_locator.get_storage_client()
+        if not event_manager:
+            event_manager = service_locator.get_event_manager()
 
-        config = service_locator.get_configuration()
+        self._service_locator = ServiceLocator(
+            configuration=configuration, storage_client=storage_client, event_manager=event_manager
+        )
+
+        config = self._service_locator.get_configuration()
 
         # Core components
         self._request_manager = request_manager
@@ -548,7 +553,7 @@ class BasicCrawler(Generic[TCrawlingContext, TStatisticsState]):
     async def get_request_manager(self) -> RequestManager:
         """Return the configured request manager. If none is configured, open and return the default request queue."""
         if not self._request_manager:
-            self._request_manager = await RequestQueue.open()
+            self._request_manager = await RequestQueue.open(storage_client=self._service_locator.get_storage_client())
 
         return self._request_manager
 
@@ -559,7 +564,7 @@ class BasicCrawler(Generic[TCrawlingContext, TStatisticsState]):
         name: str | None = None,
     ) -> Dataset:
         """Return the `Dataset` with the given ID or name. If none is provided, return the default one."""
-        return await Dataset.open(id=id, name=name)
+        return await Dataset.open(id=id, name=name, storage_client=self._service_locator.get_storage_client())
 
     async def get_key_value_store(
         self,
@@ -568,7 +573,7 @@ class BasicCrawler(Generic[TCrawlingContext, TStatisticsState]):
         name: str | None = None,
     ) -> KeyValueStore:
         """Return the `KeyValueStore` with the given ID or name. If none is provided, return the default KVS."""
-        return await KeyValueStore.open(id=id, name=name)
+        return await KeyValueStore.open(id=id, name=name, storage_client=self._service_locator.get_storage_client())
 
     def error_handler(
         self, handler: ErrorHandler[TCrawlingContext | BasicCrawlingContext]
@@ -684,7 +689,7 @@ class BasicCrawler(Generic[TCrawlingContext, TStatisticsState]):
         return final_statistics
 
     async def _run_crawler(self) -> None:
-        event_manager = service_locator.get_event_manager()
+        event_manager = self._service_locator.get_event_manager()
 
         self._crawler_state_rec_task.start()
 
@@ -1520,7 +1525,7 @@ class BasicCrawler(Generic[TCrawlingContext, TStatisticsState]):
 
     async def _crawler_state_task(self) -> None:
         """Emit a persist state event with the given migration status."""
-        event_manager = service_locator.get_event_manager()
+        event_manager = self._service_locator.get_event_manager()
 
         current_state = self.statistics.state
 
--- a/src/crawlee/crawlers/_basic/_basic_crawler.py
+++ b/src/crawlee/crawlers/_basic/_basic_crawler.py
@@ -347,21 +347,16 @@ class BasicCrawler(Generic[TCrawlingContext, TStatisticsState]):
             _logger: A logger instance, typically provided by a subclass, for consistent logging labels.
                 Intended for use by subclasses rather than direct instantiation of `BasicCrawler`.
         """
-        global_configuration: None | Configuration = None
-
         if not configuration:
-            global_configuration = service_locator.get_configuration()
-            configuration = global_configuration
+            configuration = service_locator.get_configuration()
 
         if not storage_client:
-            if global_configuration:
-                # If global configuration was used, reuse its storage client too
-                storage_client = service_locator.get_storage_client()
-            else:
-                # If unique configuration was used, create a unique storage client based on such configuration
-                storage_client = service_locator.get_storage_client().create_client(configuration)
+            storage_client = service_locator.get_storage_client()
 
         if not event_manager:
+            # This is weird if someone passes configuration and its event manager related stuff gets ignored as the
+            # event manager will be used from service_locator. Maybe keep the was created flag for it? It does not have
+            # the use cases like the storages
             event_manager = service_locator.get_event_manager()
 
         self._service_locator = ServiceLocator(
@@ -563,7 +558,10 @@ class BasicCrawler(Generic[TCrawlingContext, TStatisticsState]):
     async def get_request_manager(self) -> RequestManager:
         """Return the configured request manager. If none is configured, open and return the default request queue."""
         if not self._request_manager:
-            self._request_manager = await RequestQueue.open(storage_client=self._service_locator.get_storage_client())
+            self._request_manager = await RequestQueue.open(
+                storage_client=self._service_locator.get_storage_client(),
+                configuration=self._service_locator.get_configuration(),
+            )
 
         return self._request_manager
 
@@ -574,7 +572,12 @@ class BasicCrawler(Generic[TCrawlingContext, TStatisticsState]):
         name: str | None = None,
     ) -> Dataset:
         """Return the `Dataset` with the given ID or name. If none is provided, return the default one."""
-        return await Dataset.open(id=id, name=name, storage_client=self._service_locator.get_storage_client())
+        return await Dataset.open(
+            id=id,
+            name=name,
+            storage_client=self._service_locator.get_storage_client(),
+            configuration=self._service_locator.get_configuration(),
+        )
 
     async def get_key_value_store(
         self,
@@ -583,7 +586,12 @@ class BasicCrawler(Generic[TCrawlingContext, TStatisticsState]):
         name: str | None = None,
     ) -> KeyValueStore:
         """Return the `KeyValueStore` with the given ID or name. If none is provided, return the default KVS."""
-        return await KeyValueStore.open(id=id, name=name, storage_client=self._service_locator.get_storage_client())
+        return await KeyValueStore.open(
+            id=id,
+            name=name,
+            storage_client=self._service_locator.get_storage_client(),
+            configuration=self._service_locator.get_configuration(),
+        )
 
     def error_handler(
         self, handler: ErrorHandler[TCrawlingContext | BasicCrawlingContext]
--- a/src/crawlee/storage_clients/_base/_storage_client.py
+++ b/src/crawlee/storage_clients/_base/_storage_client.py
@@ -55,6 +55,10 @@ class StorageClient(ABC):
     ) -> RequestQueueClient:
         """Create a request queue client."""
 
+    @abstractmethod
+    def create_client(self, configuration: Configuration) -> StorageClient:
+        """Create a storage client from an existing storage."""
+
     def get_rate_limit_errors(self) -> dict[int, int]:
         """Return statistics about rate limit errors encountered by the HTTP client in storage client."""
         return {}
--- a/src/crawlee/storage_clients/_base/_storage_client.py
+++ b/src/crawlee/storage_clients/_base/_storage_client.py
@@ -34,7 +34,6 @@ class StorageClient(ABC):
         *,
         id: str | None = None,
         name: str | None = None,
-        configuration: Configuration | None = None,
     ) -> DatasetClient:
         """Create a dataset client."""
 
@@ -44,7 +43,6 @@ class StorageClient(ABC):
         *,
         id: str | None = None,
         name: str | None = None,
-        configuration: Configuration | None = None,
     ) -> KeyValueStoreClient:
         """Create a key-value store client."""
 
@@ -54,7 +52,6 @@ class StorageClient(ABC):
         *,
         id: str | None = None,
         name: str | None = None,
-        configuration: Configuration | None = None,
     ) -> RequestQueueClient:
         """Create a request queue client."""
 
--- a/src/crawlee/storage_clients/_base/_storage_client.py
+++ b/src/crawlee/storage_clients/_base/_storage_client.py
@@ -6,6 +6,8 @@ from typing import TYPE_CHECKING
 from crawlee._utils.docs import docs_group
 
 if TYPE_CHECKING:
+    from collections.abc import Hashable
+
     from crawlee.configuration import Configuration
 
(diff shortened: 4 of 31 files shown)


janbuchar:
Can you please expand the PR description with an explanation of how this updated logic works with the Apify SDK? Namely I'm interested in the way it overrides the global storage client. For instance, 

- If I reconfigure the storage client in a crawler constructor, will that one be preserved after `Actor.init`? 
- If I reconfigure the storage client in the global service locator, will `Actor.init` keep it that way? Will it not crash?

Judging from https://github.com/apify/apify-sdk-python/pull/576, it should be fine. But we should make sure that this is covered by tests.

src/crawlee/storages/_base.py
```diff
@@ -36,15 +35,13 @@ async def open(
         *,
         id: str | None = None,
         name: str | None = None,
-        configuration: Configuration | None = None,
```
vdusek: Is this intentional? Doesn't look like.

src/crawlee/storages/_base.py
```diff
@@ -36,15 +35,13 @@ async def open(
...
         storage_client: StorageClient | None = None,
     ) -> Storage:
         """Open a storage, either restore existing or create a new one.
 
         Args:
             id: The storage ID.
             name: The storage name.
-            configuration: Configuration object used during the storage creation or restoration process.
```
vdusek: Is this intentional? Doesn't look like.

tests/unit/storages/test_storage_instance_manager.py
```diff
@@ -0,0 +1,123 @@
+from pathlib import Path
```
vdusek, line 1: Maybe we could use a parametrized fixture to test it across all storages (dataset, kvs, rq).

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -559,7 +572,12 @@ async def get_dataset(
         name: str | None = None,
     ) -> Dataset:
         """Return the `Dataset` with the given ID or name. If none is provided, return the default one."""
-        return await Dataset.open(id=id, name=name)
+        return await Dataset.open(
+            id=id,
+            name=name,
+            storage_client=self._service_locator.get_storage_client(),
+            configuration=self._service_locator.get_configuration(),
```
vdusek, line 586: okay, so we're passing this from basic crawler so that the storage does not use the value from the global service locator, correct?

src/crawlee/storage_clients/_base/_storage_client.py
```diff
@@ -28,6 +30,13 @@ class StorageClient(ABC):
     (where applicable), and consistent access patterns across all storage types it supports.
     """
 
+    def get_additional_cache_key(self, configuration: Configuration) -> Hashable:  # noqa: ARG002
```
vdusek, line 33:
```suggestion
    def get_additional_cache_key(self, _: Configuration) -> Hashable: 
```

src/crawlee/storages/_dataset.py
```diff
@@ -106,12 +106,16 @@ async def open(
...
+        additional_cache_key = storage_client.get_additional_cache_key(configuration=configuration)
+
         return await service_locator.storage_instance_manager.open_storage_instance(
             cls,
             id=id,
             name=name,
-            configuration=configuration,
-            client_opener=storage_client.create_dataset_client,
+            client_opener=client_opener,
+            storage_client_type=storage_client.__class__,
```
vdusek, line 121: okay - because we use the storage client class for caching, right?

src/crawlee/storages/_dataset.py
```diff
@@ -317,7 +319,6 @@ async def export_to(
         to_kvs_id: str | None = None,
         to_kvs_name: str | None = None,
         to_kvs_storage_client: StorageClient | None = None,
-        to_kvs_configuration: Configuration | None = None,
```
vdusek: Why?

tests/unit/storages/test_request_queue.py
```diff
@@ -23,30 +23,30 @@
...
         purge_on_start=True,
     )
+    service_locator.set_configuration(configuration)
+    return configuration
 
 
 @pytest.fixture
 async def rq(
     storage_client: StorageClient,
-    configuration: Configuration,
```
vdusek, line 44: So the `configuration` fixture is now being used implicitly even if we don't specify it as an argument? (maybe it is my limited knowledge of pytest)

tests/unit/storages/test_request_queue.py
```diff
@@ -23,30 +23,30 @@
...
 def storage_client(request: pytest.FixtureRequest) -> StorageClient:
     """Parameterized fixture to test with different storage clients."""
-    if request.param == 'memory':
-        return MemoryStorageClient()
-
-    return FileSystemStorageClient()
+    storage_client: StorageClient
+    storage_client = MemoryStorageClient() if request.param == 'memory' else FileSystemStorageClient()
+    service_locator.set_storage_client(storage_client)
+    return storage_client
```
vdusek: Don't do this as we will have the `sql` client

tests/unit/storages/test_request_queue.py
Pijukatel: Actually, I forgot to delete the configuration fixture, which is not needed at all, because in global autouse fixture `_isolate_test_environment` we set `monkeypatch.setenv('CRAWLEE_STORAGE_DIR', str(tmp_path))` and this ensures that any implicitly created configuration will point to this temp path - which is unique per test case.

src/crawlee/storages/_dataset.py
Pijukatel: Good catch. This was deleted by mistake. I reverted it and added test to cover those parameters.

src/crawlee/storages/_base.py
Pijukatel: Good catch, thanks. I am actually disappointed that mypy was not complaining about this.

src/crawlee/storages/_base.py
Pijukatel: Good catch, thanks.

tests/unit/storages/test_storage_instance_manager.py
Pijukatel: Ok, parametrized where it made sense.

src/crawlee/crawlers/_basic/_basic_crawler.py
Pijukatel: Exactly. And if there was no custom configuration/storage_client, then the crawler will have the same as the global service_locator has.

src/crawlee/storage_clients/_base/_storage_client.py
Pijukatel: I can't use it here. It can be called as named argument and child classes, for example, FilesystemStorageClient will use it.

src/crawlee/storages/_dataset.py
Pijukatel: Exactly

tests/unit/storages/test_request_queue.py
Pijukatel: This is kind of an edit by the linter, not really important. Once SQL client is merged, we can just expand the argument and add one more `elif`.

Pijukatel:
> Can you please expand the PR description with an explanation of how this updated logic works with the Apify SDK? Namely I'm interested in the way it overrides the global storage client. For instance,
> 
> Judging from [apify/apify-sdk-python#576](https://github.com/apify/apify-sdk-python/pull/576), it should be fine. But we should make sure that this is covered by tests.

The SDK PR includes the tests that cover the interaction of BasicCrawler and Actor with respect to init and services, the description is also included in the linked PR.

>     * If I reconfigure the storage client in a crawler constructor, will that one be preserved after `Actor.init`?

No, configuring the crawler with a custom storage_client will not set it in the global servcice_locator. This side effect was removed. 

>     * If I reconfigure the storage client in the global service locator, will `Actor.init` keep it that way? Will it not crash?

It will work and Actor.init will take it from global service locator, but if you set (explicitly or implicitly)  storage client in global service locator and then you try to use different storage client in Actor it will raise ServiceConflictError. See tests from linked PR:

```
async def test_existing_apify_config_respected_by_actor() -> None:
    """Set Apify Configuration in service_locator and verify that Actor respects it."""
    max_used_cpu_ratio = 0.123456  # Some unique value to verify configuration
    apify_config = ApifyConfiguration(max_used_cpu_ratio=max_used_cpu_ratio)
    service_locator.set_configuration(apify_config)
    async with Actor:
        pass

    returned_config = service_locator.get_configuration()
    assert returned_config is apify_config


async def test_existing_apify_config_throws_error_when_set_in_actor() -> None:
    """Test that passing explicit configuration to actor after service locator configuration was already set,
    raises exception."""
    service_locator.set_configuration(ApifyConfiguration())
    with pytest.raises(ServiceConflictError):
        async with Actor(configuration=ApifyConfiguration()):
            pass
```

janbuchar:
> > ```
> > * If I reconfigure the storage client in a crawler constructor, will that one be preserved after `Actor.init`?
> > ```
> 
> No, configuring the crawler with a custom storage_client will not set it in the global servcice_locator. This side effect was removed.

That means that this:

```py

crawler = BasicCrawler(storage_client=MemoryStorageClient())
# ...
await crawler.run()
dataset = await Dataset.open()
await dataset.export_data()`
```

will behave differently after this PR, correct? While I agree that it is better to be explicit about this, I'm pretty sure that it will surprise someone.

Pijukatel:
> That means that this:
> 
> ```python
> crawler = BasicCrawler(storage_client=MemoryStorageClient())
> # ...
> await crawler.run()
> dataset = await Dataset.open()
> await dataset.export_data()`
> ```
>
> will behave differently after this PR, correct? While I agree that it is better to be explicit about this, I'm pretty sure that it will surprise someone.

Yes, it can surprise people who are used to the old behavior. But I think this makes more sense, especially since we have public methods for getting storages on BasicCrawler. 
Example for dataset (Same situation for RQ and KVS):
```
# Depends on the global service_locator
default_dataset = await Dataset.open()
# Depends on the crawler service_locator (it can be the same as the previous, but it can be custom)
crawler_default_dataset = await crawler.get_dataset() 
```

Alternativelly you can set the storage_client globally and not pass it to the crawler `service_locator.set_storage_client(MemoryStorageClient())`

janbuchar:
> > That means that this:
> > ```python
> > crawler = BasicCrawler(storage_client=MemoryStorageClient())
> > # ...
> > await crawler.run()
> > dataset = await Dataset.open()
> > await dataset.export_data()
> > ```
> >   
> > will behave differently after this PR, correct? While I agree that it is better to be explicit about this, I'm pretty sure that it will surprise someone.
> 
> Yes, it can surprise people who are used to the old behavior. But I think this makes more sense, especially since we have public methods for getting storages on BasicCrawler. Example for dataset (Same situation for RQ and KVS):

Can we come up with some warning if this happens?

Another concern, if a crawler uses a slightly modified instance of `FilesystemStorageClient` (e.g., different path), `Actor.init` won't replace that. While it's true that it's more predictable, I believe it might worsen the DX... Also that way the user will have to use the specialized ApifyFilesystemStorageClient explicitly if they want to handle input.json as expected.

Pijukatel:
> ...
> Can we come up with some warning if this happens?


Adding warning and test to SDK PR https://github.com/apify/apify-sdk-python/pull/576/commits/5bf51f768e68bde620d81a13dff77ca7a4cddac5

docs/upgrading/upgrading_to_v1.md
```diff
@@ -5,6 +5,95 @@ title: Upgrading to v1
 
 This page summarizes the breaking changes between Crawlee for Python v0.6 and v1.0.
 
+## ServiceLocator changes
+
+### ServiceLocator is stricter with registering services.
```
vdusek:
```suggestion
### ServiceLocator is stricter with registering services
```

docs/upgrading/upgrading_to_v1.md
```diff
@@ -5,6 +5,95 @@ title: Upgrading to v1
...
+
+```python
+from crawlee import service_locator
+from crawlee.storage_clients import MemoryStorageClient
+
+service_locator.set_storage_client(MemoryStorageClient())
+service_locator.set_storage_client(MemoryStorageClient())  # Raises an error
+```
+
+### BasicCrawler has own instance of ServiceLocator to track its own services.
```
vdusek:
```suggestion
### BasicCrawler has its own instance of ServiceLocator to track its own services
```

docs/upgrading/upgrading_to_v1.md
```diff
@@ -5,6 +5,95 @@ title: Upgrading to v1
...
+
+
+async def main() -> None:
+    custom_storage_client = MemoryStorageClient()
+    crawler = BasicCrawler(storage_client=custom_storage_client)
+
+    assert service_locator.get_storage_client() is not custom_storage_client
+    assert await crawler.get_dataset() is not await Dataset.open()
+```
+### There can be two crawlers with different services at the same time.
```
vdusek:
```suggestion
### There can be two crawlers with different services at the same time
```

docs/upgrading/upgrading_to_v1.md
```diff
@@ -5,6 +5,95 @@ title: Upgrading to v1
...
+from crawlee.storage_clients import MemoryStorageClient
+
+service_locator.set_storage_client(MemoryStorageClient())
+service_locator.set_storage_client(MemoryStorageClient())  # Raises an error
+```
+
+### BasicCrawler has own instance of ServiceLocator to track its own services.
+Explicitly passed services to the crawler can be different the global ones accessible in `crawlee.service_locator`. `BasicCrawler` no longer causes the global services in `service_locator` to be set to the crawler's explicitly passed services.
+
+```python
```
vdusek:
```suggestion

**Before (v0.6):**
```python
```

docs/upgrading/upgrading_to_v1.md
vdusek: Could we merge the last 2 subsections? They are both related to the same change.

Basically thanks to the: "BasicCrawler has own instance of ServiceLocator to track its own services" it is possible to do "There can be two crawlers with different services at the same time".

docs/upgrading/upgrading_to_v1.md
```diff
@@ -5,6 +5,95 @@ title: Upgrading to v1
 
 This page summarizes the breaking changes between Crawlee for Python v0.6 and v1.0.
 
+## ServiceLocator changes
```
vdusek: Could we move this section after "New architecture of storage clients" and before "Other smaller updates"? The service locator is more of an edge component and isn't used that often.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -346,14 +347,23 @@ def __init__(
...
+            configuration = service_locator.get_configuration()
 
-        config = service_locator.get_configuration()
+        if not storage_client:
+            storage_client = service_locator.get_storage_client()
+
+        if not event_manager:
+            # This is weird if someone passes configuration and its event manager related stuff gets ignored as the
+            # event manager will be used from service_locator. Maybe keep the was created flag for it? It does not have
+            # the use cases like the storages
```
vdusek: what are we gonna do with this? :) maybe warning log?

src/crawlee/storages/_dataset.py
```diff
@@ -107,13 +107,17 @@ async def open(
...
+
         return await service_locator.storage_instance_manager.open_storage_instance(
             cls,
             id=id,
             name=name,
             alias=alias,
-            configuration=configuration,
-            client_opener=storage_client.create_dataset_client,
+            client_opener=client_opener,
+            storage_client_type=storage_client.__class__,
```
vdusek, line 121: Maybe just a string literal would also be good, definitely not a request, probably subjective

src/crawlee/storages/_key_value_store.py
```diff
@@ -119,13 +119,17 @@ async def open(
...
         storage_client = service_locator.get_storage_client() if storage_client is None else storage_client
 
+        client_opener = storage_client.create_kvs_client(id=id, name=name, alias=alias, configuration=configuration)
+        additional_cache_key = storage_client.get_additional_cache_key(configuration=configuration)
+
         return await service_locator.storage_instance_manager.open_storage_instance(
             cls,
             id=id,
             name=name,
+            client_opener=client_opener,
```
vdusek: This is really misleading name since you reworked it and it's not an opener (function) anymore. This should be more `client_coro` or something.

src/crawlee/storages/_storage_instance_manager.py
```diff
@@ -1,21 +1,49 @@
...
+        default_factory=lambda: defaultdict(lambda: defaultdict(lambda: defaultdict()))
+    )
+    """Cache for storage instances by alias, separated by storage type and additional hash key."""
+
+
 StorageClientType = DatasetClient | KeyValueStoreClient | RequestQueueClient
 """Type alias for the storage client types."""
 
-ClientOpener = Callable[..., Awaitable[StorageClientType]]
+ClientOpener = Coroutine[None, None, StorageClientType]
```
vdusek: ClientOpener - rename it please (all occurrences), as I pointed out somewhere else

src/crawlee/storages/_storage_instance_manager.py
```diff
@@ -48,100 +66,113 @@ async def open_storage_instance(
...
+
+            # Validate input parameters.
+            specified_params = sum(1 for param in [id, name, alias] if param is not None)
+            if specified_params > 1:
+                raise ValueError('Only one of "id", "name", or "alias" can be specified, not multiple.')
+
+            # Auto-set alias='default' when no parameters are specified.
+            # Default unnamed storage is equal to alias=default unnamed storage.
+            if specified_params == 0:
+                alias = self._DEFAULT_STORAGE_ALIAS
```
vdusek, line 105: We should probably normalize this in an opposite direction - when alias=default, let's use alias=None. Because of the implementation in the SDK - to not to create a record with the default mapping in the default KVS, because this is resolved via the DEFAULT_... env vars. Could you do it here please?

tests/unit/crawlers/_basic/test_basic_crawler.py
```diff
@@ -1051,9 +1052,125 @@ async def test_sets_services() -> None:
...
+    service_locator.set_event_manager(custom_event_manager)
+
+    custom_storage_client = MemoryStorageClient()
+    service_locator.set_storage_client(custom_storage_client)
+
+    basic_crawler = BasicCrawler()
+
+    assert basic_crawler._service_locator.get_configuration() is custom_configuration
+    assert basic_crawler._service_locator.get_event_manager() is custom_event_manager
+    assert basic_crawler._service_locator.get_storage_client() is custom_storage_client
```
vdusek, line 1074: we're touching the internal attr, but i'm ok with that

tests/unit/storage_clients/_memory/test_memory_kvs_client.py
```diff
@@ -5,7 +5,6 @@
 
 import pytest
 
-from crawlee.configuration import Configuration
```
vdusek, line 8: so all these configurations removal are possible because they are set in the conftest now, correct?

src/crawlee/storages/_storage_instance_manager.py
```diff
@@ -48,100 +66,113 @@ async def open_storage_instance(
...
+        storage_client_type: type[StorageClient],
         client_opener: ClientOpener,
+        additional_cache_key: Hashable = '',
     ) -> T:
         """Open a storage instance with caching support.
 
         Args:
             cls: The storage class to instantiate.
             id: Storage ID.
-            name: Storage name (global scope, persists across runs).
```
Mantisus, line 59: Should this really be removed?

src/crawlee/configuration.py
```diff
@@ -28,7 +28,7 @@ class Configuration(BaseSettings):
     Settings can also be configured via environment variables, prefixed with `CRAWLEE_`.
     """
 
-    model_config = SettingsConfigDict(populate_by_name=True)
+    model_config = SettingsConfigDict(validate_by_name=True, validate_by_alias=True)
```
Mantisus, line 31: `validate_by_name` was only added in pydantic v2.11 - https://docs.pydantic.dev/latest/api/config/?query=validate_by_name#pydantic.config.ConfigDict.validate_by_name

This update is not compatible with the current pydantic versions in `pyproject.toml` - “pydantic>=2.8.0,!=2.10.0,!=2.10.1,!=2.10.2”,

tests/unit/crawlers/_basic/test_basic_crawler.py
Mantisus: I would consider creating `@property` for `service_locator` in `BasicCrawler`.

Mantisus (commented):
I like it.

Just a few comments.

src/crawlee/configuration.py
vdusek: Let's use `>= 2.11` then

docs/upgrading/upgrading_to_v1.md
Pijukatel: Ok

src/crawlee/configuration.py
Pijukatel: Good catch, I think I do not need this change.

src/crawlee/crawlers/_basic/_basic_crawler.py
Pijukatel: Added warning.

src/crawlee/storages/_key_value_store.py
Pijukatel: I am not sure about this. I would like to avoid putting type into the name. It changed from an opener async function to an opener coroutine, but the purpose is the same.

It was not introduced in any release yet, so there is no one to be confused by this change.

Anyway, if you have a good name suggestion, I am open to it.

src/crawlee/storages/_storage_instance_manager.py
Pijukatel: I will keep it.

src/crawlee/storages/_storage_instance_manager.py
Pijukatel: Will be handled in the SDK.

tests/unit/crawlers/_basic/test_basic_crawler.py
Pijukatel: Yes, touching internal in test is not great, but making it public property is a bigger commitment. I am not sure we should make it public just for the sake of the unit tests. Let's start like this and update it if the need arises.

tests/unit/storage_clients/_memory/test_memory_kvs_client.py
Pijukatel: Yes

docs/upgrading/upgrading_to_v1.md
vdusek: This is not resolved

src/crawlee/storages/_storage_instance_manager.py
```diff
@@ -26,20 +54,10 @@ class StorageInstanceManager:
     and provides a unified interface for opening and managing storage instances.
     """
 
-    _DEFAULT_STORAGE = 'default'
+    _DEFAULT_STORAGE_ALIAS = '__default__'
```
vdusek, line 57:
```suggestion
    _DEFAULT_STORAGE_ALIAS = '__default__'
    """Reserved alias for default unnamed storage."""
```

src/crawlee/storages/_storage_instance_manager.py
vdusek: Okay

docs/upgrading/upgrading_to_v1.md
Pijukatel: Done

pyproject.toml
```diff
@@ -40,7 +40,7 @@ dependencies = [
     "protego>=0.5.0",
     "psutil>=6.0.0",
     "pydantic-settings>=2.2.0,!=2.7.0,!=2.7.1,!=2.8.0",
-    "pydantic>=2.8.0,!=2.10.0,!=2.10.1,!=2.10.2",
+    "pydantic>= 2.11.0",
```
vdusek: extra white space
```suggestion
    "pydantic>=2.11.0",
```

PR #1401, opened 2025-09-09
feat: Add support for NDU storages

vdusek:
### Description

- The alias version was implemented, see the https://github.com/apify/crawlee-python/issues/1175#issuecomment-3244739703 for more context.
- Storage with `alias='default'` is the default unnamed storage.

### Issues

- Closes: #1175

### Testing

- New tests were implemented.

### Checklist

- [x] CI passed

### DIFF
--- a/src/crawlee/storage_clients/_memory/_key_value_store_client.py
+++ b/src/crawlee/storage_clients/_memory/_key_value_store_client.py
@@ -59,6 +59,9 @@ class MemoryKeyValueStoreClient(KeyValueStoreClient):
         memory KVS don't check for existing stores with the same name or ID since all data exists only in memory
         and is lost when the process terminates.
 
+        Alias does not have any effect on the memory storage client implementation, because unnamed storages
+        are supported by default, since data are not persisted.
+
         Args:
             id: The ID of the key-value store. If not provided, a random ID will be generated.
             name: The name of the key-value store for named (global scope) storages.
--- a/src/crawlee/storage_clients/_memory/_dataset_client.py
+++ b/src/crawlee/storage_clients/_memory/_dataset_client.py
@@ -61,6 +61,9 @@ class MemoryDatasetClient(DatasetClient):
         datasets don't check for existing datasets with the same name or ID since all data exists only in memory
         and is lost when the process terminates.
 
+        Alias does not have any effect on the memory storage client implementation, because unnamed storages
+        are supported by default, since data are not persisted.
+
         Args:
             id: The ID of the dataset. If not provided, a random ID will be generated.
             name: The name of the dataset for named (global scope) storages.
--- a/src/crawlee/storage_clients/_memory/_request_queue_client.py
+++ b/src/crawlee/storage_clients/_memory/_request_queue_client.py
@@ -71,6 +71,9 @@ class MemoryRequestQueueClient(RequestQueueClient):
         memory queues don't check for existing queues with the same name or ID since all data exists only
         in memory and is lost when the process terminates.
 
+        Alias does not have any effect on the memory storage client implementation, because unnamed storages
+        are supported by default, since data are not persisted.
+
         Args:
             id: The ID of the request queue. If not provided, a random ID will be generated.
             name: The name of the request queue for named (global scope) storages.


vdusek:
cc @Mantisus in context of https://github.com/apify/crawlee-python/pull/1339

janbuchar:
Can you add some docs/examples please? :slightly_smiling_face:

tests/unit/storages/test_dataset.py
```diff
@@ -582,3 +583,493 @@ async def test_purge(
...
+    """Test alias parameter validation."""
+    # Should not allow both name and alias
+    with pytest.raises(ValueError, match='Only one of'):
+        await Dataset.open(
+            name='test',
+            alias='test',
+            storage_client=storage_client,
+            configuration=configuration,
+        )
+
```
Mantisus: This test essentially duplicates the tests `test_alias_with_name_error` and `test_alias_parameter`

tests/unit/storages/test_request_queue.py
```diff
@@ -642,3 +642,681 @@ async def test_purge(
...
+    await rq_2.add_request('https://example2.com')
+
+    # Verify data isolation
+    request_1 = await rq_1.fetch_next_request()
+    request_2 = await rq_2.fetch_next_request()
+
+    assert request_1 is not None
+    assert request_2 is not None
+    assert request_1.url == 'https://example1.com'
+    assert request_2.url == 'https://example2.com'
```
Mantisus: If both requests were placed in the same queue, the result would be the same. 🙂

It might be worth adding two requests to `rq1` to clearly show that `rq_2.fetch_next_request()` gets a request from another isolated queue.

tests/unit/storages/test_key_value_store.py
```diff
@@ -598,3 +599,528 @@ async def test_record_exists_after_purge(kvs: KeyValueStore) -> None:
...
+) -> None:
+    """Test alias parameter validation."""
+    # Should not allow both name and alias
+    with pytest.raises(ValueError, match='Only one of'):
+        await KeyValueStore.open(
+            name='test',
+            alias='test',
+            storage_client=storage_client,
+            configuration=configuration,
+        )
```
Mantisus: This test essentially duplicates the tests `test_alias_with_name_error` and `test_alias_parameter`

tests/unit/storages/test_request_queue.py
```diff
@@ -642,3 +642,682 @@ async def test_purge(
...
+
+    # Data should be shared
+    await default_rq.add_request('https://default.example.com')
+    metadata = await alias_default_rq.get_metadata()
+    assert metadata.pending_request_count == 1
+
+    await default_rq.drop()
+
+
+async def test_alias_parameter_validation(
```
Mantisus: This test essentially duplicates the tests `test_alias_with_name_error` and `test_alias_parameter`

vdusek:
> Can you add some docs/examples please? 🙂

I extended the Storages guide.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -781,20 +784,22 @@ async def get_data(
 
         Args:
             dataset_id: The ID of the `Dataset`.
-            dataset_name: The name of the `Dataset`.
+            dataset_name: The name of the `Dataset` (global scope).
```
Mantisus: Perhaps, instead of using `run` and `global` scopes, it would be better to use the same approach as in clients and documentation, `named` and `unnamed` storages?

```
name: The name of the dataset for named storages. Mutually exclusive with alias.
alias: The alias of the dataset for unnamed storages. Mutually exclusive with name.
```

docs/guides/storages.mdx
```diff
@@ -186,13 +205,7 @@ Crawlee provides the following helper function to simplify interactions with the
...
 ## Cleaning up the storages
 
-By default, Crawlee automatically cleans up **default storages** before each crawler run to ensure a clean state. This behavior is controlled by the <ApiLink to="class/Configuration#purge_on_start">`Configuration.purge_on_start`</ApiLink> setting (default: `True`).
-
-### What gets purged
-
-- **Default storages** are completely removed and recreated at the start of each run, ensuring that you start with a clean slate.
-- **Named storages** are never automatically purged and persist across runs.
-- The behavior depends on the storage client implementation.
+By default, Crawlee cleans up all unnamed storages (including the default one) at the start of each run, so every crawl begins with a clean state. This behavior is controlled by <ApiLink to="class/Configuration#purge_on_start">`Configuration.purge_on_start`</ApiLink> (default: True). In contrast, named storages are never purged automatically and persist across runs. The exact behavior may vary depending on the storage client implementation.
```
Pijukatel, line 208: Maybe it would be worth mentioning that the `INPUT.json` in the default kvs is locally not purged as an example of a specific storage client convenience.

src/crawlee/storage_clients/_file_system/_dataset_client.py
```diff
@@ -104,17 +102,23 @@ async def open(
...
 
         Raises:
-            ValueError: If a dataset with the specified ID is not found, or if metadata is invalid.
+            ValueError: If a dataset with the specified ID is not found, if metadata is invalid,
+                or if both name and alias are provided.
         """
-        storage_dir = Path(configuration.storage_dir)
-        dataset_base_path = storage_dir / cls._STORAGE_SUBDIR
+        # Validate input parameters.
+        specified_params = sum(1 for param in [id, name, alias] if param is not None)
```
Pijukatel, line 117: Looks like a counter with this condition:
Counter([id, name, alias])[None] < 2

Whatever you prefer.

But more importantly, this exact piece of code is repeated 7 times. I think it would be worth extracting to its own method/function and reuse.

src/crawlee/storages/_storage_instance_manager.py
```diff
@@ -82,8 +93,38 @@ async def open_storage_instance(
...
+                cached_instance = type_cache_by_alias[alias]
+                if isinstance(cached_instance, cls):
+                    return cached_instance
+
+        # Check for conflicts between named and alias storages
+        if name is not None:
+            # Check if there's already an alias storage with the same identifier
+            type_cache_by_alias = self._cache_by_alias.get(cls, {})
+            if name in type_cache_by_alias:
+                raise ValueError(
```
Pijukatel, line 108: I do not think we should have this constraint. Name and alias are different things as we try to tell the user, so they should not have collisions. I think it should be up to the `StorageClient` to ensure they do not collide.

For example, I would suggest adding this test:

```
async def test_memory_named_unnamed_collision() -> None:
    """Test that named dataset and unnamed dataset does not collide on the filesystem."""
    test_data = {"a":"b"}
    named_dataset_client = await FileSystemStorageClient().create_dataset_client(
        name='test_dataset',
    )

    unnamed_dataset_client = await FileSystemStorageClient().create_dataset_client(
        alias='test_dataset',
    )
    assert named_dataset_client is not unnamed_dataset_client
    assert (await unnamed_dataset_client.get_data()).items == []
    assert (await named_dataset_client.get_data()).items == []

    await named_dataset_client.push_data(test_data)
    assert (await named_dataset_client.get_data()).items == [test_data]
    assert (await unnamed_dataset_client.get_data()).items == [], "Changes in named dataset should not affect unnamed dataset"
```

vdusek: We discussed this with Honza when you were on vacation and came to an agreement that they should not collide. cc @janbuchar

Pijukatel:
Should `StorageMetadata` contain some information regarding the alias or scope? How will, for example Apify platform know what kind of storage it is?

docs/guides/storages.mdx
vdusek: The `INPUT.json` is handled specifically in the SDK, not here :slightly_smiling_face:.

src/crawlee/storage_clients/_file_system/_dataset_client.py
vdusek: Using:
```
Counter([id, name, alias])[None] < 2
```
results in Ruff complaining about using magic values :smile:.

> But more importantly, this exact piece of code is repeated 7 times. I think it would be worth extracting to its own method/function and reuse.

TBH, I don't see any benefit here; it's really clear, and effectively two lines of code.

vdusek:
> Should StorageMetadata contain some information regarding the alias or scope? How will, for example Apify platform know what kind of storage it is?

I don't think so. On the Apify platform, the distinction between global scope and run scope storage is based just on naming - unnamed versus named. The alias is just for us (FS and Apify clients use it, or will use it).

src/crawlee/storage_clients/_file_system/_dataset_client.py
Pijukatel: It is reused 7x times just in Crawlee for now, but it will be repeated in SDK 3x when updating the ApifyStorageClient,  3x for new SQL client, and probably 3x for the ApifyStorageClient that does not use locks...

That is 16 uses of the same 3 lines of code. I think that such an amount of re-use justifies the extraction.

src/crawlee/storages/_storage_instance_manager.py
Pijukatel: I am just worried that it further blurs the difference between the name and the alias and makes it more confusing.

Especially when looking at the following:

- It is not possible to have unnamed storage of the same alias as the existing name of named storage because they conflict in the `StorageInstanceManager`
- `StorageInstanceManager` knows there can be no two storage instances of the same alias and name. This effectively makes the name and alias the same caching key, but not really, as we still want to distinguish between them.

Pijukatel: Also this constraint is enforced only on the `StorageInstanceManager` level as shown with test I pasted in previous comment, the behavior will be different if using `FileSystemStorageClient` directly, where it will not be enforced and will lead to strange behavior.

janbuchar: Yeah, the possible drawbacks (somebody has to pick a different name/alias after getting an error message) are IMO smaller than in the opposite case (every storage client has to resolve this, filesystem storage has to invent some way to distinguish name/alias).

docs/guides/storages.mdx
Pijukatel: I think we already discussed that the `FileSystemKeyValueStoreClient` should not clean the `INPUT.json` as it is convenient for any local development, but I am afraid it had slipped away. That should work independently of the SDK I believe.

vdusek: Check [this](https://github.com/apify/apify-sdk-python/blob/master/src/apify/storage_clients/_file_system/_key_value_store_client.py) out.

src/crawlee/storages/_storage_instance_manager.py
vdusek: > the behavior will be different if using FileSystemStorageClient directly, where it will not be enforced and will lead to strange behavior.

Btw. this is not how it should be used, you should always use `Storage.open(storage_client=...)`.

src/crawlee/storage_clients/_file_system/_dataset_client.py
vdusek: I can shorten it to two lines if you'd prefer it:

```python
    if sum(1 for param in [id, name, alias] if param is not None) > 1:
        raise ValueError('Only one of "id", "name", or "alias" can be specified, not multiple.')
```

Really, there's no need to hide this trivial check in a helper function; I even consider it less readable to push a simple argument check into a separate function and raise an exception from there. It should be the function's responsibility to verify its inputs. Don't over-engineer, and please let's move on to more serious stuff; this is not productive at all.

docs/guides/storages.mdx
Pijukatel: Yes, I forgot about that, sorry

src/crawlee/storage_clients/_memory/_key_value_store_client.py
```diff
@@ -51,6 +51,7 @@ async def open(
         *,
         id: str | None,
         name: str | None,
+        alias: str | None,
```
janbuchar, line 54: um, shouldn't this actually be, you know, used somewhere in the method? I imagine you could just do `if alias: name = alias`, but I don't see that here - am I missing something?

I guess this might work that to the `StorageInstanceManager` doing the actual work needed to distinguish the instances, but it's kinda hard to see at first, if that's the case. Also we might want to put the `alias` in the metadata?

tests/unit/storages/test_dataset.py
```diff
@@ -582,3 +583,469 @@ async def test_purge(
...
+        storage_client=storage_client,
+        configuration=configuration,
+    )
+
+    # Should be same instance
+    assert dataset_1 is dataset_2
+    assert dataset_1.id == dataset_2.id
+
+    # Clean up
+    await dataset_1.drop()
```
janbuchar, line 652: Doesn't the `storage_client` fixture take care of this? And if it doesn't, it definitely should...

vdusek:
> Looks pretty good to me! I have some nitpicky comments, plus I'd like to ask if you already tried implementing this in ApifyStorageClient. It would be good to see that before merging this part (not mandatory I guess...).

I've got it working, it just needs some more polishing and tests. Anyway, there were no issues with the implementation, and it's pretty straightforward. Also, no BCs, as expected.

tests/unit/storages/test_dataset.py
vdusek: `StorageClient` class does not have access, and so cannot drop the specific storage clients that were created using it.

janbuchar: I see. I guess it's fine then, even though you could theoretically set up some monkey patching to get that access

src/crawlee/storage_clients/_memory/_key_value_store_client.py
vdusek: As we discussed on Slack, the alias does not have any effect on the memory storage client implementation. I added additional explanatory text to the docstrings.

> Also we might want to put the alias in the metadata?

We might, but I'm not 100% sure about it. I would say we can better add it later than remove it later. So I would suggest keeping it as it is and adding it later if we have any reasoning/request for it.

vdusek:
I'm merging this as all the comments were clarified and/or addressed, so that we can adopt this in https://github.com/apify/crawlee-python/pull/1386 and https://github.com/apify/crawlee-python/pull/1339.

PR #1406, opened 2025-09-13
feat:  Add `RedisStorageClient`

Mantisus:
### Description

- TODO

### Testing

- TODO

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -91,6 +91,7 @@ dev = [
     "apify_client", # For e2e tests.
     "build~=1.3.0", # For e2e tests.
     "dycw-pytest-only~=2.1.0",
+    "fakeredis[probabilistic,json,lua]>=2.31.0",
     "mypy~=1.17.0",
     "pre-commit~=4.3.0",
     "proxy-py~=2.4.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,7 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-all = ["crawlee[adaptive-crawler,beautifulsoup,cli,curl-impersonate,httpx,parsel,playwright,otel]"]
+all = ["crawlee[adaptive-crawler,beautifulsoup,cli,curl-impersonate,httpx,parsel,playwright,otel,redis]"]
 adaptive-crawler = [
     "jaro-winkler>=2.0.3",
     "playwright>=1.27.0",
@@ -71,6 +71,7 @@ otel = [
     "opentelemetry-semantic-conventions>=0.54",
     "wrapt>=1.17.0",
 ]
+redis = ["redis[hiredis] >= 6.4.0"]
 
 [project.scripts]
 crawlee = "crawlee._cli:cli"
--- a/src/crawlee/storage_clients/_redis/_client_mixin.py
+++ b/src/crawlee/storage_clients/_redis/_client_mixin.py
@@ -1,0 +1,90 @@
+from __future__ import annotations
+
+import asyncio
+from abc import ABC
+from contextlib import asynccontextmanager
+from logging import getLogger
+from pathlib import Path
+from typing import TYPE_CHECKING, ClassVar, TypedDict
+
+from ._utils import await_redis_response, read_lua_script
+
+if TYPE_CHECKING:
+    from collections.abc import AsyncIterator
+
+    from redis.asyncio import Redis
+    from redis.asyncio.client import Pipeline
+    from redis.commands.core import AsyncScript
+    from typing_extensions import NotRequired
+
+
+logger = getLogger(__name__)
+
+
+class MetadataUpdateParams(TypedDict, total=False):
+    """Parameters for updating metadata."""
+
+    update_accessed_at: NotRequired[bool]
+    update_modified_at: NotRequired[bool]
+    force: NotRequired[bool]
+
+
+class RedisClientMixin(ABC):
+    """Mixin class for SQL clients.
+
+    This mixin provides common SQL operations and basic methods for SQL storage clients.
+    """
+
+    _DEFAULT_NAME = 'default'
+
+    _MAIN_KEY: ClassVar[str]
+
+    def __init__(self, *, storage_name: str, redis: Redis) -> None:
+        self._storage_name = storage_name
+        self._redis = redis
+
+    @classmethod
+    async def _get_metadata_by_name(cls, name: str, redis: Redis) -> dict | None:
+        response = await await_redis_response(redis.json().get(f'{cls._MAIN_KEY}:{name}:metadata'))
+        data = response[0] if response is not None and isinstance(response, list) else response
+        if data is not None and not isinstance(data, dict):
+            raise TypeError('The metadata data was received in an incorrect format.')
+        return data
+
+    @classmethod
+    async def _get_metadata_name_by_id(cls, id: str, redis: Redis) -> str | None:
+        return await await_redis_response(redis.get(f'{cls._MAIN_KEY}:id_to_name:{id}'))
+
+    @asynccontextmanager
+    async def _get_pipeline(self, *, with_execute: bool = True) -> AsyncIterator[Pipeline]:
+        """Create a new Redis pipeline for this storage."""
+        async with self._redis.pipeline() as pipe:
+            try:
+                pipe.multi()  # type: ignore[no-untyped-call]
+                yield pipe
+            finally:
+                if with_execute:
+                    await pipe.execute()
+
+    async def _create_storage(self, pipeline: Pipeline) -> None:
+        _pipeline = pipeline  # To avoid unused variable mypy error
+
+    async def _create_script(self, script_name: str) -> AsyncScript:
+        """Load a Lua script from a file and return a Script object."""
+        script_path = Path(__file__).parent / 'lua_scripts' / script_name
+        script_content = await asyncio.to_thread(read_lua_script, script_path)
+
+        return self._redis.register_script(script_content)
+
+    async def _create_metadata_and_storage(self, metadata: dict) -> None:
+        metadata_key = f'{self._MAIN_KEY}:{self._storage_name}:metadata'
+        index_id_to_name = f'{self._MAIN_KEY}:id_to_name:{metadata["id"]}'
+        metadata['created_at'] = metadata['created_at'].isoformat()
+        metadata['accessed_at'] = metadata['accessed_at'].isoformat()
+        metadata['modified_at'] = metadata['modified_at'].isoformat()
+        name = metadata['name'] if metadata['name'] is not None else self._DEFAULT_NAME
+        # Use a transaction to ensure atomicity
+        async with self._get_pipeline() as pipe:
+            await await_redis_response(pipe.json().set(metadata_key, '$', metadata, nx=True))
+            await await_redis_response(pipe.set(index_id_to_name, name, nx=True))
+            await self._create_storage(pipe)
--- a/src/crawlee/storage_clients/_redis/_dataset_client.py
+++ b/src/crawlee/storage_clients/_redis/_dataset_client.py
@@ -1,0 +1,348 @@
+from __future__ import annotations
+
+from datetime import datetime, timezone
+from logging import getLogger
+from typing import TYPE_CHECKING, Any, cast
+
+from typing_extensions import override
+
+from crawlee._utils.crypto import crypto_random_object_id
+from crawlee.storage_clients._base import DatasetClient
+from crawlee.storage_clients.models import DatasetItemsListPage, DatasetMetadata
+
+from ._client_mixin import RedisClientMixin
+from ._utils import await_redis_response
+
+if TYPE_CHECKING:
+    from collections.abc import AsyncIterator
+
+    from redis.asyncio import Redis
+    from redis.asyncio.client import Pipeline
+
+logger = getLogger(__name__)
+
+
+class RedisDatasetClient(DatasetClient, RedisClientMixin):
+    """Memory implementation of the dataset client.
+
+    This client stores dataset items in memory using Python lists and dictionaries. No data is persisted
+    between process runs, meaning all stored data is lost when the program terminates. This implementation
+    is primarily useful for testing, development, and short-lived crawler operations where persistent
+    storage is not required.
+
+    The memory implementation provides fast access to data but is limited by available memory and
+    does not support data sharing across different processes. It supports all dataset operations including
+    sorting, filtering, and pagination, but performs them entirely in memory.
+    """
+
+    _DEFAULT_NAME = 'default'
+
+    _MAIN_KEY = 'dataset'
+
+    def __init__(
+        self,
+        dataset_name: str,
+        redis: Redis,
+    ) -> None:
+        """Initialize a new instance.
+
+        Preferably use the `MemoryDatasetClient.open` class method to create a new instance.
+        """
+        super().__init__(storage_name=dataset_name, redis=redis)
+
+    @override
+    async def get_metadata(self) -> DatasetMetadata:
+        metadata_dict = await self._get_metadata_by_name(name=self._storage_name, redis=self._redis)
+        if metadata_dict is None:
+            raise ValueError(f'Dataset with name "{self._storage_name}" does not exist.')
+        return DatasetMetadata.model_validate(metadata_dict)
+
+    @classmethod
+    async def open(
+        cls,
+        *,
+        id: str | None,
+        name: str | None,
+        redis: Redis,
+    ) -> RedisDatasetClient:
+        """Open or create a new Redis dataset client.
+
+        This method creates a new Redis dataset instance. Unlike persistent storage implementations, Redis
+        datasets don't check for existing datasets with the same name or ID since all data exists only in memory
+        and is lost when the process terminates.
+
+        Args:
+            id: The ID of the dataset. If not provided, a random ID will be generated.
+            name: The name of the dataset. If not provided, the dataset will be unnamed.
+            redis: Redis client instance.
+
+        Returns:
+            An instance for the opened or created storage client.
+        """
+        if id:
+            dataset_name = await cls._get_metadata_name_by_id(id=id, redis=redis)
+            if dataset_name is None:
+                raise ValueError(f'Dataset with ID "{id}" does not exist.')
+        else:
+            search_name = name or cls._DEFAULT_NAME
+            metadata_data = await cls._get_metadata_by_name(name=search_name, redis=redis)
+            dataset_name = search_name if metadata_data is not None else None
+        if dataset_name:
+            client = cls(dataset_name=dataset_name, redis=redis)
+            async with client._get_pipeline() as pipe:
+                await client._update_metadata(pipe, update_accessed_at=True)
+        else:
+            now = datetime.now(timezone.utc)
+            metadata = DatasetMetadata(
+                id=crypto_random_object_id(),
+                name=name,
+                created_at=now,
+                accessed_at=now,
+                modified_at=now,
+                item_count=0,
+            )
+            dataset_name = name or cls._DEFAULT_NAME
+            client = cls(dataset_name=dataset_name, redis=redis)
+            await client._create_metadata_and_storage(metadata.model_dump())
+        return client
+
+    @override
+    async def _create_storage(self, pipeline: Pipeline) -> None:
+        items_key = f'{self._MAIN_KEY}:{self._storage_name}:items'
+        await await_redis_response(pipeline.json().set(items_key, '$', []))
+
+    @override
+    async def drop(self) -> None:
+        storage_id = (await self.get_metadata()).id
+        async with self._get_pipeline() as pipe:
+            await pipe.delete(f'{self._MAIN_KEY}:{self._storage_name}:metadata')
+            await pipe.delete(f'{self._MAIN_KEY}:{self._storage_name}:items')
+            await pipe.delete(f'{self._MAIN_KEY}:id_to_name:{storage_id}')
+
+    @override
+    async def purge(self) -> None:
+        async with self._get_pipeline() as pipe:
+            await self._create_storage(pipe)
+
+            await self._update_metadata(
+                pipe,
+                update_accessed_at=True,
+                update_modified_at=True,
+                new_item_count=0,
+            )
+
+    @override
+    async def push_data(self, data: list[dict[str, Any]] | dict[str, Any]) -> None:
+        if isinstance(data, dict):
+            data = [data]
+
+        async with self._get_pipeline() as pipe:
+            # Incorrect signature for args type in redis-py
+            pipe.json().arrappend(f'{self._MAIN_KEY}:{self._storage_name}:items', '$', *data)  # type: ignore[arg-type]
+            delta_item_count = len(data)
+            await self._update_metadata(
+                pipe, update_accessed_at=True, update_modified_at=True, delta_item_count=delta_item_count
+            )
+
+    @override
+    async def get_data(
+        self,
+        *,
+        offset: int = 0,
+        limit: int | None = 999_999_999_999,
+        clean: bool = False,
+        desc: bool = False,
+        fields: list[str] | None = None,
+        omit: list[str] | None = None,
+        unwind: list[str] | None = None,
+        skip_empty: bool = False,
+        skip_hidden: bool = False,
+        flatten: list[str] | None = None,
+        view: str | None = None,
+    ) -> DatasetItemsListPage:
+        # Check for unsupported arguments and log a warning if found
+        # When implementing, explore the capabilities of jsonpath to determine what can be done at the Redis level.
+        unsupported_args: dict[str, Any] = {
+            'clean': clean,
+            'fields': fields,
+            'omit': omit,
+            'unwind': unwind,
+            'skip_hidden': skip_hidden,
+            'flatten': flatten,
+            'view': view,
+        }
+        unsupported = {k: v for k, v in unsupported_args.items() if v not in (False, None)}
+
+        if unsupported:
+            logger.warning(
+                f'The arguments {list(unsupported.keys())} of get_data are not supported '
+                f'by the {self.__class__.__name__} client.'
+            )
+
+        metadata = await self.get_metadata()
+
+        total = metadata.item_count
+        items_key = f'{self._MAIN_KEY}:{self._storage_name}:items'
+        json_path = '$'
+
+        # Apply sorting and pagination
+        if desc:
+            if offset and limit is not None:
+                json_path += f'[-{offset + limit}:-{offset}]'
+            elif limit is not None:
+                json_path += f'[-{limit}:]'
+            elif offset:
+                json_path += f'[:-{offset}]'
+        else:  # noqa: PLR5501 # not a mistake, just to please the linter
+            if offset and limit is not None:
+                json_path += f'[{offset}:{offset + limit}]'
+            elif limit is not None:
+                json_path += f'[:{limit}]'
+            elif offset:
+                json_path += f'[{offset}:]'
+
+        if json_path == '$':
+            json_path = '$[*]'
+
+        data = await await_redis_response(self._redis.json().get(items_key, json_path))
+
+        if data is None:
+            data = []
+
+        if skip_empty:
+            data = [item for item in data if item]
+
+        if desc:
+            data = list(reversed(data))
+
+        async with self._get_pipeline() as pipe:
+            await self._update_metadata(pipe, update_accessed_at=True)
+
+        return DatasetItemsListPage(
+            count=len(data),
+            offset=offset,
+            limit=limit or (total - offset),
+            total=total,
+            desc=desc,
+            items=data,
+        )
+
+    @override
+    async def iterate_items(
+        self,
+        *,
+        offset: int = 0,
+        limit: int | None = None,
+        clean: bool = False,
+        desc: bool = False,
+        fields: list[str] | None = None,
+        omit: list[str] | None = None,
+        unwind: list[str] | None = None,
+        skip_empty: bool = False,
+        skip_hidden: bool = False,
+    ) -> AsyncIterator[dict[str, Any]]:
+        """Iterate over dataset items one by one.
+
+        This method yields items individually instead of loading all items at once,
+        which is more memory efficient for large datasets.
+        """
+        # Log warnings for unsupported arguments
+        unsupported_args: dict[str, Any] = {
+            'clean': clean,
+            'fields': fields,
+            'omit': omit,
+            'unwind': unwind,
+            'skip_hidden': skip_hidden,
+        }
+        unsupported = {k: v for k, v in unsupported_args.items() if v not in (False, None)}
+
+        if unsupported:
+            logger.warning(
+                f'The arguments {list(unsupported.keys())} of iterate_items are not supported '
+                f'by the {self.__class__.__name__} client.'
+            )
+
+        metadata = await self.get_metadata()
+        total_items = metadata.item_count
+        items_key = f'{self._MAIN_KEY}:{self._storage_name}:items'
+
+        # Calculate actual range based on parameters
+        start_idx = offset
+        end_idx = min(total_items, offset + limit) if limit is not None else total_items
+
+        # Update accessed_at timestamp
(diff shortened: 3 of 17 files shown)


Mantisus:
Performance test.

1 client

Code to run
```python
import asyncio

from crawlee import ConcurrencySettings
from crawlee.crawlers import ParselCrawler, ParselCrawlingContext
from crawlee.http_clients import HttpxHttpClient
from crawlee.storage_clients import RedisStorageClient

CONNECTION = 'redis://localhost:6379'


async def main() -> None:
    storage_client = RedisStorageClient(connection_string=CONNECTION)
    http_client = HttpxHttpClient()

    crawler = ParselCrawler(
        storage_client=storage_client,
        http_client=http_client,
        concurrency_settings=ConcurrencySettings(desired_concurrency=20),
    )

    async def request_handler(context: ParselCrawlingContext) -> None:
        context.log.info(f'Processing URL: {context.request.url}...')
        data = {
            'url': context.request.url,
            'title': context.selector.css('title::text').get(),
        }
        await context.push_data(data)
        await context.enqueue_links()

    await crawler.run(['https://crawlee.dev'])


if __name__ == '__main__':
    asyncio.run(main())
```

```bash
[ParselCrawler] INFO  Final request statistics:
┌───────────────────────────────┬────────────┐
│ requests_finished             │ 2363       │
│ requests_failed               │ 0          │
│ retry_histogram               │ [2363]     │
│ request_avg_failed_duration   │ None       │
│ request_avg_finished_duration │ 358.4ms    │
│ requests_finished_per_minute  │ 3545       │
│ requests_failed_per_minute    │ 0          │
│ request_total_duration        │ 14min 6.8s │
│ requests_total                │ 2363       │
│ crawler_runtime               │ 39.99s     │
└───────────────────────────────┴────────────┘
```

3 clients

Code to run
```python
import asyncio
from concurrent.futures import ProcessPoolExecutor

from crawlee import ConcurrencySettings, service_locator
from crawlee.crawlers import ParselCrawler, ParselCrawlingContext
from crawlee.http_clients import HttpxHttpClient
from crawlee.storage_clients import RedisStorageClient

CONNECTION = 'redis://localhost:6379'

async def run(queue_name: str) -> None:
    storage_client = RedisStorageClient(connection_string=CONNECTION)

    service_locator.set_storage_client(storage_client)
    queue = await RequestQueue.open(name=queue_name)

    http_client = HttpxHttpClient()

    crawler = ParselCrawler(
        http_client=http_client,
        request_manager=queue,
        concurrency_settings=ConcurrencySettings(desired_concurrency=20),
    )

    async def request_handler(context: ParselCrawlingContext) -> None:
        context.log.info(f'Processing URL: {context.request.url}...')
        data = {
            'url': context.request.url,
            'title': context.selector.css('title::text').get(),
        }
        await context.push_data(data)
        await context.enqueue_links()

    await crawler.run(['https://crawlee.dev'])

def process_run(queue_name: str) -> None:
    asyncio.run(run(queue_name))

def multi_run(queue_name: str = 'multi') -> None:
    workers = 3
    with ProcessPoolExecutor(max_workers=workers) as executor:
        executor.map(process_run, [queue_name for i in range(workers)])

if __name__ == '__main__':
    multi_run()
```

```bash
[ParselCrawler] INFO  Final request statistics:
┌───────────────────────────────┬────────────┐
│ requests_finished             │ 779        │
│ requests_failed               │ 0          │
│ retry_histogram               │ [779]      │
│ request_avg_failed_duration   │ None       │
│ request_avg_finished_duration │ 356.9ms    │
│ requests_finished_per_minute  │ 2996       │
│ requests_failed_per_minute    │ 0          │
│ request_total_duration        │ 4min 38.0s │
│ requests_total                │ 779        │
│ crawler_runtime               │ 15.60s     │
└───────────────────────────────┴────────────┘
[ParselCrawler] INFO  Final request statistics:
┌───────────────────────────────┬────────────┐
│ requests_finished             │ 762        │
│ requests_failed               │ 0          │
│ retry_histogram               │ [762]      │
│ request_avg_failed_duration   │ None       │
│ request_avg_finished_duration │ 360.0ms    │
│ requests_finished_per_minute  │ 2931       │
│ requests_failed_per_minute    │ 0          │
│ request_total_duration        │ 4min 34.3s │
│ requests_total                │ 762        │
│ crawler_runtime               │ 15.60s     │
└───────────────────────────────┴────────────┘
[ParselCrawler] INFO  Final request statistics:
┌───────────────────────────────┬────────────┐
│ requests_finished             │ 822        │
│ requests_failed               │ 0          │
│ retry_histogram               │ [822]      │
│ request_avg_failed_duration   │ None       │
│ request_avg_finished_duration │ 342.2ms    │
│ requests_finished_per_minute  │ 3161       │
│ requests_failed_per_minute    │ 0          │
│ request_total_duration        │ 4min 41.3s │
│ requests_total                │ 822        │
│ crawler_runtime               │ 15.60s     │
└───────────────────────────────┴────────────┘
```

Mantisus:
In `RedisRequestQueueClient`, I used a [Bloom filter](https://en.wikipedia.org/wiki/Bloom_filter) in [`Redis`](https://redis.io/docs/latest/develop/data-types/probabilistic/bloom-filter/) for deduplication and tracking handled requests. This approach differs from what we used in other clients (using `set`). My main motivation was that Redis is an in-memory database, and memory consumption can be critical in some cases.

Since a Bloom filter is a probabilistic data structure, the final data structure size is affected by the error probability; I used `1e-7`. This means that with a probability of `1e-7`, we may get a false positive when checking the filter. In our case, this translates to a probability of skipping a request: with probability `1e-7`, a request that wasn't added to the queue will be considered as already added. Similarly, with probability `1e-7`, a request that hasn't been handled yet will be considered as already processed.

Memory consumption for records in the format 'https://crawlee.dev/{i}' (record size doesn't matter for Bloom filters):

**Redis Bloom filter:**
- 100,000 - 427 KB
- 1,000,000 - 4 MB  
- 10,000,000 - 42 MB

**Redis set:**
- 100,000 - 6 MB
- 1,000,000 - 61 MB
- 10,000,000 - 662 MB

Discussion about whether it's worth pursuing this approach is welcome!

janbuchar:
I haven't read the PR yet, but I did look into bloom filters for request deduplication in the past and what you wrote piqued my interest 🙂 I am a little worried about the chance of dropping a URL completely, even with a super small probability.

Perhaps we should default to a solution that tolerates some percentage of the "opposite" errors and allows a URL to get processed multiple times in rare cases. A fixed size hash table is an example of such data structure. I don't know if anything more sophisticated exists.

But maybe I have an irrational fear of probabilistic stuff 🙂

Mantisus:
> I am a little worried about the chance of dropping a URL completely, even with a super small probability.

Yes, I agree that this may be a little disturbing. And if we go down this route, it will need to be highlighted separately for the user.

But perhaps I am not sufficiently afraid of probabilistic structures, as I have used them before. 🙂

Mantisus:
Since we have already added the ability to parameterize queue behavior in the SDK (`single`, `shared`), I updated `RedisStorageClient` with a parameter defining the deduplication strategy `queue_dedup_strategy`. If set to `default`, it will use Set, `bloom` Bloom filters.

Pijukatel:
It also works with:
-  [Valkey](https://valkey.io/) from following docker image:[valkey/valkey-bundle ](https://hub.docker.com/r/valkey/valkey-bundle) 
-  [Dragonfly](https://www.dragonflydb.io/) from following docker image:[docker.dragonflydb.io/dragonflydb/dragonfly](https://hub.docker.com/r/dragonflydb/dragonfly)

Mantisus:
> It also works with:


Thanks for checking that out!

It's a real surprise to me that the Redis client is fully compatible with these.

src/crawlee/storage_clients/_redis/_client_mixin.py
```diff
@@ -0,0 +1,297 @@
...
+    @classmethod
+    async def _get_metadata_by_name(cls, name: str, redis: Redis, *, with_wait: bool = False) -> dict | None:
+        """Retrieve metadata by storage name.
+
+        Args:
+            name: The name of the storage.
+            redis: The Redis client instance.
+            with_wait: Whether to wait for the storage to be created if it doesn't exist.
+        """
+        if with_wait:
```
Pijukatel, line 75: Could you please explain the usecases?
Is there any scenario where someone would call this with `with_wait=False` for non-existent storage?

src/crawlee/storage_clients/_redis/_client_mixin.py
```diff
@@ -0,0 +1,297 @@
...
+                id=crypto_random_object_id(),
+                name=name,
+                created_at=now,
+                accessed_at=now,
+                modified_at=now,
+                **extra_metadata_fields,
+            )
+            client = cls(storage_name=internal_name, storage_id=metadata.id, redis=redis, **instance_kwargs)
+            created = await client._create_metadata_and_storage(internal_name, metadata.model_dump())
+            # The client was probably not created due to a race condition. Let's try to open it using the name.
```
Pijukatel, line 158: This can happen for two attempts to create same storage at the same time?

src/crawlee/storage_clients/_redis/_dataset_client.py
```diff
@@ -0,0 +1,330 @@
...
+            ),
+        )
+
+    @override
+    async def push_data(self, data: list[dict[str, Any]] | dict[str, Any]) -> None:
+        if isinstance(data, dict):
+            data = [data]
+
+        async with self._get_pipeline() as pipe:
+            # Incorrect signature for args type in redis-py
```
Pijukatel: Is there an existing issue about this?

src/crawlee/storage_clients/_redis/_dataset_client.py
```diff
@@ -0,0 +1,330 @@
...
+            pipe.json().arrappend(self._items_key, '$', *data)  # type: ignore[arg-type]
+            delta_item_count = len(data)
```
Pijukatel: I would probably just inline this since the argument name is same as variable name:
`delta_item_count=len(data)`

src/crawlee/storage_clients/_redis/_dataset_client.py
```diff
@@ -0,0 +1,330 @@
...
+        fields: list[str] | None = None,
+        omit: list[str] | None = None,
+        unwind: list[str] | None = None,
+        skip_empty: bool = False,
+        skip_hidden: bool = False,
+        flatten: list[str] | None = None,
+        view: str | None = None,
+    ) -> DatasetItemsListPage:
+        # Check for unsupported arguments and log a warning if found
+        # When implementing, explore the capabilities of jsonpath to determine what can be done at the Redis level.
```
Pijukatel: This seems more like a dev comment to himself, is it supposed to stay in the code?

src/crawlee/storage_clients/_redis/_dataset_client.py
```diff
@@ -0,0 +1,330 @@
...
+                f'by the {self.__class__.__name__} client.'
+            )
+
+        metadata = await self.get_metadata()
+
+        total = metadata.item_count
+        json_path = '$'
+
+        # Apply sorting and pagination
+        if desc:
```
Pijukatel: Since we have python = ">=3.10", I think this could be more readable with [pattern matching](https://peps.python.org/pep-0636/)

src/crawlee/storage_clients/_redis/_dataset_client.py
```diff
@@ -0,0 +1,330 @@
...
+            # Get batch of items
+            batch_items = await await_redis_response(self._redis.json().get(self._items_key, json_path))
+
+            # Handle case where batch_items might be None or not a list
+            if batch_items is None:
+                continue
+
+            # Reverse batch if desc order (since we got items in normal order but need desc)
+            if desc:
+                batch_items = list(reversed(batch_items))
```
Pijukatel: Do we need to do list(...)  here? We are going to yield from it anyway

src/crawlee/storage_clients/_redis/_request_queue_client.py
```diff
@@ -0,0 +1,578 @@
...
+        if self._dedup_strategy == 'bloom':
+            self._add_requests_script = await self._create_script('atomic_bloom_add_requests.lua')
+        elif self._dedup_strategy == 'default':
+            self._add_requests_script = await self._create_script('atomic_set_add_requests.lua')
+
+    @override
+    async def _create_storage(self, pipeline: Pipeline) -> None:
+        # Create Bloom filters for added and handled requests
+        if self._dedup_strategy == 'bloom':
+            await await_redis_response(pipeline.bf().create(self._added_filter_key, 1e-7, 100000, expansion=10))  # type: ignore[no-untyped-call]
```
Pijukatel: Could you please save these constants with a meaningful name:
??? =1e-7
??? = 100000

src/crawlee/storage_clients/_redis/_request_queue_client.py
```diff
@@ -0,0 +1,578 @@
...
+    _MAIN_KEY = 'request_queues'
+    """Main Redis key prefix for Request Queue."""
+
+    _CLIENT_TYPE = 'Request queue'
+    """Human-readable client type for error messages."""
+
+    _MAX_BATCH_FETCH_SIZE = 10
+    """Maximum number of requests to fetch in a single batch operation."""
+
+    _BLOCK_REQUEST_TIME = 300_000  # milliseconds
```
Pijukatel, line 81: Could you please explain this a bit?

Based on the description, I get the impression that I can't call reclaim_request for 300 s, but that is probably not the case.

src/crawlee/storage_clients/_redis/_request_queue_client.py
```diff
@@ -0,0 +1,578 @@
...
+                ],
+                args=[int(forefront), json.dumps(new_unique_keys), json.dumps(new_request_data)],
+            )
+            actually_added = set(json.loads(script_results))
+
+            delta_pending = len(actually_added)
+            delta_total = len(actually_added)
+
+            for unique_key in new_unique_keys:
+                if unique_key in actually_added:
```
Pijukatel: Maybe simplify to
```
  processed_requests.append(
      ProcessedRequest(
          unique_key=unique_key,
          was_already_present=not unique_key in actually_added,
          was_already_handled=False,
      )
  )
```

src/crawlee/storage_clients/_redis/_storage_client.py
```diff
@@ -0,0 +1,140 @@
...
+        Args:
+            connection_string: Redis connection string (e.g., "redis://localhost:6379").
+                Supports standard Redis URL format with optional database selection.
+            redis: Pre-configured Redis client instance.
+            queue_dedup_strategy: Strategy for request queue deduplication. Options are:
+                - 'default': Uses Redis sets for exact deduplication.
+                - 'bloom': Uses Redis Bloom filters for probabilistic deduplication with lower memory usage. When using
+                    this approach, there is a possibility 1e-7 that requests will be skipped in the queue.
+        """
+        if redis is not None and connection_string is not None:
```
Pijukatel: Maybe pattern matching again would be more readable

tests/unit/storages/conftest.py
```diff
@@ -1,17 +1,42 @@
...
+    StorageClient,
+)
+
+if TYPE_CHECKING:
+    from fakeredis import FakeAsyncRedis
 
 
-@pytest.fixture(params=['memory', 'file_system', 'sql'])
-def storage_client(request: pytest.FixtureRequest) -> StorageClient:
+@pytest.fixture(params=['memory', 'file_system', 'sql', ('redis', 'default'), ('redis', 'bloom')])
```
Pijukatel: Maybe just `redis_default` and `redis_bloom` would make it simpler
```
    elif storage_type == 'redis_default':
        storage_client = RedisStorageClient(redis=redis_client, queue_dedup_strategy='default')
    elif storage_type == 'redis_bloom':
        storage_client = RedisStorageClient(redis=redis_client, queue_dedup_strategy='bloom')
```

tests/unit/storage_clients/_redis/test_redis_dataset_client.py
```diff
@@ -0,0 +1,179 @@
...
+    await client.drop()
+
+
+async def test_base_keys_creation(dataset_client: RedisDatasetClient) -> None:
+    """Test that Redis dataset client creates proper keys."""
+    metadata = await dataset_client.get_metadata()
+    name = await await_redis_response(dataset_client.redis.hget('datasets:id_to_name', metadata.id))
+
+    assert name is not None
+    assert (name.decode() if isinstance(name, bytes) else name) == 'test_dataset'
```
Pijukatel, line 38: Why can it be bytes or string?

Pijukatel (commented):
Just some small comments and questions. I am really looking forward to using this. Good work.

src/crawlee/storage_clients/_redis/_client_mixin.py
Mantisus: Essentially, this is a synchronizer at the Redis level for working with multiple clients.
If, when opening the storage, we find that the `name_to_id` index already contains the name/alias of the storage, we cannot be sure whether it has already been completely created or not. In this case, the `with_wait` synchronizer is used.

src/crawlee/storage_clients/_redis/_client_mixin.py
Mantisus: Yes, that's why the `with_wait` synchronizer is used. So that when the client encounters a race condition, it waits until the first client finishes creating the storage and reports it.

src/crawlee/storage_clients/_redis/_dataset_client.py
Mantisus: Yes, we can expect this to be fixed in the next release.
https://github.com/redis/redis-py/issues/3780

src/crawlee/storage_clients/_redis/_request_queue_client.py
Mantisus: This is the time for which the request is blocked in the storage for other clients. After it expires, it will be returned to the queue. This protects against queue freezing if any of the clients crash.

I will improve the documentation.

docs/guides/storage_clients.mdx
janbuchar, line 1: I think we should mention that Redis persistence is unlike that of filesystem or SQL storage and link to https://redis.io/docs/latest/operate/oss_and_stack/management/persistence/

src/crawlee/storage_clients/_redis/_utils.py
```diff
@@ -0,0 +1,22 @@
...
+@overload
+async def await_redis_response(response: T) -> T: ...
+
+
+async def await_redis_response(response: Awaitable[T] | T) -> T:
+    """Solve the problem of ambiguous typing for redis."""
+    return await response if isinstance(response, Awaitable) else response
+
+
+def read_lua_script(file_path: Path) -> str:
```
janbuchar: Couldn't this accept just the file name and prepend the path to the `lua_scripts` directory automatically?

src/crawlee/storage_clients/_redis/_storage_client.py
```diff
@@ -0,0 +1,140 @@
...
+        """Initialize the Redis storage client.
+
+        Args:
+            connection_string: Redis connection string (e.g., "redis://localhost:6379").
+                Supports standard Redis URL format with optional database selection.
+            redis: Pre-configured Redis client instance.
+            queue_dedup_strategy: Strategy for request queue deduplication. Options are:
+                - 'default': Uses Redis sets for exact deduplication.
+                - 'bloom': Uses Redis Bloom filters for probabilistic deduplication with lower memory usage. When using
+                    this approach, there is a possibility 1e-7 that requests will be skipped in the queue.
```
janbuchar:
```suggestion
                - 'bloom': Uses Redis Bloom filters for probabilistic deduplication with lower memory usage. When using
                    this approach, approximately 1 in 1e-7 requests will be falsely considered duplicate.
```

BTW, is there a way to configure the probability?

tests/unit/storage_clients/_redis/test_redis_dataset_client.py
Mantisus: Due to the `decode_responses` parameter in the Redis constructor.

src/crawlee/storage_clients/_redis/_storage_client.py
```diff
@@ -0,0 +1,140 @@
...
+            self._redis = redis
+
+        elif connection_string is not None:
+            self._redis = Redis.from_url(connection_string)
+
+        self._queue_dedup_strategy = queue_dedup_strategy
+
+        # Call the notification only once
+        warnings.warn(
+            'The RedisStorageClient is experimental and may change or be removed in future releases.',
```
janbuchar: I don't think we'll want to remove it 🙂 The storage "schema" could change, though - perhaps we should mention that.

src/crawlee/storage_clients/_redis/_client_mixin.py
```diff
@@ -0,0 +1,297 @@
...
+            alias: The alias of the storage for unnamed (run scope) storages.
+            redis: Redis client instance.
+            metadata_model: Pydantic model for metadata validation.
+            extra_metadata_fields: Storage-specific metadata fields.
+            instance_kwargs: Additional arguments for the client constructor.
+
+        Returns:
+            An instance for the opened or created storage client.
+        """
+        internal_name = name or alias or cls._DEFAULT_NAME
```
janbuchar, line 128: I don't understand this - does it mean that there is no difference in behavior of named and aliased storages?

janbuchar (commented):
I think we should reach out to somebody who knows Redis better to review this for us. Or do you feel confident @Pijukatel? 😁

src/crawlee/storage_clients/_redis/_client_mixin.py
Mantisus: for alias in the metadata object `name=None`.

internal_name is used to form the key prefix. As in `FileSystemStorageClient`, the folder name is formed.

src/crawlee/storage_clients/_redis/_dataset_client.py
Pijukatel: It seems to be part of the v7 release done 2 days ago. Maybe bump the version to `redis = ["redis[hiredis] >= 7.0.0"] ` and if ok, remove the `# type: ignore[arg-type]`

Pijukatel:
> I think we should reach out to somebody who knows Redis better to review this for us. Or do you feel confident @Pijukatel? 😁

Feel free to invite someone more experienced with Redis to review; I was mainly focusing on the Python part in the review. 

Anyway, I tried running it a little and did not see anything wrong. It is also optional and experimental, so I would not be afraid to release it and improve it as we go. No existing user should be affected by this, even if there is some hidden bug.

docs/guides/storage_clients.mdx
```diff
@@ -27,6 +29,7 @@ Crawlee provides three main storage client implementations:
 - <ApiLink to="class/FileSystemStorageClient">`FileSystemStorageClient`</ApiLink> - Provides persistent file system storage with in-memory caching.
 - <ApiLink to="class/MemoryStorageClient">`MemoryStorageClient`</ApiLink> - Stores data in memory with no persistence.
 - <ApiLink to="class/SqlStorageClient">`SqlStorageClient`</ApiLink> – Provides persistent storage using a SQL database ([SQLite](https://sqlite.org/) or [PostgreSQL](https://www.postgresql.org/)). Requires installing the extra dependency: 'crawlee[sql_sqlite]' for SQLite or 'crawlee[sql_postgres]' for PostgreSQL.
+- <ApiLink to="class/RedisStorageClient">`RedisStorageClient`</ApiLink> – Provides persistent storage using a [Redis](https://redis.io/) database v8.0+. Requires installing the extra dependency: 'crawlee[redis]'.
```
vdusek: There are different dash chars for Redis and SQL storage clients, please unify them.

> The character U+2013 "–" could be confused with the ASCII character U+002d "-", which is more common in source code.

image

vdusek: Could we please use backticks for pkg+extra name? It looks better when rendered. Same for the SQL storage client.

```suggestion
- <ApiLink to="class/RedisStorageClient">`RedisStorageClient`</ApiLink> – Provides persistent storage using a [Redis](https://redis.io/) database v8.0+. Requires installing the extra dependency `crawlee[redis]`.
```

docs/guides/storage_clients.mdx
```diff
@@ -313,6 +319,172 @@ For advanced scenarios, you can configure <ApiLink to="class/SqlStorageClient">`
     {SQLStorageClientConfigurationExample}
 </CodeBlock>
 
+### Redis storage client
+
+:::warning Experimental feature
+The <ApiLink to="class/RedisStorageClient">`RedisStorageClient`</ApiLink> is experimental. Its API and behavior may change in future releases.
+:::
+
+The <ApiLink to="class/RedisStorageClient">`RedisStorageClient`</ApiLink> provides persistent storage using Redis database. It supports concurrent access from multiple independent clients or processes and uses Redis native data structures for efficient operations.
```
vdusek:
```suggestion
The <ApiLink to="class/RedisStorageClient">`RedisStorageClient`</ApiLink> provides persistent storage using [Redis](https://redis.io/) database. It supports concurrent access from multiple independent clients or processes and uses Redis native data structures for efficient operations.
```

docs/guides/storage_clients.mdx
```diff
@@ -313,6 +319,172 @@ For advanced scenarios, you can configure <ApiLink to="class/SqlStorageClient">`
...
+```
+
+Configuration options for the <ApiLink to="class/RedisStorageClient">`RedisStorageClient`</ApiLink> can be set through environment variables or the <ApiLink to="class/Configuration">`Configuration`</ApiLink> class:
+
+- **`purge_on_start`** (env: `CRAWLEE_PURGE_ON_START`, default: `True`) - Whether to purge default storages on start.
+
+Configuration options for the <ApiLink to="class/RedisStorageClient">`RedisStorageClient`</ApiLink> can be set via constructor arguments:
+
+- **`connection_string`** – Redis connection string, e.g. `redis://localhost:6379/0`.
+- **`redis`** – Pre-configured Redis client instance (optional).
```
vdusek: There are various dash characters in use; please unify them.

> The character U+2013 "–" could be confused with the ASCII character U+002d "-", which is more common in source code.

pyproject.toml
```diff
@@ -98,6 +99,7 @@ dev = [
     "apify_client", # For e2e tests.
     "build~=1.3.0", # For e2e tests.
     "dycw-pytest-only~=2.1.0",
+    "fakeredis[probabilistic,json,lua]>=2.31.0",
```
vdusek: please use tilde for dev deps
```suggestion
    "fakeredis[probabilistic,json,lua]~=2.31.0",
```

docs/guides/storage_clients.mdx
```diff
@@ -313,6 +319,172 @@ For advanced scenarios, you can configure <ApiLink to="class/SqlStorageClient">`
...
+
+class RedisDatasetClient {
+    <<Dataset>>
+}
+
+%% ========================
+%% Dataset Keys
+%% ========================
+
+class Dataset_Keys {
```
vdusek: Do we follow `PascalCase` combined with `snake_case` here? 🙂

tests/unit/storages/conftest.py
```diff
@@ -1,17 +1,40 @@
...
+    StorageClient,
+)
+
+if TYPE_CHECKING:
+    from fakeredis import FakeAsyncRedis
 
 
-@pytest.fixture(params=['memory', 'file_system', 'sql'])
-def storage_client(request: pytest.FixtureRequest) -> StorageClient:
+@pytest.fixture(params=['memory', 'file_system', 'sql', 'redis_default', 'redis_bloom'])
```
vdusek: I'm not sure if it makes sense to have two Redis storage clients here, as that would result in many more test runs covering the same functionality, just due to a different type of deduplication in Redis RQ. Maybe we could test Redis RQ with both the default and bloom deduplication in a more targeted way instead.

tests/unit/storage_clients/_redis/test_redis_dataset_client.py
```diff
@@ -0,0 +1,179 @@
+from __future__ import annotations
```
vdusek, line 1: I'm a bit surprised by the size of the test files in `tests/unit/storage_clients/_redis`. Most of the storage client usage should already be covered by the tests in `tests/unit/storages`, which use different storage clients as fixtures.

Are you aware of this overlap? Are the tests in `tests/unit/storage_clients/_redis` meant to cover only additional, Redis-specific cases that aren't already included in `tests/unit/storages`?

If not, could you please consider reducing the tests here to avoid duplication?

src/crawlee/storage_clients/_redis/_storage_client.py
```diff
@@ -0,0 +1,146 @@
...
+    to a Redis database v8.0+. Each storage type uses Redis-specific data structures and key patterns for
+    efficient storage and retrieval.
+
+    The client accepts either a Redis connection string or a pre-configured Redis client instance.
+    Exactly one of these parameters must be provided during initialization.
+
+    Storage types use the following Redis data structures:
+    - **Datasets**: Redis JSON arrays for item storage with metadata in JSON objects
+    - **Key-Value Stores**: Redis hashes for key-value pairs with separate metadata storage
+    - **Request Queues**: Redis lists for FIFO queuing, hashes for request data and in-progress tracking,
```
vdusek:
```suggestion
    - **Key-value stores**: Redis hashes for key-value pairs with separate metadata storage
    - **Request queues**: Redis lists for FIFO queuing, hashes for request data and in-progress tracking,
```

janbuchar:
@JuanGalilea did you have a chance to look into this?

tests/unit/storages/conftest.py
Mantisus: That makes sense. We could test deduplication at the `SqlRequestQueueClient` level.

tests/unit/storage_clients/_redis/test_redis_dataset_client.py
Mantisus: I will review the tests and remove the unnecessary ones.

Mantisus: Not many tests could be removed. This is because client tests check interaction with Redis at a lower level than storage tests.

vdusek: Thank you

tests/unit/storages/conftest.py
vdusek: thanks

vdusek:
@Mantisus Could you please resolve the conflicts? Once that's done, we'll merge it.

docs/examples/playwright_crawler_with_fingerprint_generator.mdx
janbuchar: um? 🙂

pyproject.toml
```diff
@@ -98,6 +99,7 @@ dev = [
     "apify_client", # For e2e tests.
     "build<2.0.0", # For e2e tests.
     "dycw-pytest-only<3.0.0",
+    "fakeredis[probabilistic,json,lua]~=2.32.0",
```
vdusek:
```suggestion
    "fakeredis[probabilistic,json,lua]<3.0.0",
```

uv.lock
```diff
@@ -870,13 +876,14 @@ requires-dist = [
...
 ]
-provides-extras = ["all", "adaptive-crawler", "beautifulsoup", "cli", "curl-impersonate", "httpx", "parsel", "playwright", "otel", "sql-postgres", "sql-sqlite"]
+provides-extras = ["all", "adaptive-crawler", "beautifulsoup", "cli", "curl-impersonate", "httpx", "parsel", "playwright", "otel", "sql-postgres", "sql-sqlite", "redis"]
 
 [package.metadata.requires-dev]
 dev = [
     { name = "apify-client" },
     { name = "build", specifier = "<2.0.0" },
     { name = "dycw-pytest-only", specifier = "<3.0.0" },
+    { name = "fakeredis", extras = ["probabilistic", "json", "lua"], specifier = "~=2.32.0" },
```
vdusek:
```suggestion
    { name = "fakeredis", extras = ["probabilistic", "json", "lua"], specifier = "<3.0.0" },
```

PR #1423, opened 2025-09-23
refactor: Add util function for checking kwarg count

Pijukatel:
### Description

- Add a simple util function to limit the number of kwargs and reuse it in the code.

### Testing

- Added unit tests.

### DIFF
--- a/src/crawlee/_utils/raise_if_too_many_kwargs.py
+++ b/src/crawlee/_utils/raise_if_too_many_kwargs.py
@@ -1,13 +1,12 @@
-from collections import Counter
 from typing import Any
 
 
 def raise_if_too_many_kwargs(max_kwargs: int = 1, **kwargs: Any) -> None:
     """Raise ValueError if there are more non-None kwargs then max_kwargs."""
-    if len(kwargs) - Counter(kwargs.values())[None] > max_kwargs:
-        kwargs_names = [f'"{kwarg_name}"' for kwarg_name in kwargs]
-        used_kwargs_names = [f'"{kwarg_name}"' for kwarg_name, value in kwargs.items() if value is not None]
+    none_kwargs_names = [f'"{kwarg_name}"' for kwarg_name, value in kwargs.items() if value is not None]
+    if len(none_kwargs_names) > max_kwargs:
+        all_kwargs_names = [f'"{kwarg_name}"' for kwarg_name in kwargs]
         raise ValueError(
-            f'Only one of {", ".join(kwargs_names)} can be specified, but following arguments were '
-            f'specified: {", ".join(used_kwargs_names)}.'
+            f'Only one of {", ".join(all_kwargs_names)} can be specified, but following arguments were '
+            f'specified: {", ".join(none_kwargs_names)}.'
         )


src/crawlee/storage_clients/_file_system/_dataset_client.py
```diff
@@ -114,9 +115,7 @@ async def open(
                 or if both name and alias are provided.
         """
         # Validate input parameters.
-        specified_params = sum(1 for param in [id, name, alias] if param is not None)
-        if specified_params > 1:
-            raise ValueError('Only one of "id", "name", or "alias" can be specified, not multiple.')
+        limit_kwarg_count(alias=alias, name=name, id=id)
```
janbuchar: I'm not sure about the name, `limit_kwarg_count` doesn't really suggest that this line will throw if there's more than one non-None value passed to the function. Also having to repeat each kwarg twice is pretty unwieldy.

Pijukatel: Using more explicit name. 

>  Also having to repeat each kwarg twice is pretty unwieldy.

That is to have the names of the arguments available in the raised error. Is there a more clever way?

janbuchar: > Is there a more clever way?

I don't think so, unless you want to do some real voodoo (like stack frame inspection).

src/crawlee/_utils/raise_if_too_many_kwargs.py
```diff
@@ -0,0 +1,9 @@
+from collections import Counter
+from typing import Any
+
+
+def raise_if_too_many_kwargs(max_kwargs: int = 1, **kwargs: Any) -> None:
+    """Raise ValueError if there are more kwargs then max_kwargs."""
+    if len(kwargs) - Counter(kwargs.values())[None] > max_kwargs:
+        kwargs_names = [f'"{kwarg_name}"' for kwarg_name in kwargs]
+        raise ValueError(f'Only one of {", ".join(kwargs_names)} can be specified, not multiple.')
```
janbuchar: Can you also tell the user which ones were specified?

Pijukatel: Ok, added to the exception text

src/crawlee/storage_clients/_file_system/_dataset_client.py
Pijukatel: In that case, let's keep it this way for now. This is clear and readable.

src/crawlee/_utils/raise_if_too_many_kwargs.py
```diff
@@ -0,0 +1,13 @@
+from collections import Counter
+from typing import Any
+
+
+def raise_if_too_many_kwargs(max_kwargs: int = 1, **kwargs: Any) -> None:
+    """Raise ValueError if there are more non-None kwargs then max_kwargs."""
+    if len(kwargs) - Counter(kwargs.values())[None] > max_kwargs:
+        kwargs_names = [f'"{kwarg_name}"' for kwarg_name in kwargs]
+        used_kwargs_names = [f'"{kwarg_name}"' for kwarg_name, value in kwargs.items() if value is not None]
```
Mantisus: Since we need to use list comprehension for `used_kwargs_names` anyway, do we really need Counter?

maybe just
```python
used_kwargs_names = [f'"{kwarg_name}"' for kwarg_name, value in kwargs.items() if value is not None]
if len(used_kwargs_names ) > max_kwargs:
```

PR #1425, opened 2025-09-24
refactor: Give more instance caching flexibility to the storage clients

Pijukatel:
### Description

- `StorageClient` type is not an individual storage key and is instead used in an additional cache key (can be overridden).
- This allows `StorageClients` to have better control over caching.

### Issues

- Enables: https://github.com/apify/apify-sdk-python/issues/513

### Testing

- Unit tests

### DIFF
--- a/src/crawlee/storage_clients/_base/_storage_client.py
+++ b/src/crawlee/storage_clients/_base/_storage_client.py
@@ -30,7 +30,7 @@ class StorageClient(ABC):
     (where applicable), and consistent access patterns across all storage types it supports.
     """
 
-    def get_additional_cache_key(self, configuration: Configuration) -> Hashable:  # noqa: ARG002
+    def get_storage_client_cache_key(self, configuration: Configuration) -> Hashable:  # noqa: ARG002
         """Return a cache key that can differentiate between different storages of this and other clients.
 
         Can be based on configuration or on the client itself. By default, returns a module and name of the client
--- a/src/crawlee/storage_clients/_base/_storage_client.py
+++ b/src/crawlee/storage_clients/_base/_storage_client.py
@@ -31,11 +31,11 @@ class StorageClient(ABC):
     """
 
     def get_additional_cache_key(self, configuration: Configuration) -> Hashable:  # noqa: ARG002
-        """Return a cache key that can differentiate between different storages of this client.
+        """Return a cache key that can differentiate between different storages of this and other clients.
 
-        Can be based on configuration or on the client itself. By default, returns an empty string.
+        Can be based on configuration or on the client itself. By default, returns a name of the client class.
         """
-        return ''
+        return self.__class__.__name__
 
     @abstractmethod
     async def create_dataset_client(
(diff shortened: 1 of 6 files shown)


src/crawlee/storage_clients/_base/_storage_client.py
```diff
@@ -31,11 +31,11 @@ class StorageClient(ABC):
...
 
     def get_additional_cache_key(self, configuration: Configuration) -> Hashable:  # noqa: ARG002
-        """Return a cache key that can differentiate between different storages of this client.
+        """Return a cache key that can differentiate between different storages of this and other clients.
 
-        Can be based on configuration or on the client itself. By default, returns an empty string.
+        Can be based on configuration or on the client itself. By default, returns a name of the client class.
         """
-        return ''
+        return self.__class__.__name__
```
janbuchar: Two points regarding this method:

- `self.__class__.__name__` is not unique enough - could you at least attach `__module__` as well?
- the name of the method will not be very fitting after this PR

src/crawlee/storages/_storage_instance_manager.py
```diff
@@ -187,23 +157,22 @@ def remove_from_cache(self, storage_instance: Storage) -> None:
...
+        # Remove from name cache or alias cache. It can never be in both.
+        if storage_instance.name is not None:
+            for additional_key in self._cache.by_name[storage_type][storage_instance.name]:
+                del self._cache.by_name[storage_type][storage_instance.name][additional_key]
+                break
+        else:
+            for alias_key in self._cache.by_alias[storage_type]:
+                for additional_key in self._cache.by_alias[storage_type][alias_key]:
+                    del self._cache.by_alias[storage_type][alias_key][additional_key]
                     break
```
janbuchar: Perhaps this logic could be moved into the `_StorageCache` class now?

janbuchar (commented):
LGTM, found nits

src/crawlee/storages/_storage_instance_manager.py
Pijukatel: Makes sense. Done.

src/crawlee/storage_clients/_base/_storage_client.py
Pijukatel: Using module as well now.

But not sure about renaming, I am quite content with it.  Do you have any better name in mind?

janbuchar: Maybe something like `get_storage_client_cache_key`?

Pijukatel: `StorageClient.get_storage_client_cache_key`  feels kind of redundant, since the class is `StorageClient` and its children

janbuchar: I get it, but most people don't need to use it, so I wouldn't be too sad about that

Pijukatel: Ok, renamed

Issue #1071, opened 2025-03-11
Add chrome option for PlaywrightCrawler


Pijukatel:
Currently when using `PlaywrightCrawler` the three default browser types are `'chromium', 'firefox', 'webkit'`

Add native support for actual `Chrome` as well. (consider if that is supposed to be in the same optional group as playwright or even addtional optional group as it will require download and installation of Chrome browser as well)

Pijukatel:
> But `Chrome` is a [`channel`](https://playwright.dev/python/docs/api/class-browsertype#browser-type-launch-option-channel) not a `browser_type`.
> 
> It is already supported if Chrome browser is installed on the system and pass the appropriate parameter to the `browser_launch_options`

@B4nan best if you answer this question. No point in me being the middleman. What exactly was your idea for "chrome support"?

PR #1426, opened 2025-09-24
docs: Add an example of running `PlaywrightCrawler` using local `Chrome` and `Firefox` profiles.

Mantisus:
### Description

- Add an example of running `PlaywrightCrawler` using local `Chrome` and `Firefox` profiles.

### DIFF
--- a/docs/examples/using_browser_profile.mdx
+++ b/docs/examples/using_browser_profile.mdx
@@ -18,6 +18,8 @@ Using browser profiles allows you to leverage existing login sessions, saved pas
 
 To run <ApiLink to="class/PlaywrightCrawler">`PlaywrightCrawler`</ApiLink> with your Chrome profile, you need to know the path to your profile files. You can find this information by entering `chrome://version/` as a URL in your Chrome browser. If you have multiple profiles, pay attention to the profile name - if you only have one profile, it's always `Default`.
 
+You also need to use the [`channel`](https://playwright.dev/python/docs/api/class-browsertype#browser-type-launch-option-channel) parameter in `browser_launch_options` to use the Chrome browser installed on your system instead of Playwright's Chromium.
+
 :::warning Profile access limitation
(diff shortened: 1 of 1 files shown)


Mantisus:
> Is it safer even for Firefox to copy the profile to temdir?  I guess the profile can get changed during crawling? Is that desired side effect, or bad side effect?

It depends on the goal. If it should not affect the main profile, then yes, it is safer to use a copy.

But I can imagine a use case where all changes made during scanning must be synchronized with the profile.

Mantisus:
This doc may also solve issue #1071. Since it shows how to use `channel`, use the installed Chrome.

@B4nan, what do you think?

Mantisus:
> do you think you could make a JS counterpart to the example as well?

I think I'll try to do that. 🙂

B4nan:
> This doc may also solve issue https://github.com/apify/crawlee-python/issues/1071. Since it shows how to use channel, use the installed Chrome.

In JS version we have `useChrome` flag under `launchContext` in browser crawlers, and a default mechanism for inference of the browser executable path:

```ts
const crawler = new PlaywrightCrawler({
    launchContext: {
        useChrome: true,
    },
    requestHandler() { ... },
});
```

https://github.com/apify/crawlee/blob/07cd2c43fcbf21c6917b5bf55cd1af64c4bdcb00/packages/browser-crawler/src/internals/browser-launcher.ts#L202-L204
https://github.com/apify/crawlee/blob/07cd2c43fcbf21c6917b5bf55cd1af64c4bdcb00/packages/browser-crawler/src/internals/browser-launcher.ts#L220

Not saying we need to port this 1:1, but IMO it's very common to use chrome for scraping, so having a bit more native support than just a guide on how to use it with Playwright feels right to me.

And more importantly, we have a ready made docker image with preinstalled chrome:

https://github.com/apify/apify-actor-docker/tree/master/node-playwright-chrome

PR #1432, opened 2025-09-29
chore: Prevent request manager change after crawler restart

Pijukatel:
### Description

- Prevent `request_manager` change in `BasicCrawler` after crawler restart.

### Testing

- Added unit test


Pijukatel:
Fix extracted from https://github.com/apify/crawlee-python/pull/1418

PR #1438, opened 2025-09-30
refactor: Minimize global service locator side effects

Pijukatel:
### Description

- Reduce the amount of global side effects in `service_locator` by using an explicit KVS factory in `RecoverableState`.
- Fix `KeyValueStore.auto_saved_value` not working properly if the global storage_client was different from the current kvs storage client.
- Improve test isolation.

### Issues

- Closes: https://github.com/apify/crawlee-python/issues/1354

### Testing

- Added tests for some edge cases.


Pijukatel:
This PR replaces https://github.com/apify/crawlee-python/pull/1368. That PR included many changes that were already released in Crawlee v1.0, and continuing in the previous PR would not be good for review.

src/crawlee/storage_clients/_file_system/_request_queue_client.py
```diff
@@ -254,6 +271,7 @@ async def open(
                     metadata=metadata,
                     path_to_rq=path_to_rq,
                     lock=asyncio.Lock(),
+                    recoverable_state=await cls._create_recoverable_state(id=metadata.id, configuration=configuration),
```
vdusek, line 274: Instead of creating it three times, we can create it once, store it in a variable, and just pass it where needed.

vdusek (commented):
I wouldn't be afraid to call it a fix if it's (according to the description) fixing something, and then it should be included in the changelog.

But also, I'm not 100% I understand why we're doing this:

> Reduce the amount of global side effects in service_locator by using an explicit KVS factory in RecoverableState.

What side effects?

> Fix KeyValueStore.auto_saved_value not working properly if the global storage_client was different from the current kvs storage client.

:+1:

src/crawlee/storage_clients/_file_system/_request_queue_client.py
```diff
@@ -136,6 +133,22 @@ def path_to_metadata(self) -> Path:
         """The full path to the request queue metadata file."""
         return self.path_to_rq / METADATA_FILENAME
 
+    @classmethod
+    async def _create_recoverable_state(cls, id: str, configuration: Configuration) -> RecoverableState:
+        async def kvs_factory() -> KeyValueStore:
+            from crawlee.storage_clients import FileSystemStorageClient  # noqa: PLC0415 avoid circular import
+            from crawlee.storages import KeyValueStore  # noqa: PLC0415 avoid circular import
+
+            return await KeyValueStore.open(storage_client=FileSystemStorageClient(), configuration=configuration)
```
janbuchar, line 142: Creating a fresh filesystem storage client to open a request queue feels _wrong_ - at this point, we can be pretty sure that another one already exists. Is there a specific reason to do this or is it just because you don't have access to the existing one?

tests/unit/storage_clients/_file_system/test_fs_rq_client.py
```diff
@@ -78,6 +78,12 @@ async def test_request_file_persistence(rq_client: FileSystemRequestQueueClient)
             assert request_data['url'].startswith('https://example.com/')
 
 
+async def test_opening_rq_does_not_have_side_effect_on_service_locator(
```
janbuchar: Can you add some explanation here? Also, inlining the `rq_client` fixture could lead to better readable code

tests/unit/conftest.py
```diff
@@ -69,6 +70,10 @@ def _prepare_test_env() -> None:
         # Verify that the test environment was set up correctly.
         assert os.environ.get('CRAWLEE_STORAGE_DIR') == str(tmp_path)
 
+        # Reset global class variables to ensure test isolation.
+        KeyValueStore._autosaved_values = {}
+        Statistics._Statistics__next_id = 0  # type:ignore[attr-defined] # Mangled attribute
```
janbuchar, line 78: How does this contribute towards test isolation? Is there anything that depends on the persist state key that is derived from the ID?

Pijukatel: The answer is I am not sure. But we have so many tests, so I think it is best if we restore all we can to the same state at the beginning of the test. This reduces the chance of some weird behavior based on the order of the test execution.

tests/unit/storage_clients/_file_system/test_fs_rq_client.py
Pijukatel: Ok

src/crawlee/storage_clients/_file_system/_request_queue_client.py
Pijukatel: At that point, we are not sure it exists. It could have been created through a class method:
`
await FileSystemStorageClient.create_rq_client(name='test_request_queue', configuration=configuration)
`

And why not open the KVS through such a class method as well? Because that way, you bypass the storage instance manager - and that is generally something we do not want.

Anyway, that instance is super light and will get garbage collected pretty quickly.

src/crawlee/storage_clients/_file_system/_request_queue_client.py
Pijukatel: That creation requires metadata to get the RQ.id, so we have to repeat this call, as in all three branches, we get metadata in a different way.

src/crawlee/storage_clients/_file_system/_request_queue_client.py
Pijukatel: At that point, we are not sure it exists. It could have been created through a class method without the client and even when created with the help of client, it is out of scope:
`
await FileSystemRequestQueueClient.open(...)
`

And why not open the KVS through such a class method as well? Because that way, you bypass the storage instance manager - and that is generally something we do not want.

FileSystemStorageClient is just a helper factory class, which is mainly for convenience and for registering the storage instance manager.

Pijukatel:
> I wouldn't be afraid to call it a fix if it's (according to the description) fixing something, and then it should be included in the changelog.
> 
> But also, I'm not 100% I understand why we're doing this:
> 
> > Reduce the amount of global side effects in service_locator by using an explicit KVS factory in RecoverableState.
> 
> What side effects?
> 
> > Fix KeyValueStore.auto_saved_value not working properly if the global storage_client was different from the current kvs storage client.
> 
> 👍


Each time you use `RecoverableState` and you do not pass an explicit `persist_state_kvs_factory` the service_locator will be used to get the current global storage_client. If there is no global storage_client set yet, then it is created and set - and this is the side effect I want to minimize. In many cases, a specific `RecoverableState` does not really want to use the global `storage_client` as in many cases that does not make sense at all.

There are different cases, where I am not sure, for example, `Statistics`, so I did not touch those; I changed only those that I am 100% sure about.

vdusek (approved):
LGTM

Issue #1434, opened 2025-09-30
Validate storage names before passing them to the storage client


vdusek:
The Apify platform has these limitations:

```
apify_client.errors.ApifyApiError: Invalid value provided in store: name can only contain letters 'a' through 'z', the digits '0' through '9', and the hyphen ('-') but only in the middle of the string (e.g. 'my-value-1')
```

We should validate and enforce this rule on the storage level - before passing it to the storage clients.

Pijukatel:
Related to: https://github.com/apify/crawlee-python/issues/1354

I am not sure about validation at the storage level. This naming rule is specific to one client from the downstream repo. I think the constraint should be validated in [that specific client ](https://github.com/apify/apify-sdk-python/blob/master/src/apify/storage_clients/_apify/_storage_client.py)

vdusek:
Isn't that weird? I'd say storage names (and other attributes) should follow the same constraints regardless of the storage client being used. Otherwise, you could have a valid storage setup locally, but when deploying it to Apify, it might raise an error. I'd prefer to keep it consistent.

What do you think @B4nan?

B4nan:
Yeah, definitely, while it sucks that the only real problem is the Apify platform imposing restrictions, we should keep those rules global, otherwise you would only find out about it once you try to deploy.

B4nan:
@Mantisus please try to priotize this, its technically a small BC for people not targetting the apify platform, lets try to ship it later this week.

Pijukatel:
> Yeah, definitely, while it sucks that the only real problem is the Apify platform imposing restrictions, we should keep those rules global, otherwise you would only find out about it once you try to deploy.

I think the proper level is implementing it at the SDK level. This way it is not affecting other clients and it solves the problem of "otherwise you would only find out about it once you try to deploy." 
https://github.com/apify/apify-sdk-python/tree/master/src/apify/storage_clients/_apify
https://github.com/apify/apify-sdk-python/tree/master/src/apify/storage_clients/_smart_apify

B4nan:
Not sure I follow, you need the same restrictions during local development, what you suggest means using the Apify Client, so the API, so not local implementations, right?

edit: or that "smart apify client" is the "run things locally if not on platform" one? then I guess it sounds all right

PR #1457, opened 2025-10-06
fix: Add storages name validation

Mantisus:
### Description

- Added name validation for storages in accordance with Apify platform restrictions

### Issues

- Closes: #1434
- Relates: #1354 

### Testing

- Added new tests to verify correct validation.

### DIFF
--- a/src/crawlee/storages/_base.py
+++ b/src/crawlee/storages/_base.py
@@ -46,7 +46,7 @@ class Storage(ABC):
             id: The storage ID.
             name: The storage name (global scope, persists across runs). Name can only contain letters "a" through "z",
                 the digits "0" through "9", and the hyphen ("-") but only in the middle of the string
-                (e.g. "my-value-1")
+                (e.g. "my-value-1").
             alias: The storage alias (run scope, creates unnamed storage).
             configuration: Configuration object used during the storage creation or restoration process.
             storage_client: Underlying storage client to use. If not provided, the default global storage client
--- a/src/crawlee/storages/_base.py
+++ b/src/crawlee/storages/_base.py
@@ -44,7 +44,9 @@ class Storage(ABC):
 
         Args:
             id: The storage ID.
-            name: The storage name (global scope, persists across runs).
+            name: The storage name (global scope, persists across runs). Name can only contain letters "a" through "z",
+                the digits "0" through "9", and the hyphen ("-") but only in the middle of the string
+                (e.g. "my-value-1")
             alias: The storage alias (run scope, creates unnamed storage).
             configuration: Configuration object used during the storage creation or restoration process.
             storage_client: Underlying storage client to use. If not provided, the default global storage client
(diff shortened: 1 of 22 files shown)


vdusek:
assigning @janbuchar as @Pijukatel is off for this week

src/crawlee/storages/_base.py
```diff
@@ -44,7 +44,9 @@ async def open(
 
         Args:
             id: The storage ID.
-            name: The storage name (global scope, persists across runs).
+            name: The storage name (global scope, persists across runs). Name can only contain letters "a" through "z",
+                the digits "0" through "9", and the hyphen ("-") but only in the middle of the string
+                (e.g. "my-value-1")
```
vdusek:
```suggestion
            name: The storage name (global scope, persists across runs). Name can only contain letters "a" through "z",
                the digits "0" through "9", and the hyphen ("-") but only in the middle of the string
                (e.g. "my-value-1").
```

src/crawlee/storages/_storage_instance_manager.py
```diff
@@ -90,7 +92,9 @@ async def open_storage_instance(
         Args:
             cls: The storage class to instantiate.
             id: Storage ID.
-            name: Storage name. (global scope, persists across runs).
+            name: Storage name. (global scope, persists across runs). Name can only contain letters "a" through "z",
+                the digits "0" through "9", and the hyphen ("-") but only in the middle of the string
+                (e.g. "my-value-1")
```
vdusek:
```suggestion
            name: Storage name. (global scope, persists across runs). Name can only contain letters "a" through "z",
                the digits "0" through "9", and the hyphen ("-") but only in the middle of the string
                (e.g. "my-value-1").
```

tests/unit/storages/test_dataset.py
```diff
@@ -1053,3 +1053,35 @@ async def test_name_default_not_allowed(storage_client: StorageClient) -> None:
...
+        pytest.param('value-1', True, id='number end'),
+        pytest.param('test-1-value', True, id='number middle'),
+        pytest.param('test-------value', True, id='multiple-dashes'),
+        pytest.param('test-VALUES-test', True, id='multiple-cases'),
+        pytest.param('name_with_underscores', False, id='underscores'),
+        pytest.param('name with spaces', False, id='spaces'),
+        pytest.param('-test', False, id='dashes start'),
+        pytest.param('test-', False, id='dashes end'),
+    ],
+)
```
vdusek, line 1077: Does this also correspond to the naming constraints in `SqlStorageClient`?

vdusek (commented):
Thanks. I'm wondering, shouldn't we add this new enforcement on this level to https://crawlee.dev/python/docs/upgrading/upgrading-to-v1?

janbuchar (approved):

tests/unit/storages/test_dataset.py
Mantisus: Yep.

vdusek:
@Mantisus sorry did you overlook this:

> Thanks. I'm wondering, shouldn't we add this new enforcement on this level to https://crawlee.dev/python/docs/upgrading/upgrading-to-v1?

What do you think about that?

Mantisus:
Sorry, yes

> What do you think about that?

I'm not sure, because it's a fix. And we didn't include fixes in the upgrading guide, if I'm not mistaken.

Mantisus:
The upgrading guide was updated after discussing on Slack

Issue #1464, opened 2025-10-07
Updating to Pydantic >3.11.9 causes TypeError on import


Mycsina:
After updating to Pydantic 3.12.0, when importing the AdaptivePlaywrightCrawler I get the following exception:

`TypeError: cannot specify both default and default_factory`

After inspecting more closely, the exception is triggered when these BaseModels are created:

- [HTTPHeaders](https://github.com/apify/crawlee-python/blob/5581ed6cdb2c3fc9cb804891c616ebabf60f62bd/src/crawlee/_types.py#L69C1-L78C11)
- [Request](https://github.com/apify/crawlee-python/blob/5581ed6cdb2c3fc9cb804891c616ebabf60f62bd/src/crawlee/_request.py#L188C1-L211C80)

Removing the three assignments stops the exception from being raised, albeit it causes these warnings to appear when they previously didn't, seems to be related to my configuration:
```
[crawlee._service_locator] WARN  No configuration set, implicitly creating and using default Configuration.
[crawlee._service_locator] WARN  No storage client set, implicitly creating and using default FileSystemStorageClient. 
[crawlee._service_locator] WARN  No event manager set, implicitly creating and using default LocalEventManager.
```

vdusek:
Thank you, this seems like two independent problems (the warnings and the Pydantic issues). We're gonna pin Pydantic version to <2.12.0 (https://github.com/apify/crawlee-python/pull/1467) and release a new patch version ASAP. The rest will be resolved later. I also opened a separate issue for the warnings https://github.com/apify/crawlee-python/issues/1468.

PR #1471, opened 2025-10-08
fix: Add support pydantic v2.12

Mantisus:
### Description

- Add support pydantic v2.12. This uses a workaround to ensure compatibility with `Annotated[..., Field(default_factory=...)]` and with type hints.

### Issues

- Closes: #1464


Mantisus:
> shouldn't we wait for a fix on the Pydantic side instead?

I'm not sure that Pydantic will fix this. Since the default value was factually ignored before 2.12, we set it for the type checker.

An example of a discussion in the Pydantic repository - https://github.com/pydantic/pydantic/issues/6713#issuecomment-1813374273

PR #1474, opened 2025-10-10
fix: Only apply requestHandlerTimeout to request handler

janbuchar:
I stumbled upon this when working on ContextPipeline for the JS version. I'm eager to hear your thoughts 🙂


janbuchar:
> Okay, so the `requestHandlerTimeout` is now applied only to the request handler (`router(final_context)`) and not the whole context pipeline.

This is correct.

> Do we have any examples of where the previous behavior caused any troubles?

In the JS version, the browser crawler (playwright/puppeteer ancestor) has two kinds of timeout - `navigationTimeout` and `requestHandlerTimeout`. Then it does this: https://github.com/apify/crawlee/blob/master/packages/browser-crawler/src/internals/browser-crawler.ts#L395

This is super awkward and not really what the interface promises. However, the Python version doesn't have a navigation timeout. I believe I should add that as part of this PR.

Pijukatel (commented):
Could you please add some test for the correct application of the timeout.

src/crawlee/http_clients/_impit.py
```diff
@@ -189,14 +200,18 @@ async def stream(
...
+            response = await client.request(
+                method=method,
+                url=url,
+                content=payload,
+                headers=dict(headers) if headers else None,
+                timeout=timeout.total_seconds() if timeout else None,
+                stream=True,
+            )
+        except TimeoutException as exc:
+            raise asyncio.TimeoutError from exc
```
janbuchar, line 213: @Mantisus do you recall if this is the only place where an unwrapped `TimeoutError` can pop out of impit? Or should we also handle it somewhere else?

src/crawlee/crawlers/_playwright/_playwright_crawler.py
```diff
@@ -134,6 +137,8 @@ def __init__(
             use_incognito_pages: By default pages share the same browser context. If set to True each page uses its
                 own context that is destroyed once the page is closed or crashes.
                 This option should not be used if `browser_pool` is provided.
+            navigation_timeout: Timeout for navigation (the process between opening a Playwright page and calling
```
janbuchar, line 143: Open question - should the `navigation_timeout` also apply to pre-navigation hooks? Should they have their own limit? Should they share a limit with the request handler?

@B4nan your opinion on this is also welcome. This is due to change in crawlee js v4, so we should be aligned.

B4nan: I would include the hooks, otherwise we would need another two options for timeouts of pre and post nav hooks. They need to be part of some timeout handler, and having 3 options just for the navigation feels too much.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1507,12 +1507,15 @@ async def __run_task_function(self) -> None:
             raise
 
     async def _run_request_handler(self, context: BasicCrawlingContext) -> None:
-        await wait_for(
```
Pijukatel, line 1510: I am just wondering. Can the context pipeline now get stuck?

navigation_timeout(some parts of context pipeline) -> time unlimited parts of context pipeline -> request_handler_timeout(request_hanlder)

janbuchar: Realistically, that has always been the case, but yeah, this increases the odds by a bit. The hooks (discussed in https://github.com/apify/crawlee-python/pull/1474#discussion_r2571298194) are probably the biggest risk.

Perhaps we could add some comically large timeout to the context pipeline execution as a whole, too.

src/crawlee/http_clients/_impit.py
Mantisus: Only here 🙂

src/crawlee/crawlers/_basic/_basic_crawler.py
Pijukatel: In that case, it makes even more sense to include the hooks in the timeout. 

Other steps do not appear to be a problem in our pipelines for now.

src/crawlee/crawlers/_playwright/_playwright_crawler.py
janbuchar: @B4nan include them in `navigation_timeout`, you mean?

B4nan: Yes, either that, or separate timeouts over pre and post nav hooks. User code needs to be wrapped in timeout handlers.

vdusek: I would say we can start with one shared timeout for all, and adjust later if needed.

src/crawlee/http_clients/_base.py
```diff
@@ -144,6 +147,7 @@ async def send_request(
             payload: The data to be sent as the request body.
             session: The session associated with the request.
             proxy_info: The information about the proxy to be used.
+            timeout: Request timeout
```
vdusek: Something more descriptive? Although not sure whether process is the correct word, I cannot find a better one.

```suggestion
            timeout: Maximum time allowed to process the request.
```

src/crawlee/http_clients/_base.py
```diff
@@ -114,6 +115,7 @@ async def crawl(
             session: The session associated with the request.
             proxy_info: The information about the proxy to be used.
             statistics: The statistics object to register status codes.
+            timeout: Request timeout
```
vdusek: Something more descriptive? Although not sure whether process is the correct word, I cannot find a better one.

```suggestion
            timeout: Maximum time allowed to process the request.
```

src/crawlee/crawlers/_playwright/_playwright_crawler.py
janbuchar: Well, that's how we got here :grin: but yeah, including hook in navigation timeout is fine with me.

Mantisus (approved):

src/crawlee/_utils/time.py
```diff
@@ -35,6 +38,40 @@ def measure_time() -> Iterator[TimerResult]:
         result.cpu = after_cpu - before_cpu
 
 
+class SharedTimeout:
+    """Keeps track of a time budget shared by multiple independent async operations."""
```
Pijukatel: I would mention that this is [“reusable, but not reentrant”](https://docs.python.org/3/library/contextlib.html#reusable-context-managers) context manager as classified in official Python docs.

src/crawlee/_utils/time.py
```diff
@@ -35,6 +38,40 @@ def measure_time() -> Iterator[TimerResult]:
...
+        self._remaining_timeout = timeout
+        self._active_timeout: Timeout | None = None
+        self._activation_timestamp: float | None = None
+
+    async def __aenter__(self) -> timedelta:
+        if self._active_timeout is not None or self._activation_timestamp is not None:
+            raise RuntimeError('A shared timeout context cannot be entered twice at the same time')
+
+        self._activation_timestamp = time.monotonic()
+        self._active_timeout = new_timeout = timeout(self._remaining_timeout.total_seconds())
```
Pijukatel, line 57: This `new_timeout` seems little redundant

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
```diff
@@ -128,8 +132,12 @@ def _create_static_content_crawler_pipeline(self) -> ContextPipeline[ParsedHttpC
     async def _execute_pre_navigation_hooks(
         self, context: BasicCrawlingContext
     ) -> AsyncGenerator[BasicCrawlingContext, None]:
+        self._shared_navigation_timeouts[context] = SharedTimeout(self._navigation_timeout)
```
Pijukatel: I am not sure about this. Context is expanded by many pipeline steps, so this mapping is not very flexible as it will break if you try to apply SharedTimeout over pipeline steps that work on expanded context.

Maybe it would be better to keep the timeout on the context itself, instead of this mapping on the crawler? I think it belongs there also conceptually. The crawler sets the timeout, but it belongs to the context. Also only the consumers of this specific context should have access to the timeout, there is no reason to keep it global for the whole Crawler.

src/crawlee/_utils/time.py
```diff
@@ -35,6 +38,40 @@ def measure_time() -> Iterator[TimerResult]:
         result.cpu = after_cpu - before_cpu
 
 
+class SharedTimeout:
```
Pijukatel, line 41: Nice idea. I like this a lot. 

Just thinking out loud, where this could lead to:
In case we need to create more granular timeouts for specific steps, I think this class could be easily expanded to support that. I am wondering if even the final context consumer (request handler) could just use timeout from here if the timeout is set on the context (my other comment)

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
janbuchar: Same here :slightly_smiling_face: I want to avoid exposing the timeout to the request handler, mostly because it just doesn't make sense to me to do so.

Since the tests are now failing because of this, I guess we can agree that this approach is not optimal - I'll iterate on it for a while.

src/crawlee/_utils/time.py
janbuchar: It makes the type checker happier :man_shrugging:

src/crawlee/_utils/time.py
janbuchar: I'm not sure I follow, do you think that the request handler should be limited by a shared timeout? Or that it should have access to the remaining timeout "budget"?

Mantisus (approved):
An excellent solution with context for a general navigation timeout!

src/crawlee/_utils/time.py
Pijukatel: Please do not focus on any specific interface example in my example. It is just about capability.

Imagine that the context would be created something like this:
```
pipeline_timeout = SharedTimeout(...)

BasicCrawlingContext(.....,
 timeouts={
"WholePipeline": pipeline_timeout,   # Maybe the other timeouts could be somehow limited by this one? 
"Navigation": pipeline_timeout.limited_to(NAVIGATION_LIMIT),
"RequestHandler": pipeline_timeout.limited_to(HANDLER_LIMIT)
})
```

And each timeout-protected context consumer would pick the relevant timeout from the context and apply it. Context consumers could even modify the timeouts of the steps that follow them.

For example, users could mutate "RequestHandler" timeout in pre-navigation hooks.

Pijukatel: No need to block this change for this. If needed, we can discuss here: https://github.com/apify/crawlee-python/issues/1596

Pijukatel (approved):

vdusek (approved):

Issue #1469, opened 2025-10-08
Test `test_crawler_instrumentor_capability` is flaky


vdusek:
- On Windows.
- Code [`test_crawler_instrumentor_capability`](https://github.com/apify/crawlee-python/blob/v1.0.2/tests/unit/otel/test_crawler_instrumentor.py#L23).
- Failing log run [here](https://github.com/apify/crawlee-python/actions/runs/18276318550/job/52029163683).

Issue #1460, opened 2025-10-07
Failing tests on Windows + Python 3.10/3.11


vdusek:
After the recent lockfile maintenance ([#1455](https://github.com/apify/crawlee-python/pull/1455)), a few tests started failing on Windows with Python 3.10 and 3.11:
- [test_gz_sitemap_with_non_gzipped](https://github.com/apify/crawlee-python/actions/runs/18303257601/job/52115062290)
- [test_crawler_instrumentor_capability](https://github.com/apify/crawlee-python/actions/runs/18276318550/job/52029163683)

vdusek:
still flaky, https://github.com/apify/crawlee-python/actions/runs/18586313750/job/52990892997

Pijukatel:
I had hoped that [this](https://github.com/apify/crawlee-python/pull/1493) was the root cause, but unfortunately not. Anyway, I still believe it was caused by some dependency bump

PR #1478, opened 2025-10-13
chore: fix flaky tests

Mantisus:
### Issues

- Closes: #1469
- Closes: #1460

### DIFF
--- a/tests/unit/otel/test_crawler_instrumentor.py
+++ b/tests/unit/otel/test_crawler_instrumentor.py
@@ -47,7 +47,7 @@ async def test_crawler_instrumentor_capability(server_url: URL) -> None:
     crawler = ParselCrawler(
         max_requests_per_crawl=1,
         request_handler=mock.AsyncMock(),
-        concurrency_settings=ConcurrencySettings(desired_concurrency=1),
+        concurrency_settings=ConcurrencySettings(desired_concurrency=1, max_concurrency=1),
     )
 
     # Run crawler and generate more telemetry data.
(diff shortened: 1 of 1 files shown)


tests/unit/otel/test_crawler_instrumentor.py
```diff
@@ -91,8 +91,16 @@ async def test_crawler_instrumentor_capability(server_url: URL) -> None:
     assert telemetry_data[-1]['attributes']['code.function.name'] == 'BasicCrawler.__run_task_function'
     assert telemetry_data[-1]['resource']['attributes'] == dict(resource.attributes)
 
+    last_run_fuction_data = None
```
Pijukatel: Is this needed?  Maybe setting `max_requests_per_crawl=1` and `ConcurrencySettings(desired_concurrency=1, max_concurrency=1)` should be enough

Mantisus: I added this because I found some strange snapshots in the latest indexes a couple of times. Unfortunately, I didn't save the examples.

Pijukatel: It would be good to fully understand the issue; otherwise, I am worried it could be hiding some other problem.
Why are there some "strange snapshots in latest indexes" sometimes?

Mantisus: Hmm. I probably encountered either a very rare case or a side effect caused by something else. Since I couldn't reproduce it. I'm removing this workaround for now, let's see if anything strange happens in CI.

PR #1487, opened 2025-10-15
feat: add `chrome` `BrowserType` for `PlaywrightCrawler` to use the Chrome browser

Mantisus:
### Description

- add `chrome` `BrowserType` for `PlaywrightCrawler` to use the Chrome browser

### Issues

- Closes: #1071

### DIFF
--- a/src/crawlee/browsers/_browser_pool.py
+++ b/src/crawlee/browsers/_browser_pool.py
@@ -118,8 +118,10 @@ class BrowserPool:
         """Initialize a new instance with a single `PlaywrightBrowserPlugin` configured with the provided options.
 
         Args:
-            browser_type: The type of browser to launch ('chromium', 'firefox', 'webkit' or 'chrome'). Use `chrome` to
-                use the installed Chrome browser instead of Chromium.
+            browser_type: The type of browser to launch:
+                - 'chromium', 'firefox', 'webkit': Use Playwright-managed browsers
+                - 'chrome': Use your locally installed Google Chrome browser. Requires Google Chrome to be installed on
+                    the system.
             user_data_dir: Path to a user data directory, which stores browser session data like cookies
                 and local storage.
             browser_launch_options: Keyword arguments to pass to the browser launch method. These options are provided
--- a/src/crawlee/browsers/_playwright_browser_plugin.py
+++ b/src/crawlee/browsers/_playwright_browser_plugin.py
@@ -55,8 +55,10 @@ class PlaywrightBrowserPlugin(BrowserPlugin):
         """Initialize a new instance.
 
         Args:
-            browser_type: The type of browser to launch ('chromium', 'firefox', 'webkit' or 'chrome'). Use `chrome` to
-                use the installed Chrome browser instead of Chromium.
+            browser_type: The type of browser to launch:
+                - 'chromium', 'firefox', 'webkit': Use Playwright-managed browsers
+                - 'chrome': Use your locally installed Google Chrome browser. Requires Google Chrome to be installed on
+                    the system.
             user_data_dir: Path to a User Data Directory, which stores browser session data like cookies and local
                 storage.
             browser_launch_options: Keyword arguments to pass to the browser launch method. These options are provided
@@ -86,8 +88,10 @@ class PlaywrightBrowserPlugin(BrowserPlugin):
                 'Cannot use browser_type `chrome` with `Configuration.default_browser_path` or `executable_path` set.'
             )
 
+        # Map 'chrome' to 'chromium' with the 'chrome' channel.
         if browser_type == 'chrome':
             browser_type = 'chromium'
+            # Chromium parameter 'channel' set to 'chrome' enables using installed Google Chrome.
             default_launch_browser_options['channel'] = 'chrome'
 
         self._browser_type: BrowserType = browser_type
--- a/src/crawlee/crawlers/_playwright/_playwright_crawler.py
+++ b/src/crawlee/crawlers/_playwright/_playwright_crawler.py
@@ -114,8 +114,10 @@ class PlaywrightCrawler(BasicCrawler[PlaywrightCrawlingContext, StatisticsState]
             browser_pool: A `BrowserPool` instance to be used for launching the browsers and getting pages.
             user_data_dir: Path to a user data directory, which stores browser session data like cookies
                 and local storage.
-            browser_type: The type of browser to launch ('chromium', 'firefox', 'webkit' or 'chrome'). Use `chrome` to
-                use the installed Chrome browser instead of Chromium.
+            browser_type: The type of browser to launch:
+                - 'chromium', 'firefox', 'webkit': Use Playwright-managed browsers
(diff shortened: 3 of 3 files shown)


tests/unit/browsers/test_playwright_browser_plugin.py
```diff
@@ -69,3 +69,22 @@ async def test_methods_raise_error_when_not_active() -> None:
...
     async with plugin:
         assert plugin.active is True
+
+
+async def raise_error_if_use_chrome_and_executable_path() -> None:
+    with pytest.raises(
+        ValueError, match=r'Cannot use `use_chrome` with `Configuration.default_browser_path` or `executable_path` set.'
+    ):
+        PlaywrightBrowserPlugin(
+            browser_type='chromium',
```
Pijukatel: To me, this seems a little strange from the user perspective. It is defining one type of browser in one argument and then overriding it in another argument, which is valid only for a specific value of the first argument. I would prefer to just add browser_type=`chrome` and connect any logic to such a specific browser_type internally.

Mantisus: Since `browser_type` in Crawlee is a parameter that directly corresponds to a similar parameter in Playwright, I think adding a new option to it will confuse users familiar with Playwright.

Pijukatel: I think that users familiar with Playwright are probably OK with the current state of defining `channel` in `browser_launch_options` - as that is the Playwright way of using Chrome and offers the best flexibility without us needing to change anything.

Experienced Playwright users aside, when considering anyone else, I think it is not a very intuitive way to select Chrome with this change.

Just consider the following snippets:

To use Firefox:
`PlaywrightCrawler(browser_type="firefox")`

But to use Chrome
`PlaywrightCrawler(browser_type="chromium", use_chrome=True)`

I think that most users would expect just this instead:
`PlaywrightCrawler(browser_type="chrome")`

Anyway, I do not have a very strong opinion about this, but I would suggest introducing new arguments only if we see big improvements in readability, which I do not think is the case now.

Mantisus: Maybe @vdusek, @janbuchar have some ideas about which approach would be best to choose?

vdusek: I agree with @Pijukatel

Mantisus: Updated

src/crawlee/browsers/_browser_pool.py
```diff
@@ -118,7 +118,8 @@ def with_default_plugin(
         """Initialize a new instance with a single `PlaywrightBrowserPlugin` configured with the provided options.
 
         Args:
-            browser_type: The type of browser to launch ('chromium', 'firefox', or 'webkit').
+            browser_type: The type of browser to launch ('chromium', 'firefox', 'webkit' or 'chrome'). Use `chrome` to
+                use the installed Chrome browser instead of Chromium.
```
vdusek: If I understand correctly, `chromium`, `firefox`, and `webkit` are Playwright's bundled browsers, but when the `chrome` option is used, it relies on the user's installed Chrome instance. If that's correct, could you please update the docstring to clarify this?

src/crawlee/browsers/_playwright_browser_plugin.py
```diff
@@ -55,7 +55,8 @@ def __init__(
         """Initialize a new instance.
 
         Args:
-            browser_type: The type of browser to launch ('chromium', 'firefox', or 'webkit').
+            browser_type: The type of browser to launch ('chromium', 'firefox', 'webkit' or 'chrome'). Use `chrome` to
+                use the installed Chrome browser instead of Chromium.
```
vdusek: If I understand correctly, `chromium`, `firefox`, and `webkit` are Playwright's bundled browsers, but when the `chrome` option is used, it relies on the user's installed Chrome instance. If that's correct, could you please update the docstring to clarify this?

src/crawlee/browsers/_playwright_browser_plugin.py
```diff
@@ -80,6 +81,15 @@ def __init__(
...
         }
 
+        if browser_type == 'chrome' and default_launch_browser_options['executable_path']:
+            raise ValueError(
+                'Cannot use browser_type `chrome` with `Configuration.default_browser_path` or `executable_path` set.'
+            )
+
+        if browser_type == 'chrome':
+            browser_type = 'chromium'
+            default_launch_browser_options['channel'] = 'chrome'
```
vdusek, line 95: Could we please add a comment explaining this?

src/crawlee/crawlers/_playwright/_playwright_crawler.py
```diff
@@ -114,7 +114,8 @@ def __init__(
             browser_pool: A `BrowserPool` instance to be used for launching the browsers and getting pages.
             user_data_dir: Path to a user data directory, which stores browser session data like cookies
                 and local storage.
-            browser_type: The type of browser to launch ('chromium', 'firefox', or 'webkit').
+            browser_type: The type of browser to launch ('chromium', 'firefox', 'webkit' or 'chrome'). Use `chrome` to
+                use the installed Chrome browser instead of Chromium.
```
vdusek: If I understand correctly, `chromium`, `firefox`, and `webkit` are Playwright's bundled browsers, but when the `chrome` option is used, it relies on the user's installed Chrome instance. If that's correct, could you please update the docstring to clarify this?

src/crawlee/crawlers/_playwright/_playwright_crawler.py
```diff
@@ -153,7 +154,7 @@ def __init__(
             ):
                 raise ValueError(
                     'You cannot provide `headless`, `browser_type`, `browser_launch_options`, '
-                    '`browser_new_context_options`, `use_incognito_pages`, `user_data_dir`  or'
+                    '`browser_new_context_options`, `use_incognito_pages`, `user_data_dir` or'
```
vdusek:
```suggestion
                    '`browser_new_context_options`, `use_incognito_pages`, `user_data_dir` or '
```

src/crawlee/crawlers/_playwright/_playwright_crawler.py
```diff
@@ -494,7 +495,8 @@ class _PlaywrightCrawlerAdditionalOptions(TypedDict):
     """A `BrowserPool` instance to be used for launching the browsers and getting pages."""
 
     browser_type: NotRequired[BrowserType]
-    """The type of browser to launch ('chromium', 'firefox', or 'webkit').
+    """The type of browser to launch ('chromium', 'firefox', 'webkit' or 'chrome'). Use `chrome` to
+    use the installed Chrome browser instead of Chromium.
```
vdusek: If I understand correctly, `chromium`, `firefox`, and `webkit` are Playwright's bundled browsers, but when the `chrome` option is used, it relies on the user's installed Chrome instance. If that's correct, could you please update the docstring to clarify this?

Issue #1483, opened 2025-10-14
Updating request inside "default_handler", "error_handler" and "failed_request_handler"


ericvg97:
Is there a way to update the request.user_data inside these handlers? I would like to store info about the last error, or the timestamp of the different retries. 

This is my handler 

    async def default_handler(context: AdaptivePlaywrightCrawlingContext) -> None:
            context.log.info(f"Processing {context.request.url} ...")
            context.request.user_data["something"] = "something"
            await context.enqueue_links(strategy="same-domain")
    
    crawler.router.default_handler(default_handler)

but looking at the db this field is not being saved. I just see ""user_data":{"__crawlee":{"state":3}"

I guess the context is either not saved after that or it is sent as a copy, but not sure from looking at the code. Any other way of doing this that I am not seeing?

Mantisus:
I checked it, and the code provided should be processed correctly.

```python
import asyncio

from crawlee import ConcurrencySettings
from crawlee.crawlers import ParselCrawler, ParselCrawlingContext
from crawlee.http_clients import HttpxHttpClient
from crawlee.storage_clients import SqlStorageClient


async def main() -> None:
    async with (
        SqlStorageClient(connection_string='postgresql+asyncpg://myuser:mypassword@localhost:5432/postgres') as storage_client
            ):
        http_client = HttpxHttpClient()

        crawler = ParselCrawler(
            storage_client=storage_client,
            http_client=http_client,
            concurrency_settings=ConcurrencySettings(desired_concurrency=20),
        )

        @crawler.router.default_handler
        async def request_handler(context: ParselCrawlingContext) -> None:
            context.log.info(f'Processing URL: {context.request.url}...')
            context.request.user_data['something'] = 'something'

        await crawler.run(['https://crawlee.dev'])


if __name__ == '__main__':
    asyncio.run(main())
```

<img width="1423" height="213" alt="Image" src="https://github.com/user-attachments/assets/1a5c3a37-46ea-4c51-a6f2-ef05bd8ce04f" />

Could you double-check this and let me know if the data is not being updated?

ericvg97:
The example you sahred works, but just changing ParselCrawler to AdaptativePlaywrightCrawler breaks it. This is the new code (that doesn't work)


```
async def main() -> None:
    async with SqlStorageClient(
        connection_string="postgresql+asyncpg://postgres:password@localhost:5432/lighthouse"
    ) as storage_client:
        http_client = HttpxHttpClient()

[...]

PR #1488, opened 2025-10-15
fix: save context state in result for `AdaptivePlaywrightCrawler` after isolated processing in `SubCrawler`

Mantisus:
### Description

- save context state in result for `AdaptivePlaywrightCrawler` after isolated processing in `SubCrawler`

### Issues

- Closes: #1483

### DIFF
--- a/src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
+++ b/src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
@@ -458,24 +458,22 @@ class AdaptivePlaywrightCrawler(
         Uses object.__setattr__ to bypass frozen dataclass restrictions,
         allowing state synchronization after isolated crawler execution.
         """
-        updating_attributes = ('request', 'session')
+        updating_attributes = {
+            'request': ('headers', 'user_data'),
+            'session': ('_user_data', '_usage_count', '_error_score', '_cookies'),
+        }
 
-        for attr_name in updating_attributes:
-            original_obj = getattr(context, attr_name, None)
-            updated_obj = getattr(context_copy, attr_name, None)
+        for attr, sub_attrs in updating_attributes.items():
+            original_sub_obj = getattr(context, attr)
+            copy_sub_obj = getattr(context_copy, attr)
 
-            # Skip if either object is None or they are the same object
-            if (
-                original_obj is None
-                or updated_obj is None
-                or original_obj is updated_obj
-                or not hasattr(updated_obj, '__dict__')
-            ):
+            # Check that both sub objects are not None
+            if original_sub_obj is None or copy_sub_obj is None:
                 continue
 
-            # Copy all attributes using __dict__
-            for obj_attr_name, obj_attr_value in updated_obj.__dict__.items():
-                object.__setattr__(original_obj, obj_attr_name, obj_attr_value)
+            for sub_attr in sub_attrs:
+                new_value = getattr(copy_sub_obj, sub_attr)
+                object.__setattr__(original_sub_obj, sub_attr, new_value)
 
 
 @dataclass(frozen=True)


Pijukatel:
Please add the test that will cover the reported issue.

Mantisus:
> Please add the test that will cover the reported issue.

Added

tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
```diff
@@ -727,3 +728,40 @@ async def request_handler(context: AdaptivePlaywrightCrawlingContext) -> None:
     await crawler.run(test_urls[:1])
 
     mocked_h3_handler.assert_called_once_with(None)
+
+
+async def test_change_context_state_after_handling(server_url: URL) -> None:
+    """Test that context state is saved after handling the request."""
+    static_only_predictor_no_detection = _SimpleRenderingTypePredictor(detection_probability_recommendation=cycle([0]))
```
janbuchar: I think that the test should cover all three possible cases - static crawl, browser crawl and rendering type detection (both crawl types).

Also, please make sure that the modifications to user data are not done twice in the case of a detection. Adding something to a list-typed user data item should do nicely.

tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
```diff
@@ -727,3 +728,40 @@ async def request_handler(context: AdaptivePlaywrightCrawlingContext) -> None:
...
+
+    test_mapper = {}
```
janbuchar: I believe this could be just a single variable, making the code a bit easier to understand.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -455,8 +456,34 @@ def track_browser_request_handler_runs(self) -> None:
     def track_rendering_type_mispredictions(self) -> None:
         self.statistics.state.rendering_type_mispredictions += 1
 
+    def _update_context_from_copy(self, context: BasicCrawlingContext, context_copy: BasicCrawlingContext) -> None:
+        """Update mutable fields of `context` from `context_copy`.
+
+        Uses object.__setattr__ to bypass frozen dataclass restrictions,
+        allowing state synchronization after isolated crawler execution.
+        """
+        updating_attributes = ('request', 'session', 'proxy_info')
```
janbuchar: I don't think that the user should be able to change anything in the `proxy_info` - what am I missing?

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -455,8 +456,34 @@ def track_browser_request_handler_runs(self) -> None:
...
+            if (
+                original_obj is None
+                or updated_obj is None
+                or original_obj is updated_obj
+                or not hasattr(updated_obj, '__dict__')
+            ):
+                continue
+
+            # Copy all attributes using __dict__
+            for obj_attr_name, obj_attr_value in updated_obj.__dict__.items():
```
janbuchar: This is quite an overkill. I'd much rather just enumerate the attributes for which it makes sense to mutate them from the request handler.

Mantisus: Yes, but this is consistent with other crawlers. The user can change the attributes of `Request`, even if it doesn't make sense or is better not to do so, for example, changing `context.request.unique_key`.
We would probably like to change this, but that's a task for a separate PR.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
Mantisus: Good point

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
janbuchar: I mean, yeah, the user can do it, but if it's something that we'd like to prohibit in the future, we don't need to write code to support it here 🤷

Mantisus: Updated using specific attributes

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -417,6 +415,9 @@ async def _run_request_handler(self, context: BasicCrawlingContext) -> None:
                 context.log.debug(f'Detected rendering type {detection_result} for {context.request.url}')
                 self.rendering_type_predictor.store_result(context.request, detection_result)
 
+            self._update_context_from_copy(context, pw_run.run_context)
```
janbuchar, line 418: I think that we should make the changes made to the context part of the `RequestHandlerResult` data structure at some point. What do you think @Mantisus @Pijukatel?

No need to stall this PR because of that, it can be done later.

janbuchar: https://github.com/apify/crawlee-python/issues/1514

PR #1490, opened 2025-10-15
fix: Fix `BasicCrawler` statistics persistence

Pijukatel:
### Description

- Ensure that  `BasicCrawler` is persisting statistics by default.
- Ensure that  `BasicCrawler` is recovering existing statistics by default if `Configuration.purge_on_start ` is False.
- Let the `BasicCrawler` emit `Event.PERSIST_STATE` when finishing.

### Issues

- Closes: #1501

### Testing

- Added unit test
- Tested on SDK level: https://github.com/apify/apify-sdk-python/pull/629


tests/unit/test_configuration.py
```diff
@@ -41,10 +41,10 @@ async def test_storage_not_persisted_when_disabled(tmp_path: Path, server_url: U
...
     storage_client = MemoryStorageClient()
 
-    crawler = HttpCrawler(
-        configuration=configuration,
-        storage_client=storage_client,
-    )
+    service_locator.set_configuration(configuration)
+    service_locator.set_storage_client(storage_client)
+
+    crawler = HttpCrawler()
```
vdusek: Why?

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -440,6 +440,7 @@ def __init__(
         self._statistics = statistics or cast(
             'Statistics[TStatisticsState]',
             Statistics.with_default_state(
+                persistence_enabled=True,
```
vdusek: I suppose changing the default values (`persistence_enabled: bool = True`) is a no-go in patch releases, right?

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -742,7 +740,20 @@ async def _run_crawler(self) -> None:
             for context in contexts_to_enter:
                 await exit_stack.enter_async_context(context)  # type: ignore[arg-type]
 
-            await self._autoscaled_pool.run()
+            self._crawler_state_rec_task.start()
+            try:
+                await self._autoscaled_pool.run()
+            finally:
+                await self._crawler_state_rec_task.stop()
```
vdusek: Maybe `RecurringTask` could also be an async context manager (two options of usage - `start`/`stop` and `aenter`/`aexit`) and then it can be just another `contexts_to_enter`.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -742,7 +740,20 @@ async def _run_crawler(self) -> None:
...
+
+            # Emit PERSIST_STATE event when crawler is finishing to allow listeners to persist their state if needed
+            if not self.statistics.state.crawler_last_started_at:
+                raise RuntimeError('Statistics.state.crawler_last_started_at not set.')
+            run_duration = datetime.now(timezone.utc) - self.statistics.state.crawler_last_started_at
+            self._statistics.state.crawler_runtime = self.statistics.state.crawler_runtime + run_duration
+            self._service_locator.get_event_manager().emit(
+                event=Event.PERSIST_STATE, event_data=EventPersistStateData(is_migrating=False)
+            )
```
```suggestion
            event_manager.emit(
                event=Event.PERSIST_STATE,
                event_data=EventPersistStateData(is_migrating=False),
            )
```

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -742,7 +740,20 @@ async def _run_crawler(self) -> None:
...
+            try:
+                await self._autoscaled_pool.run()
+            finally:
+                await self._crawler_state_rec_task.stop()
+
+            # Emit PERSIST_STATE event when crawler is finishing to allow listeners to persist their state if needed
+            if not self.statistics.state.crawler_last_started_at:
+                raise RuntimeError('Statistics.state.crawler_last_started_at not set.')
+            run_duration = datetime.now(timezone.utc) - self.statistics.state.crawler_last_started_at
+            self._statistics.state.crawler_runtime = self.statistics.state.crawler_runtime + run_duration
```
vdusek: This doesn't seem to belong here - this part handles high-level crawler behavior, so computing the run duration directly here feels wrong - Maybe move it to a separate method or to `Statistics`?

vdusek (changes requested):
I'm surprised we use the `SDK_CRAWLER_STATISTICS_...` key for state persistence. Why is the SDK prefix in Crawlee? Also, since this is internal, we use a double-underscore prefix (`__STORAGE_ALIASES_MAPPING`, `__RQ_STATE_...`) for other cases. Could we update the key name, please?

tests/unit/test_configuration.py
Pijukatel: This is because `RecoverableState` of statistics persists to/recovers from global `storage_client`. And since statistics is persisted by default now, it will try to persist to default global service_client, which is FileSystem... regardless of the crawler-specific storage_client

Mentioned here:
https://github.com/apify/crawlee-python/pull/1438#issuecomment-3359681615

I am open to discussion about this.

src/crawlee/crawlers/_basic/_basic_crawler.py
Pijukatel: Well, it is changing the default value of the internal attribute... since we consider previous behavior a bug, this is probably ok?

src/crawlee/crawlers/_basic/_basic_crawler.py
Pijukatel: Done

src/crawlee/crawlers/_basic/_basic_crawler.py
Pijukatel: Done

src/crawlee/crawlers/_basic/_basic_crawler.py
vdusek: I meant changing the default value in the `Statistics.with_default_state`:

```python
Statistics.with_default_state(persistence_enabled: bool = True)
```

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -742,7 +740,11 @@ async def _run_crawler(self) -> None:
             for context in contexts_to_enter:
                 await exit_stack.enter_async_context(context)  # type: ignore[arg-type]
 
-            await self._autoscaled_pool.run()
+            async with self._crawler_state_rec_task:
```
vdusek: `self._crawler_state_rec_task` can't be part of `contexts_to_enter`? It's `aexit` must be called before the `event_manager.emit`?

tests/unit/test_configuration.py
vdusek: Couldn't we use the storage client passed to the crawler?

src/crawlee/crawlers/_basic/_basic_crawler.py
Pijukatel: I see. Yes, that is why I changed it only in `BasicCrawler` and not in `Statistics.with_default_state`

tests/unit/test_configuration.py
Pijukatel: We could, but do we want to? I had an inconclusive discussion about this with @janbuchar
I am still not sure about this.

src/crawlee/crawlers/_basic/_basic_crawler.py
Pijukatel: Good point, done.

vdusek (approved):
LGTM

tests/unit/test_configuration.py
janbuchar: Honestly, I'm kinda disappointed with the amount of edge cases that arose from having a separate service locator for crawlers. 

From a "common sense" perspective, the `RecoverableState` is owned by the crawler and it doesn't make much sense to put the serialized state in a different storage (the global one). Then again, there's a good chance that the crawler-wide storage client will be a memory storage, which is not a great fit for `RecoverableState`.

But, unless I'm missing something, it should be super rare that somebody will do this intentionally. In my opinion, we should pick one of these options and just show a warning if both the global and crawler-specific storage client are configured.

PR #1493, opened 2025-10-16
chore: Fix increased flakinnes of curl_cffi based tests

Pijukatel:
### Description

- Limit `cffi` to version <2.0.0 to avoid increased test flakiness on Windows and Python 3.10

### Issues

- Closes: #1494

### Testing

- Running tests in CI and monitoring for flakiness

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,6 @@ dependencies = [
     "pydantic-settings>=2.2.0,!=2.7.0,!=2.7.1,!=2.8.0",
     "pydantic>=2.11.0",
     "pyee>=9.0.0",
-    "pytest-randomly>=4.0.1",
     "tldextract>=5.1.0",
     "typing-extensions>=4.1.0",
     "yarl>=1.18.0",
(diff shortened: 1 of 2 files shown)


Pijukatel:
Running on Windows with Python3.10 I got a lot of event loop warnings and maybe this one is specifically for curl based tests

> tests/unit/http_clients/test_curl_impersonate.py: 4 warnings
>   D:\Workspace\Python\crawlee-python\.venv\lib\site-packages\curl_cffi\aio.py:144: CurlCffiWarning:
>       Proactor event loop does not implement add_reader family of methods required.
>       Registering an additional selector thread for add_reader support.
>       To avoid this warning use:
>           asyncio.set_event_loop_policy(WindowsSelectorEventLoopPolicy())

Given that, maybe it would be best to use `WindowsSelectorEventLoopPolicy` all the time in our tests

Pijukatel:
Randomizing test order is enough to make UTs stable. This means that some tests are not properly isolated. This could cause some race conditions for shared resources. Maybe the session-scoped test server together with xdist is the problem, too hard to debug and most likely not a product  issue, but test setup issue.

Issue #1495, opened 2025-10-16
Crawler's storage_client errors when using SqlStorageClient


ericvg97:
This crawler
```
class Crawler:
    def __init__(self, storage_client: StorageClient):
        self.storage_client = storage_client

    async def crawl(
        self,
        seed_url: str,
        purge_on_start: bool = False,
    ) -> None:
        print(f"Crawling {seed_url} ...")

        config = service_locator.get_configuration()
        config.purge_on_start = purge_on_start
        # service_locator.set_storage_client(storage_client=self.storage_client)

        concurrency_settings = ConcurrencySettings(
            min_concurrency=1,
            max_concurrency=1,
            desired_concurrency=1,
            max_tasks_per_minute=200,
        )
        statistics = Statistics.with_default_state(log_interval=timedelta(minutes=1), statistics_log_format="table")

        crawler = AdaptivePlaywrightCrawler.with_beautifulsoup_static_parser(
            max_requests_per_crawl=10000,
            playwright_crawler_specific_kwargs={
                "browser_type": "chromium",
                "headless": True,
            },
            max_session_rotations=10,
            retry_on_blocked=True,
            concurrency_settings=concurrency_settings,
            statistics=statistics,
            configure_logging=True,
            use_session_pool=True,
            max_request_retries=5,
            request_handler_timeout=timedelta(seconds=30),
            keep_alive=False,
            respect_robots_txt_file=True,
            storage_client=self.storage_client,
        )

        crawler.pre_navigation_hook(pre_navigation_hook)
        crawler.router.default_handler(default_handler)
        crawler.error_handler(error_handler)
        crawler.failed_request_handler(failed_request_handler)

        await crawler.run([seed_url])

        logger.info("Finished crawling")
```

breaks when storage client is of type SqlStorageClient with error:
```
scripts/crawling/crawl_url.py:13: UserWarning: The SqlStorageClient is experimental and may change or be removed in future releases.
  sql_client = SqlStorageClient(engine=engine)
Crawling http://httpbingo.org/status/429 ...
No configuration set, implicitly creating and using default Configuration.
Traceback (most recent call last):
  File "crawl_url.py", line 21, in <module>
    tyro.cli(crawl_entrypoint)
  File ".venv/lib/python3.12/site-packages/tyro/_cli.py", line 238, in cli
    return run_with_args_from_cli()
           ^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1496, opened 2025-10-17
fix: Resolve compatibility issue between `SqlStorageClient` and `AdaptivePlaywrightCrawler`

Mantisus:
### Description

- Remove the use of `deepcopy` for `kwargs` in `AdaptivePlaywrightCrawler` for compatibility with `SqlStorageClient`

### Issues

- Closes: #1495

### DIFF
--- a/tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
+++ b/tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
@@ -30,10 +30,12 @@ from crawlee.crawlers._adaptive_playwright._adaptive_playwright_crawling_context
     AdaptiveContextError,
 )
 from crawlee.statistics import Statistics
+from crawlee.storage_clients import SqlStorageClient
 from crawlee.storages import KeyValueStore
 
 if TYPE_CHECKING:
     from collections.abc import AsyncGenerator, Iterator
+    from pathlib import Path
 
     from yarl import URL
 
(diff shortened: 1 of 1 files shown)


src/crawlee/storage_clients/_sql/_storage_client.py
```diff
@@ -94,6 +94,29 @@ async def __aexit__(
         """Async context manager exit."""
         await self.close()
 
+    def __deepcopy__(self, memo: dict[int, Any] | None) -> SqlStorageClient:
```
vdusek: Could you please add a docstring or a comment explaining the reason why we must implement it?

janbuchar:
Huh. _Huh_. While this does fix the issue, I think the root cause is the unfortunate way `deepcopy` is used when constructing the subcrawlers in adaptive crawler. Could we fix that instead?

Mantisus:
> Huh. Huh. While this does fix the issue, I think the root cause is the unfortunate way deepcopy is used when constructing the subcrawlers in adaptive crawler. Could we fix that instead?

Updated.

@Pijukatel could you take a look at this as well? You are the most familiar with the `AdaptivePlaywrightCrawler`. I just want to make sure I'm not missing any important cases where `deepcopy` was used.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -25,6 +25,7 @@
     PlaywrightCrawlingContext,
     PlaywrightPreNavCrawlingContext,
 )
+from crawlee.crawlers._basic._basic_crawler import _BasicCrawlerOptions
```
vdusek: I'm sorry, but could you please rename `_BasicCrawlerOptions` to `BasicCrawlerOptions` since we're importing it from another module?

janbuchar: Or we could just skip the "conversion" because `TypedDict` doesn't really do anything at runtime, so there is no benefit

src/crawlee/storage_clients/_sql/_storage_client.py
janbuchar, line 1: I believe we can now remove these changes (and the test). I don't see a good reason to support deep-copying a storage client.

Mantisus: Yeah, that makes sense.

janbuchar (approved):
LGTM

janbuchar:
Perhaps we could add a test that shows that adaptive crawler can be instantiated with sql storage client, that wouldn't be too hard :slightly_smiling_face:

Pijukatel (approved):

PR #1525, opened 2025-10-30
feat: Add `transform_request_function` parameter for `SitemapRequestLoader`

Mantisus:
### Description

This PR is inspired by this [discussion](https://github.com/apify/crawlee-python/discussions/1517).
Add support `transform_request_function` for `SitemapRequestLoader`, which works the same way as in `EnqueueLinksFunction`. This can be useful for setting `label` for correct routing or `user_data` with custom data.

### DIFF
--- a/docs/examples/code_examples/using_sitemap_request_loader.py
+++ b/docs/examples/code_examples/using_sitemap_request_loader.py
@@ -17,6 +17,8 @@ def create_transform_request(
     def transform_request(
         request_options: RequestOptions,
     ) -> RequestOptions | RequestTransformAction:
+        # According to the Sitemap protocol, all URLs in a Sitemap must be from a single
+        # host.
         request_host = URL(request_options['url']).host
 
         if request_host and (mapping_data := data_mapper.get(request_host)):


tests/unit/request_loaders/test_sitemap_request_loader.py
```diff
@@ -172,3 +173,24 @@ async def test_recovery_data_persistence_for_sitemap_loading(
...
+        request_options['user_data'] = {'transformed': True}
+        return request_options
+
+    sitemap_loader = SitemapRequestLoader(
+        [str(sitemap_url)],
+        http_client=http_client,
+        transform_request_function=transform_request,
+    )
+
+    while not await sitemap_loader.is_finished():
```
Pijukatel, line 193: Can we add an assertion about the correct number of requests extracted from the BASIC_SITEMAP?

if sitemap_loader.is_finished() returns True and no request was extracted, then this test still passes

src/crawlee/request_loaders/_sitemap_request_loader.py
```diff
@@ -112,6 +113,7 @@ def __init__(
         exclude: list[re.Pattern[Any] | Glob] | None = None,
         max_buffer_size: int = 200,
         persist_state_key: str | None = None,
+        transform_request_function: Callable[[RequestOptions], RequestOptions | RequestTransformAction] | None = None,
```
janbuchar, line 116: Buuut... shouldn't this also receive an URL of the origin sitemap?

Mantisus: I don't think that makes sense. A sitemap cannot contain links to another domain.
This way, users can easily create a mapping between the original link to the sitemap and the link inside `transform_request_function`. From my point of view, the most valuable thing that adding `transform_request_function` gives is the ability to add a `label` so that the request is processed by the appropriate handler.

janbuchar: That makes a lot of sense, thanks. But I'm afraid that this won't "click" for a lot of people. Perhaps we could add an example that showcases this?

docs/examples/code_examples/using_sitemap_request_loader.py
```diff
@@ -0,0 +1,99 @@
...
+
+# Create a transform_request_function that maps request options based on the host in
+# the URL
+def create_transform_request(
+    data_mapper: dict[str, dict],
+) -> Callable[[RequestOptions], RequestOptions | RequestTransformAction]:
+    def transform_request(
+        request_options: RequestOptions,
+    ) -> RequestOptions | RequestTransformAction:
+        request_host = URL(request_options['url']).host
```
janbuchar, line 22: Maybe we should mention that a sitemap should only contain links to the same host here

Mantisus: Good point. Added.

Issue #1526, opened 2025-10-31
Fetching new requests in Postgres does full table scan


ericvg97:
I am using version 1.0.4 with SQL storage in Postgres. My queue has around 2 millions rows (only approx 10% have been handled yet) and I am using 10 crawler instances. At some point, all my instances stopped crawling. I found out it is because the query to fetch new requests is timing out (more than 1 minute). The query is:
```
SELECT *
FROM request_queue_records
where request_queue_id = [queue id]
and is_handled = false
  AND (time_blocked_until IS NULL OR time_blocked_until < now())
ORDER BY sequence_number ASC
LIMIT 10
FOR UPDATE SKIP LOCKED;
```
It seems that postgres isn't using the index (run with _EXPLAIN (ANALYZE, BUFFERS, VERBOSE)_ to see) and it does a full table scan. I think it is because of the filtering done on "time_blocked_until IS NULL" but I am not sure. I manually added my own index `CREATE INDEX sorting_by_sequence_number ON public.request_queue_records USING btree (sequence_number)` and now the query runs through the index (no table scan) and takes only a few milliseconds. Could you add this index (or fix the idx_fetch_available one)? 

Also, as a side issue, when this query was timing out, crawlee didn't fire any warnings nor errors, just kept scheduling more tasks that eventually timed out but when looking to the logs, even in debug mode, nothing helped find the issue (I had to fork the project and add logs to see the root problem) I just saw no new pages were being crawled.

Thanks a lot!

PR #1527, opened 2025-10-31
fix: Improve indexing of the `request_queue_records` table for `SqlRequestQueueClient`

Mantisus:
### Description

- This PR improves the index for the `request_queue_records` table, as the previous version of the index was not efficient for retrieving requests.

EXPLAIN result for over a million requests
Old index

image

New index

image

### Issues

- Closes: #1526

### DIFF
--- a/src/crawlee/storage_clients/_sql/_db_models.py
+++ b/src/crawlee/storage_clients/_sql/_db_models.py
@@ -206,7 +206,7 @@ class RequestDb(Base):
             'request_queue_id',
             'is_handled',
             'sequence_number',
-            postgresql_where=text('is_handled = false'),
+            postgresql_where=text('is_handled is false'),
         ),
     )
 


Mantisus:
@ericvg97, thank you for your analysis and for finding this. Is it possible for you to check the new index in your environment and confirm that the queue is working correctly with it?

ericvg97:
> @ericvg97, thank you for your analysis and for finding this. Is it possible for you to check the new index in your environment and confirm that the queue is working correctly with it?

It doesn't work in my environment. The query I shared was incorrect, the actual query crawlee is doing is
SELECT * FROM request_queue_records 
WHERE request_queue_id = 'QXxcAytWcFELg1yFR' 
AND is_handled IS false 
AND (time_blocked_until IS NULL OR time_blocked_until < now()) 
ORDER BY sequence_number ASC LIMIT 10 FOR UPDATE SKIP LOCKED

Notice the condition "is_handled IS false" vs "is_handled = false" that I shared before. When doing the first, the planner doesn't use the index you are proposing. When doing the second, the planner does use the index. I didn't know this could happen in postgres TBH. Can you check this also happens in your environmnet

ericvg97:
Changing  "postgresql_where=text('is_handled = false')", to "postgresql_where=text('is_handled is false') fixes it in my environment.

Mantisus:
> Changing "postgresql_where=text('is_handled = false')", to "postgresql_where=text('is_handled is false') fixes it in my environment.

@ericvg97 Thank you, I double-checked the query and updated the index. It works the same way in my environment.

> I didn't know this could happen in postgres TBH. 

Yes, I think it's because it's a partition index. Therefore, it is important that the expression fully matches the index.

Thanks again for double-checking this and helping to better identify the error. I really appreciate it.

janbuchar (approved):

PR #1540, opened 2025-11-07
fix: Fix `crawler_runtime` not being updated during run and only in the end

Pijukatel:
### Description

- Fix `BasicCrawler.statistics.state.crawler_runtime` to be properly updated on each `BasicCrawler.statistics.calculate()` call when the statistics are still active.
- Do not update `BasicCrawler.statistics.state.crawler_runtime` on `BasicCrawler.statistics.calculate()` when the statistics are already deactivated - to avoid confusing differences between logged and persisted state.

### Issues

Closes: #1541

### Testing

- Added test.

### Checklist

- [ ] CI passed

### DIFF
--- a/src/crawlee/statistics/_statistics.py
+++ b/src/crawlee/statistics/_statistics.py
@@ -110,7 +110,7 @@ class Statistics(Generic[TStatisticsState]):
         # Flag to indicate the context state.
         self._active = False
 
-        # Pre-existing runtime offset when importing existing statistics.
+        # Pre-existing runtime offset, that can be non-zero when restoring serialized state from KVS.
         self._runtime_offset = timedelta(seconds=0)
 
     def replace_state_model(self, state_model: type[TNewStatisticsState]) -> Statistics[TNewStatisticsState]:


src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -71,7 +71,6 @@ def __init__(self) -> None:
     async def __aenter__(self) -> Self:
         self._active = True
         await self._state.initialize()
-        self._after_initialize()
```
Pijukatel, line 74: I am not sure what this was for these dummy statistics. Could you please double-check @janbuchar ?

janbuchar: It was probably so that `record_*` methods wouldn't randomly fail during context pipeline execution because of incorrectly initialized state. I assume it's not necessary anymore?

src/crawlee/statistics/_statistics.py
```diff
@@ -111,6 +110,9 @@ def __init__(
         # Flag to indicate the context state.
         self._active = False
 
+        # Pre-existing runtime offset when importing existing statistics.
```
janbuchar: What does importing existing statistics mean here? Like restoring serialized state from KVS?

Pijukatel: Yes, updated comment.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
Pijukatel: I haven't seen any failure in tests, and those methods execute. But the same can be seen on master when deleting this line, so I guess it was made redundant by some other change?

janbuchar: Probably :crossed_fingers:

PR #1544, opened 2025-11-10
chore: Increase e2e Pytest timeout

Pijukatel:
### Description

- Camoufox templates fail due to Pytest timeout from time to time. This is mainly due to the Actor build taking too long, and there is not enough time left for the run. Increase the timeout to make enough time for both run and build.


Pijukatel:
E2E tests run on this branch pass: https://github.com/apify/crawlee-python/actions/runs/19226428933

PR #1513, opened 2025-10-23
chore: improve `HttpClient` isolating in tests

Mantisus:
### Issues

- Closes: #1441
- Closes: #1494
- Closes: #1460

Issue #1441, opened 2025-09-30
Fix root cause of flaky test crash [gw0] node down: Not properly terminated


Pijukatel:
Seems like this happens only on Windows and Python3.12, always the last test crahes with 
` [gw0] node down: Not properly terminated`

Seems like it was introduced when bumping uvicorn
https://github.com/apify/crawlee-python/pull/1421

vdusek:
I'm reopening this because https://github.com/apify/crawlee-python/pull/1447 did not fix the root cause. We should make our setup work even with uvicorn > 0.35.

PR #1552, opened 2025-11-11
chore: Relax uvicorn ~=0.35.0 constraint

Mantisus:
### Description

I think the issue in #1441 was related to isolating HTTP clients, and it was resolved in #1513. I don't think we should limit the version for `uvicorn` anymore.


Mantisus:
Close, as the issue is reproduced on uvicorn 0.38

PR #1553, opened 2025-11-13
feat: Add support for Python 3.14

Mantisus:
### Description

- Add support for Python 3.14

At this stage, it is not possible to provide PostgreSQL support for `SqlStorageClient` on Python 3.14, as `asyncpg` does not yet support Python 3.14.

### Testing

- Some tests for Redis are limited because `fakeredis` does not yet support Python 3.14, which causes problems for Windows.
- Updated the `test_memory_estimation_does_not_overestimate_due_to_shared_memory` test to use `multiprocessing.get_context('fork')` due to the update of `multiprocessing.Process` in Python 3.14/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -74,7 +74,7 @@ otel = [
 ]
 sql_postgres = [
     "sqlalchemy[asyncio]>=2.0.0,<3.0.0",
-    "asyncpg>=0.24.0; python_version < '3.14'" # Asyncpg has not yet been released with support for Python 3.14.
+    "asyncpg>=0.24.0; python_version < '3.14'" # TODO: https://github.com/apify/crawlee-python/issues/1555
 ]
 sql_sqlite = [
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -73,7 +74,7 @@ otel = [
 ]
 sql_postgres = [
     "sqlalchemy[asyncio]>=2.0.0,<3.0.0",
-    "asyncpg>=0.24.0"
+    "asyncpg>=0.24.0; python_version < '3.14'" # Asyncpg has not yet been released with support for Python 3.14.
```
vdusek: Do we also need to add a constraint on our side? I believe it will fail otherwise as well.

Mantisus: With `; python_version < '3.14'`, this package will be skipped during installation on Python 3.14.
I added an error to `SqlStorageClient` to make the situation clearer for the user.

vdusek: Okay, thank you

Issue #1560, opened 2025-11-14
Test `test_crawler_statistics_persistence` is flaky


vdusek:
- Linux Python 3.13 / 3.14.
- Code [test_crawler_statistics_persistence](https://github.com/apify/crawlee-python/blob/v1.0.4/tests/unit/crawlers/_basic/test_basic_crawler.py#L1676).
- Failing log run e.g. [here](https://github.com/apify/crawlee-python/actions/runs/19358516306/job/55385289123).

PR #1562, opened 2025-11-14
fix: Ensure persist state event emission when exiting `EventManager` context

Pijukatel:
### Description

- Fixes root cause for flakiness in `test_crawler_statistics_persistence`
- Previously, `BasicCrawler` was manually emitting a persist state event when exiting. Now this should be done by the `EventManager`. 
- Statistics used to be double-persisted previously, and due to a race condition, the old state could be persisted sometimes. Now the statistics are persisted by themselves when exiting their own context, and not by the Crawler-emitted event
- `StatisticsState.crawler_runtime` changed to a computed field in a backwards compatible way. This allows pushing the runtime calculation to the state and ensures consistency between attribute access and persistence. (This prevents a theoretical race condition when automatic persistence triggers almost at the same time as the crawler finishes, and again causes the scenario described above).

### Issues

- Closes: #1560

### Testing

- Stress testing in CI
- Added unit test

### Checklist

- [x] CI passed

### DIFF
--- a/src/crawlee/statistics/_models.py
+++ b/src/crawlee/statistics/_models.py
@@ -109,6 +109,7 @@ class StatisticsState(BaseModel):
     @crawler_runtime.setter
     def crawler_runtime(self, value: timedelta) -> None:
         # Setter for backwards compatibility only, the crawler_runtime is now computed_field, and cant be set manually.
+        # To be removed in v2 release https://github.com/apify/crawlee-python/issues/1567
         warnings.warn(
             f"Setting 'crawler_runtime' is deprecated and will be removed in a future version."
             f' Value {value} will not be used.',


src/crawlee/events/_event_manager.py
```diff
@@ -130,11 +130,13 @@ async def __aexit__(
         if not self._active:
             raise RuntimeError(f'The {self.__class__.__name__} is not active.')
 
+        # Stop persist state event periodic emission and manually emit last one to ensure latest state is saved.
+        await self._emit_persist_state_event_rec_task.stop()
+        await self._emit_persist_state_event()
```
vdusek, line 135: 👍

src/crawlee/statistics/_models.py
```diff
@@ -93,6 +93,36 @@ class StatisticsState(BaseModel):
...
+
+    @crawler_runtime.setter
+    def crawler_runtime(self, value: timedelta) -> None:
+        # Setter for backwards compatibility only, the crawler_runtime is now computed_field, and cant be set manually.
+        warnings.warn(
+            f"Setting 'crawler_runtime' is deprecated and will be removed in a future version."
+            f' Value {value} will not be used.',
+            DeprecationWarning,
+            stacklevel=2,
+        )
```
vdusek, line 118: Could you please open an issue for milestone v2 with a link?

vdusek (approved):
Nice 👍, LGTM

Mantisus (approved):
LGTM

src/crawlee/statistics/_models.py
Pijukatel: https://github.com/apify/crawlee-python/issues/1567

Issue #1532, opened 2025-11-05
failed_request_handler runs and logs but context.push_data(...) does not write to dataset (PlaywrightCrawler)


loic-bellinger:
When a request fails in PlaywrightCrawler, my `@crawler.failed_request_handler` runs (I can see `context.log` output) but `await context.push_data(...)` does not create any dataset rows. Logging works (I can see the `context.log.info(...)` lines),  only the dataset push appears to be ignored.

Minimal repro attached below.

```python
import asyncio

from crawlee.crawlers import (
    PlaywrightCrawler,
    PlaywrightCrawlingContext,
    BasicCrawlingContext,
)


async def main() -> None:
    crawler = PlaywrightCrawler(
        max_requests_per_crawl=10,
        max_request_retries=2,
        headless=True,
        browser_type='chromium',
    )

    @crawler.router.default_handler
    async def request_handler(context: PlaywrightCrawlingContext) -> None:
        context.log.info(f'Processing {context.request.url} ...')

    @crawler.failed_request_handler
    async def failed_request_handler(context: BasicCrawlingContext, error: Exception) -> None:
        context.log.info(f'failed_request_handler: processing {context.request.url} ...')
        await context.push_data(
            dataset_name="failed_request_handler_errors",
            data={
                "failed_url": context.request.url,
                "label": context.request.label,
                "error_type": type(error).__name__,
                "error_message": str(error),
                "retry_count": context.request.retry_count,
                "status": "failed",
            },
        )

    await crawler.run(['https://www.info.gouv.fr/totalnonsense'])


if __name__ == '__main__':
    asyncio.run(main())
```

I'm using `crawlee[playwright]>=1.0.4`

Relevant part of the documentation is [https://crawlee.dev/python/docs/guides/request-router#failed-request-handler](https://crawlee.dev/python/docs/guides/request-router#failed-request-handler)

janbuchar:
Hi @loic-bellinger, thanks for reporting this! The issue is that we "buffer" any storage-related calls on the context object so that partial results of failing request handler runs are not stored. This is an unintended side effect of this feature. We'll do our best to fix this soon 🙂 

As a temporary workaround, you may try this:

```py
dataset = await Dataset.open(name="failed_request_handler_errors")
await dataset.push_data({...})
```

It should bypass the "buffering" mechanism.

loic-bellinger:
[...]

PR #1570, opened 2025-11-18
fix: Make context helpers work in `FailedRequestHandler` and `ErrorHandler`

Pijukatel:
### Description

- Make context helpers `push_data`,`get_key_value_store`, and `add_requests` work in `FailedRequestHandler` and `ErrorHandler`. Previously, they had no effect as the result of a failed request was never committed. 

### Issues

- Closes: #1532

### Testing

- Added unit tests

### Checklist

- [ ] CI passed

### DIFF
--- a/src/crawlee/_types.py
+++ b/src/crawlee/_types.py
@@ -33,7 +33,6 @@ else:
     from pydantic import JsonValue as JsonSerializable
 
 T = TypeVar('T')
-TCrawlingContext = TypeVar('TCrawlingContext')
 
 HttpMethod = Literal['GET', 'HEAD', 'POST', 'PUT', 'DELETE', 'CONNECT', 'OPTIONS', 'TRACE', 'PATCH']
 


src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1256,52 +1262,61 @@ def _convert_url_to_request_iterator(self, urls: Sequence[str | Request], base_u
...
+        filter_requests_iterator = self._enqueue_links_filter_iterator(requests_iterator, context.request.url, **kwargs)
+        for dst_request in filter_requests_iterator:
+            # Update the crawl depth of the request.
+            dst_request.crawl_depth = context.request.crawl_depth + 1
+
+            if self._max_crawl_depth is None or dst_request.crawl_depth <= self._max_crawl_depth:
+                context_aware_requests.append(dst_request)
+        return context_aware_requests
+
+    async def _create_context_aware_add_requests(self, context: BasicCrawlingContext) -> AddRequestsFunction:
```
janbuchar: nit - this function does not have to be `async`.

Also, maybe we could just make an `_add_requests` method that would directly accept the context and then use `partial` to bind it to the failed request handler context when needed?

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1256,52 +1262,61 @@ def _convert_url_to_request_iterator(self, urls: Sequence[str | Request], base_u
             else:
                 yield Request.from_url(url)
 
+    async def _open_suitable_request_manager(
```
janbuchar: nit - this is only used in one place, isn't it? I'd probably inline it.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1166,7 +1167,12 @@ async def _handle_failed_request(self, context: TCrawlingContext | BasicCrawling
 
         if self._failed_request_handler:
             try:
-                await self._failed_request_handler(context, error)
+                error_context = context.create_modified_copy(
+                    push_data=self._push_data,
+                    get_key_value_store=self.get_key_value_store,
+                    add_requests=await self._create_context_aware_add_requests(context),
+                )
+                await self._failed_request_handler(error_context, error)
```
janbuchar: Shouldn't we do the same with error handler?

src/crawlee/crawlers/_basic/_basic_crawler.py
Pijukatel: Ok

src/crawlee/crawlers/_basic/_basic_crawler.py
Pijukatel: Ok, added.

src/crawlee/crawlers/_basic/_basic_crawler.py
Pijukatel: Ok

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1256,52 +1278,51 @@ def _convert_url_to_request_iterator(self, urls: Sequence[str | Request], base_u
...
-
-            for dst_request in filter_requests_iterator:
-                # Update the crawl depth of the request.
-                dst_request.crawl_depth = context.request.crawl_depth + 1
-
-                if self._max_crawl_depth is None or dst_request.crawl_depth <= self._max_crawl_depth:
-                    requests.append(dst_request)
-
-            await request_manager.add_requests(requests)
+            await functools.partial(self._add_requests, context)(**add_requests_call)
```
janbuchar:
```suggestion
            await self._add_requests(context, **add_requests_call)
```

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1256,52 +1278,51 @@ def _convert_url_to_request_iterator(self, urls: Sequence[str | Request], base_u
             else:
                 yield Request.from_url(url)
 
+    def _get_context_aware_requests(
```
janbuchar: day 4237 of me hating methods that start with `get_` 😁 the method does not really retrieve anything without modifying it, the logic is much more sophisticated. I'd either rename it to signal that the `context` is used to handle relative URLs, or just inline it in the `_add_requests` body with a comment where you have more creative freedom than in a method name

janbuchar (commented):
It's starting to grow on me 🤷 What do you think @Pijukatel?

src/crawlee/crawlers/_basic/_basic_crawler.py
Pijukatel: Ok

src/crawlee/_types.py
```diff
@@ -33,6 +33,7 @@
     from pydantic import JsonValue as JsonSerializable
 
 T = TypeVar('T')
+TCrawlingContext = TypeVar('TCrawlingContext')
```
janbuchar: Is this used here?

Pijukatel: Removed

Issue #1579, opened 2025-11-25
user_data deforms non-latin letters when passed in


DoctorEvil92:
I'm trying to pass some data into a handler function with user_data, I hope I'm doing it correctly. So, the data that is passed is originally:
`{'pagination_state_key': 'run0https://www.supermaxi.com/catalogo/.json', 'desired_store_name': 'Supermaxi El Jardín'}`

And what I get after loading it with `dict(context.request.user_data)` is this:
`{'pagination_state_key': 'run0https://www.supermaxi.com/catalogo/.json', 'desired_store_name': 'Supermaxi El JardÃ\xadn'}`

This is the log attached.

<img width="1347" height="568" alt="Image" src="https://github.com/user-attachments/assets/b1fa7160-024e-4ef0-b720-2454f4f5fcd7" />
I'm not sure if I'm doing something wrong, the encoding on the original string should be utf-8 and seems to work fine by itself.

Pijukatel:
Hello, 

Could you please create a minimal reproducible example? I am not sure how to reproduce it.

Maybe modify this:
```
from crawlee.crawlers import ParselCrawler, ParselCrawlingContext

async def main():
    crawler = ParselCrawler()

    @crawler.router.default_handler
    async def handler(context: ParselCrawlingContext):
        context.log.info("Supermaxi El Jardín")
        context.request.user_data.update(key='desired_store_name', value='Supermaxi El Jardín')
        context.log.info(context.request.user_data)

    await crawler.run(["https://crawlee.dev"])

if __name__ == "__main__":
    import asyncio
    asyncio.run(main())
```

It could be something with your setup as I can see that even the Crawlee coloured logs are not properly decoded

Pijukatel:
Also could please answer following:
- what OS are you using?
- which version of Crawlee?
- what version of Python?

DoctorEvil92:
I'm using Windows 10, Crawlee 1.1.0, Python 3.14.0.
The red letters seem to be standard when running it with the native Python Shell, the request does seem to work and get data from the site, encoding with soup works just fine with displaying characters, but with user_data it doesn't seem to work properly.

Here's the minimal example
```
from crawlee.crawlers import BeautifulSoupCrawler, BeautifulSoupCrawlingContext
from crawlee import Request
import asyncio


async def main():
    crawler = BeautifulSoupCrawler()

    @crawler.router.default_handler
    async def handler(context: BeautifulSoupCrawlingContext):
[...]

PR #1580, opened 2025-11-26
fix: Fix non-ASCII character corruption in `FileSystemStorageClient` on systems without UTF-8 default encoding

Mantisus:
### Description

- Fix non-ASCII character corruption in `FileSystemStorageClient` on systems without UTF-8 default encoding

### Issues

- Closes: #1579

### Testing

- Add new tests for storage

### DIFF
--- a/tests/unit/storages/test_dataset.py
+++ b/tests/unit/storages/test_dataset.py
@@ -1082,3 +1082,20 @@ async def test_validate_name(storage_client: StorageClient, name: str, *, is_val
     else:
         with pytest.raises(ValueError, match=rf'Invalid storage name "{name}".*'):
             await Dataset.open(name=name, storage_client=storage_client)
+
+
+async def test_record_with_noascii_chars(dataset: Dataset) -> None:
+    """Test handling record with non-ASCII characters."""
+    init_value = {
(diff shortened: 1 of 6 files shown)


tests/unit/storages/test_dataset.py
```diff
@@ -1082,3 +1082,20 @@ async def test_validate_name(storage_client: StorageClient, name: str, *, is_val
     else:
         with pytest.raises(ValueError, match=rf'Invalid storage name "{name}".*'):
             await Dataset.open(name=name, storage_client=storage_client)
+
+
+async def test_record_with_noascii_chars(dataset: Dataset) -> None:
```
janbuchar, line 1087: The fix is in file system storage only. We can either move this check to file system storage client tests or test with all possible storage clients - your call.

Pijukatel (approved):

tests/unit/storages/test_dataset.py
Mantisus: I would keep the tests for all possible storage clients. 🙂

janbuchar: Ah right, the `dataset` fixture uses the parametrized storage client fixture. Makes sense.

PR #1594, opened 2025-12-04
chore: Reduce the number of warnings in tests

Mantisus:
### Description

- Reduce the number of warnings in tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -221,6 +221,9 @@ timeout = 300
 markers = [
     "run_alone: marks tests that must run in isolation",
 ]
+# Ignore DeprecationWarnings coming from Uvicorn's internal imports. Uvicorn relies on deprecated
+# modules from `websockets`, which triggers warnings during tests. These are safe to ignore until
+# Uvicorn updates its internals.
 filterwarnings = [
     "ignore:websockets.legacy is deprecated:DeprecationWarning",
     "ignore:websockets.server.WebSocketServerProtocol is deprecated:DeprecationWarning",


pyproject.toml
```diff
@@ -221,6 +221,10 @@ timeout = 300
 markers = [
     "run_alone: marks tests that must run in isolation",
 ]
+filterwarnings = [
+    "ignore:websockets.legacy is deprecated:DeprecationWarning",
+    "ignore:websockets.server.WebSocketServerProtocol is deprecated:DeprecationWarning",
+]
```
vdusek, line 230: These are from uvicorn, right? Not sure if we wanna just surpass them completely on this level.

tests/unit/_utils/test_system.py
```diff
@@ -26,6 +26,9 @@ def test_get_cpu_info_returns_valid_values() -> None:
     assert 0 <= cpu_info.used_ratio <= 1
 
 
+# Suppress the warning because there were no problems with using `fork' in our test environment.
+# However, there are serialization issues in other cases.
+@pytest.mark.filterwarnings('ignore:This process .* is multi-threaded, use of fork:DeprecationWarning')
```
vdusek: The warning log message:

```
DeprecationWarning: This process (pid=15537) is multi-threaded, use of fork() may lead to deadlocks in the child.
    self.pid = os.fork()
```

Not sure if we wanna just silence it.

pyproject.toml
Mantisus: These are warnings related to internal imports in Uvicorn. They will disappear when Uvicorn finally abandons deprecated imports.

Our environment uses `ws=‘websockets-sansio’`, an implementation that should replace the deprecated one - https://github.com/Kludex/uvicorn/issues/1908.

vdusek, line 230: Could we please add a comment like this explaining why it was added?

```suggestion
# Ignore DeprecationWarnings coming from Uvicorn's internal imports. Uvicorn relies on deprecated
# modules from `websockets`, which triggers warnings during tests. These are safe to ignore until
# Uvicorn updates its internals.
filterwarnings = [
    "ignore:websockets.legacy is deprecated:DeprecationWarning",
    "ignore:websockets.server.WebSocketServerProtocol is deprecated:DeprecationWarning",
]
```

vdusek: Ok, let's add at least a comment there -> separate thread with suggestion.

vdusek (approved):

PR #1595, opened 2025-12-04
fix: Handle the case when `error_handler` returns `Request`

Mantisus:
### Description

- This PR fixes the behavior of `crawler` when `error_handler` returns `Request`. The old behavior resulted in the queue never reaching the empty state.

### Testing

- Add new test

### DIFF
--- a/src/crawlee/crawlers/_basic/_basic_crawler.py
+++ b/src/crawlee/crawlers/_basic/_basic_crawler.py
@@ -1137,26 +1137,12 @@ class BasicCrawler(Generic[TCrawlingContext, TStatisticsState]):
                 else:
                     if new_request is not None and new_request != request:
                         await request_manager.add_request(new_request)
-                        await wait_for(
-                            lambda: request_manager.mark_request_as_handled(request),
-                            timeout=self._internal_timeout,
-                            timeout_message='Marking request as handled timed out after '
-                            f'{self._internal_timeout.total_seconds()} seconds',
-                            logger=self._logger,
-                            max_retries=3,
-                        )
+                        await self._mark_request_as_handled(request)
                         return
 
             await request_manager.reclaim_request(request)
         else:
-            await wait_for(
-                lambda: request_manager.mark_request_as_handled(context.request),
-                timeout=self._internal_timeout,
-                timeout_message='Marking request as handled timed out after '
-                f'{self._internal_timeout.total_seconds()} seconds',
-                logger=self._logger,
-                max_retries=3,
-            )
+            await self._mark_request_as_handled(request)
             await self._handle_failed_request(context, error)
             self._statistics.record_request_processing_failure(request.unique_key)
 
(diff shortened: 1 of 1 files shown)


Mantisus:
I made a fix in accordance with the current signature for `ErrorHandler`.

But the signature of `ErrorHandler` does not match the TS version.  In TS, it only returns 'None'.

tests/unit/crawlers/_basic/test_basic_crawler.py
```diff
@@ -1768,3 +1768,37 @@ async def handler(_: BasicCrawlingContext) -> None:
...
+    await crawler.run([request])
+
+    check_original_request = await queue.get_request(request.unique_key)
+    check_error_request = await queue.get_request(f'{request.unique_key}|test')
+
+    assert check_original_request is not None
+    assert check_original_request.state == RequestState.ERROR_HANDLER
+
+    assert check_error_request is not None
+    assert check_error_request.state == RequestState.REQUEST_HANDLER
```
Pijukatel:
```suggestion
    original_request = await queue.get_request(request.unique_key)
    error_request = await queue.get_request(f'{request.unique_key}|test')

    assert original_request is not None
    assert original_request.state == RequestState.ERROR_HANDLER
    assert original_request.was_already_handled

    assert error_request is not None
    assert error_request.state == RequestState.REQUEST_HANDLER
    assert error_request.was_already_handled
```

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1135,8 +1135,17 @@ async def _handle_request_retries(
...
+                    if new_request is not None and new_request != request:
+                        await request_manager.add_request(new_request)
+                        await wait_for(
+                            lambda: request_manager.mark_request_as_handled(request),
+                            timeout=self._internal_timeout,
+                            timeout_message='Marking request as handled timed out after '
+                            f'{self._internal_timeout.total_seconds()} seconds',
+                            logger=self._logger,
+                            max_retries=3,
+                        )
```
Pijukatel: Now this is repeated 6x times in BasicCrawler in some variants. I would consider creating a utility function for it. 

```
async def some_very_good_name(self, request: Request)->None:
            request_manager = await self.get_request_manager()
            await wait_for(
                lambda: request_manager.mark_request_as_handled(request),
                timeout=self._internal_timeout,
                timeout_message='Marking request as handled timed out after '
                f'{self._internal_timeout.total_seconds()} seconds',
                logger=self._logger,
                max_retries=3,
            )
```

(no need to worry about `request_manager = await self.get_request_manager()`, it will resolve to `self._request_manager`)

vdusek: IMO we can use just a private method `async def _mark_request_as_handled(self, request: Request) -> None:`.

vdusek (approved):
LGTM

Pijukatel (approved):

PR #1601, opened 2025-12-05
fix: Align `Request.state` transitions with `Request` lifecycle

Mantisus:
### Description

- Set the appropriate statuses for `Request.state` at the appropriate stages. 
- Set the correct final status for `Request`.

### Testing

- Add new tests


vdusek (approved):
Thanks

PR #1603, opened 2025-12-06
fix: Freeze core `Request` fields

Mantisus:
### Description

- Ensures that core `Request` fields such as `unique_key`, `method`, and others will not be changed.

### DIFF
--- a/src/crawlee/_request.py
+++ b/src/crawlee/_request.py
@@ -34,7 +34,7 @@ class RequestState(IntEnum):
 class CrawleeRequestData(BaseModel):
     """Crawlee-specific configuration stored in the `user_data`."""
 
-    max_retries: Annotated[int | None, Field(alias='maxRetries')] = None
+    max_retries: Annotated[int | None, Field(alias='maxRetries', frozen=True)] = None
     """Maximum number of retries for this request. Allows to override the global `max_request_retries` option of
     `BasicCrawler`."""
 
@@ -137,6 +137,8 @@ class RequestOptions(TypedDict):
     always_enqueue: NotRequired[bool]
     user_data: NotRequired[dict[str, JsonSerializable]]
     no_retry: NotRequired[bool]
+    enqueue_strategy: NotRequired[EnqueueStrategy]
+    max_retries: NotRequired[int | None]
 
 
 @docs_group('Storage data')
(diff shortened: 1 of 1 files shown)


janbuchar:
> I agree in principle, but I'm not entirely sure we can do it. We should definitely test this change on the platform - run integration and E2E tests in the SDK.

I completely agree with this. I don't see a good reason why it shouldn't work, but better safe than sorry.

Once this is ready, no need to wait for my review unless you decide to change the PR substantially.

Pijukatel:
I can not start the E2E tests directly on the branch from forked repo. I created a copy of the branch and started the tests here: https://github.com/apify/crawlee-python/actions/runs/20028369926

vdusek:
tests in the SDK - https://github.com/apify/apify-sdk-python/actions/runs/20029622226/job/57435192024?pr=702

Edit: they're failing

vdusek:
Also do we want to freeze only these 4 fields?

List of all from docs:

image

What about `label`, `max_retries`, `no_retry`, `enqueue_strategy`, `forefront`?

Mantisus:
> What about label, max_retries, no_retry, enqueue_strategy, forefront?

`no_retry` - we directly change it in `__run_task_function` in some situations.
`label` - I can think of some use cases where you might want to change it. For example, in `error_handler`.
`enqueue_strategy` - has a setter, so I think it was also considered a mutable attribute.
`forefront` - same thing.
`max_retries` - same thing.

I'm not sure that protecting these attributes is as critical as, for example, `unique_key`. But now is a good time to decide.

> Edit: they're failing

I'll check why it's falling.

Edit: Now the tests are passed without falling

janbuchar:
> `no_retry` - we directly change it in `__run_task_function` in some situations. 

makes sense to change it in user handler as well - e.g., you detect a weird error and want to tell crawlee not to retry the request anymore

> `label` - I can think of some use cases where you might want to change it. For example, in `error_handler`. 

yeah, I guess that it could make sense to change it to "redirect" the request to a different branch of the router :shrug: 

> `enqueue_strategy` - has a setter, so I think it was also considered a mutable attribute. 

This actually doesn't make much sense - it's there for checking if the URL still matches the `enqueue_strategy` after a redirect

> `forefront` - same thing. 

You may want to change this when re-enqueueing a request.

> `max_retries` - same thing.

This is debatable, maybe it's enough to allow setting it when creating the request?

Mantisus:
> it's there for checking if the URL still matches the enqueue_strategy after a redirect

This setter is required for fix #1606.

Since `EnqueueLinksFunction` can accept a `requests` parameter with a list of `Request` objects, this setter is required to set the appropriate strategy for these requests so that the check works correctly after redirection.

vdusek (approved):
I tested it again (https://github.com/apify/apify-sdk-python/pull/702) with the current state, and it passes.

LGTM and thanks for thinking through all the options.

PR #1609, opened 2025-12-11
docs: add segment analytics [internal]

katzino:
requested on [slack](https://apify.slack.com/archives/C04BTPAF0V8/p1765380246880559) by @mr-rajce as it was missing for python subpages

### DIFF
--- a/.github/workflows/build_and_deploy_docs.yaml
+++ b/.github/workflows/build_and_deploy_docs.yaml
@@ -54,6 +54,7 @@ jobs:
         run: make build-docs
         env:
           APIFY_SIGNING_TOKEN: ${{ secrets.APIFY_SIGNING_TOKEN }}
+          SEGMENT_TOKEN: ${{ secrets.SEGMENT_TOKEN }}
 
       - name: Set up GitHub Pages
         uses: actions/configure-pages@v5
--- a/website/src/plugins/docusaurus-plugin-segment/index.js
+++ b/website/src/plugins/docusaurus-plugin-segment/index.js
@@ -1,0 +1,37 @@
+const path = require('path');
+
+module.exports = function (context, options) {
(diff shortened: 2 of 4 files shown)


.github/workflows/build_and_deploy_docs.yaml
```diff
@@ -54,6 +54,7 @@ jobs:
         run: make build-docs
         env:
           APIFY_SIGNING_TOKEN: ${{ secrets.APIFY_SIGNING_TOKEN }}
+          SEGMENT_TOKEN: ${{ secrets.SEGMENT_TOKEN }}
```
katzino, line 57: Not sure whether it will be populated from doppler out-of-the-box 🤔

website/src/plugins/docusaurus-plugin-segment/index.js
katzino, line 1: This is just copy pasted logic from other repositories

website/src/plugins/docusaurus-plugin-segment/segment.js
```diff
@@ -0,0 +1,18 @@
+import ExecutionEnvironment from '@docusaurus/ExecutionEnvironment';
+
+export default ExecutionEnvironment.canUseDOM ? {
+    onRouteUpdate() {
+        // this forces deferred execution that ensures `window.location` is in sync
+        setTimeout(() => {
+            // Don't track page views on development
+            if (process.env.NODE_ENV === 'production' && window.analytics) {
```
katzino, line 8: There is [an issue](https://github.com/apify/apify-web/issues/5405) in the backlog not to use `process.env` here - looking at the logic, I actually think we don't need that check at all, since segment shouldn't get initialised based on this condition in a file above
```
if (process.env.NODE_ENV !== 'production' && !allowedInDev) {
```

vdusek (approved):

PR #1613, opened 2025-12-13
chore(ci): Align GitHub workflows across Python projects

vdusek:


.github/workflows/manual_release_stable.yaml
```diff
@@ -26,8 +26,21 @@ concurrency:
...
+    name: Lint check
+    uses: apify/workflows/.github/workflows/python_lint_check.yaml@main
+    with:
+      python-versions: '["3.10", "3.11", "3.12", "3.13", "3.14"]'
+
+  type_check:
+    name: Type check
+    uses: apify/workflows/.github/workflows/python_type_check.yaml@main
+    with:
+      python-versions: '["3.10", "3.11", "3.12", "3.13", "3.14"]'
```
janbuchar: Shouldn't this delegate to `.github/workflows/code_checks.yaml` instead?

.github/workflows/manual_release_stable.yaml
janbuchar, line 1: This workflow doesn't check if the target ref is green, and that kinda worries me. I don't trust maintainers to be responsible and check that before hitting "Release" 😁

.github/workflows/manual_release_stable.yaml
vdusek: great catch, thank you

.github/workflows/manual_release_stable.yaml
vdusek: Yeah, let's utilize https://github.com/apify/workflows/pull/238 for this, but please in a separate PR

PR #1625, opened 2025-12-17
chore: Migrate to `ty` type checker

vdusek:


Mantisus (approved):
Great job!

pyproject.toml
```diff
@@ -230,57 +230,24 @@ filterwarnings = [
...
-disallow_untyped_defs = true
-no_implicit_optional = true
-warn_redundant_casts = true
-warn_return_any = true
-warn_unreachable = true
-warn_unused_ignores = true
-
-[[tool.mypy.overrides]]
-# Example codes are sometimes showing integration of crawlee with external tool, which is not dependency of crawlee.
-module = [
```
Pijukatel, line 254: I think, that replacing per module explicit override by blanket file override is downgrade.

previously we would catch typos for example like this:
`from crawle.crawlers import BasicCrawler`

now it will just ignore it as another missing import. Is **ty** configurable like **mypy** to handle it the same way as before?

I looked in docs, but could not find it. Question raised: https://github.com/astral-sh/ty/issues/2337

src/crawlee/_utils/recurring_task.py
```diff
@@ -25,7 +25,7 @@ class RecurringTask:
     """
 
     def __init__(self, func: Callable, delay: timedelta) -> None:
-        logger.debug(f'Calling RecurringTask.__init__(func={func.__name__}, delay={delay})...')
+        logger.debug(f'Calling RecurringTask.__init__(func={func.__name__}, delay={delay})...')  # ty: ignore[unresolved-attribute]
```
Pijukatel: This is actually valid point, but rare edgecase. Maybe we should do something like this and remove the ignore:
`func.__name__ if hasattr(func,"__name__") else func.__class__.__name__`

src/crawlee/_utils/recurring_task.py
```diff
@@ -55,7 +55,7 @@ async def _wrapper(self) -> None:
 
     def start(self) -> None:
         """Start the recurring task execution."""
-        self.task = asyncio.create_task(self._wrapper(), name=f'Task-recurring-{self.func.__name__}')
+        self.task = asyncio.create_task(self._wrapper(), name=f'Task-recurring-{self.func.__name__}')  # ty: ignore[possibly-missing-attribute]
```
Pijukatel: Same as above

src/crawlee/_utils/sitemap.py
```diff
@@ -430,10 +430,10 @@ async def parse_sitemap(
...
     """
     # Set default options
-    options = options or {}
-    emit_nested_sitemaps = options.get('emit_nested_sitemaps', False)
-    max_depth = options.get('max_depth', float('inf'))
-    sitemap_retries = options.get('sitemap_retries', 3)
+    options = options or {}  # ty: ignore[invalid-assignment]
+    emit_nested_sitemaps = options.get('emit_nested_sitemaps', False)  # ty: ignore[possibly-missing-attribute]
+    max_depth = options.get('max_depth', float('inf'))  # ty: ignore[possibly-missing-attribute]
+    sitemap_retries = options.get('sitemap_retries', 3)  # ty: ignore[possibly-missing-attribute]
```
Pijukatel: Can be rewritten to avoid ignores.

For example
```
    default_timeout = timedelta(seconds=30)
    if not options:
        emit_nested_sitemaps = False
        max_depth = float('inf')
        sitemap_retries = 3
        timeout = default_timeout
    else:
        emit_nested_sitemaps = options['emit_nested_sitemaps']
        max_depth = options['max_depth']
        sitemap_retries = options['sitemap_retries']
        timeout = options.get('timeout', default_timeout)
```

src/crawlee/_utils/sitemap.py
```diff
@@ -472,7 +472,7 @@ async def parse_sitemap(
                 sitemap_retries,
                 emit_nested_sitemaps=emit_nested_sitemaps,
                 proxy_info=proxy_info,
-                timeout=options.get('timeout', timedelta(seconds=30)),
+                timeout=options.get('timeout', timedelta(seconds=30)),  # ty: ignore[possibly-missing-attribute]
```
Pijukatel: Solved above

src/crawlee/_utils/system.py
```diff
@@ -41,11 +41,19 @@ class CpuInfo(BaseModel):
...
-    """The time at which the measurement was taken."""
+    # Workaround for Pydantic and type checkers when using Annotated with default_factory
+    if TYPE_CHECKING:
+        created_at: datetime = datetime.now(timezone.utc)
+        """The time at which the measurement was taken."""
+    else:
+        created_at: Annotated[
+            datetime,
+            Field(
+                alias='createdAt',
```
Pijukatel, line 52: This is not very nice, but I failed to find any elegant solution. The only workaround that seems to work exactly the same is

```
    created_at: datetime = Field(
        validation_alias='createdAt',
        serialization_alias='createdAt',
        default_factory=lambda: datetime.now(timezone.utc),
    )
    """The time at which the measurement was taken."""
```

But you have to repeat alias twice. If we never serialize this into alias, we can omit `serialization_alias`...

(Same for other similar workarounds usages)

src/crawlee/browsers/_browser_pool.py
```diff
@@ -142,7 +142,7 @@ def with_default_plugin(
         plugin_options['browser_new_context_options'] = browser_new_context_options or {}
 
         if headless is not None:
-            plugin_options['browser_launch_options']['headless'] = headless
+            plugin_options['browser_launch_options']['headless'] = headless  # ty: ignore[invalid-assignment]
```
Pijukatel: This ignore can be avoided if we do this on line 141
`plugin_options['browser_launch_options'] = dict(browser_launch_options) if browser_launch_options else {}`

src/crawlee/browsers/_playwright_browser.py
```diff
@@ -78,7 +78,7 @@ async def new_context(self, **context_options: Any) -> BrowserContext:
 
     async def _delete_temp_dir(self, _: BrowserContext | None) -> None:
         if self._temp_dir and self._temp_dir.exists():
-            await asyncio.to_thread(shutil.rmtree, self._temp_dir, ignore_errors=True)
+            await asyncio.to_thread(lambda: shutil.rmtree(self._temp_dir, ignore_errors=True))  # ty: ignore[invalid-argument-type]
```
Pijukatel: I would not be so sure about ignoring this. This is pointing to very hard to catch potential bug.

**if self._temp_dir** only ensures that the `self._temp_dir` is not `None` when calling `asyncio.to_thread`, but it does not guarantee that it will not be mutated to `None` when `shutil.rmtree` is finally executed. This will most likely not happen, but type checker is right.

We can fix it by saving it to immutable local variable.

```
  temp_dir = self._temp_dir
  await asyncio.to_thread(lambda: shutil.rmtree(temp_dir, ignore_errors=True))
```

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
```diff
@@ -102,7 +102,7 @@ def create_parsed_http_crawler_class(
 
         class _ParsedHttpCrawler(
             AbstractHttpCrawler[ParsedHttpCrawlingContext[TParseResult], TParseResult, TSelectResult]
-        ):
+        ):  # ty: ignore[invalid-generic-class]
```
Pijukatel: After all the casting in context pipeline, it still throws errors when changing type checker :-(

I am not sure there is an easy fix for this

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -160,7 +160,7 @@ def __init__(
         super().__init__(statistics=statistics, **kwargs)
 
         # Sub crawlers related.
-        playwright_crawler_specific_kwargs = playwright_crawler_specific_kwargs or {}
+        playwright_crawler_specific_kwargs = playwright_crawler_specific_kwargs or {}  # ty: ignore[invalid-assignment]
```
Pijukatel: Can be avoided by

`playwright_crawler_specific_kwargs = playwright_crawler_specific_kwargs or _PlaywrightCrawlerAdditionalOptions()`

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -181,7 +181,7 @@ def __init__(
         )
         playwright_crawler = PlaywrightCrawler(
             statistics=_NonPersistentStatistics(),
-            **playwright_crawler_specific_kwargs,
+            **playwright_crawler_specific_kwargs,  # ty: ignore[invalid-argument-type]
```
Pijukatel: Not needed when using fix from above

src/crawlee/crawlers/_playwright/_playwright_crawler.py
```diff
@@ -183,7 +183,7 @@ def __init__(
                     generator_browser_type = [fingerprint_browser_type_from_playwright_browser_type(browser_type)]
 
                 fingerprint_generator = DefaultFingerprintGenerator(
-                    header_options=HeaderGeneratorOptions(browsers=generator_browser_type)
+                    header_options=HeaderGeneratorOptions(browsers=generator_browser_type)  # ty: ignore[invalid-argument-type]
```
Pijukatel: Can be avoided by adding 
this below line 179
`generator_browser_type: list[Literal["chrome", "firefox", "safari", "edge"]] | None`

src/crawlee/crawlers/_playwright/_playwright_crawler.py
```diff
@@ -516,7 +516,7 @@ async def _get_cookies(self, page: Page) -> list[PlaywrightCookieParam]:
 
     async def _update_cookies(self, page: Page, cookies: list[PlaywrightCookieParam]) -> None:
         """Update the cookies in the page context."""
-        await page.context.add_cookies([{**cookie} for cookie in cookies])
+        await page.context.add_cookies([{**cookie} for cookie in cookies])  # ty: ignore[invalid-argument-type]
```
Pijukatel, line 516: Maybe add link to `ty` issue, I think our code is correct.
Probably this https://github.com/astral-sh/ty/issues/1493,

but there whole label for `TypedDict` issues in `ty` https://github.com/astral-sh/ty/issues?q=is%3Aissue%20state%3Aopen%20TypedDict%20label%3Atypeddict

src/crawlee/events/_event_manager.py
```diff
@@ -178,7 +178,7 @@ async def listener_wrapper(event_data: EventData) -> None:
                 else asyncio.to_thread(cast('Callable[..., None]', listener), *bound_args.args, **bound_args.kwargs)
             )
 
-            listener_task = asyncio.create_task(coro, name=f'Task-{event.value}-{listener.__name__}')
+            listener_task = asyncio.create_task(coro, name=f'Task-{event.value}-{listener.__name__}')  # ty: ignore[invalid-argument-type, unresolved-attribute]
```
Pijukatel: Again

`listener.__name__ if hasattr(listener,"__name__") else listener.__class__.__name__`

src/crawlee/events/_event_manager.py
```diff
@@ -189,7 +189,7 @@ async def listener_wrapper(event_data: EventData) -> None:
                 # We need to swallow the exception and just log it here, otherwise it could break the event emitter
                 logger.exception(
                     'Exception in the event listener',
-                    extra={'event_name': event.value, 'listener_name': listener.__name__},
+                    extra={'event_name': event.value, 'listener_name': listener.__name__},  # ty: ignore[unresolved-attribute]
```
Pijukatel: Again

`listener.__name__ if hasattr(listener,"__name__") else listener.__class__.__name__`

src/crawlee/http_clients/_curl_impersonate.py
```diff
@@ -93,12 +93,12 @@ async def read(self) -> bytes:
         return self._response.content
 
     async def read_stream(self) -> AsyncGenerator[bytes, None]:
-        if not self._response.astream_task or self._response.astream_task.done():  # type: ignore[attr-defined]
+        if not self._response.astream_task or self._response.astream_task.done():  # ty: ignore[possibly-missing-attribute]
```
Pijukatel: This can be solved like

```
        if not self._response.astream_task:
            raise RuntimeError(
                'Cannot read stream, Response not obtained from `stream` method.'
            )

        if isinstance(self._response.astream_task, Future) and self._response.astream_task.done():
            raise RuntimeError(
                'Cannot read stream, it was already consumed.'
            )
```

src/crawlee/http_clients/_curl_impersonate.py
```diff
@@ -156,7 +156,7 @@ async def crawl(
         try:
             response = await client.request(
                 url=request.url,
-                method=request.method.upper(),  # type: ignore[arg-type] # curl-cffi requires uppercase method
+                method=request.method.upper(),  # ty: ignore[invalid-argument-type]
```
Pijukatel: This is actually a valid error.

Our set of methods is different from curl-cffi
Extra methods in our set:
`Literal['CONNECT', 'PATCH', ...]`
Extra methods in their set
`HttpMethod = Literal["TRACE", "QUERY", ...]`

src/crawlee/http_clients/_curl_impersonate.py
```diff
@@ -203,7 +203,7 @@ async def send_request(
         try:
             response = await client.request(
                 url=url,
-                method=method.upper(),  # type: ignore[arg-type] # curl-cffi requires uppercase method
+                method=method.upper(),  # ty: ignore[invalid-argument-type]
```
Pijukatel: same as above

src/crawlee/http_clients/_curl_impersonate.py
```diff
@@ -244,7 +244,7 @@ async def stream(
...
```
Pijukatel: same as above

src/crawlee/browsers/_playwright_browser.py
janbuchar: What happened here? If you're ignoring the type error anyway, why add the lambda?

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
```diff
@@ -122,9 +122,9 @@ def _create_static_content_crawler_pipeline(self) -> ContextPipeline[ParsedHttpC
             ContextPipeline()
             .compose(self._execute_pre_navigation_hooks)
             .compose(self._make_http_request)
-            .compose(self._handle_status_code_response)
+            .compose(self._handle_status_code_response)  # ty: ignore[invalid-argument-type]
```
janbuchar, line 123: It's strange to see type ignore directives on two out of five similar lines - aren't we covering some actual issue that `mypy` failed to notice?

src/crawlee/storage_clients/_base/_dataset_client.py
```diff
@@ -87,8 +87,8 @@ async def iterate_items(
 
         The backend method for the `Dataset.iterate_items` call.
         """
-        # This syntax is to make mypy properly work with abstract AsyncIterator.
+        # This syntax is to make type checker properly work with abstract AsyncIterator.
         # https://mypy.readthedocs.io/en/stable/more_types.html#asynchronous-iterators
         raise NotImplementedError
-        if False:  # type: ignore[unreachable]
+        if False:
```
janbuchar, line 93: Can't the `if False` block be removed instead?

src/crawlee/storage_clients/_file_system/_key_value_store_client.py
```diff
@@ -212,7 +212,7 @@ async def purge(self) -> None:
             for file_path in self.path_to_kvs.glob('*'):
                 if file_path.name == METADATA_FILENAME:
                     continue
-                await asyncio.to_thread(file_path.unlink, missing_ok=True)
+                await asyncio.to_thread(lambda f=file_path: f.unlink(missing_ok=True))
```
janbuchar: It'd have been more readab le to just keep the original `asyncio.to_thread` call - does `ty` have problems understanding that?

.gitignore
```diff
@@ -1,6 +1,6 @@
 # Cache
 __pycache__
-.mypy_cache
+.uv_cache
```
janbuchar: Maybe this was supposed to be `.ty_cache`? Also, there is `.uv-cache` a couple of lines below.

janbuchar (approved):
Couple of nits but probably OK. Will `ty` let us know when a type-ignore comment becomes obsolete?

src/crawlee/http_clients/_curl_impersonate.py
```diff
@@ -309,8 +309,8 @@ def _is_proxy_error(error: CurlRequestError) -> bool:
     @staticmethod
     def _get_cookies(curl: Curl) -> list[Cookie]:
         cookies: list[Cookie] = []
-        for curl_cookie in curl.getinfo(CurlInfo.COOKIELIST):  # type: ignore[union-attr]
-            curl_morsel = CurlMorsel.from_curl_format(curl_cookie)  # type: ignore[arg-type]
+        for curl_cookie in curl.getinfo(CurlInfo.COOKIELIST):  # ty: ignore[not-iterable]
+            curl_morsel = CurlMorsel.from_curl_format(curl_cookie)  # ty: ignore[invalid-argument-type]
```
Pijukatel: Probably better to be explicit about it

```
        cookie_list = cast(list[bytes], curl.getinfo(CurlInfo.COOKIELIST))  # Implementation of getinfo always returns list[bytes] for CurlInfo.COOKIELIST
        for curl_cookie in cookie_list:
            curl_morsel = CurlMorsel.from_curl_format(curl_cookie)
```

src/crawlee/statistics/_models.py
```diff
@@ -77,9 +77,20 @@ class StatisticsState(BaseModel):
...
+        retry_errors: dict[str, Any] = {}
+        requests_with_status_code: dict[str, int] = {}
+    else:
+        errors: Annotated[dict[str, Any], Field(default_factory=dict)]
+        retry_errors: Annotated[dict[str, Any], Field(alias='retryErrors', default_factory=dict)]
+        requests_with_status_code: Annotated[
+            dict[str, int],
+            Field(alias='requestsWithStatusCode', default_factory=dict),
+        ]
+
```
Pijukatel, line 93: Probably better just:

```
    errors: dict[str, Any] = Field(default_factory=dict)
    retry_errors: dict[str, Any] = Field(alias='retryErrors', default_factory=dict)
    requests_with_status_code: dict[str, int] = Field(alias='requestsWithStatusCode', default_factory=dict)
```

src/crawlee/storage_clients/_redis/_dataset_client.py
```diff
@@ -179,13 +179,13 @@ async def get_data(
             case (True, int(), None):
                 json_path += f'[:-{offset}]'
             case (True, int(), int()):
-                json_path += f'[-{offset + limit}:-{offset}]'
+                json_path += f'[-{offset + limit}:-{offset}]'  # ty: ignore[unsupported-operator]
```
Pijukatel, line 183: `ty` is not so good with pattern matching yet. Could you please add this issue to comment
https://github.com/astral-sh/ty/issues/887

src/crawlee/storage_clients/_redis/_dataset_client.py
```diff
@@ -179,13 +179,13 @@ async def get_data(
...
             case (False, 0, int()):
                 json_path += f'[:{limit}]'
             case (False, int(), None):
                 json_path += f'[{offset}:]'
             case (False, int(), int()):
-                json_path += f'[{offset}:{offset + limit}]'
+                json_path += f'[{offset}:{offset + limit}]'  # ty: ignore[unsupported-operator]
```
Pijukatel, line 190: `ty` is not so good with pattern matching yet. Could you please add this issue to comment
https://github.com/astral-sh/ty/issues/887

src/crawlee/storage_clients/_redis/_dataset_client.py
```diff
@@ -210,7 +210,7 @@ async def get_data(
             limit=limit or (total - offset),
             total=total,
             desc=desc,
-            items=data,
+            items=data,  # ty: ignore[invalid-argument-type]
```
Pijukatel: Such type assumption looks dangerous unless explained (type hints show many different types compared to what is expected)

src/crawlee/storage_clients/models.py
```diff
@@ -127,8 +127,13 @@ class DatasetItemsListPage(BaseModel):
     desc: Annotated[bool, Field(default=False)]
     """Indicates if the returned list is in descending order."""
 
-    items: Annotated[list[dict], Field(default_factory=list)]
-    """The list of dataset items returned on this page."""
+    # Workaround for Pydantic and type checkers when using Annotated with default_factory
```
Pijukatel, line 130: See preferable workaround from above

tests/unit/_utils/test_html_to_text.py
```diff
@@ -191,7 +191,7 @@ def test_html_to_text(source: str, expected_text: str, html_to_text: Callable[[s
 @pytest.mark.parametrize('html_to_text', [html_to_text_parsel, html_to_text_beautifulsoup])
 def test_html_to_text_raises_on_wrong_input_type(html_to_text: Callable[[str], str]) -> None:
     with pytest.raises(TypeError):
-        html_to_text(1)  # type: ignore[arg-type]  # Intentional wrong type test.
+        html_to_text(1)  # ty: ignore[invalid-argument-type]
```
Pijukatel, line 195: I would prefer to keep the explanation of the ignore 
`# Intentional wrong type test.`

tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
```diff
@@ -115,7 +115,7 @@ class TestInput:
             TestInput(
                 expected_pw_count=0,
                 expected_static_count=2,
-                rendering_types=cycle(['static']),
+                rendering_types=cycle(['static']),  # ty: ignore[invalid-argument-type]
```
Pijukatel, line 120: Code is correct. Seems like another lack of support in **ty**. I raised a question:
https://github.com/astral-sh/ty/issues/2348

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
Pijukatel: It happens only on compose call that changes the crawling context type.

.gitignore
vdusek: Thanks

pyproject.toml
vdusek: Resolving as the question was answered.

vdusek:
> Will ty let us know when a type-ignore comment becomes obsolete?

Not now, see https://github.com/astral-sh/ty/issues/278 and https://github.com/astral-sh/ty/issues/1501 for more context.

src/crawlee/http_clients/_curl_impersonate.py
vdusek: alright, let's write a convertor

src/crawlee/storage_clients/_redis/_dataset_client.py
vdusek: sure

src/crawlee/storage_clients/_redis/_dataset_client.py
vdusek: The codebase was not changed. What do you suggest? If you want this to be further investigated, let's open a dedicated issue.

src/crawlee/storage_clients/_base/_dataset_client.py
vdusek: Unfortunately, no

```
error[not-iterable]: Object of type `Unknown | CoroutineType[Any, Any, AsyncIterator[KeyValueStoreRecordMetadata]]` may not be async-iterable
   --> src/crawlee/storages/_key_value_store.py:234:30
    |
232 |           return [
233 |               key
234 |               async for key in self._client.iterate_keys(
    |  ______________________________^
235 | |                 exclusive_start_key=exclusive_start_key,
236 | |                 limit=limit,
237 | |             )
    | |_____________^
238 |           ]
    |
info: Its `__aiter__` attribute (with type `Unknown`) may not be callable
info: rule `not-iterable` is enabled by default
```

src/crawlee/storage_clients/_redis/_dataset_client.py
Pijukatel: @Mantisus , could you please check this one? I do not mind the type ignore if there is some reasonable explanation next to it.

src/crawlee/browsers/_playwright_browser.py
vdusek: The type error is resolved, and the lambda thing I will answer in the other thread

src/crawlee/storage_clients/_file_system/_key_value_store_client.py
vdusek: I really think there was a reason for these changes, but I don't exactly remember it, and with the current `ty` version (0.0.9), it no longer seems necessary. I'll revert these changes and keep the original `asyncio.to_thread` calls.

src/crawlee/storage_clients/_redis/_dataset_client.py
Mantisus: `ty` correctly show an error, since `json().get` can return the field value directly for specific query.

To avoid ignoring this, we can use

```python
if data is None:
    data = []

data = [item for item in data if isinstance(item, dict)]
```

janbuchar:
> > Will ty let us know when a type-ignore comment becomes obsolete?
> 
> Not now, see [astral-sh/ty#278](https://github.com/astral-sh/ty/issues/278) and [astral-sh/ty#1501](https://github.com/astral-sh/ty/issues/1501) for more context.

Can we make an issue to revisit the ignores in ~2 months or something then?

vdusek:
> > > Will ty let us know when a type-ignore comment becomes obsolete?
> > 
> > Not now, see [astral-sh/ty#278](https://github.com/astral-sh/ty/issues/278) and [astral-sh/ty#1501](https://github.com/astral-sh/ty/issues/1501) for more context.
> 
> Can we make an issue to revisit the ignores in ~2 months or something then?

Sure, -> https://github.com/apify/crawlee-python/issues/1664.

src/crawlee/_utils/globs.py
```diff
@@ -36,7 +36,7 @@ def _translate(
     if not seps:
         seps = (os.path.sep, os.path.altsep) if os.path.altsep else os.path.sep
 
-    escaped_seps = ''.join(map(re.escape, seps))
+    escaped_seps = ''.join(map(re.escape, seps))  # ty: ignore[invalid-argument-type]
```
Mantisus: Can be resolved

```python
_seps = ((os.path.sep, os.path.altsep) if os.path.altsep else (os.path.sep,)) if seps is None else seps

escaped_seps = ''.join(map(re.escape, _seps))
```

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
Mantisus:
```python
class _ParsedHttpCrawler(AbstractHttpCrawler):
```

Work here

src/crawlee/sessions/_cookies.py
```diff
@@ -68,7 +68,7 @@ def __init__(self, cookies: SessionCookies | CookieJar | dict[str, str] | list[C
 
         if isinstance(cookies, dict):
             for key, value in cookies.items():
-                self.set(key, value)
+                self.set(key, value)  # ty: ignore[invalid-argument-type]
```
Mantisus: Can be resolved

```python
self._jar = CookieJar()

if isinstance(cookies, list):
    for item in cookies:
        self.set(**item)

elif isinstance(cookies, SessionCookies):
    for cookie in cookies.jar:
        self._jar.set_cookie(cookie)

elif isinstance(cookies, dict):
    cookies_dict: dict[str, str] = cookies
    for key, value in cookies_dict.items():
        self.set(key, value)
```

src/crawlee/storage_clients/_redis/_dataset_client.py
vdusek: Implemented, resolving

src/crawlee/_utils/globs.py
vdusek: It does not work for me:

> Argument to function `__new__` is incorrect: Expected `Iterable[Never]`, found `(Sequence[str] & ~AlwaysFalsy) | LiteralString` ty [invalid-argument-type](https://ty.dev/rules#invalid-argument-type)

Issue #1621, opened 2025-12-16
Dataset get_data() returns fewer items than successfully pushed with context.push_data() in PlaywrightCrawler


mabreuortega:
# GitHub Issue for Crawlee Team

## Title
Dataset `get_data()` returns fewer items than successfully pushed with `context.push_data()` in PlaywrightCrawler

## Labels
- `t-tooling`
- `bug`

### Problem
When using `context.push_data()` in a PlaywrightCrawler's default request handler, all `push_data()` calls complete successfully (no errors thrown), but subsequent calls to `crawler.get_data()` or `Dataset.get_data()` return only a subset of the pushed items.

This appears to be related to the storage buffering mechanism mentioned in issue #1532, but affects **success handlers** rather than error handlers.

### Expected Behavior
- Push 4 items with `context.push_data()` → All 4 succeed
- Query with `crawler.get_data()` → Should return all 4 items

### Actual Behavior
- Push 4 items with `context.push_data()` → All 4 succeed
- Query with `crawler.get_data()` → Returns only 2 items (non-deterministic which ones)
- Crawler statistics show `requests_finished: 4`
- No errors or warnings logged

### Impact
- Integration tests become flaky and unreliable
- Cannot verify all crawled data was persisted
- Affects production data quality assurance
- Makes it impossible to know if data loss occurred

### Minimal Reproduction

```python
import asyncio
from pathlib import Path
import tempfile
from crawlee.crawlers import PlaywrightCrawler
from crawlee.configuration import Configuration

async def main():
    # Create unique storage directory
    with tempfile.TemporaryDirectory(prefix="crawlee_test_") as tmpdir:
        storage_path = Path(tmpdir)

        # Configure crawler with isolated storage
        config = Configuration(storage_dir=str(storage_path))
        crawler = PlaywrightCrawler(configuration=config)

        # Simple handler that pushes data
        @crawler.router.default_handler
        async def handler(context):
            await context.push_data({
                "url": context.request.url,
                "title": "Test",
                "timestamp": str(context.request.loaded_url),
            })

        # Crawl 4 URLs (same page with different query params to avoid deduplication)
        test_url = "https://crawlee.dev"
        await crawler.run([
            f"{test_url}?test=1",
            f"{test_url}?test=2",
            f"{test_url}?test=3",
            f"{test_url}?test=4",
        ])

        # Poll for data (allow time for persistence)
        await asyncio.sleep(2.0)

        # Query dataset
[...]

PR #1626, opened 2025-12-17
fix: Prevent race condition in concurrent storage creation

Mantisus:
### Description

- Fix race condition when concurrent creating storage.

### Issues

- Closes: #1621

### Testing

- Add new tests for `StorageInstanceManager`

### DIFF
--- a/src/crawlee/storages/_storage_instance_manager.py
+++ b/src/crawlee/storages/_storage_instance_manager.py
@@ -122,7 +122,7 @@ class StorageInstanceManager:
             if not any([name, alias, id]):
                 alias = self._DEFAULT_STORAGE_ALIAS
 
-            # Check cache
+            # Check cache without lock first for performance.
             if cached_instance := self._get_from_cache(
                 cls,
                 id=id,
@@ -131,14 +131,6 @@ class StorageInstanceManager:
                 storage_client_cache_key=storage_client_cache_key,
             ):
                 return cached_instance
-
-            # Check for conflicts between named and alias storages
-            self._check_name_alias_conflict(
-                cls,
-                name=name,
-                alias=alias,
-                storage_client_cache_key=storage_client_cache_key,
-            )
 
             # Validate storage name
             if name is not None:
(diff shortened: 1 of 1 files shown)


src/crawlee/storages/_storage_instance_manager.py
```diff
@@ -120,62 +123,74 @@ async def open_storage_instance(
...
-            client: KeyValueStoreClient | DatasetClient | RequestQueueClient
-            client = await client_opener_coro
+            # Acquire lock for this opener
+            opener_lock_key = (cls, str(id or name or alias), storage_client_cache_key)
+            if not (lock := self._opener_locks.get(opener_lock_key)):
+                lock = Lock()
+                self._opener_locks[opener_lock_key] = lock
+
+            async with lock:
+                # Re-check cache inside the lock
```
janbuchar:
```suggestion
                # Another task could have created the storage while we were waiting for the lock - check if that happened
```

src/crawlee/storages/_storage_instance_manager.py
```diff
@@ -120,62 +123,74 @@ async def open_storage_instance(
...
-                raise RuntimeError('Cached instance type mismatch.')
-
-            if name is not None and (cached_instance := self._cache.by_name[cls][name].get(storage_client_cache_key)):
-                if isinstance(cached_instance, cls):
-                    return cached_instance
-                raise RuntimeError('Cached instance type mismatch.')
-
-            if alias is not None and (
-                cached_instance := self._cache.by_alias[cls][alias].get(storage_client_cache_key)
+            if cached_instance := self._get_from_cache(
```
janbuchar, line 126: Maybe we should add a comment here that explains that we first check the cache without acquiring the lock as an optimization

src/crawlee/storages/_storage_instance_manager.py
```diff
@@ -120,62 +123,74 @@ async def open_storage_instance(
...
-            return instance
+                # Cache named storage.
+                if instance_name is not None:
+                    self._cache.by_name[cls][instance_name][storage_client_cache_key] = instance
+
+                # Cache unnamed storage.
+                if alias is not None:
+                    self._cache.by_alias[cls][alias][storage_client_cache_key] = instance
+
+                return instance
```
janbuchar, line 189: This has to happen without interruption. Otherwise, the "cache pre-check" without a lock could fail in weird ways.

Can you add comments for our future selves that there should be no `await`s in this range?

janbuchar (commented):
Nice catch!

PR #1635, opened 2025-12-20
feat: Expose `AdaptivePlaywrightCrawlerStatisticState` for `AdaptivePlaywrightCrawler`

Mantisus:


Mantisus:
> We need to expose this to support cases where we want to implement a new with_<NAME>_static_parser method on AdaptivePlaywrightCrawler - to allow custom parsers other than BeautifulSoup and Parsel.

> The generic StatisticsState is already exposed, but for the AdaptivePlaywrightCrawler, the method needs to use AdaptivePlaywrightCrawlerStatisticState instead. That makes sense.

All right 🙂 

And this class must be public according to our current documentation - https://crawlee.dev/python/api/class/AdaptivePlaywrightCrawlerStatisticState.

Issue #1627, opened 2025-12-18
Port fix of multiple crawler instances sharing state from JS version


janbuchar:
- https://github.com/apify/crawlee/issues/3024 and  https://github.com/apify/crawlee/pull/3309

Pijukatel:
I think that the issue #1653  was something different.

Pijukatel:
While adding this to Python. It can be added in an already improved way described here:
https://github.com/apify/crawlee/issues/3330#issuecomment-3737537019

Crawler can get an explicit ID (or a predictably automatically generated ID)
This ID is used for:

- Statistics
- UseState

Implementation has to ensure the following scenarios:
- **1 Crawler with migration**
Crawler uses the same statistics and state even after migration (Apify SDK context or running locally twice without purging the default kvs)
- **2 Crawlers at the same time (automatic ids) and migration**
Each crawler should by default have its own statistics and state (breaking backwards compatibility bugfix???)
After migration, each crawler is re-using the matching statistics and state
- **2 Crawlers at the same time with explicitly same id**
They use the same statistics and state.

PR #1669, opened 2026-01-12
fix: Do not share state between different crawlers unless requested

Pijukatel:
### Description

- Introduces a new argument `id` for `BasicCrawler`. This argument controls the shared state.
- Each new instance of the `BasicCrawler` gets an automatically incremented `id` to avoid unintentional sharing of state between crawlers.
- If two crawlers should use the same state, then it is possible to pass the same `id` to the crawler `__init__`.

### Issues

Closes: #1627

### Testing

- Added tests.

### DIFF
--- a/src/crawlee/crawlers/_basic/_basic_crawler.py
+++ b/src/crawlee/crawlers/_basic/_basic_crawler.py
@@ -355,8 +355,8 @@ class BasicCrawler(Generic[TCrawlingContext, TStatisticsState]):
             status_message_logging_interval: Interval for logging the crawler status messages.
             status_message_callback: Allows overriding the default status message. The default status message is
                 provided in the parameters. Returning `None` suppresses the status message.
-            id: Id of the crawler used for state tracking. You can use same explicit id to share state and between two
-                crawlers. By default, each crawler will use own state.
+            id: Identifier used for crawler state tracking. Use the same id across multiple crawlers to share state
+                between them.
             _context_pipeline: Enables extending the request lifecycle and modifying the crawling context.
                 Intended for use by subclasses rather than direct instantiation of `BasicCrawler`.
             _additional_context_managers: Additional context managers used throughout the crawler lifecycle.
--- a/tests/unit/crawlers/_basic/test_basic_crawler.py
+++ b/tests/unit/crawlers/_basic/test_basic_crawler.py
@@ -1722,7 +1722,7 @@ async def test_add_requests_with_rq_param(queue_name: str | None, queue_alias: s
     crawler = BasicCrawler()
     rq = await RequestQueue.open(name=queue_name, alias=queue_alias)
     if by_id:
-        queue_id = rq._id
+        queue_id = rq.id
         queue_name = None
     else:
         queue_id = None
@@ -1816,6 +1816,7 @@ async def _run_crawler(crawler_id: int | None, requests: list[str], storage_dir:
         context.log.info(f'State {state}')
 
     crawler = BasicCrawler(
+        id=crawler_id,
         request_handler=request_handler,
         concurrency_settings=ConcurrencySettings(max_concurrency=1, desired_concurrency=1),
         configuration=Configuration(


Pijukatel:
Better discuss this. After implementing this draft, I am leaning towards **alternative 1** (see description)

@janbuchar , @barjin, @B4nan. Could you please share your point of view?

You can see the usage in code in the updated and new test in this PR.

barjin:
> Explicitly passing `state_kvs` instead of `crawler_id`

The `state` is just one key in the KVS though, it feels weird to me to make the `state` this prominent in our API. If it's about the entire KVS (so e.g. [get_key_value_store](https://github.com/apify/crawlee-python/blob/2d5a5719f9545c403e55f52c49a60fedc25cd14c/src/crawlee/crawlers/_basic/_basic_crawler.py#L625) will also return this KVS), then it makes a bit more sense to me. 

Maybe it's unclear that the "crawler state" is actually stored in KVS - this we should IMO communicate better in the docs.


Having thought about this a bit more, I see the "state sharing" as a bug again :) Different crawler instances IMO shouldn't influence each other just because they are touching the same storage implementation (if this is intentional, it should be explicit).

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -297,6 +298,7 @@ def __init__(
         status_message_logging_interval: timedelta = timedelta(seconds=10),
         status_message_callback: Callable[[StatisticsState, StatisticsState | None, str], Awaitable[str | None]]
         | None = None,
+        crawler_id: int | None = None,
```
B4nan: the option should be called just `id` as in the JS version, right?

janbuchar: Maybe Pepa chose that name because `id` is a bultin function in Python. If that's the case, I think we can safely shadow it.

Pijukatel: @janbuchar exactly. But just id is fine for me as well.

Pijukatel: The point why I dislike it is, that "crawler.id" is very ambiguous, and it can be implemented to control:

- only the state - people might get confused, why does `id` control only sharing of the state and not statistics? In that case it is not explicitly obvious from the argument and it would be better called something like `state_id`. 
- state and statistics - this would cause a situation where you can either share both or none and also `Statistics` can already be shared explicitly.

Pijukatel:
> > Explicitly passing `state_kvs` instead of `crawler_id`
> 
> The `state` is just one key in the KVS though, it feels weird to me to make the `state` this prominent in our API. If it's about the entire KVS (so e.g. [get_key_value_store](https://github.com/apify/crawlee-python/blob/2d5a5719f9545c403e55f52c49a60fedc25cd14c/src/crawlee/crawlers/_basic/_basic_crawler.py#L625) will also return this KVS), then it makes a bit more sense to me.
> 
> Maybe it's unclear that the "crawler state" is actually stored in KVS - this we should IMO communicate better in the docs.
> 
> Having thought about this a bit more, I see the "state sharing" as a bug again :) Different crawler instances IMO shouldn't influence each other just because they are touching the same storage implementation (if this is intentional, it should be explicit).

What about having an optional argument `use_state`?  The default will be a function that saves to the default kvs under an automatically incremented id and the user can pass whatever custom implementation if they want something custom, like sharing the same state between two crawlers.

This will be an easy and clear default without the need for extra arguments and maximum flexibility for custom use cases.

janbuchar:
> What about having an optional argument `use_state`? The default will be a function that saves to the default kvs under an automatically incremented id and the user can pass whatever custom implementation if they want something custom, like sharing the same state between two crawlers.

I have a hard time imagining that, could you sketch out some code samples?

Pijukatel:
> > What about having an optional argument `use_state`? The default will be a function that saves to the default kvs under an automatically incremented id and the user can pass whatever custom implementation if they want something custom, like sharing the same state between two crawlers.
> 
> I have a hard time imagining that, could you sketch out some code samples?

Please check the latest commit. I added an example of how this could be done. (Please do not focus on that specific example; it is just to demonstrate the idea. The question is whether the `use_state` should be some hardcoded internal that can be parametrized, or if it should be a component of the crawler that can be fully replaced by a custom implementation. )

tests/unit/crawlers/_basic/test_basic_crawler.py
```diff
@@ -810,11 +810,65 @@ async def handler(context: BasicCrawlingContext) -> None:
...
     assert value == {'hello': 'world'}
 
 
+async def test_context_use_state_crawlers_share_custom_state() -> None:
+    custom_state_dict = {}
+
+    async def custom_use_state(default_state: dict[str, JsonSerializable]) -> dict[str, JsonSerializable]:
+        if not custom_state_dict:
+            custom_state_dict.update(default_state)
+        return custom_state_dict
```
janbuchar: The fact that this is not persistent by default will definitely surprise someone.

Pijukatel: This is just a totally basic **custom** implementation of `use_state` for the sake of the test. By no means to be used anywhere outside of this test.

This is not the default implementation.

janbuchar: Sure, what I mean is that users could be surprised by the fact that if they supply a custom use_state, we don't handle persistence for them.

src/crawlee/crawlers/_basic/_basic_crawler.py
B4nan: IMO it should cover both state and stats (and whatever else we identify). 

I don't like `state_id` much, since that sounds like something only related to `useState`. And I do like plain `id` since we are talking about a unique identifier for a crawler instance.

This is all rather internal to me. I wouldn't bike shed on this topic too much, 99% people won't use this explicitly.

tests/unit/crawlers/_basic/test_basic_crawler.py
B4nan: Maybe I missed something, but why are we even discussing "custom use state implementation"? That was never the goal, we just wanted to make sure two crawler instances don't share stuff unless explicitly instructed to. I actually doubt someone will even use that, I don't see the why.

B4nan:
Regarding [this comment](https://github.com/apify/apify-docs/pull/2176#discussion_r2687099830) about `Actor.useState` vs `Crawler.useState`, I'd actually say they should point to the same state by default, let's call it `id: 0`. If you would create two crawlers in one context, the first would use the same state as the SDK. 

Now the reasoning - we also have `context.useState` and I'd say it makes sense to allow people use `Actor.useState` and `context.useState` together.

If we don't do this, it feels like a nasty BC that might surprise many people. Imagine you init the state before you run the crawler via `Actor.useState` and then you use `context.useState` in the handler, but the initial state is not there, forcing you to define it in the handler again.

src/crawlee/crawlers/_basic/_basic_crawler.py
Pijukatel: Ok, based on the [discussion](https://apify.slack.com/archives/C02JQSN79V4/p1768467566940549), the id could be used for more than just the state. Lets use it than.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -213,6 +213,10 @@ class _BasicCrawlerOptions(TypedDict):
     """Allows overriding the default status message. The default status message is provided in the parameters.
     Returning `None` suppresses the status message."""
 
+    id: NotRequired[int]
```
janbuchar, line 216: This is a string in JS and it makes more sense that way

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -213,6 +213,10 @@ class _BasicCrawlerOptions(TypedDict):
...
+    """Id of the crawler used for state tracking. You can use same explicit id to share state between two crawlers.
```
janbuchar: Please ask an LLM to improve the wording here 🙂

src/crawlee/crawlers/_basic/_basic_crawler.py
Pijukatel: This is done the same way as in [Statistics](https://github.com/apify/crawlee-python/blob/master/src/crawlee/statistics/_statistics.py#L85) and [StatisticsState](https://github.com/apify/crawlee-python/blob/master/src/crawlee/statistics/_models.py#L62). I would like to stay consistent with those types in this repo in the first place. Changing them to a string would be a breaking change.

So keeping this int for now and creating V2 issue to change them to a string?

janbuchar: Dunno, not a fan of "for now" solutions, but if V2 happens this year, we can do it that way.

Pijukatel: https://github.com/apify/crawlee-python/issues/1699

src/crawlee/crawlers/_basic/_basic_crawler.py
Pijukatel: reworded

PR #1670, opened 2026-01-12
test: Use rerun instead skip for flaky tests

Mantisus:
### Description

- Perform 3 launch attempts for tests marked as flaky, instead of skipping them.


Pijukatel:
If used carefully, it can be ok. However, we should be cautious not to overuse it, as it can hide the real issues.

Issue #1673, opened 2026-01-15
The enqueue_links function misinterprets the limit keyword argument


honzajavorek:
Minimal reproducible code:

```python
import asyncio
from urllib.parse import quote_plus

from crawlee import Request
from crawlee.crawlers import BeautifulSoupCrawler, BeautifulSoupCrawlingContext


async def main() -> None:
    crawler = BeautifulSoupCrawler()

    @crawler.router.default_handler
    async def handle_listing(context: BeautifulSoupCrawlingContext) -> None:
        await context.enqueue_links(
            selector="h4 a", label="DETAIL", limit=1
        )

    @crawler.router.handler("DETAIL")
    async def handle_detail(context: BeautifulSoupCrawlingContext) -> None:
        print("Detail page URL:", context.request.url)

    await crawler.run(["https://honzajavorek.cz/blog/"])


if __name__ == "__main__":
    asyncio.run(main())

```

Run like this:

```
uv run --with crawlee[beautifulsoup] python crawlee_bug.py
```

Expected output:

```
Detail page URL: https://honzajavorek.cz/blog/tydenni-poznamky-vanoce-a-tak/
```

Actual output:

```
Detail page URL: https://honzajavorek.cz/blog/tydenni-poznamky-vanoce-a-tak/
Detail page URL: https://honzajavorek.cz/blog/tydenni-poznamky-vylepsovani-seznamu-kandidatu-nove-bydleni-a-odpocinek-v-mlze/
```

If I change the limit to 5, I get 6 links on the output.

I tried to nail down the problem, but I spent 20 minutes drowning in the sea of type definitions, kwargs passed down as is, and nested `_create_...` functions. It was impossible to me to follow the flow of the keyword arguments and find the place where the limit gets actually applied.

honzajavorek:
Thanks for the fix!

PR #1674, opened 2026-01-15
fix: Fix adding extra link for `EnqueueLinksFunction` with `limit`

Mantisus:
### Description

- fix adding extra link for `EnqueueLinksFunction` with `limit`

### Issues

- Closes: #1673 

### Testing

- Add tests for `enqueue_links` with `limit` argument

### DIFF
--- a/src/crawlee/crawlers/_basic/_basic_crawler.py
+++ b/src/crawlee/crawlers/_basic/_basic_crawler.py
@@ -1056,9 +1056,10 @@ class BasicCrawler(Generic[TCrawlingContext, TStatisticsState]):
             ) and self._check_url_patterns(target_url, kwargs.get('include'), kwargs.get('exclude')):
                 yield request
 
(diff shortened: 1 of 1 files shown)


Pijukatel (approved):

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1057,7 +1057,7 @@ def _enqueue_links_filter_iterator(
                 yield request
 
                 limit = limit - 1 if limit is not None else None
-                if limit and limit <= 0:
+                if limit is not None and limit <= 0:
                     break
```
Pijukatel: Consider using a nested condition here. It seems to me slightly more readable, and we do not need to check the limit for None twice. (Just a weak recommendation)

```
                if limit is not None:
                    limit -= 1
                    if limit <= 0:
                        break
```

PR #1688, opened 2026-01-26
docs: Relax website dependency constraints

vdusek:
- Relax website dependency constraints to reduce the number of unnecessary Renovate PRs
  - For example https://github.com/apify/crawlee-python/pull/1518, https://github.com/apify/crawlee-python/pull/1479, ...
- Switch GitHub Actions to major-version tags for the same reason (only `typos`, for `actionlint` it isn't possible, unfortunately).

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -13,26 +13,14 @@
         "prettify": "prettier --write --config ./tools/docs-prettier.config.js ../docs/guides/*.md",
         "swizzle": "docusaurus swizzle",
         "deploy": "rimraf .docusaurus && node --max_old_space_size=16000 node_modules/@docusaurus/core/bin/docusaurus.mjs deploy",
-        "docusaurus": "docusaurus"
-    },
-    "devDependencies": {
-        "@apify/eslint-config-ts": ">=0.4.0 <1.0.0",
-        "@apify/tsconfig": ">=0.1.0 <1.0.0",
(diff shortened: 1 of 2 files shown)


website/package.json
```diff
@@ -16,45 +16,45 @@
...
+        "@docusaurus/core": "^3.9.2",
+        "@docusaurus/faster": "^3.9.2",
+        "@docusaurus/mdx-loader": "^3.9.2",
+        "@docusaurus/plugin-client-redirects": "^3.9.2",
+        "@docusaurus/preset-classic": "^3.9.2",
+        "@docusaurus/theme-mermaid": "^3.9.2",
         "@giscus/react": "^3.0.0",
         "@mdx-js/react": "^3.0.1",
-        "@mermaid-js/layout-elk": "^0.2.0",
+        "@mermaid-js/layout-elk": ">=0.2.0 <1.0.0",
```
barjin, line 36: These are not that safe; any version bump in `0.x.y` versions can contain breaking changes ([see here](https://semver.org/#spec-item-4)).

vdusek: Yeah, I know - but we still have a lockfile that pins the exact version, and we should find out if there is an issue during a lock file maintenance.

This is the only production dependency affected. If you think the lockfile is not sufficient and that it would be better to keep it as it was (using `^` constraint), I'll revert it.

For dev dependencies, I think it's really OK.

PR #1693, opened 2026-01-26
ci: Increase unit tests concurrency from 2 to 8

vdusek:
Increase unit tests concurrency from 2 to 8, to make them faster and identify all the weak points.

### DIFF
--- a/.github/workflows/_tests.yaml
+++ b/.github/workflows/_tests.yaml
@@ -17,4 +17,4 @@ jobs:
       operating_systems: '["ubuntu-latest", "windows-latest", "macos-latest"]'
       python_version_for_codecov: "3.14"
       operating_system_for_codecov: ubuntu-latest
-      tests_concurrency: "1"
+      tests_concurrency: "8"


vdusek:
> I think this is risky and can make us less aware of parallelism-related issues. We already have run_alone marker which is to be used explicitly to make tests run alone for code we know is not supposed to be run in parallel. The rest of the code should be able to run in parallel; if it does not, it is either a bug or our wrong assumption - in either case, that is something we need to expose and not hide.

Alright then, let's increase the concurrency to 8 and identify all the weak points :slightly_smiling_face:

Pijukatel:
> > I think this is risky and can make us less aware of parallelism-related issues. We already have run_alone marker which is to be used explicitly to make tests run alone for code we know is not supposed to be run in parallel. The rest of the code should be able to run in parallel; if it does not, it is either a bug or our wrong assumption - in either case, that is something we need to expose and not hide.
> 
> Alright then, let's increase the concurrency to 8 and identify all the weak points 🙂

If it causes too much flakiness in the CI I would propose to mark flaky tests as `run_alone` (first `run_alone`,  skip if it does not help) and create an issue to be investigated later. This should offer a nice balance between stable CI and exposing issues I hope.

vdusek:
> If it causes too much flakiness in the CI I would propose to mark flaky tests as run_alone (first run_alone, skip if it does not help) and create an issue to be investigated later. This should offer a nice balance between stable CI and exposing issues I hope.

Agreed.

cc @Mantisus

Issue #861, opened 2025-01-05
Is there an option to cache downloaded pages locally?


matecsaj:
It would be a huge time saver during development if the website only needed to be crawled once. While iterating on my data extraction code, I wish I could load the pages from my local disk instead of the Internet.

During development, repeatedly crawling the website to test data extraction code is time-consuming and inefficient. A caching mechanism that stores previously downloaded pages on the local disk would be a huge time saver. Instead of fetching the pages from the Internet, the development environment could simply load them from local storage.

This approach would not only speed up the iterative development process but also reduce load on the website's server. Such a feature would be particularly useful for debugging and refining data extraction scripts.

vdusek:
Hi, this sounds reasonable, while it is not currently on our roadmap. We will revise it in the future.

Pijukatel:
I think this can already be achieved by the approach described in the documentation related to web archiving.
During the first crawl, follow these steps: 
[Crawling through proxy recording server](https://crawlee.dev/python/docs/guides/creating-web-archive#crawling-through-proxy-recording-server)
To just re-use already archived pages, follow these steps:
[Using the archived data](https://crawlee.dev/python/docs/guides/creating-web-archive#using-the-archived-data)

PR #1698, opened 2026-01-28
feat: Add `response_cache` argument for HTTP-based crawlers for caching HTTP responses

Mantisus:
### Description

- Add `response_cache` argument to HTTP-based crawlers for optional response caching
- Add `ContextPipeline.compose_with_skip()` for conditional middleware skipping

`response_cache` accepts a `KeyValueStore` instance. When enabled, successful responses are cached on the first request and reused on subsequent runs. Useful during development to avoid excessive load on target sites.

Inspired by #801 and #861.

### Testing

- Add new tests for `ContextPipeline` and `HttpCrawler`.

### DIFF
--- a/src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
+++ b/src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
@@ -5,6 +5,7 @@ import logging
 import sys
 from abc import ABC
 from datetime import timedelta
+from hashlib import sha256
 from typing import TYPE_CHECKING, Any, Generic
 
 if sys.version_info >= (3, 14):
@@ -342,7 +343,7 @@ class AbstractHttpCrawler(
         if not self._response_cache:
             raise RuntimeError('Response cache is not configured.')
 
-        key = f'response_{context.request.unique_key}'
+        key = self._get_cache_key(context.request.unique_key)
         raw = await self._response_cache.get_value(key)
 
         if raw is None:
@@ -375,7 +376,13 @@ class AbstractHttpCrawler(
 
         compressed = _compressor.compress(cached.model_dump_json().encode())
 
-        key = f'response_{context.request.unique_key}'
+        key = self._get_cache_key(context.request.unique_key)
         await self._response_cache.set_value(key, compressed)
 
         yield context
+
+    @staticmethod
+    def _get_cache_key(unique_key: str) -> str:
+        """Generate a deterministic cache key for a unique_key."""
+        hashed_key = sha256(unique_key.encode('utf-8')).hexdigest()
+        return f'response_{hashed_key[:15]}'


Pijukatel:
Hi, thanks for the PR.

I am not sure we need this, though. My reasoning is that this can already be achieved by [the external archiving tool](https://crawlee.dev/python/docs/guides/creating-web-archive#crawling-through-proxy-recording-server), and that works both for simple `HTTP-based` crawlers and `PlaywrightCrawler`. No need for any change in Crawlee code.

This solution adds another way how to do it for the HTTP-based crawler only. It has some advantages, like easier setup, but the disadvantages are additional code that increases the complexity of the pipeline and the limitation to the HTTP-based crawlers only.

Mantisus:
> I am not sure we need this, though. My reasoning is that this can already be achieved by [the external archiving tool](https://crawlee.dev/python/docs/guides/creating-web-archive#crawling-through-proxy-recording-server), and that works both for simple HTTP-based crawlers and PlaywrightCrawler. No need for any change in Crawlee code.

Hi.

Yes, I have considered this. However, I decided to try to implement this solution because there are nuances when using the archiving tool that may make it less convenient for users.

1. When archiving through a proxy, the user loses the ability to use external proxy servers.
2. The user needs to create two different workflows. One for recording, the second for working with recorded data.

However, I agree with the shortcomings of the proposed solution.

janbuchar:
What would be the minimal change to crawlee required to allow downstream users to implement this on their own?

Mantisus:
> What would be the minimal change to crawlee required to allow downstream users to implement this on their own?

The minimal change is an extension of the `ContextPipeline` functionality, allowing for skipping middleware.

However, to implement caching, the pipeline in `AbstractHttpCrawler` needs to be updated.

Therefore, users will need to implement their own subclass of `AbstractHttpCrawler` with a new pipeline and, based on the subclass, create `HttpCrawler` or a crawler with a parser.

UPD: Alternatively, we need to consider refactoring `ContextPipeline` to make it easier for users to add custom middleware to the middle of the pipeline. But probably this is for v2

Pijukatel:
> * When archiving through a proxy, the user loses the ability to use external proxy servers.
> 
> * The user needs to create two different workflows. One for recording, the second for working with recorded data.


I am not 100% sure, but I believe that both points could be solved by the correct setting of the archiving server.

janbuchar:
By the way, couldn't the same behavior be achieved by modifying the HttpClient? If so, then we could probably keep the ContextPipeline simple(r)

Mantisus:
> By the way, couldn't the same behavior be achieved by modifying the HttpClient?

It is preferable that caching occurs after `_handle_status_code_response`. This is to avoid caching pages that require repeated requests.

> we could probably keep the ContextPipeline simple(r)

Yes, I understand. I agree with @Pijukatel that this update will significantly complicate the code.
At least partially, this task can be solved using [the external archiving tool](https://crawlee.dev/python/docs/guides/creating-web-archive#crawling-through-proxy-recording-server).

For now, I would suggest closing this PR. If we encounter future use cases that require similar updates, we can revisit it.

Mantisus:
I am closing this PR so as not to complicate the creation of pipelines until there is a more compelling reason to do so.

PR #1702, opened 2026-01-30
fix: Prevent mutation of default URL patterns list in `block_requests`

vdusek:
When `url_patterns` was `None`, the code assigned it to the module-level `_DEFAULT_BLOCK_REQUEST_URL_PATTERNS` constant and then called `extend()` on it, permanently mutating the shared default list. This caused `extra_url_patterns` to accumulate across all subsequent calls that used the default patterns.

Fixed by creating a copy of the list before extending it.

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>


Mantisus (approved):
Good catch

PR #1703, opened 2026-01-30
fix: Prevent list modification during iteration in BrowserPool

vdusek:
Both `_identify_inactive_browsers` and `_close_inactive_browsers` methods were modifying their respective lists (`_active_browsers` and `_inactive_browsers`) while iterating over them. This is a classic bug that can cause items to be skipped or raise `RuntimeError`.

The fix iterates over a copy of the list (`list(...)`) while modifying the original, ensuring all items are properly processed.

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

### DIFF
--- a/src/crawlee/browsers/_browser_pool.py
+++ b/src/crawlee/browsers/_browser_pool.py
@@ -346,14 +346,14 @@ class BrowserPool:
 
     def _identify_inactive_browsers(self) -> None:
         """Identify inactive browsers and move them to the inactive list if their idle time exceeds the threshold."""
-        for browser in self._active_browsers:
+        for browser in list(self._active_browsers):
             if browser.idle_time >= self._browser_inactive_threshold:
                 self._active_browsers.remove(browser)
                 self._inactive_browsers.append(browser)
 
     async def _close_inactive_browsers(self) -> None:
         """Close the browsers that have no active pages and have been idle for a certain period."""
-        for browser in self._inactive_browsers:
+        for browser in list(self._inactive_browsers):
             if not browser.pages:
                 await browser.close()
                 self._inactive_browsers.remove(browser)


Pijukatel:
Would this ever cause any problem? I think it is no issue in the current implementation, but it could be a hidden bug if the implementation changes. Not sure if that is good enough justification for the change :-D

`_identify_inactive_browsers` is sync, so it will run through the iteration without anyone modifying its list.
`_close_inactive_browsers` is async, so it can yield in the middle of the iteration and the `_identify_inactive_browsers` can start changing the list. But the worst-case scenario is that it will miss some freshly added `inactive_browsers`, which does not matter at all, because it is a `RecurringTask` that runs periodically, so they will be closed in the next run anyway.

vdusek:
The point is, when you remove an item from a list during iteration, the iterator's index doesn't adjust, and the next item will be skipped.

```python
lst = [1, 2, 3, 4]
for x in lst:
    print(x)
    if x == 2:
        lst.remove(x)
print(lst)
```

```
$ python run.py 
1
2
4
[1, 3, 4]
```

See e.g. this https://stackoverflow.com/questions/1637807/modifying-list-while-iterating

So this is clearly a bug.

Pijukatel:
> The point is, when you remove an item from a list during iteration, the iterator's index doesn't adjust, and the next item will be skipped.
...

Sure, but that is not the case in our code. The case in our code is that an item can be added to the list during the iteration, so such an item will be missed in the current loop, but since this function runs periodically, it does not matter, as it will be handled next time the function runs.

vdusek:
> Sure, but that is not the case in our code.

It is exactly the case in our code - both methods remove from the same list they iterate over.

I understand that the practical impact is minor as skipped items get caught on the next periodic run, but it's still a bug, and this 2-line fix is trivially safe. The PR should be merged.

Pijukatel (approved):

Issue #1706, opened 2026-02-02
user_data drops null values that are passed in


DoctorEvil92:
Hi, it seems like `user_data` will drop dict keys that have null values under them. See the example
```
from crawlee.router import Router
from crawlee.crawlers import BeautifulSoupCrawler, BeautifulSoupCrawlingContext
from crawlee import Request
import asyncio


async def main():
    # define links
    target_url = "https://example.com/" # target site

    # define router
    router = Router[BeautifulSoupCrawlingContext]()
    @router.handler("MAIN")
    async def main_handler(context : BeautifulSoupCrawlingContext) -> None:
        loaded_user_data = dict(context.request.user_data)
        print("loaded user_data:", loaded_user_data)
        return
    
    # then try to do a request with crawlee with the same proxy
    crawler = BeautifulSoupCrawler(request_handler=router)

    # run it
    await crawler.run( [Request.from_url(target_url, label='MAIN', user_data={"a":1, "b":"2", "c":None, "d":False})] )

    
    

if __name__ == '__main__':
    asyncio.run(main())

```

<img width="504" height="650" alt="Image" src="https://github.com/user-attachments/assets/e883fa91-01e1-412f-bc88-95718f757a44" />

"c" key got dropped as I load it into a dictionary in the handler.

PR #1707, opened 2026-02-02
fix: Keep None values for `user_data` in `Request`

Mantisus:
### Description

- This PR fixes loss of `None` values in `user_data` during `Request` serialization

### Issues

- Closes: #1706

### DIFF
--- a/src/crawlee/_request.py
+++ b/src/crawlee/_request.py
@@ -217,7 +217,7 @@ class Request(BaseModel):
                 lambda instance: user_data_adapter.dump_python(
                     instance,
                     by_alias=True,
-                    exclude_none=True,
+                    exclude_none=False,
                     exclude_unset=True,
                     exclude_defaults=True,
                 )


src/crawlee/_request.py
```diff
@@ -217,7 +217,7 @@ class Request(BaseModel):
                 lambda instance: user_data_adapter.dump_python(
                     instance,
                     by_alias=True,
-                    exclude_none=True,
```
Pijukatel, line 220: @janbuchar do you remember if this had any specific reason?

janbuchar: No, I don't. The new behavior seems more explicit and I can't think of a problem that I could cause, so let's do this

janbuchar: No, I don't. The new behavior seems more explicit and I can't think of a problem that I could cause, so let's do this

Issue #1704, opened 2026-02-01
Snapshotter does not account for dynamic memory scaling (e.g., K8s burstable QoS)


erandagan:
## Description
In environments like Kubernetes, workloads often scale vertically. A Pod may start with a specific memory request (e.g., 1GiB) but be allowed to burst up to a limit (e.g., 8GiB) if cluster resources permit.

Currently, Crawlee's Snapshotter captures the memory limit only once during initialization. If the runtime allocates more memory to the process later, the Snapshotter remains "locked" to the initial baseline, leading to inaccurate autoscaling decisions and under-utilization of available resources.

## Root Cause
- max_memory_size is set once in the [constructor](https://github.com/apify/crawlee-python/blob/fcf9adb6a0cfeaa87ca482372d4e066584eb28d6/src/crawlee/_autoscaling/snapshotter.py#L72).

- Subsequent [snapshot creation](https://github.com/apify/crawlee-python/blob/fcf9adb6a0cfeaa87ca482372d4e066584eb28d6/src/crawlee/_autoscaling/snapshotter.py#L285) uses this cached value rather than querying the current system state.

## Proposed Solution
Modify `_snapshot_memory` to fetch the max_memory_size value dynamically when a memory event is processed, ensuring the autoscaler respects the current ceiling.

Pijukatel:
Hi, thanks for raising the issue @erandagan.

Could you please test, if the following approach works for your setup?
https://github.com/apify/crawlee-python/pull/1715

PR #1715, opened 2026-02-05
feat: Dynamic memory snapshots 

Pijukatel:
### Description

- Add `Ratio` type to represent the maximum relative available memory of the system.
- Allow to initialize the `Snapshotter.max_memory_size` with either `Ratio` (dynamic memory) or `ByteSize` (fixed memory)
- When `Ratio` is used, `Snapshotter` will take into account the current available memory. (Previously, it would take into account only the initial available memory.)

**Top level usage in Crawlers:**
Fixed memory 
```
BasicCrawler(configuration=Configuration(memory_mbytes=1024))
```

Dynamic memory
```
BasicCrawler(configuration=Configuration(available_memory_ratio=0.5))
```

### Issues

- Closes: #1704

### Testing

- Unit test

### Checklist

- [ ] CI passed

### DIFF
--- a/src/crawlee/_autoscaling/_types.py
+++ b/src/crawlee/_autoscaling/_types.py
@@ -2,10 +2,13 @@ from __future__ import annotations
 
 from dataclasses import dataclass, field
 from datetime import datetime, timedelta, timezone
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Annotated
+
+from pydantic import Field
+from pydantic.dataclasses import dataclass as pydantic_dataclass
 
 if TYPE_CHECKING:
-    from crawlee._utils.byte_size import ByteSize, Ratio
+    from crawlee._utils.byte_size import ByteSize
 
 SYSTEM_WIDE_MEMORY_OVERLOAD_THRESHOLD = 0.97
 
@@ -96,12 +99,8 @@ class MemorySnapshot:
     system_wide_used_size: ByteSize | None
     """Memory usage of all processes, system-wide."""
 
-    max_memory_size: ByteSize | Ratio
-    """The maximum memory that can be used by `AutoscaledPool`.
-
-    When of type `ByteSize` then it is used as fixed memory size. When of type `Ratio` then it allows for dynamic memory
-    scaling based on the available system memory.
-    """
+    max_memory_size: ByteSize
+    """The maximum memory that can be used by `AutoscaledPool`."""
 
     system_wide_memory_size: ByteSize | None
     """Total memory available in the whole system."""
@@ -170,3 +169,10 @@ class ClientSnapshot:
 
 
 Snapshot = MemorySnapshot | CpuSnapshot | EventLoopSnapshot | ClientSnapshot
+
+
+@pydantic_dataclass
+class Ratio:
+    """Represents ratio of memory."""
+
+    value: Annotated[float, Field(gt=0.0, le=1.0)]
--- a/src/crawlee/_autoscaling/snapshotter.py
+++ b/src/crawlee/_autoscaling/snapshotter.py
@@ -3,13 +3,14 @@
 from __future__ import annotations
 
 import bisect
+import functools
 from datetime import datetime, timedelta, timezone
 from logging import getLogger
 from typing import TYPE_CHECKING, TypeVar, cast
 
 from crawlee import service_locator
-from crawlee._autoscaling._types import ClientSnapshot, CpuSnapshot, EventLoopSnapshot, MemorySnapshot, Snapshot
-from crawlee._utils.byte_size import ByteSize, Ratio
+from crawlee._autoscaling._types import ClientSnapshot, CpuSnapshot, EventLoopSnapshot, MemorySnapshot, Ratio, Snapshot
+from crawlee._utils.byte_size import ByteSize
 from crawlee._utils.context import ensure_context
 from crawlee._utils.docs import docs_group
 from crawlee._utils.recurring_task import RecurringTask
@@ -24,6 +25,12 @@ if TYPE_CHECKING:
 logger = getLogger(__name__)
 
 T = TypeVar('T', bound=Snapshot)
+
+
+@functools.lru_cache
+def _warn_once(warning_message: str) -> None:
+    """Log a warning message only once."""
+    logger.warning(warning_message)
 
 
 class SortedSnapshotList(list[T]):
@@ -289,9 +296,13 @@ class Snapshotter:
                 system_wide_memory_size = memory_info.total_size
 
             case MemoryUsageInfo(), Ratio() as ratio:
-                # This is just hypothetical case, that should not happen in practice.
-                # `LocalEvenManager` should always provide `MemoryInfo` in the event data.
+                # This is just hypothetical case, that will most likely not happen in practice.
+                # `LocalEventManager` should always provide `MemoryInfo` in the event data.
                 # When running on Apify, `self._max_memory_size` is always `ByteSize`, not `Ratio`.
+                _warn_once(
+                    'It is recommended that a custom implementation of `LocalEventManager` emits `SYSTEM_INFO` events '
+                    'with `MemoryInfo` and not just `MemoryUsageInfo`.'
+                )
                 max_memory_size = get_memory_info().total_size * ratio.value
                 system_wide_used_size = None
                 system_wide_memory_size = None
--- a/src/crawlee/_utils/byte_size.py
+++ b/src/crawlee/_utils/byte_size.py
@@ -1,20 +1,12 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Annotated, Any
-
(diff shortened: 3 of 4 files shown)


Pijukatel:
E2E tests: https://github.com/apify/crawlee-python/actions/runs/21871925606

src/crawlee/_autoscaling/_types.py
```diff
@@ -97,8 +95,12 @@ class MemorySnapshot:
     system_wide_used_size: ByteSize | None
     """Memory usage of all processes, system-wide."""
 
-    max_memory_size: ByteSize
-    """The maximum memory that can be used by `AutoscaledPool`."""
+    max_memory_size: ByteSize | Ratio
+    """The maximum memory that can be used by `AutoscaledPool`.
+
+    When of type `ByteSize` then it is used as fixed memory size. When of type `Ratio` then it allows for dynamic memory
+    scaling based on the available system memory.
```
Mantisus: I believe the `available_memory_ratio` docstring in `Configuration` should also mention this dynamic scaling behavior.

src/crawlee/_utils/byte_size.py
```diff
@@ -1,14 +1,22 @@
...
 _BYTES_PER_KB = 1024
 _BYTES_PER_MB = _BYTES_PER_KB**2
 _BYTES_PER_GB = _BYTES_PER_KB**3
 _BYTES_PER_TB = _BYTES_PER_KB**4
 
 
+class Ratio(BaseModel):
+    """Represents ratio of memory."""
+
+    value: Annotated[float, Field(gt=0.0, le=1.0)]
```
Mantisus: I believe we could add `gt=0.0` and `le=1.0` constraints to `available_memory_ratio` in `Configuration` for consistency.

src/crawlee/_autoscaling/_types.py
```diff
@@ -117,7 +119,15 @@ def is_overloaded(self) -> bool:
             if system_wide_utilization > SYSTEM_WIDE_MEMORY_OVERLOAD_THRESHOLD:
                 return True
 
-        return (self.current_size / self.max_memory_size) > self.max_used_memory_ratio
+        if isinstance(self.max_memory_size, Ratio):
+            # The snapshot overload is decided not when the snapshot was taken, but when `is_overload` property is
+            # accessed. This allows for dynamic memory scaling. The same memory snapshot that used to be overloaded in
+            # the past can become non-overloaded if the available memory was increased.
+            max_memory_size = ByteSize(int(get_memory_info().total_size.bytes * self.max_memory_size.value))
```
Mantisus: I believe `get_memory_info()` should be wrapped in `asyncio.to_thread` since it uses `psutil`.

Also, I don't think historical snapshots should be affected by memory scaling - the overload was real at the moment the snapshot was taken. Once it's older than `_SNAPSHOT_HISTORY`, it won't influence the `Snapshotter` anyway. The 30-second inertia from old memory values seems reasonable to me.

src/crawlee/_autoscaling/snapshotter.py
```diff
@@ -296,7 +298,17 @@ def _snapshot_memory(self, event_data: EventSystemInfoData) -> None:
         snapshots = cast('list[Snapshot]', self._memory_snapshots)
         self._prune_snapshots(snapshots, snapshot.created_at)
         self._memory_snapshots.add(snapshot)
-        self._evaluate_memory_load(event_data.memory_info.current_size, event_data.memory_info.created_at)
+
+        if isinstance(self._max_memory_size, Ratio):
+            max_memory_size = ByteSize(int(get_memory_info().total_size.bytes * self._max_memory_size.value))
```
Mantisus: I think we could skip calling `get_memory_info()` when `event_data.memory_info` is `MemoryInfo` and just use `memory_info.total_size.bytes` instead.

Mantisus (changes requested):
Looks good! I have a few comments below

src/crawlee/_autoscaling/_types.py
Pijukatel: Updated

src/crawlee/_utils/byte_size.py
Pijukatel: Added

src/crawlee/_autoscaling/_types.py
Pijukatel: Good point. No need to be so "realtime-ish"

src/crawlee/_autoscaling/snapshotter.py
Pijukatel: Updated

src/crawlee/_autoscaling/_types.py
```diff
@@ -97,8 +96,12 @@ class MemorySnapshot:
     system_wide_used_size: ByteSize | None
     """Memory usage of all processes, system-wide."""
 
-    max_memory_size: ByteSize
-    """The maximum memory that can be used by `AutoscaledPool`."""
+    max_memory_size: ByteSize | Ratio
```
vdusek: In `_snapshot_memory`, we now always resolve `Ratio` to `ByteSize`, so I believe there should be only `ByteSize`. Also, don't forget to update the docstring.


```suggestion
    max_memory_size: ByteSize
```

src/crawlee/_utils/byte_size.py
```diff
@@ -1,14 +1,22 @@
 from __future__ import annotations
```
vdusek, line 1: I don't think the `Ratio` belongs here. A ratio is not a byte size - it's a proportion of memory. We should place `Ration` in the `_autoscaling/_types.py` where it is (only) used.

src/crawlee/_utils/byte_size.py
```diff
@@ -1,14 +1,22 @@
...
+
+from pydantic import BaseModel, Field
 
 _BYTES_PER_KB = 1024
 _BYTES_PER_MB = _BYTES_PER_KB**2
 _BYTES_PER_GB = _BYTES_PER_KB**3
 _BYTES_PER_TB = _BYTES_PER_KB**4
 
 
+class Ratio(BaseModel):
```
vdusek: This should probably be only a dataclass instead of a Pydantic model

src/crawlee/_autoscaling/snapshotter.py
```diff
@@ -280,23 +282,51 @@ def _snapshot_memory(self, event_data: EventSystemInfoData) -> None:
...
+        match event_data.memory_info, self._max_memory_size:
+            case MemoryInfo() as memory_info, Ratio() as ratio:
+                max_memory_size = memory_info.total_size * ratio.value
+                system_wide_used_size = memory_info.system_wide_used_size
+                system_wide_memory_size = memory_info.total_size
+
+            case MemoryUsageInfo(), Ratio() as ratio:
+                # This is just hypothetical case, that should not happen in practice.
+                # `LocalEvenManager` should always provide `MemoryInfo` in the event data.
+                # When running on Apify, `self._max_memory_size` is always `ByteSize`, not `Ratio`.
```
vdusek: Please log warning then

tests/unit/_autoscaling/test_snapshotter.py
```diff
@@ -369,3 +378,75 @@ def test_sorted_snapshot_list_add_maintains_order() -> None:
             prev_time = sorted_list[i - 1].created_at
             curr_time = snapshot.created_at
             assert prev_time <= curr_time, f'Items at indices {i - 1} and {i} are not in chronological order'
+
+
+_initial_memory_info = get_memory_info()
```
vdusek: Can we use a fixture so that this does not run at import time?

janbuchar:
Once you folks are done here, we need to make sure that this is also implemented in the JS port

tests/unit/_autoscaling/test_snapshotter.py
Pijukatel: Reworked to avoid this.

src/crawlee/_autoscaling/snapshotter.py
Pijukatel: Added a warning once.

src/crawlee/_utils/byte_size.py
Pijukatel: Ok, but using `pydantic.dataclasses.dataclass` to get the range validation

src/crawlee/_utils/byte_size.py
Pijukatel: Moved

src/crawlee/_autoscaling/_types.py
Pijukatel: Removed

Pijukatel:
> Once you folks are done here, we need to make sure that this is also implemented in the JS port

https://github.com/apify/crawlee/issues/3408

src/crawlee/_utils/byte_size.py
vdusek: yeah, it probably makes sense

Issue #1734, opened 2026-02-11
Flaky test: test_snapshot_pruning_removes_outdated_records fails with assert 3 == 2


vdusek:
## Description

`test_snapshot_pruning_removes_outdated_records` in `tests/unit/_autoscaling/test_snapshotter.py` is consistently failing across multiple CI runs on Ubuntu (Python 3.13 and 3.14), even when running in isolation (`run_alone` marker with `numprocesses=1`).

## Failure

```
FAILED tests/unit/_autoscaling/test_snapshotter.py::test_snapshot_pruning_removes_outdated_records - assert 3 == 2
```

The test expects that after emitting 4 events with timestamps `[now-5h, now-3h, now-2h, now]` and a `_SNAPSHOT_HISTORY` of 2 hours, only 2 snapshots should remain (the ones within the 2-hour window). However, 3 snapshots are retained instead of 2.

## Root cause hypothesis

The snapshot pruning logic likely has an off-by-one or boundary condition issue when determining which records fall outside the history window. The `now-2h` timestamp sits exactly at the boundary of the 2-hour history window, and depending on timing/precision of `datetime.now()` between the test setup and the pruning execution, it may or may not be pruned.

## Observed in

- CI run on `ubuntu-latest` with Python 3.13 and 3.14 (PR #1733)
- CI run on `ubuntu-latest` with Python 3.13 (commit 9d3da48)
- Also seen as `test_get_cpu_sample` failure (same file, similar pattern) on Windows with Python 3.13 (commit df1347a)

## Steps to reproduce

Run the test repeatedly:
```bash
pytest tests/unit/_autoscaling/test_snapshotter.py::test_snapshot_pruning_removes_outdated_records -x --count=50
```

Pijukatel:
I think that the problem is probably in some race condition in sorting/pruning snapshots. Where both are modifying the list in place and most likely are running at the same time.

PR #1735, opened 2026-02-11
fix: Fix `Snapshotter` handling of out of order samples

Pijukatel:
### Description:
- Update `test_snapshot_pruning_removes_outdated_records` to remove the expected source of flaky behavior
- The root cause of the flaky test behavior is a design flaw in snapshot handling:
`event_manager.on(event=Event.SYSTEM_INFO, listener=self._snapshot_cpu)` adds `self._snapshot_cpu` listener, but event manager runs sync listeners through `asyncio.to_thread`. `self._snapshot_cpu` modifies in-place instance list (for example `self._cpu_snapshots`) - it does `bisect.insort` and `del` operations on the same list from several threads, which creates oportunity for a race condition.
- Changed order of operations. The new `Snapshot` is now added first, and the pruning of the snapshot list is done based on `Snapshot` with the newest date in the list (the newest date does not have to be the last added Snapshot). This fixes the bug when the last out-of-order `Snapshot` could cause wrong pruning.

### Issues:
Closes: #1734

### DIFF
--- a/tests/unit/_autoscaling/test_snapshotter.py
+++ b/tests/unit/_autoscaling/test_snapshotter.py
@@ -242,7 +242,7 @@ async def test_snapshot_pruning_removes_outdated_records(
                 cpu_info=CpuInfo(used_ratio=0.5, created_at=now - timedelta(hours=delta)),
                 memory_info=default_memory_info,
             )
-            for delta in [3, 2, 5, 0]  # Out of order timestamps. Snapshotter can not rely on natural ordering.
+            for delta in [0, 3, 2, 5]  # Out of order timestamps. Snapshotter can not rely on natural ordering.
         ]
 
         for event_data in events_data:
--- a/src/crawlee/_autoscaling/snapshotter.py
+++ b/src/crawlee/_autoscaling/snapshotter.py
@@ -279,8 +279,8 @@ class Snapshotter:
         )
 
         snapshots = cast('list[Snapshot]', self._cpu_snapshots)
-        self._prune_snapshots(snapshots, event_data.cpu_info.created_at)
         self._cpu_snapshots.add(snapshot)
+        self._prune_snapshots(snapshots, self._cpu_snapshots[-1].created_at)
 
     async def _snapshot_memory(self, event_data: EventSystemInfoData) -> None:
         """Capture a snapshot of the current memory usage.
@@ -334,8 +334,8 @@ class Snapshotter:
         )
 
         snapshots = cast('list[Snapshot]', self._memory_snapshots)
-        self._prune_snapshots(snapshots, snapshot.created_at)
         self._memory_snapshots.add(snapshot)
+        self._prune_snapshots(snapshots, self._memory_snapshots[-1].created_at)
 
         self._evaluate_memory_load(
             event_data.memory_info.current_size,
@@ -361,8 +361,8 @@ class Snapshotter:
             snapshot.delay = event_loop_delay
 
         snapshots = cast('list[Snapshot]', self._event_loop_snapshots)
-        self._prune_snapshots(snapshots, snapshot.created_at)
         self._event_loop_snapshots.add(snapshot)
+        self._prune_snapshots(snapshots, self._event_loop_snapshots[-1].created_at)
 
     async def _snapshot_client(self) -> None:
         """Capture a snapshot of the current API state by checking for rate limit errors (HTTP 429).
@@ -385,8 +385,8 @@ class Snapshotter:
         )
 
         snapshots = cast('list[Snapshot]', self._client_snapshots)
-        self._prune_snapshots(snapshots, snapshot.created_at)
         self._client_snapshots.add(snapshot)
+        self._prune_snapshots(snapshots, self._client_snapshots[-1].created_at)
 
     def _prune_snapshots(self, snapshots: list[Snapshot], now: datetime) -> None:
         """Remove snapshots that are older than the `self._snapshot_history`.


Pijukatel:
@Mantisus , @vdusek, please check the root cause in the issue description. The suggested fix should be ok, but I fear it is too implicit and relies on `EventManager` non-obvious behavior. Would you be in favor of a more explicit(probably more complex) solution that would involve explicit threading locks that would prevent such a race condition?

Mantisus:
> @Mantisus, @vdusek, please check the root cause in the issue description. 

The root cause hypothesis seems incorrect. If the cause was an error processing in the 2-hour snapshot, then there would be 1 snapshot in the failed tests, not 3.
 
It might be worth trying 

```python
for event_data in events_data:
    event_manager.emit(event=Event.SYSTEM_INFO, event_data=event_data)
    await event_manager.wait_for_all_listeners_to_complete()
```

Perhaps the reason is that `event_manager` generates events out of chronological order.

Pijukatel:
> The root cause hypothesis seems incorrect.

Yes, the hypothesis in the issue description seems incorrect. I was referring to the root cause description in this PR

vdusek:
> Please check the root cause in the issue description. The suggested fix should be ok, but I fear it is too implicit and relies on EventManager non-obvious behavior. Would you be in favor of a more explicit(probably more complex) solution that would involve explicit threading locks that would prevent such a race condition?

I'd leave it up to you. If you believe the current approach is stable, I'm okay with it. Otherwise, let's implement the thread locks.

Mantisus:
> I was referring to the root cause description in this PR

Got it. Perhaps the race condition also affects this.

I hypothesize that when `event_manager.wait_for_all_listeners_to_complete()` is executed, all 5 events are executed simultaneously without any guarantee that they will be executed in the same order as they were created. If the result is [`now-3`, `now-2`, `now`, `now-5`], the last snapshot will not be deleted.
This is not a problem in normal work, because if this happened, it would be deleted when the next snapshot was recorded. But it is a problem for this test.

Pijukatel:
> > I was referring to the root cause description in this PR
> 
> Got it. Perhaps the race condition also affects this.
> 
> I hypothesize that when `event_manager.wait_for_all_listeners_to_complete()` is executed, all 5 events are executed simultaneously without any guarantee that they will be executed in the same order as they were created. If the result is [`now-3`, `now-2`, `now`, `now-5`], the last snapshot will not be deleted. This is not a problem in normal work, because if this happened, it would be deleted when the next snapshot was recorded. But it is a problem for this test.

Different insertion order is not a problem, as it should be handled by using `SortedSnapshotList` (which orders automatically when calling `add`), so we do not need to rely on the snapshots arriving time-ordered. The problem is that the same `SortedSnapshotList` can be in-place sorted/deleted in different threads. This is happening often in tests, but it can also happen in a real scenario, although it is extremely unlikely due to the usual time difference between different snapshots. But there is no guarantee that the snapshots will arrive "slowly enough" to guarantee processing of only one snapshot at a time.

Mantisus:
> Different insertion order is not a problem, as it should be handled by using SortedSnapshotList

The snapshot is added to `SortedSnapshotList` after `_prune_snapshots`.
https://github.com/apify/crawlee-python/blob/master/src/crawlee/_autoscaling/snapshotter.py#L272

Therefore, it is important which snapshot will be processed last.

In the next test, 0 snapshots will be deleted. This is because the last snapshot processed is `now-3`.
```python
async def test_without_pruning(
    snapshotter: Snapshotter, event_manager: LocalEventManager, default_memory_info: MemoryInfo
) -> None:
    snapshotter._SNAPSHOT_HISTORY = timedelta(hours=2)
    now = datetime.now(timezone.utc)

    snapshots = [
        CpuSnapshot(used_ratio=0.5, max_used_ratio=0.95, created_at=now - timedelta(hours=delta)) for delta in [5, 2, 0]
    ]

    for snapshot in snapshots: # create `SortedSnapshotList` with 3 elems
        snapshotter._cpu_snapshots.add(snapshot)

    event_manager.emit(event=Event.SYSTEM_INFO, event_data=EventSystemInfoData(
        cpu_info=CpuInfo(used_ratio=0.5, created_at=now - timedelta(hours=3)),
        memory_info=default_memory_info,
    ))

    await event_manager.wait_for_all_listeners_to_complete()

    cpu_snapshots = cast('list[CpuSnapshot]', snapshotter.get_cpu_sample())

    assert len(cpu_snapshots) == 4 # Not one was removed
```

Pijukatel:
@Mantisus good point! 

To ensure that the `Different insertion order is not a problem`, we need to run the `_snapshot_cpu` uninterrupted (without anyone else modifying the `self._cpu_snapshots` ). That will guarantee the assumption that pruning is happening on a sorted list.

tests/unit/_autoscaling/test_snapshotter.py
```diff
@@ -231,17 +231,23 @@ async def test_snapshot_pruning_removes_outdated_records(
...
-    for event_data in events_data:
-        event_manager.emit(event=Event.SYSTEM_INFO, event_data=event_data)
-    await event_manager.wait_for_all_listeners_to_complete()
+    with mock.patch('crawlee._autoscaling.snapshotter.insort', side_effect=randomly_delayed_insort):
+        events_data = [
+            EventSystemInfoData(
+                cpu_info=CpuInfo(used_ratio=0.5, created_at=now - timedelta(hours=delta)),
+                memory_info=default_memory_info,
+            )
+            for delta in [3, 2, 5, 0]  # Out of order timestamps. Snapshotter can not rely on natural ordering.
```
Pijukatel: @Mantisus  I meant that this order should not matter

Pijukatel:
Updated to handle the scenario mentioned by @Mantisus when the last snapshot is out of order.

vdusek (approved):
LGTM

PR #1751, opened 2026-02-16
docs: Update docs about alias storages to match the implementation

Pijukatel:
### Description

- Current implementation does not give `alias='default'` any special meaning.
- Unnamed default storage should be accessed without mentioning the alias.


Pijukatel:
> When did we change this?

https://github.com/apify/crawlee-python/pull/1386#discussion_r2351724370

Issue #1396, opened 2025-09-05
Add support for crawl-delay robots.txt directive


vdusek:
https://en.wikipedia.org/wiki/Robots.txt#Crawl-delay_directive

janbuchar:
This is tricky due to the nature of the Apify request queue which we need to support. It is not possible to set a "don't process this before <timestamp>" attribute there, so you cannot check the URL of a freshly dequeued request and re-insert it at the start of the queue with this set up. Also, you cannot consume requests from the queue selectively by domain.

An approach that I already suggested for the JS version is to make a separate non-default unnamed queue for each domain where we want to allow a delay. Then we can implement the "cooldown period" on the request queue client so that it only yields a request once every N seconds. Finally, we'll wrap these request queue clients with one that will check all of them for requests ready for processing and redirect new requests to appropriate queues.

vdusek:
Moving to the icebox for now, as the Cloudflare Pay Per Crawl is still in private beta, let's see how it evolves.

PR #1762, opened 2026-02-20
feat: Add opt-in per-domain request throttling for HTTP 429 backoff

### LINKED ISSUE #1437: Fix autoscaled pool scaling behavior on 429 Too Many Requests
### Description

- Crawlee does not currently handle 429 Too Many Requests responses correctly.
- When a target server starts returning 429s, Crawlee does not slow down.
- Instead, due to the current autoscaled pool logic, Crawlee may actually scale concurrency up when responses get slower (because of less CPU work).
- This creates a "death spiral" - the slower the server, the faster Crawlee increases concurrency, which can quickly overwhelm small websites.

### Proposed solution

- Detect 429 responses and implement proper backoff logic (reducing concurrency of autoscaled pool, cooldown period, ...).
- Ensure the autoscaled pool does not interpret slow responses or 429s as a signal to increase concurrency.
- Consider respecting `Retry-After` headers if present.

### References

- This was originally discussed on Slack https://apify.slack.com/archives/CD0SF6KD4/p1756993901117969.

MrAliHasan:
### Problem

When target websites return HTTP 429 (Too Many Requests), requests get retried without any **per-domain delay** — potentially making rate limiting worse.

### Solution

Introduces the `ThrottlingRequestManager`, an **opt-in** request manager wrapper that enforces per-domain delays at the scheduling layer.

> **Note:** Users must explicitly pass a `ThrottlingRequestManager` as the `request_manager` to enable throttling. There is no auto-wrapping or implicit behavior change.

**Key features:**

- **Per-domain sub-queues** — requests for configured domains are routed to dedicated sub-queues at insertion time
- **HTTP 429 backoff** — `record_domain_delay()` sets a per-domain `throttled_until` timestamp based on `Retry-After` headers
- **robots.txt crawl-delay** — `BasicCrawler` automatically calls `set_crawl_delay()` when `respect_robots_txt_file` is enabled and the request manager is a `ThrottlingRequestManager`
- **Crawl-delay warning** — logs a warning if `respect_robots_txt_file` is enabled but the request manager is **not** a `ThrottlingRequestManager`
- **Delay-aware scheduling** — `fetch_next_request()` skips throttled domains, falls back to the inner queue, and sleeps only when all sub-queues are throttled and the inner queue is empty
- **Persistence support** — sub-queues use the same storage backend as the inner queue; `recreate_purged()` handles queue reconstruction across crawler restarts

### How it works

1. Requests are routed to **domain-specific sub-queues** at insertion time
2. If a domain is throttled (`throttled_until`), `fetch_next_request()` skips it and falls back to the inner queue
3. `record_domain_delay()` updates per-domain backoff on HTTP 429 responses, respecting `Retry-After` headers
4. `set_crawl_delay()` integrates robots.txt crawl-delay when enabled
5. On successful requests, backoff counters reset

### Usage

```python
from crawlee.request_loaders import ThrottlingRequestManager
from crawlee.crawlers import BasicCrawler

queue = await RequestQueue.open()
manager = ThrottlingRequestManager(
    queue,
    domains=["example.com", "api.example.com"]
)
crawler = BasicCrawler(request_manager=manager)
```

### Files changed

| File                                   | Change                                                                                  |
| -------------------------------------- | --------------------------------------------------------------------------------------- |
| `_throttling_request_manager.py`     | **NEW** — Per-domain throttling request manager                                  |
| `http.py`                            | `parse_retry_after_header` utility                                                    |
| `_basic_crawler.py`                  | `recreate_purged()` integration, crawl-delay warning                                  |
| `_playwright_crawler.py`             | Pass `Retry-After` header                                                             |
| `test_throttling_request_manager.py` | **NEW** — 30 unit tests using real `RequestQueue` with `MemoryStorageClient` |

### Tests

- **30 new tests** covering: domain routing, throttling, delay scheduling, sleep behavior, delegation, `recreate_purged()`, and edge cases
- **All 1648 existing tests** pass with zero regressions

### Future work

This is a focused first step toward a more complete `RequestAnalyzer` that may include:

- robots.txt integration for multiple domains
- URL group management
- Enhanced per-domain scheduling and analytics


vdusek:
Hi @MrAliHasan, thanks for your contribution! We'll try to review this soon.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1542,22 +1556,74 @@ def _raise_for_error_status_code(self, status_code: int) -> None:
...
+            self._request_throttler.record_rate_limit(url, retry_after=retry_after)
+
         if session is not None and session.is_blocked_status_code(
             status_code=status_code,
             ignore_http_error_status_codes=self._ignore_http_error_status_codes,
         ):
             raise SessionError(f'Assuming the session is blocked based on HTTP status code {status_code}')
 
+    @staticmethod
+    def _parse_retry_after_header(value: str | None) -> timedelta | None:
```
janbuchar: This has no business being in `BasicCrawler`. Better put it in the `_utils` module.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1542,22 +1556,74 @@ def _raise_for_error_status_code(self, status_code: int) -> None:
         if is_status_code_server_error(status_code) and not is_ignored_status:
             raise HttpStatusCodeError('Error status code returned', status_code)
 
-    def _raise_for_session_blocked_status_code(self, session: Session | None, status_code: int) -> None:
+    def _raise_for_session_blocked_status_code(
+        self,
+        session: Session | None,
+        status_code: int,
```
janbuchar, line 1549: Maybe this could just receive the entire `HttpResponse` object?

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1396,6 +1398,15 @@ async def __run_task_function(self) -> None:
...
             return
 
+        # Check if this domain is currently rate-limited (429 backoff).
+        if self._request_throttler.is_throttled(request.url):
+            self._logger.debug(
+                f'Request to {request.url} delayed - domain is rate-limited '
+                f'(retry in {self._request_throttler.get_delay(request.url).total_seconds():.1f}s)'
+            )
+            await request_manager.reclaim_request(request)
+            return
```
janbuchar: If, at some point, the request queue contains **only** requests from a throttled domain, this will become a busy wait with extra steps. If you're using the Apify platform, this will cost a lot in request queue writes.

I'm afraid that this means we cannot accept the PR in the current state. See the main review comments for possible next steps.

janbuchar (changes requested):
As mentioned in https://github.com/apify/crawlee-python/pull/1762#discussion_r2841515928, the approach of reclaiming throttled requests is not optimal.

On top of that, the solution to #1437 should probably also be extensible enough to also cover #1396 without much tweaking.

I believe that such solution could be implemented in crawlee-python quite easily. See [similar issue](https://github.com/apify/crawlee/issues/3148#issuecomment-3262126226) for crawlee-js. The Python version already supports multiple "unnamed queues" via `RequestQueue.open(alias="...")`, so you'd only need to implement a `ThrottlingRequestManager` (implementation of the `RequestManager` interface) that would keep track of the per-domain queues and their delays.

Do you want to try it?

MrAliHasan:
Thanks for the detailed review. That makes sense regarding the busy-wait behavior and queue writes.
I’ll refactor this into a ThrottlingRequestManager implementation so that the throttling logic lives in the request scheduling layer rather than in BasicCrawler.
I’ll push an updated version soon. Appreciate the guidance.

src/crawlee/crawlers/_basic/_basic_crawler.py
MrAliHasan: Moved to crawlee._utils.http.parse_retry_after_header in the refactor commit.

src/crawlee/crawlers/_basic/_basic_crawler.py
MrAliHasan: I considered this, but I kept the current signature. AbstractHttpCrawler passes Crawlee’s HttpResponse while PlaywrightCrawler passes Playwright’s native Response, which are different types. Since the method only needs status_code, url, and Retry-After, extracting those at the call site keeps the abstraction simpler. Happy to revisit if you’d prefer a unified response layer.

src/crawlee/crawlers/_basic/_basic_crawler.py
MrAliHasan: Fully addressed in the refactor. The reclaim-based throttle block was removed. ThrottlingRequestManager.fetch_next_request() now handles scheduling and awaits asyncio.sleep() when all domains are throttled, eliminating busy-wait and extra queue writes.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1442,6 +1451,10 @@ async def __run_task_function(self) -> None:
 
             await self._mark_request_as_handled(request)
 
+            # Record successful request to reset rate limit backoff for this domain.
+            if self._throttling_manager:
+                self._throttling_manager.record_success(request.url)
```
janbuchar: I'd rather use `isinstance` so that an explicitly configured `ThrottlingRequestManager` also works.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1542,22 +1555,41 @@ def _raise_for_error_status_code(self, status_code: int) -> None:
...
+    ) -> None:
         """Raise an exception if the given status code indicates the session is blocked.
 
+        If the status code is 429 (Too Many Requests), the domain is recorded as
+        rate-limited in the `RequestThrottler` for per-domain backoff.
+
         Args:
             session: The session used for the request. If None, no check is performed.
             status_code: The HTTP status code to check.
+            url: The request URL, used for per-domain rate limit tracking.
```
janbuchar: Perhaps the parameter should be renamed to `request_url` so that it's not ambiguous with a URL after redirects.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1542,22 +1555,41 @@ def _raise_for_error_status_code(self, status_code: int) -> None:
...
+        session: Session | None,
+        status_code: int,
+        *,
+        url: str = '',
+        retry_after_header: str | None = None,
+    ) -> None:
         """Raise an exception if the given status code indicates the session is blocked.
 
+        If the status code is 429 (Too Many Requests), the domain is recorded as
+        rate-limited in the `RequestThrottler` for per-domain backoff.
```
janbuchar: Outdated name

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1542,22 +1555,41 @@ def _raise_for_error_status_code(self, status_code: int) -> None:
...
+                self._throttling_manager.record_domain_delay(url, retry_after=retry_after)
+
         if session is not None and session.is_blocked_status_code(
             status_code=status_code,
             ignore_http_error_status_codes=self._ignore_http_error_status_codes,
         ):
             raise SessionError(f'Assuming the session is blocked based on HTTP status code {status_code}')
 
+    # NOTE: _parse_retry_after_header has been moved to crawlee._utils.http.parse_retry_after_header
+
```
janbuchar:
```suggestion
```

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1582,7 +1614,16 @@ async def _is_allowed_based_on_robots_txt_file(self, url: str) -> bool:
         if not self._respect_robots_txt_file:
             return True
         robots_txt_file = await self._get_robots_txt_file_for_url(url)
-        return not robots_txt_file or robots_txt_file.is_allowed(url)
+        if not robots_txt_file:
+            return True
+
+        # Wire robots.txt crawl-delay into ThrottlingRequestManager (#1396).
```
janbuchar:
```suggestion
        # Wire robots.txt crawl-delay into ThrottlingRequestManager

File: src/crawlee/crawlers/_basic/_basic_crawler.py [hunk 9]
```diff
@@ -611,12 +614,18 @@ async def _get_proxy_info(self, request: Request, session: Session | None) -> Pr
...
+        per-domain delays from 429 responses and robots.txt crawl-delay directives.
+        """
         if not self._request_manager:
-            self._request_manager = await RequestQueue.open(
+            inner = await RequestQueue.open(
                 storage_client=self._service_locator.get_storage_client(),
                 configuration=self._service_locator.get_configuration(),
             )
+            self._throttling_manager = ThrottlingRequestManager(inner)
+            self._request_manager = self._throttling_manager
```
janbuchar: I'm not sure if we should use `ThrottlingRequestManager` by default - thoughts, @vdusek @Pijukatel, @Mantisus?

tests/unit/test_throttling_request_manager.py
janbuchar, line 1: Please make the tests conform to the existing test structure. Perhaps @vdusek could provide further guidance here?

src/crawlee/request_loaders/_throttling_request_manager.py
```diff
@@ -0,0 +1,340 @@
+"""A request manager wrapper that enforces per-domain delays.
+
+Handles both HTTP 429 backoff and robots.txt crawl-delay at the scheduling layer,
+eliminating the busy-wait problem described in https://github.com/apify/crawlee-python/issues/1437.
+
+Also addresses https://github.com/apify/crawlee-python/issues/1396 by providing a unified
+delay mechanism for crawl-delay directives.
+"""
```
janbuchar: No need to mention github issues here if they will be resolved by this PR.

src/crawlee/request_loaders/_throttling_request_manager.py
```diff
@@ -0,0 +1,340 @@
...
+
+    def __init__(self, inner: RequestManager) -> None:
+        """Initialize the throttling manager.
+
+        Args:
+            inner: The underlying request manager to wrap (typically a RequestQueue).
+        """
+        self._inner = inner
+        self._domain_states: dict[str, _DomainState] = {}
+        self._buffered_requests: list[Request] = []
```
janbuchar: This is a fundamental deviation from what I described - instead of storing the requests in memory, we should create a special request queue in memory (see https://github.com/apify/crawlee-python/pull/1762#pullrequestreview-3841605220) for each domain. The `ThrottlingRequestManager` will then delegate `fetch_next_request` calls to sub-queues that are not currently delayed.

janbuchar (changes requested):

src/crawlee/crawlers/_basic/_basic_crawler.py
Mantisus: I think it makes sense to enable it by default. When the crawler runs without a proxy, 429 will only increase the load on the site. This will not benefit either the site or the crawler.
 
Using a proxy requires a little configuration, so I think that an additional parameter to disable throttling for 429 will not complicate this.

MrAliHasan:
Heads up @janbuchar @vdusek  @Mantisus: I've pushed a significant refactor based on the latest feedback.

Sub-queues over memory buffer: ThrottlingRequestManager now delegates to persistent per-domain sub-queues via RequestQueue.open(alias=f"throttled-{domain}") instead of keeping throttled requests in memory.

Test Structure: Completely rewrote test_throttling_request_manager.py to drop the Test... classes and conform to Crawlee's standard test structure.

BasicCrawler fixes: Addressed all inline nits (used isinstance(), renamed url to request_url in _raise_for_session_blocked_status_code, updated docstrings/comments).

The tests track the routing origin and safely aggregate get_handled_count and is_empty metrics across the main queue and sub-queues. All 24 tests pass, and Ruff and Pytest issues have been resolved. Let me know if the updated delegation architecture feels right!

src/crawlee/request_loaders/_throttling_request_manager.py
MrAliHasan: You're absolutely right. This has been fully refactored. ThrottlingRequestManager now dynamically instantiates RequestQueue.open(alias=f"throttled-{domain}") for throttled domains. Requests are routed to these separate sub-queues to avoid memory pressure, and fetch_next_request pulls from them seamlessly when domain delays expire. State operations (like is_empty and get_handled_count) now aggregate cleanly across the inner queue and all sub-queues. This aligns with Crawlee's distributed design and ensures request persistence.

src/crawlee/crawlers/_basic/_basic_crawler.py
MrAliHasan: Fixed! Updated to check isinstance(manager, ThrottlingRequestManager) accurately.

src/crawlee/crawlers/_basic/_basic_crawler.py
MrAliHasan: Done. Renamed to request_url across BasicCrawler, AbstractHttpCrawler, and PlaywrightCrawler.

tests/unit/test_throttling_request_manager.py
MrAliHasan: Good call! I've removed all the Test... classes and rewritten the entirety of test_throttling_request_manager.py to use standard top-level async def test_...() functions with pytest fixtures to match test_request_list.py.

MrAliHasan:
Update: I just pushed a small follow-up commit fixing the MyPy typing and Ruff linting errors in the test suite that were causing the CI to fail. All local checks for ThrottlingRequestManager are now passing 100%. Ready for review whenever you have time!

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -707,12 +714,21 @@ async def run(
...
+                        configuration=self._service_locator.get_configuration(),
+                    )
+                elif isinstance(request_manager, ThrottlingRequestManager):
+                    await request_manager.drop()
+                    inner = await RequestQueue.open(
+                        storage_client=self._service_locator.get_storage_client(),
+                        configuration=self._service_locator.get_configuration(),
+                    )
+                    self._throttling_manager = ThrottlingRequestManager(inner)
+                    self._request_manager = self._throttling_manager
```
Pijukatel:
```suggestion
                    self._request_manager = ThrottlingRequestManager(inner)
```

src/crawlee/request_loaders/_throttling_request_manager.py
```diff
@@ -0,0 +1,356 @@
...
+    @staticmethod
+    def _extract_domain(url: str) -> str:
+        """Extract the domain (hostname) from a URL."""
+        parsed = urlparse(url)
+        return parsed.hostname or ''
+
+    async def _get_or_create_sub_queue(self, domain: str) -> RequestQueue:
+        """Get or create a per-domain sub-queue."""
+        if domain not in self._sub_queues:
+            self._sub_queues[domain] = await RequestQueue.open(alias=f'throttled-{domain}')
```
Pijukatel: We should think this through. Calling `self._sub_queues[domain]=await RequestQueue.open(alias=f'throttled-{domain}')` will use the global `service_locator` to get the configuration and storage client that will be used to create this RQ. On the other hand, inner was created from crawler specific `service-locator`:
```
inner = await RequestQueue.open(
                storage_client=self._service_locator.get_storage_client(),
                configuration=self._service_locator.get_configuration(),
            )
```

So this could lead to unexpected behavior when these two service locators are not the same and thus they could use different `configuration` and different `storage_client`.

What are other options?

### Hard-code `MemoryStorageClient`
Hard-code `MemoryStorageClient` for all `self._sub_queues[domain]` and use crawler-specific configuration. This would probably work for the majority of scenarios. It would be fast and cheap. But what about a heavily throttled crawler? Imagine a crawler that is crawling only one site, has massive RQ, and that one site has `crawl_delay` 60s. In such a scenario, the massive RQ would just be loaded to one of the `self._sub_queues` until the memory limit of the crawler is used. Would it then stay in a deadlock due to memory load? 

### Use crawler-specific `service_locator` as init argument to `ThrottlingRequestManager` and create all `self._sub_queues` like this 
```
self._sub_queues[domain]=await RequestQueue.open(
    alias=f'throttled-{domain}',
    storage_client=self._service_locator.get_storage_client(),
    configuration=self._service_locator.get_configuration(),
)
```
This seems to me like a safe choice, but it would probably not be the cheapest and fastest. For example, on the Apify platform, it would use Apify-based RQ, which is more expensive and slower than an in-memory one.

### Use in-memory as default, but allow custom `service_locator`?
I would probably prefer this approach. By default, no service locator would be passed to the `ThrottlingRequestManager` and it would use `MemoryStorageClient` and for special usecase you could use:

`BasicCrawler(request_manager=ThrottlingRequestManager(service_loactor=custom_service_locator))`

### Use in-memory as default, but limit the max size of the in-memory self._sub_queues?

src/crawlee/request_loaders/_throttling_request_manager.py
```diff
@@ -0,0 +1,356 @@
...
+        origin = self._dispatched_origins.get(request.unique_key)
+        if origin and origin != 'inner' and origin in self._sub_queues:
+            return await self._sub_queues[origin].reclaim_request(request, forefront=forefront)
+        return await self._inner.reclaim_request(request, forefront=forefront)
+
+    @override
+    async def mark_request_as_handled(self, request: Request) -> ProcessedRequest | None:
+        origin = self._dispatched_origins.get(request.unique_key)
+        if origin and origin != 'inner' and origin in self._sub_queues:
+            return await self._sub_queues[origin].mark_request_as_handled(request)
```
Pijukatel: Ideally, I think we do not want to just mark the request in `self._sub_queues` as handled, but completely delete it from the subqueue. There is no point in tracking such a request in a subqueue, as it can consume memory in some implementations. 
It is enough that it is marked as handled in the `inner`. That one already handles all the deduplication needed and that is the only place where we need to store handled requests.

But RQ does not define a delete method so not sure what to do with this...

src/crawlee/request_loaders/_throttling_request_manager.py
```diff
@@ -0,0 +1,356 @@
...
+        it sleeps until the earliest domain becomes available.
+        """
+        # First, check if any sub-queues have unthrottled requests.
+        for domain, sq in self._sub_queues.items():
+            if not self._is_domain_throttled(domain):
+                req = await sq.fetch_next_request()
+                if req:
+                    self._mark_domain_dispatched(req.url)
+                    self._dispatched_origins[req.unique_key] = domain
+                    return req
```
Pijukatel: Imagine a scenario where multiple domains are throttled at the same time. But the first of `self._sub_queues` has a massive amount of requests. The current approach would prefer such subqueu just due to the order in the `self._sub_queues`.

I think that the mechanism that decides which subqueue should be used to `fetch_next_request` should be based on the lowest time in `_DomainState.throttled_until`. So fetch from the longest overdue one.

Which brings me to another point. Do we even need to track `_DomainState.last_request_at`? I think we should just track `_DomainState.throttled_until` and update it on `ThrottlingRequestManager._mark_domain_dispatched` call.

Pijukatel (commented):
Hello, thanks for the work on this PR! I have just some annoying edgecase to think about. I am not sure myself what the best way is to deal with them.

MrAliHasan:
Hey @Pijukatel, thanks for looking deeply into this! Great catches all around on the edge cases. I spent some time analyzing these points, and here is how I propose we handle them. Let me know if you are aligned on this direction before I push the changes:

**1. `BasicCrawler` initialization**

Good catch. I'll simplify `_basic_crawler.py` to just directly assign `self._request_manager = ThrottlingRequestManager(inner)`.

**2. Sub-Queue storage strategy**

Regarding the `service_locator` mismatch: you are absolutely right. If a crawler relies on persistent storage (e.g. Apify Platform) and we silently default the sub-queues to a fast `MemoryStorageClient`, an interrupted crawl will lose its throttled queues entirely on restart. We would be trading determinism and correctness for raw performance.

Therefore, I propose your "Option 2" as the default, but with the flexibility of "Option 3":

- `ThrottlingRequestManager` will accept an optional `service_locator`.
- By default, `BasicCrawler` will pass its own `self._service_locator` down to the throttler so that sub-queues consistently match the persistence mechanism of the main queue.
- If a power user really wants pure in-memory sub-queues for cost/speed reasons despite a persistent main queue, they can explicitly pass a custom `ServiceLocator` into the throttler instance.

**3. Fetch priority redesign**

I completely agree. Relying on iteration order rather than the longest-overdue domain is a flaw in the scheduling logic.

- I will remove `_DomainState.last_request_at` entirely. Less state means fewer edge cases.
- On dispatch, we simply update `_DomainState.throttled_until` immediately.
- `fetch_next_request` will collect available sub-queues and explicitly sort them by `throttled_until`, so the domain that has been waiting the longest gets fetched first.

**4. Deleting Handled Requests**

Because `RequestQueue` does not currently support `.delete()`, we can't physically purge them without API additions. However, since the sub-queues only act as temporary routing buffers, this shouldn't be a problem. The source of truth remains the `inner` queue. The handled tracking duplication in the sub-queue doesn't affect the core deduplication logic at all.

Does this direction look solid to you? If so, I'll get it coded and pushed up!

Pijukatel:
...
> 
> Does this direction look solid to you? If so, I'll get it coded and pushed up!

We discussed it internally, and there are some open points. Please let me think about it over the weekend, so that I don't point you in the wrong direction.

Pijukatel:
...
> Does this direction look solid to you? If so, I'll get it coded and pushed up!

It is going in a good direction. Just a few more edgecases we discussed:

### Having ad hoc request queues being created for domains can lead to two undesired scenarios:
- If redirecting new requests to such queues, then deduplication is difficult (would have to check two queues)
- If not redirecting new requests to such queues, then requests are duplicated, and there is a lot of work done by just duplicating specific domain requests from the main queue to the specialized queue.

To deal with this, we agreed it would be best to have the `RequestThrottler` be created with specific domains as init arguments. It should be redirecting new requests to such explicitly requested sub-queues. For domains that are not explicitly mentioned in init, it would not do any request throttling.

### Preserve existing behavior and introduce `RequestThrottler` in a  safe way

It would be best if we start with this feature being optional for now, until we get more usage feedback. So `RequestThrottler` would have to be explicitly passed to the `BasicCrawler`. Could you please add a short guide on how to do it and a small code example? (See: https://github.com/apify/crawlee-python/tree/master/docs/guides)

MrAliHasan:
Hey @Pijukatel, I've pushed the refactor based on your latest feedback. Here's what changed:

**Explicit domain routing**
- `ThrottlingRequestManager` now requires `domains` at init — only listed domains are throttled.
- `add_request` / `add_requests` route listed-domain requests directly to their sub-queue at insertion time. No duplication, no transfers.
- Non-listed domains pass through to the main queue untouched.
- Sub-queues use the same storage backend as the inner queue by default to preserve persistence across restarts.

**Opt-in only**
- Removed auto-wrapping from `BasicCrawler.get_request_manager()`. The feature must be explicitly enabled by passing a `ThrottlingRequestManager` as `request_manager`.

**Simplified state**
- Removed `_DomainState.last_request_at`. Only `throttled_until` is tracked, updated on dispatch.
- Eliminated `_dispatched_origins` and `_transferred_requests_count` — no longer needed since requests start in the right queue.
- `fetch_next_request` sorts sub-queues by `throttled_until` ascending (longest-overdue domain first).

**Documentation**
- Added `docs/guides/request_throttling.mdx` with a short guide and code example showing explicit usage.

All local checks pass (1647 tests, 0 failures). Ready for review!

src/crawlee/_utils/http.py
```diff
@@ -0,0 +1,41 @@
...
+
+    # Try parsing as integer seconds first.
+    try:
+        seconds = int(value)
+        return timedelta(seconds=seconds)
+    except ValueError:
+        pass
+
+    # Try parsing as HTTP-date (e.g., "Wed, 21 Oct 2015 07:28:00 GMT").
+    from email.utils import parsedate_to_datetime  # noqa: PLC0415
```
Pijukatel: Is there any specific reason for the import not being at the top of the file?

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -707,12 +709,23 @@ async def run(
...
+                elif isinstance(request_manager, ThrottlingRequestManager):
+                    domains = list(request_manager._domains)  # noqa: SLF001
+                    await request_manager.drop()
+                    inner = await RequestQueue.open(
+                        storage_client=self._service_locator.get_storage_client(),
+                        configuration=self._service_locator.get_configuration(),
+                    )
+                    self._request_manager = ThrottlingRequestManager(
+                        inner, domains=domains, service_locator=self._service_locator
+                    )
```
Pijukatel: Now when the `ThrottlingRequestManager` is constructed outside of the Crawler, we do not have enough information to reconstruct it here. 

In order to reconstruct it properly, I think that the `ThrottlingRequestManager` should have some dedicated method specifically for this. It should also track its own `_service_locator` and then it can use own `.service_locator` and also `._domains` to return `ThrottlingRequestManager` instance with all the internal request queues emptied again.

src/crawlee/crawlers/_playwright/_playwright_crawler.py
```diff
@@ -459,7 +459,13 @@ async def _handle_status_code_response(
         """
         status_code = context.response.status
         if self._retry_on_blocked:
-            self._raise_for_session_blocked_status_code(context.session, status_code)
+            retry_after_header = context.response.headers.get('retry-after')
+            self._raise_for_session_blocked_status_code(
+                context.session,
+                status_code,
+                request_url=context.request.url,
+                retry_after_header=retry_after_header,
```
Pijukatel: Maybe just pass directly `context.response.headers.get('retry-after')` without the `retry_after_header` variable

src/crawlee/request_loaders/_throttling_request_manager.py
```diff
@@ -0,0 +1,407 @@
...
+class ThrottlingRequestManager(RequestManager):
+    """A request manager that wraps another and enforces per-domain delays.
+
+    Requests for explicitly configured domains are routed into dedicated sub-queues
+    at insertion time — each request lives in exactly one queue, eliminating
+    duplication and simplifying deduplication.
+
+    When `fetch_next_request()` is called, it returns requests from the sub-queue
+    whose domain has been waiting the longest. If all configured domains are
+    throttled, it sleeps until the earliest cooldown expires.
```
Pijukatel: `If all configured domains are throttled` , then it should fetch from the inner or?

src/crawlee/request_loaders/_throttling_request_manager.py
```diff
@@ -0,0 +1,407 @@
...
+            inner: The underlying request manager to wrap (typically a RequestQueue).
+                Requests for non-throttled domains are stored here.
+            domains: Explicit list of domain hostnames to throttle. Only requests
+                matching these domains will be routed to per-domain sub-queues.
+            service_locator: Optional service locator for creating sub-queues.
+                If provided, sub-queues will use its storage client and configuration,
+                ensuring consistency with the crawler's storage backend.
+        """
+        self._inner = inner
+        self._domains = set(domains)
```
Pijukatel: Do we need to track `._domains`? It feels a little redundant sicne we use it as keys both in `self._domain_states` and `self._sub_queues` 

Maybe `self._domain_states` can be fully populated on init and set `state.throttled_until` to now?

src/crawlee/request_loaders/_throttling_request_manager.py
```diff
@@ -0,0 +1,407 @@
...
+        self._service_locator = service_locator
```
Pijukatel: Maybe it would be simpler to just use global `service_locator` if `None` is passed explicitly. 

(Otherwise, it relies on it as well, but it is implicit and we need to check for `None` later in code)

src/crawlee/request_loaders/_throttling_request_manager.py
```diff
@@ -0,0 +1,407 @@
...
+    def _get_url_from_request(self, request: str | Request) -> str:
+        """Extract URL string from a request that may be a string or Request object."""
+        if isinstance(request, str):
+            return request
+        return request.url
+
+    async def _get_or_create_sub_queue(self, domain: str) -> RequestQueue:
+        """Get or create a per-domain sub-queue."""
+        if domain not in self._sub_queues:
+            if self._service_locator:
```
Pijukatel: See the init comment. I think it will be simpler to always have `_service_locator`

src/crawlee/request_loaders/_throttling_request_manager.py
```diff
@@ -0,0 +1,407 @@
...
+                    configuration=self._service_locator.get_configuration(),
+                )
+            else:
+                self._sub_queues[domain] = await RequestQueue.open(alias=f'throttled-{domain}')
+        return self._sub_queues[domain]
+
+    def _is_domain_throttled(self, domain: str) -> bool:
+        """Check if a domain is currently throttled."""
+        state = self._domain_states.get(domain)
+        if state is None:
```
Pijukatel: Is there any point where we would like this to be `None`? I would suggest to add all `_domain_states` on init.

src/crawlee/request_loaders/_throttling_request_manager.py
```diff
@@ -0,0 +1,407 @@
...
+    def set_crawl_delay(self, url: str, delay_seconds: int) -> None:
+        """Set the robots.txt crawl-delay for a domain.
+
+        Args:
+            url: A URL from the domain to throttle.
+            delay_seconds: The crawl-delay value in seconds.
+        """
+        domain = self._extract_domain(url)
+        if not domain:
+            return
```
Pijukatel: Maybe use the same early return in `record_success` to be consistent

src/crawlee/request_loaders/_throttling_request_manager.py
```diff
@@ -0,0 +1,407 @@
...
+
+        if domain not in self._domain_states:
```
Pijukatel: We can get rid of this, if all `_domain_states` are properly set on init

src/crawlee/request_loaders/_throttling_request_manager.py
```diff
@@ -0,0 +1,407 @@
...
+    def _mark_domain_dispatched(self, url: str) -> None:
+        """Record that a request to this domain was just dispatched.
+
+        If a crawl-delay is configured, push throttled_until forward by that amount.
+        """
+        domain = self._extract_domain(url)
+        if not domain:
+            return
+
+        if domain not in self._domain_states:
```
Pijukatel: We can get rid of this, if all _domain_states are properly set on init

src/crawlee/request_loaders/_throttling_request_manager.py
```diff
@@ -0,0 +1,407 @@
...
+            self._domain_states[domain] = _DomainState(domain=domain)
+        state = self._domain_states[domain]
+
+        # If crawl-delay is set, enforce minimum interval by pushing throttled_until.
+        if state.crawl_delay is not None:
+            state.throttled_until = datetime.now(timezone.utc) + state.crawl_delay
+
+    # ──────────────────────────────────────────────────────
+    # RequestManager interface delegation + smart scheduling
+    # ──────────────────────────────────────────────────────
```
Pijukatel: Maybe there is no need for this visually strong comment since all methods have good docstrings

src/crawlee/request_loaders/_throttling_request_manager.py
```diff
@@ -0,0 +1,407 @@
...
+        return True
+
+    @override
+    async def fetch_next_request(self) -> Request | None:
+        """Fetch the next request, respecting per-domain delays.
+
+        Sub-queues are checked in order of longest-overdue domain first
+        (sorted by `throttled_until` ascending). If all configured domains are
+        throttled, falls back to the inner queue. If everything is throttled,
+        sleeps until the earliest domain becomes available.
```
Pijukatel: `If everything is throttled` this happens only if the inner is empty and all subqueues are throttled, right?

tests/unit/test_throttling_request_manager.py
```diff
@@ -0,0 +1,471 @@
...
+from crawlee._utils.http import parse_retry_after_header
+from crawlee.request_loaders._throttling_request_manager import ThrottlingRequestManager
+
+THROTTLED_DOMAIN = 'throttled.com'
+NON_THROTTLED_DOMAIN = 'free.com'
+TEST_DOMAINS = [THROTTLED_DOMAIN]
+
+
+@pytest.fixture
+def mock_inner() -> AsyncMock:
```
Pijukatel: Instead of mocking the RQ, I think you can use RQ created with `MemoryStorageClient`, the test will be easier to write and read.

Also it can be used in `ThrottlingRequestManager._service_locator` for creation of subqueues

Pijukatel (commented):
Nice.
Now I have just a few more code-related comments.

MrAliHasan:
Hey @Pijukatel, I've pushed all the changes from your latest review. Here's what was addressed:

**Import cleanup**

- Moved `parsedate_to_datetime` import to the top of `http.py`.
- Inlined `retry_after_header` variable in `_playwright_crawler.py`.

**`ThrottlingRequestManager` simplifications**

- Removed redundant `_domains` set. All `_domain_states` are now fully populated at init with `throttled_until` set to `now`.
- `_service_locator` now defaults to the global `service_locator` when not explicitly provided, eliminating all `if self._service_locator` branches in `_get_or_create_sub_queue` and `recreate_purged`.
- Added `if not domain: return` guard in `record_success` for consistency with `set_crawl_delay` and `_mark_domain_dispatched`.
- Removed `if domain not in self._domain_states` guards in `record_domain_delay`, `set_crawl_delay`, and `_mark_domain_dispatched` since all states are initialized at init.
- Removed the `# ──────` separator comment block.

**Dedicated purge method**

- Added `recreate_purged()` method on `ThrottlingRequestManager` that drops all queues and returns a fresh instance with the same domain configuration and service locator. `BasicCrawler` now calls this directly instead of accessing private attributes.

**Docstrings**

- Clarified that `fetch_next_request` falls back to the inner queue when all sub-queues are throttled, and only sleeps when the inner queue is also empty.

**Tests**

- Replaced all `AsyncMock` queue mocks with real `RequestQueue` backed by `MemoryStorageClient`, which also provides the `service_locator` for sub-queue creation.
- Added test for `recreate_purged`.

All checks pass (1648 tests, 0 failures). Ready for review!

src/crawlee/request_loaders/_throttling_request_manager.py
```diff
@@ -0,0 +1,428 @@
...
+
+    When `fetch_next_request()` is called, it returns requests from the sub-queue
+    whose domain has been waiting the longest. If all configured domains are
+    throttled, it falls back to the inner queue for non-throttled domains. If the
+    inner queue is also empty and all sub-queues are throttled, it sleeps until the
+    earliest cooldown expires.
+
+    Delay sources:
+    - HTTP 429 responses (via `record_domain_delay`)
+    - robots.txt crawl-delay directives (via `set_crawl_delay`)
```
janbuchar, line 47: So if I want to obey crawl-delay directives from robots.txt, I just need to do `BasicCrawler(..., request_manager=ThrottingRequestManager(...))` with no extra settings, correct?

Pijukatel:
@MrAliHasan good work! Please just fix the errors reported by `Type check` and I have no more comments.

src/crawlee/request_loaders/_throttling_request_manager.py
MrAliHasan: Yes, that's correct.

`BasicCrawler` already handles parsing of `robots.txt` and checks whether the configured request manager is a `ThrottlingRequestManager`. If it is, it automatically calls `set_crawl_delay()` with the value from the `crawl-delay` directive.

So simply passing `ThrottlingRequestManager` as the `request_manager` is enough; no additional configuration is required to respect `crawl-delay`.

MrAliHasan:
> @MrAliHasan good work! Please just fix the errors reported by `Type check` and I have no more comments.


The CI type check (Python 3.10) fails with `invalid-return-type` on the two `return` statements in `add_request()`.

It seems the newer `ty` version used in CI resolves `RequestQueue.add_request()` as returning `ProcessedRequest | None` instead of `ProcessedRequest`, and `RequestManager.add_request()` as `Unknown | ProcessedRequest | None`.

One possible fix would be to narrow the type explicitly:

```python
result = await sq.add_request(request, forefront=forefront)
if result is None:
    raise RuntimeError("Unexpected None from add_request()")
return result
```

However, this case should never occur in practice, so I wanted to ask first: Is there a preferred approach for handling this kind of `ty` false positive? For example, a suppression comment or a different typing pattern.

I also noticed that `RequestManagerTandem.add_request()` uses the same delegation pattern and does not seem to trigger this issue on CI.

src/crawlee/request_loaders/_throttling_request_manager.py
janbuchar: Perfect, thank you. Could we also log a warning if `BasicCrawler` has `respect_robots_txt_file` enabled, but the request manager is not explicitly set to `ThrottlingRequestManager`? It should say something along the lines of "The `respect_robots_txt_file` has been turned on, but the crawler was not configured to use `ThrottlingRequestManager`. Unless you do so, `crawl-delay` directives will not be respected."

Mantisus:
Hi, @MrAliHasan 

Thank you for your inspiring work on this PR.

> However, this case should never occur in practice, so I wanted to ask first: Is there a preferred approach for handling this kind of ty false positive? For example, a suppression comment or a different typing pattern.

Make sure you are using `ty` 0.0.18. There are known issues with type narrowing in recent versions.

Regarding `add_request`, this is related to #1775. There is a possibility that the request will not be added to the queue with some backends.

src/crawlee/request_loaders/_throttling_request_manager.py
MrAliHasan: > Perfect, thank you. Could we also log a warning if `BasicCrawler` has `respect_robots_txt_file` enabled, but the request manager is not explicitly set to `ThrottlingRequestManager`? It should say something along the lines of "The `respect_robots_txt_file` has been turned on, but the crawler was not configured to use `ThrottlingRequestManager`. Unless you do so, `crawl-delay` directives will not be respected."

Good suggestion - implemented.

I've added a warning in `BasicCrawler.run()` that triggers when `respect_robots_txt_file` is enabled but the crawler is not configured to use `ThrottlingRequestManager`. In that case, the warning informs the user that `crawl-delay` directives from robots.txt will not be enforced.

Example warning message:

'The `respect_robots_txt_file` option is enabled, but the crawler is not using `ThrottlingRequestManager`. Crawl-delay directives from robots.txt will not be enforced. To enable crawl-delay support, configure the crawler to use `ThrottlingRequestManager` as the request manager.'

MrAliHasan:
> Hi, @MrAliHasan
> 
> Thank you for your inspiring work on this PR.
> 
> > However, this case should never occur in practice, so I wanted to ask first: Is there a preferred approach for handling this kind of ty false positive? For example, a suppression comment or a different typing pattern.
> 
> Make sure you are using `ty` 0.0.18. There are known issues with type narrowing in recent versions.
> 
> Regarding `add_request`, this is related to #1775. There is a possibility that the request will not be added to the queue with some backends.

Thanks for the guidance!

I've pinned `ty` to version `0.0.18` in `pyproject.toml` and updated `uv.lock`. This resolves the CI type check error caused by the newer `ty` versions.

Regarding #1775, understood. I'll keep the current `add_request` return type as `ProcessedRequest` for now, and it can be adjusted later if the behavior changes as part of that issue.

vdusek:
> I've pinned ty to version 0.0.18 in pyproject.toml and updated uv.lock. This resolves the CI type check error caused by the newer ty versions.

This is wrong. You should have only run `uv` to upgrade the package to its latest version that is compliant with the pin in the `pyproject.toml`. Not hard-pin one specific version there.

MrAliHasan:
> > I've pinned ty to version 0.0.18 in pyproject.toml and updated uv.lock. This resolves the CI type check error caused by the newer ty versions.
> 
> This is wrong. You should have only run `uv` to upgrade the package to its latest version that is compliant with the pin in the `pyproject.toml`. Not hard-pin one specific version there.

Thanks for the clarification.

I've reverted the hard pin in pyproject.toml and restored the original specifier, then updated the dependency using uv so the resolved version is recorded in uv.lock.

I also addressed the remaining CI issues:
- Updated add_request() to explicitly guard against a possible None return for proper type narrowing.
- Increased the retry_after delay in test_sleep_when_all_throttled to avoid timing issues observed on Windows CI.

Please let me know if anything else should be adjusted.

Pijukatel (approved):

Pijukatel:
Only last detail from me: Please update the PR description to match the latest state of the PR, especially that the `RequestThrottler` is optional.

MrAliHasan:
> Only last detail from me: Please update the PR description to match the latest state of the PR, especially that the `RequestThrottler` is optional.

Thanks! I’ve updated the PR description to reflect the latest state and clarified that the `ThrottlingRequestManager` is optional.

src/crawlee/request_loaders/_throttling_request_manager.py
```diff
@@ -0,0 +1,434 @@
...
+            sleep_duration = max(
+                (earliest - datetime.now(timezone.utc)).total_seconds(),
+                0.1,  # Minimum sleep to avoid tight loops.
+            )
+            logger.debug(
+                f'All configured domains are throttled and inner queue is empty. '
+                f'Sleeping {sleep_duration:.1f}s until earliest domain is available.'
+            )
+            await asyncio.sleep(sleep_duration)
+            return await self.fetch_next_request()
```
vdusek: Could we please use a while loop rather than recursion (because of smaller overhead)?

src/crawlee/request_loaders/_throttling_request_manager.py
```diff
@@ -0,0 +1,434 @@
...
+        )
+        crawler = BasicCrawler(request_manager=throttler)
+        ```
+    """
+
+    _BASE_DELAY = timedelta(seconds=2)
+    """Initial delay after the first 429 response from a domain."""
+
+    _MAX_DELAY = timedelta(seconds=60)
+    """Maximum delay between requests to a rate-limited domain."""
```
vdusek: Why not make these options configurable via `__init__` with these defaults?

src/crawlee/request_loaders/_throttling_request_manager.py
```diff
@@ -0,0 +1,434 @@
...
+
+        if domain in self._domain_states:
+            sq = await self._get_or_create_sub_queue(domain)
+            result = await sq.add_request(request, forefront=forefront)
+        else:
+            result = await self._inner.add_request(request, forefront=forefront)
+
+        if result is None:
+            msg = 'add_request unexpectedly returned None'
+            raise RuntimeError(msg)
```
vdusek: `RequestManager.add_request` explicitly returns `ProcessedRequest | None`. Raising on `None` is a contract violation. The `@override` signature should match the base class behavior.

From `RequestManager`'s docstring:

> Information about the request addition to the manager or None if the request was not added.

src/crawlee/request_loaders/_throttling_request_manager.py
```diff
@@ -0,0 +1,434 @@
...
+
+    async def recreate_purged(self) -> ThrottlingRequestManager:
+        """Drop all queues and return a fresh ThrottlingRequestManager with the same configuration.
+
+        This is used during crawler purge to reconstruct the throttler with empty
+        queues while preserving domain configuration and service locator.
+        """
+        await self.drop()
+
+        inner = await RequestQueue.open(
```
vdusek: It assumes the inner request manager is always a `RequestQueue`, right? But if the original inner was a `RequestManagerTandem` or another `RequestManager` subclass, this recreates it as a `RequestQueue`.

tests/unit/test_throttling_request_manager.py
```diff
@@ -0,0 +1,467 @@
...
+
+def _make_request(url: str) -> Request:
+    """Helper to create a Request object."""
+    return Request.from_url(url)
+
+
+# ── Request Routing Tests ─────────────────────────────────
+
+
+@pytest.mark.asyncio
```
vdusek: The project uses `pytest-asyncio` in `auto` mode. All `@pytest.mark.asyncio` decorators are unnecessary and inconsistent with the rest of the test suite.

uv.lock
```diff
@@ -3868,26 +3873,26 @@ wheels = [
 
 [[package]]
 name = "ty"
-version = "0.0.17"
```
vdusek: There is an even newer version of `ty` on master; I suggest undoing all updates to the lock file.

vdusek (changes requested):
Thanks for the contribution @MrAliHasan, I have a few comments 🙂...

src/crawlee/request_loaders/_throttling_request_manager.py
MrAliHasan: Good point. The current implementation assumes the inner manager is a `RequestQueue`, which covers the expected use case. I've documented this in the docstring. If a more generic approach is needed in the future, it can be extended to preserve the original manager type.

MrAliHasan:
Thanks for the detailed review @vdusek! All points have been addressed:

- **While loop instead of recursion** — replaced the recursive `fetch_next_request()` call with a `while True` loop.
- **Configurable delays** — `base_delay` and `max_delay` are now `__init__` parameters with the previous values as defaults.
- **`add_request` return type** — removed the `RuntimeError` guard. The return type now matches the base class (`ProcessedRequest`).
- **`recreate_purged` assumption** — the current implementation assumes the inner manager is a `RequestQueue`, which covers the expected use case. This is documented in the docstring and can be extended in the future if needed.
- **`@pytest.mark.asyncio` decorators** — removed all of them.
- **`uv.lock`** — reverted all changes.

MrAliHasan:
I've restored `uv.lock` from upstream master (ty 0.0.21). The previous revert went back too far to ty 0.0.17.

I did notice ty 0.0.21 flags some pre-existing errors in other files (e.g., test_basic_crawler.py, _redis/_utils.py), but none from the throttling manager code. If the type check still fails on CI, then please guide me on how to handle it.

vdusek:
Hi @MrAliHasan, there are still type check issues, could you resolve them?

MrAliHasan:
> Hi @MrAliHasan, there are still type check issues, could you resolve them?

Added an explicit None guard in `add_request` for proper type narrowing. This resolves the invalid-return-type error from ty without using RuntimeError or # type: ignore.

vdusek:
> Added an explicit None guard in add_request for proper type narrowing. This resolves the invalid-return-type error from ty without using RuntimeError or # type: ignore.

There are still typer errors. You know you can run the checks locally, right? :slightly_smiling_face:

MrAliHasan:
> > Added an explicit None guard in add_request for proper type narrowing. This resolves the invalid-return-type error from ty without using RuntimeError or # type: ignore.
> 
> There are still typer errors. You know you can run the checks locally, right? 🙂

Yes, I do run checks locally! 😄 I run `uv run poe type-check` (which runs uv run ty check) and all checks pass with ty 0.0.21, including when I manually set --python-version 3.14. I wasn't able to reproduce this error locally on macOS.

It seems the CI (Linux, ubuntu-latest) resolves the type narrowing differently for `add_request`'s return type. The root cause is that ty infers the delegated `add_request` calls as returning `ProcessedRequest | None`, while the override signature is `-> ProcessedRequest`.

Could you suggest the preferred approach here? I see a few options:

1. Suppress with `# type: ignore[return-value]` 
2. Change the return type to `ProcessedRequest | None` to match the base class behavior
3. Something else you'd prefer?

vdusek:
> Change the return type to ProcessedRequest | None to match the base class behavior

:+1:

uv.lock
vdusek, line 1: Could you please undo all changes of the `uv.lock` file?

src/crawlee/request_loaders/_throttling_request_manager.py
```diff
@@ -0,0 +1,434 @@
...
+
+    @override
+    async def drop(self) -> None:
+        await self._inner.drop()
+        for sq in self._sub_queues.values():
+            await sq.drop()
+        self._sub_queues.clear()
+
+    @override
+    async def add_request(  # type: ignore[invalid-method-override]
```
vdusek: Remove the type ignore and resolve proper typing, please.

src/crawlee/request_loaders/_throttling_request_manager.py
vdusek: I think we should resolve this right ahead in this PR. Your opinion @janbuchar?

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1582,7 +1619,16 @@ async def _is_allowed_based_on_robots_txt_file(self, url: str) -> bool:
...
         robots_txt_file = await self._get_robots_txt_file_for_url(url)
-        return not robots_txt_file or robots_txt_file.is_allowed(url)
+        if not robots_txt_file:
+            return True
+
+        # Wire robots.txt crawl-delay into ThrottlingRequestManager
+        if isinstance(self._request_manager, ThrottlingRequestManager):
+            crawl_delay = robots_txt_file.get_crawl_delay()
+            if crawl_delay is not None:
+                self._request_manager.set_crawl_delay(url, crawl_delay)
```
vdusek, line 1627: IIUC, this is called for every request, but it's redundant after the first call for a given domain. Could you improve that? (caching or checking if it was already set)

vdusek (changes requested):

src/crawlee/request_loaders/_throttling_request_manager.py
janbuchar: Agreed. IMO the most practical way to achieve this would be to require a `request_manager_opener` callback in the `__init__` method. Most often, this would be just `RequestQueue.open`. You'd need to make the `RequestManagerTandem` class generic, but that makes a lot of sense anyway.

janbuchar:
@MrAliHasan there are still some unresolved comments, mainly https://github.com/apify/crawlee-python/pull/1762#discussion_r2904227441 - can you please take care of those?

MrAliHasan:
> @MrAliHasan there are still some unresolved comments, mainly [#1762 (comment)](https://github.com/apify/crawlee-python/pull/1762#discussion_r2904227441) - can you please take care of those?

Yes, I'll work on it tomorrow.

MrAliHasan:
Thanks for the follow-up @vdusek @janbuchar! All comments have been addressed:

- **Remove type ignore, resolve proper typing**: Updated the base class `RequestManager.add_request` to return `ProcessedRequest | None` (along with `RequestManagerTandem` and `RequestQueue`), so the override in `ThrottlingRequestManager` no longer needs `# type: ignore`. Also updated `add_requests` to handle `None`.
- **`recreate_purged` assumption**: Added an explicit `isinstance` check that raises `TypeError` if the inner manager is not a `RequestQueue`, instead of silently assuming it.
- **Redundant `set_crawl_delay` calls**: Added an early return in `set_crawl_delay` if the crawl-delay is already set for the domain, making repeated calls a no-op.
- **`uv.lock`**: Reverted to match master. There are currently merge conflicts due to master moving forward

docs/guides/request_throttling.mdx
```diff
@@ -0,0 +1,47 @@
...
+import ApiLink from '@site/src/components/ApiLink';
+import RunnableCodeBlock from '@site/src/components/RunnableCodeBlock';
+
+import ThrottlingExample from '!!raw-loader!roa-loader!./code_examples/request_throttling/throttling_example.py';
+
+When crawling websites that enforce rate limits (HTTP 429) or specify `crawl-delay` in their `robots.txt`, you need a way to throttle requests per domain without blocking unrelated domains. The <ApiLink to="class/ThrottlingRequestManager">`ThrottlingRequestManager`</ApiLink> provides exactly this.
+
+## Overview
+
+The <ApiLink to="class/ThrottlingRequestManager">`ThrottlingRequestManager`</ApiLink> wraps a <ApiLink to="class/RequestQueue">`RequestQueue`</ApiLink> and manages per-domain throttling. You specify which domains to throttle at initialization, and the manager automatically:
```
janbuchar: I believe it can potentially wrap any `RequestManager`, no?

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -700,19 +702,30 @@ async def run(
...
             if self._use_session_pool:
                 await self._session_pool.reset_store()
 
             request_manager = await self.get_request_manager()
-            if purge_request_queue and isinstance(request_manager, RequestQueue):
-                await request_manager.drop()
-                self._request_manager = await RequestQueue.open(
-                    storage_client=self._service_locator.get_storage_client(),
-                    configuration=self._service_locator.get_configuration(),
-                )
```
janbuchar, line 712: Even in the state before the change, this was a code smell - shouldn't we add a "purge_on_start_hook"-like abstract method to `RequestManager` and implement it in `RequestQueue`? Or should we just call `.drop` on request manager?

This is aimed mostly at @vdusek and @Pijukatel. We definitely don't need to resolve it in this PR if you guys don't see an obvious way out.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1442,6 +1455,10 @@ async def __run_task_function(self) -> None:
 
             await self._mark_request_as_handled(request)
 
+            # Record successful request to reset rate limit backoff for this domain.
+            if isinstance(request_manager, ThrottlingRequestManager):
+                request_manager.record_success(request.url)
```
janbuchar: Can't the "record success" part be implemented in `ThrottlingRequestManager.mark_request_as_handled` instead? It could probably look at `request.state`, couldn't it?

src/crawlee/request_loaders/_throttling_request_manager.py
janbuchar: @MrAliHasan I'm afraid this is still not resolved correctly

docs/guides/request_throttling.mdx
MrAliHasan: You're right, it can wrap any `RequestManager`. I'll update the docs to reflect that.

src/crawlee/crawlers/_basic/_basic_crawler.py
MrAliHasan: Understood, happy to leave this for a follow-up.

src/crawlee/crawlers/_basic/_basic_crawler.py
MrAliHasan: Done, moved `record_success` into `ThrottlingRequestManager.mark_request_as_handled` and removed the `isinstance` check from the crawler.

MrAliHasan:
Added a `request_manager_opener` callback parameter to `__init__` (defaults to `RequestQueue.open`), as you suggested. `recreate_purged` now uses this callback instead of hardcoding `RequestQueue.open`, and preserves it across recreations.

src/crawlee/crawlers/_basic/_basic_crawler.py
janbuchar: @MrAliHasan I believe this is still unresolved?

src/crawlee/request_loaders/_throttling_request_manager.py
janbuchar: Much better now, but we still need to make `ThrottlingRequestManager` generic over the request manager type to make sure that the opener callback returns the same implementation of `RequestManager` as the one used for the "main queue" in the constructor.

janbuchar (commented):
Two more issues. Also please resolve merge conflicts and revert changes to `uv.lock`.

MrAliHasan:
@janbuchar 
Thanks for the feedback. Could you please clarify what exactly is still not resolved? I want to make sure I address the issue correctly.

janbuchar:
I meant these two comments:

- https://github.com/apify/crawlee-python/pull/1762#discussion_r3058590281
- https://github.com/apify/crawlee-python/pull/1762#discussion_r3058649697

vdusek (approved):
I did some final changes, LGTM from my side now. @janbuchar, could you take a final look?

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -697,19 +700,29 @@ async def run(
 
         self._running = True
 
+        if self._respect_robots_txt_file and not isinstance(self._request_manager, ThrottlingRequestManager):
+            self._logger.warning(
```
janbuchar, line 706: It would be nice to recommend using `ThrottlingRequestManager` even when a `Retry-After` header is encountered, but I'm not exactly sure how to wire that up 🤔 

Also, retry-after headers from domains other than the ones configured in `ThrottlingRequestManager` will be silently ignored, which is fine, but it's not optimal DX.

src/crawlee/_utils/http.py
```diff
@@ -0,0 +1,41 @@
...
+    if not value:
+        return None
+
+    try:
+        return timedelta(seconds=int(value))
+    except ValueError:
+        pass
+
+    try:
+        retry_date = parsedate_to_datetime(value)
```
janbuchar, line 33: We should ensure that the return value is timezone-aware. According to the docs, it seems that it depends on the input value and I wouldn't be surprised if servers returned naive timestamps...

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -697,19 +700,29 @@ async def run(
...
-                )
+            if purge_request_queue:
+                if isinstance(request_manager, RequestQueue):
+                    await request_manager.drop()
+                    self._request_manager = await RequestQueue.open(
+                        storage_client=self._service_locator.get_storage_client(),
+                        configuration=self._service_locator.get_configuration(),
+                    )
+                elif isinstance(request_manager, ThrottlingRequestManager):
+                    self._request_manager = await request_manager.recreate_purged()
```
janbuchar: Perhaps we should require an optional `purge`-like method on request managers instead of the `isinstance` checks?

janbuchar (approved):
I have a few potential discussion points, but I'm approving the PR so that it's not blocked anymore.

src/crawlee/crawlers/_basic/_basic_crawler.py
vdusek: Yes, good idea - `purge()` is now part of the `RequestManager` interface

src/crawlee/crawlers/_basic/_basic_crawler.py
vdusek: Good idea, implemented, both cases now emit a warning (only once) on the first 429.

src/crawlee/_utils/http.py
vdusek: OK, Claude suggested this:

> RFC 7231 specifies HTTP-dates are GMT, so treating naive values as UTC is the correct interpretation

PR #1771, opened 2026-03-01
fix(playwright): dispose of `APIResponse` body for `send_request`

kozlice:
### Description

Dispose of Playwright's APIResponse body, otherwise it stays in memory until the context closes.

If not done, this can lead to excessive memory usage during crawling when using `send_request` within a handler.

### Issues

...

### Testing

`send_request` is already covered by tests, so if they pass I think it should be OK.

### Checklist

- [x] CI passed

### DIFF
--- a/src/crawlee/crawlers/_playwright/_types.py
+++ b/src/crawlee/crawlers/_playwright/_types.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Literal, Protocol, TypedDict
 
-from playwright.async_api import APIResponse, Response
+from playwright.async_api import APIResponse
 
 from crawlee import HttpHeaders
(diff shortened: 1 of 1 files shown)


src/crawlee/crawlers/_playwright/_types.py
```diff
@@ -3,13 +3,14 @@
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Literal, Protocol, TypedDict
 
+from playwright.async_api import APIResponse, Response
```
vdusek: `Response` doesn't need to be a runtime import, right?

vdusek:
@Mantisus Could you please check this as well?

src/crawlee/crawlers/_playwright/_types.py
kozlice: Fixed this

PR #1775, opened 2026-03-03
fix: Return `None` from `add_request` when storage client fails to enqueue request

Mantisus:
### Description

- This PR updates the signature for `RequestManager.add_request` to return `None` when the storage client fails to process the request and returns it in `AddRequestsResponse.unprocessed_requests`. Previously, accessing `processed_requests[0]` on an empty list caused an `IndexError`.

### Testing

- Tests updated in accordance with the new signature

### DIFF
--- a/src/crawlee/storages/_request_queue.py
+++ b/src/crawlee/storages/_request_queue.py
@@ -166,9 +166,14 @@ class RequestQueue(Storage, RequestManager):
             return response.processed_requests[0]
 
         if response.unprocessed_requests:
-            logger.warning(f'Request {request.url} was not processed and returned as unprocessed.')
+            logger.warning(
+                f'Request {request.url} was not processed by storage client "{self._client.__class__.__name__}".'
+            )
         else:
-            logger.warning(f'Request {request.url} was not processed: empty response from the storage client.')
+            logger.warning(
+                f'Request {request.url} was not processed by storage client "{self._client.__class__.__name__}" '
+                'received empty response.'
+            )
         return None
 
     @override


src/crawlee/storages/_request_queue.py
```diff
@@ -158,10 +158,18 @@ async def add_request(
...
+    ) -> ProcessedRequest | None:
         request = self._transform_request(request)
         response = await self._client.add_batch_of_requests([request], forefront=forefront)
-        return response.processed_requests[0]
+
+        if response.processed_requests:
+            return response.processed_requests[0]
+
+        if response.unprocessed_requests:
+            logger.warning(f'Request {request.url} was not processed and returned as unprocessed.')
```
vdusek: Instead of:

> Request {request.url} was not processed and returned as unprocessed.

Could we update the warning log to something like:

> Request {request.url} was not processed by storage client {storage_client}.

This would provide additional context by explicitly identifying the storage client as the component that did not process the request.

(I am not sure if we can manage to get the storage client name here; if not, leave it without it.

src/crawlee/request_loaders/_request_manager.py
```diff
@@ -28,7 +28,7 @@ async def add_request(
         request: str | Request,
         *,
         forefront: bool = False,
-    ) -> ProcessedRequest:
+    ) -> ProcessedRequest | None:
```
janbuchar, line 31: This is breaking on the type level, but the `None` is only returned in cases that would throw an error previously, so I guess it's fine. But it'd be nice if we mentioned it in the changelog or something.

Issue #1740, opened 2026-02-12
Add discover valid sitemaps utility (port from JS)


vdusek:
## Summary

Port the `discoverValidSitemaps()` utility from Crawlee JS to Python.

**JS source:** `packages/utils/src/internals/sitemap.ts` — [#3392](https://github.com/apify/crawlee/pull/3392)

## How it works in JS

```typescript
async function* discoverValidSitemaps(
    urls: string[],
    options?: { proxyUrl?: string; httpClient?: BaseHttpClient }
): AsyncIterable<string>
```

1. **Group input URLs by hostname**
2. **For each domain**, discover sitemaps from (in order):
   - `Sitemap:` entries in robots.txt
   - Input URLs that match `/sitemap\.(xml|txt)(\.gz)?$/i`
   - HEAD-request probing of `/sitemap.xml`, `/sitemap.txt`, `/sitemap_index.xml` (fallback)
3. **Deduplicate** and **process domains concurrently**

Returns an async iterable yielding sitemap URLs as discovered.

## What Python already has

- `Sitemap.try_common_names()` — probes `/sitemap.xml` and `/sitemap.txt` for a single URL (missing `/sitemap_index.xml`)
- `RobotsTxtFile.find()` + `get_sitemaps()` — fetches and extracts `Sitemap:` entries from robots.txt

**What's missing:** the orchestrating function that combines these steps, groups by hostname, validates via HEAD requests, detects direct sitemap URLs from input, and processes domains concurrently.

PR #1777, opened 2026-03-04
feat: add `discover_valid_sitemaps` utility

Mantisus:
### Description

- Add `discover_valid_sitemaps` utility to search for sitemaps of websites for the provided URLs.

### Issues

- Closes: #1740

### Testing

- Add new unit tests

### DIFF
--- a/src/crawlee/_utils/sitemap.py
+++ b/src/crawlee/_utils/sitemap.py
@@ -32,7 +32,7 @@ logger = getLogger(__name__)
 
 VALID_CHANGE_FREQS = {'always', 'hourly', 'daily', 'weekly', 'monthly', 'yearly', 'never'}
 SITEMAP_HEADERS = {'accept': 'text/plain, application/xhtml+xml, application/xml;q=0.9, */*;q=0.8'}
-SITEMAP_URL_PATTERN = re.compile(r'sitemap\.(?:xml|txt)(?:\.gz)?$', re.IGNORECASE)
+SITEMAP_URL_PATTERN = re.compile(r'\/sitemap\.(?:xml|txt)(?:\.gz)?$', re.IGNORECASE)
 COMMON_SITEMAP_PATHS = ['/sitemap.xml', '/sitemap.txt', '/sitemap_index.xml']
 
 
@@ -520,7 +520,7 @@ async def _merge_async_generators(*generators: AsyncGenerator) -> AsyncGenerator
     finally:
         for task in tasks:
             task.cancel()
-        await asyncio.gather(*tasks)
+        await asyncio.gather(*tasks, return_exceptions=True)
 
 
 async def _discover_for_hostname(
--- a/src/crawlee/_utils/web.py
+++ b/src/crawlee/_utils/web.py
@@ -1,16 +1,18 @@
 from __future__ import annotations
 
+from http import HTTPStatus
+
 
 def is_status_code_client_error(value: int) -> bool:
     """Return `True` for 4xx status codes, `False` otherwise."""
-    return 400 <= value <= 499  # noqa: PLR2004
+    return HTTPStatus.BAD_REQUEST <= value < HTTPStatus.INTERNAL_SERVER_ERROR
 
 
 def is_status_code_server_error(value: int) -> bool:
     """Return `True` for 5xx status codes, `False` otherwise."""
-    return value >= 500  # noqa: PLR2004
+    return value >= HTTPStatus.INTERNAL_SERVER_ERROR
 
 
 def is_status_code_successful(value: int) -> bool:
     """Return `True` for 2xx and 3xx status codes, `False` otherwise."""
-    return 200 <= value < 400  # noqa: PLR2004
+    return HTTPStatus.OK <= value < HTTPStatus.BAD_REQUEST


src/crawlee/_utils/sitemap.py
```diff
@@ -484,3 +491,136 @@ async def parse_sitemap(
...
+    for sitemap_url in robots.get_sitemaps():
+        if _check_and_add(sitemap_url):
+            yield sitemap_url
+
+    # Check maybe provided URLs have sitemap url
+    sitemap_url = next((url for url in hostname_urls if SITEMAP_URL_PATTERN.search(url)), None)
+
+    if sitemap_url:
+        if _check_and_add(sitemap_url):
+            yield sitemap_url
```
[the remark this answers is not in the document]
Mantisus: We trust the sitemaps specified in robots.txt or provided by the user.

src/crawlee/_utils/sitemap.py
```diff
@@ -484,3 +491,136 @@ async def parse_sitemap(
...
+    else:
+        # Check common sitemap locations
+        base_url = URL(hostname_urls[0])
```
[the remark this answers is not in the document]
Mantisus: The order doesn't matter in this case.

src/crawlee/_utils/sitemap.py
```diff
@@ -484,3 +491,136 @@ async def parse_sitemap(
...
+            try:
+                response = await http_client.send_request(
+                    candidate, method=method_for_checking, proxy_info=proxy_info, timeout=request_timeout
+                )
+                if is_status_code_successful(response.status_code) and _check_and_add(candidate):
+                    yield candidate
+            except ProxyError:
+                logger.warning(f'Proxy error when checking {candidate} with sitemap discovery for {hostname}')
+            except asyncio.TimeoutError:
+                logger.warning(f'Timeout when checking {candidate} with sitemap discovery for {hostname}')
```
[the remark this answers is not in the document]
Mantisus: updated

src/crawlee/_utils/sitemap.py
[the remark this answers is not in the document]
Mantisus: Make sense

src/crawlee/_utils/sitemap.py
```diff
@@ -484,3 +491,140 @@ async def parse_sitemap(
...
+    hostname_urls = list(set(hostname_urls))  # Remove duplicates
+
+    def _check_and_add(url: str) -> bool:
+        if url in domain_seen:
+            return False
+        domain_seen.add(url)
+        return True
+
+    # Try getting sitemaps from robots.txt first
+    robots = await RobotsTxtFile.find(url=hostname_urls[0], http_client=http_client, proxy_info=proxy_info)
```
Pijukatel, line 548: So we are checking just the first `hostname_urls[0]` because we rely on the caller passing the same domain urls? This seems a little fragile, but it is internal and only we call it, so maybe ok.

Mantisus: Yes, in this case, I hope that only we use the private function. 🙂

src/crawlee/_utils/web.py
```diff
@@ -9,3 +9,8 @@ def is_status_code_client_error(value: int) -> bool:
 def is_status_code_server_error(value: int) -> bool:
     """Return `True` for 5xx status codes, `False` otherwise."""
     return value >= 500  # noqa: PLR2004
+
+
+def is_status_code_successful(value: int) -> bool:
+    """Return `True` for 2xx and 3xx status codes, `False` otherwise."""
+    return 200 <= value < 400  # noqa: PLR2004
```
vdusek: Maybe we can use:

```python
from http import HTTPStatus
```

instead of magic numbers, and remove the noqa.

src/crawlee/_utils/sitemap.py
```diff
@@ -484,3 +491,136 @@ async def parse_sitemap(
...
+            item = await queue.get()
+            if item is end_feed:
+                remaining_tasks -= 1
+            else:
+                yield item
+    finally:
+        for task in tasks:
+            task.cancel()
+
+
```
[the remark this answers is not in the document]
vdusek: Shouldn't we propagate the exceptions here? Is Copilot wrong?

src/crawlee/_utils/sitemap.py
vdusek: Hmm, the `hostname_urls[0]` with combination of `hostname_urls = list(set(hostname_urls))` can lead to non-deterministic behaviour. Because `set()` doesn't preserve order. Claude suggests using `list(dict.fromkeys(hostname_urls))`, or you may find even a better way.

src/crawlee/_utils/sitemap.py
```diff
@@ -27,6 +32,8 @@
 
 VALID_CHANGE_FREQS = {'always', 'hourly', 'daily', 'weekly', 'monthly', 'yearly', 'never'}
 SITEMAP_HEADERS = {'accept': 'text/plain, application/xhtml+xml, application/xml;q=0.9, */*;q=0.8'}
+SITEMAP_URL_PATTERN = re.compile(r'sitemap\.(?:xml|txt)(?:\.gz)?$', re.IGNORECASE)
```
vdusek: This can match false positives like `https://example.com/this-is-not-sitemap.xml` 😄.

vdusek (commented):
A few comments

src/crawlee/_utils/sitemap.py
Mantisus: All `hostname_urls` URLs must belong to the same hostname. `RobotsTxtFile.find` sets the path in the URL to `/robots.txt`. Thus, the order of the URLs should not affect the result.

src/crawlee/_utils/sitemap.py
Mantisus: Fair enough.

src/crawlee/_utils/web.py
Mantisus: Good point

src/crawlee/_utils/sitemap.py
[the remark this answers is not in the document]
Mantisus: `return_exceptions` does not propagate the exception; it returns it in the results. But yes, it can be useful for correctly terminating generators and requests.

PR #1781, opened 2026-03-06
feat: allow non-href links extract & enqueue

kozlice:
### Description

Right now `href` attribute is hardcoded for `enqueue_links` and `extract_links` methods.

This change would allow to:
- grab image links
- grab links from exotic attributes in SPA

Examples:

```python
await context.extract_links(selector="nav ul li", attribute="data-href")
await context.enqueue_links(selector=".gallery .item img", attribute="src", label="image")
```

### Issues

?

### Testing

Covered by new tests.

### Checklist

- [ ] CI passed

### DIFF
--- a/tests/unit/crawlers/_parsel/test_parsel_crawler.py
+++ b/tests/unit/crawlers/_parsel/test_parsel_crawler.py
@@ -51,20 +51,18 @@ async def test_enqueue_links(redirect_server_url: URL, server_url: URL, http_cli
 
     await crawler.run(requests)
 
-    first_visited = visit.call_args_list[0][0][0]
-    visited = {call[0][0] for call in visit.call_args_list}
-
-    assert first_visited == redirect_url
-    assert visited == {
-        redirect_url,
-        str(server_url / 'sub_index'),
-        str(server_url / 'page_1'),
-        str(server_url / 'page_2'),
-        str(server_url / 'page_3'),
-        str(server_url / 'page_4'),
-        str(server_url / 'base_page'),
-        str(server_url / 'base_subpath/page_5'),
-    }
+    expected_visit_calls = [
+        mock.call(redirect_url),
+        mock.call(str(server_url / 'sub_index')),
+        mock.call(str(server_url / 'page_1')),
+        mock.call(str(server_url / 'page_2')),
+        mock.call(str(server_url / 'page_3')),
+        mock.call(str(server_url / 'page_4')),
+        mock.call(str(server_url / 'base_page')),
+        mock.call(str(server_url / 'base_subpath/page_5')),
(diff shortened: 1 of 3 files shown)


tests/unit/server_endpoints.py
```diff
@@ -57,6 +57,16 @@
     <a href=https://avatars.githubusercontent.com/apify>Apify avatar/a>
 </body></html>"""
 
+NON_HREF_LINKS = b"""\
+<html><head>
+    <title>Hello</title>
+</head>
+<body>
+    <a href="/page_1"></a>
+    <li data-href="/page_2"></a>
```
vdusek: Malformed HTML

```suggestion
    <li data-href="/page_2"></li>
```

src/crawlee/_types.py
```diff
@@ -387,6 +387,7 @@ def __call__(
         self,
         *,
         selector: str | None = None,
+        attribute: str | None = None,
```
vdusek, line 390: Can we please also add tests for `enqueue_links` with `attribute`? (not just `extract_links`)

vdusek (commented):
Thank you! Two comments...

src/crawlee/_types.py
kozlice: Sure, I'll add test cases for that

tests/unit/crawlers/_parsel/test_parsel_crawler.py
```diff
@@ -67,6 +67,32 @@ async def request_handler(context: ParselCrawlingContext) -> None:
...
+    visit = mock.Mock()
+
+    @crawler.router.default_handler
+    async def request_handler(context: ParselCrawlingContext) -> None:
+        visit(context.request.url)
+        await context.enqueue_links(selector='img', attribute='src')
+
+    await crawler.run(requests)
+
+    first_visited = visit.call_args_list[0][0][0]
```
Pijukatel: Preferably, use mock assert helpers instead of digging through the call list

For example:
https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_count
https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_has_calls

kozlice: I'll fix this tomorrow, and the adjacent tests too (I followed their style)

kozlice: @Pijukatel switched to `assert_has_calls` in all 3 crawlers tests

Pijukatel: Thank you.

PR #1787, opened 2026-03-12
fix(redis): Do not remove handled request data from request queue

kozlice:
### Description

Unlike FS and SQL storages that keep handled requests data, Redis removes it.

### Issues

?

### Testing

A dedicated unit test added.

Manual testing: using RedisQueueClient, enqueue a page, run crawler, check if `:data` key holds request info.

### Checklist

- [ ] CI passed

### DIFF
--- a/tests/unit/storage_clients/_redis/test_redis_rq_client.py
+++ b/tests/unit/storage_clients/_redis/test_redis_rq_client.py
@@ -95,17 +95,6 @@ async def test_request_records_persistence(rq_client: RedisRequestQueueClient) -
         assert request_data['url'].startswith('https://example.com/')
 
 
-async def test_handled_request_records_persistence(rq_client: RedisRequestQueueClient) -> None:
-    request = Request.from_url('https://example.com/1')
-    await rq_client.add_batch_of_requests([request])
-    fetched_request = await rq_client.fetch_next_request()
-    assert isinstance(fetched_request, Request)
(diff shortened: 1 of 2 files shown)


kozlice:
Looks like an unrelated test got stuck and failed.

tests/unit/storage_clients/_redis/test_redis_rq_client.py
```diff
@@ -95,6 +95,17 @@ async def test_request_records_persistence(rq_client: RedisRequestQueueClient) -
...
 
+async def test_handled_request_records_persistence(rq_client: RedisRequestQueueClient) -> None:
+    request = Request.from_url('https://example.com/1')
+    await rq_client.add_batch_of_requests([request])
+    fetched_request = await rq_client.fetch_next_request()
+    assert isinstance(fetched_request, Request)
+    await rq_client.mark_request_as_handled(fetched_request)
+    fetched_request = await rq_client.get_request(request.unique_key)
+    assert isinstance(fetched_request, Request)
+    assert fetched_request.unique_key == request.unique_key
```
vdusek: Perhaps we should move this to `tests/unit/storages/test_request_queue.py` as a general RQ test. We can then parametrize it across all storage clients to ensure consistency among all storage clients. What do you think?

kozlice: Makes sense. I'll update the PR

Issue #1741, opened 2026-02-12
Add post-navigation hooks and browser lifecycle hooks


vdusek:
## Context

Crawlee JS provides navigation hooks and browser lifecycle hooks that Python is missing. See [parity report](https://github.com/apify/crawlee-python/issues/1739) for broader context.

### 1. Post-navigation hooks (main gap)

JS `BrowserCrawler` and `HttpCrawler` both support `postNavigationHooks` — they run after `page.goto()` / HTTP request completes but before the request handler. Useful for CAPTCHA detection, response validation, etc.

Python only has only `pre_navigation_hook`. No post-navigation equivalent exists.

### 2. Browser lifecycle hooks (BrowserPool)

JS `BrowserPool` exposes 6 hook types, these are for consideration:
- `preLaunchHooks` / `postLaunchHooks` — before/after browser launch
- `prePageCreateHooks` / `postPageCreateHooks` — before/after new page creation
- `prePageCloseHooks` / `postPageCloseHooks` — before/after page close

Python's `BrowserPool` has no lifecycle hooks.

## Reference

- JS BrowserCrawler hooks: `packages/browser-crawler/src/internals/browser-crawler.ts`
- JS BrowserPool hooks: `packages/browser-pool/src/browser-pool.ts`
- Python pre-nav hooks: `src/crawlee/crawlers/_playwright/_playwright_crawler.py`
- Python BrowserPool: `src/crawlee/browsers/_browser_pool.py`

PR #1791, opened 2026-03-13
feat: Add page lifecycle hooks to `BrowserPool`

Mantisus:
### Description

Add four page lifecycle hooks to `BrowserPool` registered as decorators:
- `pre_page_create_hook` — called before page creation; `browser_new_context_options` is mutable, so the hook can affect how the page context is configured.
- `post_page_create_hook` — called after page creation.
- `pre_page_close_hook` — called before page close.
- `post_page_close_hook` — called after page close.

### Issues

- Relates: #1741

### Testing

- Added new tests for `BrowserPool`.

### DIFF
--- a/docs/guides/code_examples/playwright_crawler/browser_pool_page_hooks_example.py
+++ b/docs/guides/code_examples/playwright_crawler/browser_pool_page_hooks_example.py
@@ -1,0 +1,77 @@
+from __future__ import annotations
+
+import asyncio
+import logging
+from typing import TYPE_CHECKING, Any
+
+from crawlee.browsers import BrowserPool
+from crawlee.crawlers import PlaywrightCrawler, PlaywrightCrawlingContext
+from crawlee.storages import KeyValueStore
+
+if TYPE_CHECKING:
+    from crawlee.browsers._browser_controller import BrowserController
+    from crawlee.browsers._types import CrawleePage
+    from crawlee.proxy_configuration import ProxyInfo
+
+logger = logging.getLogger(__name__)
+
+
+async def main() -> None:
+    async with BrowserPool() as browser_pool:
+
+        @browser_pool.pre_page_create_hook
+        async def log_page_init(
+            page_id: str,
+            _browser_controller: BrowserController,
+            _browser_new_context_options: dict[str, Any],
+            _proxy_info: ProxyInfo | None,
+        ) -> None:
+            """Log when a new page is about to be created."""
+            logger.info(f'Creating page {page_id}...')
+
+        @browser_pool.post_page_create_hook
+        async def set_viewport(
+            crawlee_page: CrawleePage, _browser_controller: BrowserController
+        ) -> None:
+            """Set a fixed viewport size on each newly created page."""
+            await crawlee_page.page.set_viewport_size({'width': 1280, 'height': 1024})
+
+        @browser_pool.pre_page_close_hook
+        async def save_screenshot(
+            crawlee_page: CrawleePage, _browser_controller: BrowserController
+        ) -> None:
+            """Save a screenshot to KeyValueStore before each page is closed."""
+            kvs = await KeyValueStore.open()
+
+            screenshot = await crawlee_page.page.screenshot()
+            await kvs.set_value(
+                key=f'screenshot-{crawlee_page.id}',
+                value=screenshot,
+                content_type='image/png',
+            )
+            logger.info(f'Saved screenshot for page {crawlee_page.id}.')
+
+        @browser_pool.post_page_close_hook
+        async def log_page_closed(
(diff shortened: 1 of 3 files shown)


src/crawlee/browsers/_browser_pool.py
```diff
@@ -357,3 +375,57 @@ async def _close_inactive_browsers(self) -> None:
...
+        """Override the page's close method to execute pre and post close hooks."""
+        if self._pre_page_close_hooks or self._post_page_close_hooks:
+            original_close = crawlee_page.page.close
+
+            async def close_with_hooks(*args: Any, **kwargs: Any) -> None:
+                await self._execute_hooks(self._pre_page_close_hooks, crawlee_page, browser_controller)
+                await original_close(*args, **kwargs)
+                await self._execute_hooks(self._post_page_close_hooks, crawlee_page.id, browser_controller)
+
+            crawlee_page.page.close: Callable[..., Awaitable[None]] = close_with_hooks
```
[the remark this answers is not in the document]
Mantisus: It is not expected that a closing hook will be added after the page has already been opened.

src/crawlee/browsers/_browser_pool.py
```diff
@@ -357,3 +375,57 @@ async def _close_inactive_browsers(self) -> None:
...
+
+    def pre_page_create_hook(
+        self, hook: Callable[[str, BrowserController, dict[str, Any], ProxyInfo | None], Awaitable[None]]
+    ) -> None:
```
vdusek: The registration methods return `None`, which breaks the decorator contract. After `@browser_pool.pre_page_create_hook`, the decorated name is rebound to `None`:

```python
@browser_pool.pre_page_create_hook
async def my_hook(...):
    ...
# my_hook is now None!
```

The tests don't catch this because the decorated names are never referenced afterward. Each method should return the hook:

```python
def pre_page_create_hook(
    self, hook: Callable[[str, BrowserController, dict[str, Any], ProxyInfo | None], Awaitable[None]]
) -> Callable[[str, BrowserController, dict[str, Any], ProxyInfo | None], Awaitable[None]]:
    self._pre_page_create_hooks.append(hook)
    return hook
```

Same applies to all four registration methods.

src/crawlee/browsers/_browser_pool.py
```diff
@@ -357,3 +375,57 @@ async def _close_inactive_browsers(self) -> None:
...
+
+    def _override_page_close(self, crawlee_page: CrawleePage, browser_controller: BrowserController) -> None:
+        """Override the page's close method to execute pre and post close hooks."""
+        if self._pre_page_close_hooks or self._post_page_close_hooks:
+            original_close = crawlee_page.page.close
+
+            async def close_with_hooks(*args: Any, **kwargs: Any) -> None:
+                await self._execute_hooks(self._pre_page_close_hooks, crawlee_page, browser_controller)
+                await original_close(*args, **kwargs)
+                await self._execute_hooks(self._post_page_close_hooks, crawlee_page.id, browser_controller)
```
vdusek, line 394: If a pre-close hook raises, `original_close` is never called, leaking the page. And if `original_close` raises, post-close hooks are skipped. Consider guarding this:

```python
async def close_with_hooks(*args: Any, **kwargs: Any) -> None:
    await self._execute_hooks(self._pre_page_close_hooks, crawlee_page, browser_controller)
    try:
        await original_close(*args, **kwargs)
    finally:
        await self._execute_hooks(self._post_page_close_hooks, crawlee_page.id, browser_controller)
```

Similarly in `_get_new_page`, if a `post_page_create_hook` raises, `_override_page_close` is never called (line 340), so close hooks won't fire for that page.

src/crawlee/browsers/_browser_pool.py
vdusek, line 396: Monkey-patching `page.close` on the Playwright `Page` object works because Python's MRO checks instance attributes before class methods, and Playwright's `Page.__aexit__` calls `self.close()`. However, this is an implementation detail of Playwright — if they ever change `__aexit__` to call an internal method directly, the hooks would silently stop firing.

Might be worth a brief comment documenting this assumption, so future maintainers know what to watch for.

src/crawlee/browsers/_browser_pool.py
Mantisus: That's a good point. The approach used follows the general pattern for hooks. But I agree that it needs to be updated for the decorator contract.

src/crawlee/browsers/_browser_pool.py
Mantisus: If a hook fails to execute, this should correctly trigger an error. Subsequent hooks should not be called.

I moved `original_close` into `finally` to prevent potential leaking the page.

PR #1796, opened 2026-03-16
fix(playwright): filter unsupported context options in persistent browser

### LINKED ISSUE #1784: PlaywrightCrawler __init__ method browser_new_context_options argument does not function
According to the documentation, the **browser_new_context_options** are the keyword arguments that are accepted by the Playwright Browser new_context method, here:

`
Keyword arguments to pass to the browser new context method. These options are provided directly to Playwright's browser.new_context method. For more details, refer to the [Playwright documentation](https://playwright.dev/python/docs/api/class-browser#browser-new-context). This option should not be used if browser_pool is provided.
`

https://crawlee.dev/python/api/class/PlaywrightCrawler#__init__

The appropriate keywords are documented on this page:
https://playwright.dev/python/docs/api/class-browser#browser-new-context

One of those keywords is ** storage_state**.

But...

When I pass the **browser_new_context_options** to the PlaywrightCrawler.__init__ method, it calls some other function called...

`
[crawlee.crawlers._playwright._playwright_crawler] WARN  Retrying request to ___ due to: BrowserType.launch_persistent_context() got an unexpected keyword argument 'storage_state'. File "./crawlee/browsers/_playwright_browser.py", line 70, in new_context,     self._context = await self._browser_type.launch_persistent_context(,                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^,         user_data_dir=user_data_dir, **launch_options,         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^,     )
`

Please update the code to match the documentation.

sushant-mutnale:
This PR fixes issue #1784, where PlaywrightCrawler would crash when passing context options (like storage_state) that are unsupported by Playwright's launch_persistent_context method.

Changes:

Implemented dynamic argument filtering in PlaywrightPersistentBrowser.new_context using inspect. signature.
Added a warning log to guide users when options are filtered out, suggesting the use of incognito pages as an alternative.
Added a unit test in 

tests/unit/browsers/test_playwright_browser.py
to verify the fix and prevent regressions.
Fixes #1784

### DIFF
--- a/src/crawlee/browsers/_playwright_browser_controller.py
+++ b/src/crawlee/browsers/_playwright_browser_controller.py
@@ -224,6 +224,31 @@ class PlaywrightBrowserController(BrowserController):
         """Handle actions after a page is closed."""
         self._pages.remove(page)
 
+    def _filter_context_options(self, options: dict[str, Any]) -> dict[str, Any]:
+        """Filter browser context options based on the current mode (incognito vs persistent).
+
+        Options that are valid only in the other mode are dropped with a warning. Unrecognized options are kept
+        and passed through so that Playwright itself can raise an appropriate error.
+        """
+        params_cache = _get_context_params_cache()
+        filtered = dict[str, Any]()
+
+        for key, value in options.items():
+            if self._use_incognito_pages and key in params_cache['persistent_unique']:
+                logger.warning(
+                    f'Option "{key}" is only supported in persistent context mode '
+                    '(use_incognito_pages=False) and will be ignored.'
+                )
+            elif not self._use_incognito_pages and key in params_cache['incognito_unique']:
+                logger.warning(
+                    f'Option "{key}" is only supported in incognito context mode '
+                    '(use_incognito_pages=True) and will be ignored.'
+                )
+            else:
+                filtered[key] = value
+
+        return filtered
+
     async def _create_browser_context(
         self,
         browser_new_context_options: Mapping[str, Any] | None = None,
@@ -238,31 +263,7 @@ class PlaywrightBrowserController(BrowserController):
         """
         browser_new_context_options = dict(browser_new_context_options) if browser_new_context_options else {}
 
-        params_cache = _get_context_params_cache()
-
-        filtered_options = {}
-        for key, value in browser_new_context_options.items():
-            if self._use_incognito_pages:
-                # Incognito mode (new_context)
-                if key in params_cache['common'] or key in params_cache['incognito_unique']:
-                    filtered_options[key] = value
-                elif key in params_cache['persistent_unique']:
-                    logger.warning(
-                        f'Option "{key}" is only supported in persistent context mode '
-                        '(use_incognito_pages=False) and will be ignored.'
-                    )
-                else:
-                    raise TypeError(f'"{key}" is not a valid Playwright context option.')
-            elif key in params_cache['common'] or key in params_cache['persistent_unique']:
-                # Persistent mode (launch_persistent_context)
-                filtered_options[key] = value
-            elif key in params_cache['incognito_unique']:
-                logger.warning(
-                    f'Option "{key}" is only supported in incognito context mode '
-                    '(use_incognito_pages=True) and will be ignored.'
-                )
-            else:
-                raise TypeError(f'"{key}" is not a valid Playwright context option.')
+        filtered_options = self._filter_context_options(browser_new_context_options)
 
         if proxy_info:
             if filtered_options.get('proxy'):
--- a/tests/unit/browsers/test_playwright_controller_validation.py
+++ b/tests/unit/browsers/test_playwright_controller_validation.py
@@ -25,9 +25,10 @@ async def browser(playwright: Playwright) -> AsyncGenerator[Browser, None]:
     await browser.close()
 
 
-async def test_controller_validation_typo(browser: Browser) -> None:
+async def test_controller_validation_typo_passed_through(browser: Browser) -> None:
+    """Invalid options (e.g. typos) are passed through so Playwright raises its own error."""
     controller = PlaywrightBrowserController(browser)
-    with pytest.raises(TypeError, match=r'"headles" is not a valid Playwright context option.'):
+    with pytest.raises(TypeError):
         await controller.new_page(browser_new_context_options={'headles': True})
     await controller.close()
 


pyproject.toml
```diff
@@ -34,11 +34,14 @@ keywords = [
     "scraping",
 ]
 dependencies = [
+    "apify-fingerprint-datapoints>=0.11.0",
```
Pijukatel: We have all these added dependencies in the optional dependencies group **playwright**. So please remove them from here.

src/crawlee/browsers/_playwright_browser.py
```diff
@@ -67,8 +68,21 @@ async def new_context(self, **context_options: Any) -> BrowserContext:
             user_data_dir = tempfile.mkdtemp(prefix=self._TMP_DIR_PREFIX)
             self._temp_dir = Path(user_data_dir)
 
+        launch_persistent_context_sig = inspect.signature(self._browser_type.launch_persistent_context)
```
Pijukatel: This is a reasonable approach, but it has some drawbacks. If user has just typo ( in otherwise valid argument name), it will just show warning in log. Same for using some completely nonsensical argument. That should raise an error and not just log a warning.

For example, this should raise (typo in **headles**):
```
    persist_browser = PlaywrightPersistentBrowser(
        playwright.chromium, browser_launch_options={'headles': True}
    )
```


Maybe this approach could be adopted one lever higher (not in `PlaywrightPersistentBrowser` - which always just calls `launch_persistent_context`), but in `PlaywrightBrowserController` - that is the class that decides about calling  `launch_persistent_context` or `new_context`, but feeds them the same arguments.

It should properly raise exceptions for bad arguments, but it could just log a warning as per your suggestion for arguments at least valid in the other method. It would have to get 3 sets of arguments to be able to do such a distinction. Something like:

```
...
    launch_persistent_context_sig = set(inspect.signature(BrowserType.launch_persistent_context).parameters)
    new_context_sig = set(inspect.signature(Browser.new_context).parameters)
    persistent_unique_options = launch_persistent_context_sig - new_context_sig
    new_context_unique_options = new_context_sig - launch_persistent_context_sig
    common_options = launch_persistent_context_sig & new_context_sig
...

```

And then raise an exception or just log based on the selected mode.

sushant-mutnale:
Hello! Thank you for the detailed feedback. I've refactored the validation logic into 

`PlaywrightBrowserController` using the suggested three-set approach with cached signatures. I also moved the dependencies back to the optional group in 
`pyproject.toml`. 

New unit tests cover both the warning and error scenarios. Ready for another look!

pyproject.toml
```diff
@@ -35,10 +35,12 @@ keywords = [
 ]
 dependencies = [
     "async-timeout>=5.0.1",
+    "browserforge>=1.2.4",
     "cachetools>=5.5.0",
     "colorama>=0.4.0",
     "impit>=0.8.0",
     "more-itertools>=10.2.0",
+    "playwright>=1.58.0",
```
vdusek: `browserforge` and `playwright` should not be part of core dependencies

pyproject.toml
```diff
@@ -55,15 +57,15 @@ adaptive-crawler = [
     "jaro-winkler>=2.0.3",
     "playwright>=1.27.0",
     "scikit-learn>=1.6.0",
-    "apify_fingerprint_datapoints>=0.0.3",
+    "apify_fingerprint_datapoints>=0.11.0",
```
vdusek: why?

src/crawlee/browsers/_playwright_browser_controller.py
```diff
@@ -27,6 +29,14 @@
 
 logger = getLogger(__name__)
 
+# Cache Playwright signatures to avoid overhead in critical path
+_launch_persistent_context_params = set(inspect.signature(PlaywrightBrowserType.launch_persistent_context).parameters)
+_new_context_params = set(inspect.signature(Browser.new_context).parameters)
```
vdusek: Is it necessary to run these at the import time of the module?

src/crawlee/browsers/_playwright_browser_controller.py
```diff
@@ -222,11 +237,38 @@ async def _create_browser_context(
...
+            elif key in params_cache['common'] or key in params_cache['persistent_unique']:
+                # Persistent mode (launch_persistent_context)
+                filtered_options[key] = value
+            elif key in params_cache['incognito_unique']:
+                logger.warning(
+                    f'Option "{key}" is only supported in incognito context mode '
+                    '(use_incognito_pages=True) and will be ignored.'
+                )
+            else:
+                raise TypeError(f'"{key}" is not a valid Playwright context option.')
```
Pijukatel: Could you please extract to a standalone private method with docstring explaining it. 
`filtered_options = self._filter_new_context_options(options=browser_new_context_options)`

src/crawlee/browsers/_playwright_browser_controller.py
```diff
@@ -222,11 +237,38 @@ async def _create_browser_context(
...
+                # Incognito mode (new_context)
+                if key in params_cache['common'] or key in params_cache['incognito_unique']:
+                    filtered_options[key] = value
+                elif key in params_cache['persistent_unique']:
+                    logger.warning(
+                        f'Option "{key}" is only supported in persistent context mode '
+                        '(use_incognito_pages=False) and will be ignored.'
+                    )
+                else:
+                    raise TypeError(f'"{key}" is not a valid Playwright context option.')
```
Pijukatel: I do not think we need to raise here; it is better for the Playwright code to raise, so that anyone can see the code where the arguments are defined.

It will be sufficient to filter out the arguments valid in the other case and warn for those, while letting the completely wrong arguments go through, and let them fail in Playwright. So it can be simplified to something like 

```
if self._use_incognito_pages and key in params_cache['persistent_unique']:
    logger.warning(
        f'Option "{key}" is only supported in persistent context mode '
        '(use_incognito_pages=False) and will be ignored.'
    )
elif not self._use_incognito_pages and key in params_cache['incognito_unique']:
    logger.warning(
        f'Option "{key}" is only supported in incognito context mode '
        '(use_incognito_pages=True) and will be ignored.'
    )
else:
    filtered_options[key] = value

```

Pijukatel (commented):
Thanks for the changes, and apologies for the delayed review. Just a few small comments, and I think it will be ready.

PR #1799, opened 2026-03-19
feat: Expose `use_state` in `BasicCrawler`

Mantisus:
### Description

- Expose `use_state` in `BasicCrawler`. This can be useful for passing data initialized during crawler setup into a handler.

Example:

```python
async def main():
    crawler = BasicCrawler()
    await crawler.use_state({'state': 'init'})

    async def handler(context):
        data = await context.use_state()

        context.log.info(f'State: {data["state"]}')

    await crawler.run(["https://www.example.com/"])


if __name__ == '__main__':
    asyncio.run(main())
```


vdusek (approved):
LGTM, all other Crawlers have it exposed, am I correct?

Mantisus:
> all other Crawlers have it exposed, am I correct?

After that PR method will be exposed in all crawlers

Issue #1802, opened 2026-03-19
`RequestList` with persistence doesn't handle malformed URLs correctly


francescoaracu:
I'm trying to implement a crawler using BS (you can find the repo [here](https://github.com/francescoaracu/crawler-test-bs)).

I tried to run the crawler, adding the persistence parameter to my `RequestList`, but it only works if it's not present.

This is the error thrown in the logs:

<img width="1999" height="1504" alt="Image" src="https://github.com/user-attachments/assets/d66452a7-ac86-42a1-9d1e-2a3b684afde0" />

Mantisus:
Hi @francescoaracu!

Thank you for being so interested in Crawlee!

In this case, the issue isn't with persistence but with validating certain URLs in your list. Without persistence, this would have caused an error, but much later.

Until a fix is released, you can handle this on your side in `load_urls_from_csv`

```python
from crawlee import Request
from pydantic import ValidationError

try:
    request = Request.from_url(row[0])
except ValidationError:
    print(f'Invalid URL encountered in CSV: {row[0]}')
    continue
```

francescoaracu:
> Thank you for being so interested in Crawlee!

Thanks man! I'm using the framework for my MSc thesis 😄 


> you can handle this on your side in `load_urls_from_csv`

I guess I should change the behaviour of my function, given that with this block I'm instantiating a Request object for each URL... or do I just need to put this block after the `if row`?

Mantisus:
> Thanks man! I'm using the framework for my MSc thesis 😄

Cool!

> I guess I should change the behaviour of my function, given that with this block I'm instantiating a Request object for each URL... or do I just need to put this block after the if row?

This should work for you

```python
from crawlee import Request
from pydantic import ValidationError

async def load_urls_from_csv(file_path: str):
    """Read URLs from CSV file asynchronously."""
    async with aiofiles.open(file_path, 'r') as file:
        reader = aiocsv.AsyncReader(file)
        # Skip header row
        await reader.__anext__()
        async for row in reader:
            if row:
                try:
                    request = Request.from_url(row[0])
                except ValidationError:
                    print(f'Invalid URL encountered in CSV: {row[0]}')
                    continue
                yield row[0]  # URL is in the first column
                # Yield control back to the event loop
                # break
                await asyncio.sleep(0)
```

francescoaracu:
Thanks a lot for your suggestion @Mantisus!
[...]

PR #1803, opened 2026-03-19
fix: Handle invalid URLs in `RequestList`

Mantisus:
### Description

- Handle invalid URLs in `RequestList`.

### Issues

- Closes: #1802 

### Testing

- Add new tests for `RequestList

### DIFF
--- a/tests/unit/request_loaders/test_request_list.py
+++ b/tests/unit/request_loaders/test_request_list.py
@@ -253,7 +253,7 @@ async def test_handle_invalid_url() -> None:
     await request_list.mark_request_as_handled(request)
 
 
-async def test_handle_invalide_url_with_persistence() -> None:
+async def test_handle_invalid_url_with_persistence() -> None:
     """Test that invalid URLs are handled gracefully even when persistence is enabled."""
     persist_key = 'test_invalid_url_persistence'
     request_list = RequestList(['invalid-url.com', 'https://valid.placeholder.com'], persist_requests_key=persist_key)


tests/unit/request_loaders/test_request_list.py
```diff
@@ -240,3 +240,31 @@ async def test_persist_requests_key_only_persists_once() -> None:
     fetched_request = await request_list_2.fetch_next_request()
     assert fetched_request is not None
     assert fetched_request.url == 'https://once2.placeholder.com'  # From original data
+
+
+async def test_handle_invalid_url() -> None:
+    """Test that invalid URLs are handled gracefully."""
+    request_list = RequestList(['invalid-url.com', 'https://valid.placeholder.com'])
```
Pijukatel, line 247: Why not running the validation even earlier?

Is there any point why should this `RequestList(['invalid-url.com'])` be even allowed?

Maybe we could warn/raise already here?

janbuchar: `RequestList` can pull the URLs from any iterable, so validation just before returning "next request" is non-negotiable. We can validate in the constructor as well though, I guess.

Mantisus: This could be a very long list, and if we iterate fully in the constructor just for validation, it feels like an overhead. However, if we wrap it in a generator that performs the validation, that would essentially be the current solution. This is because, with persistence disabled, validation occurs the moment a new URL is received from the iterator. And when persistence is enabled, we iterate through all URLs before saving in the storage.

Pijukatel: I am fine with this explanation, thanks

Issue #1808, opened 2026-03-24
When using `event_manager` to initialize the crawler, the global `event_manager` doesn't work.


Mantisus:
If `event_manager` is passed as an initialization parameter to `BasicCrawler`, the global `event_manager` in `service_locator` will never be started.

https://github.com/apify/crawlee-python/blob/master/src/crawlee/crawlers/_basic/_basic_crawler.py#L770-L778

As a result, classes that use `service_locator.get_event_manager()` do not behave as expected, because they do not receive the relevant events.

Classes affected by this issue: `Snapshotter` and `RecoverableState`.

Related: #1805

Also, if two crawlers share the global `event_manager` and one finishes before the other, the `event_manager` will be shut down for the second crawler as well.

Pijukatel:
Just for the discussion. I think we can split this into two problems.

### Reliance on global 
(We should probably first think through whether there is a scenario for multiple active `event_managers` that is valid and makes sense.)

Some parts of our code rely on the global `event_manager` no matter what. Even when they work in the context of another `event_manager`

Ideas?
- All such `event_manager` consumers should have explicit `event_manager` argument, probably on init and thus remove the reliance on the global.

### Sharing the same `event_manager`
 - Using it as a [reentrant context manager](https://docs.python.org/3/library/contextlib.html#reentrant-context-managers) despite the context manager not being designed to handle it this way
 - Letting individual consumers close it even though multiple consumers can rely on it

Ideas?
- Having isolated contexts instead of shared ones and avoiding reentrant scenarios:
```
event_manager = EventManager(...)
async with event_manager.create_context():
   ...
```
- Handle the re-entrant scenarios by counting consumers

Issue #1805, opened 2026-03-20
Running inside Docker doesn't take care of persistence


francescoaracu:
I'm implementing a crawler using BS (you can find the repo [here](https://github.com/francescoaracu/crawler-test-bs)).

Even if I set `purge_on_start` to `False` inside a Configuration object, after pausing the crawler via Ctrl-C in the attached Docker container's shell, the crawler doesn't seem to take care of the persistence status set in the `RequestList` instance.

The volume mounted in my `docker-compose.yml` is then overwritten directly when running the container again with `docker start`.

Mantisus:
Hi, @francescoaracu :)

Could you try setting up the configuration via the `service_locator` and let us know if that solves the issue?

```python
from crawlee import service_locator

config = service_locator.get_configuration()
config.available_memory_ratio = 0.7
config.purge_on_start = False
```

francescoaracu:
It seems to be working now; it starts first with the "first-level" requests from the queue and then it continues with the other "zero-level" links from the CSV.

PR #1810, opened 2026-03-24
fix: Prevent premature `EventManager` shutdown when multiple crawlers share it

Mantisus:
### Description

- Fixed a bug where the global `EventManager` was shut down prematurely when the first of multiple concurrent crawlers finished, leaving remaining crawlers with a broken event system.
- Always starts the global `event_manager`, even if the `event_manager` argument was passed in the crawler`s constructor

### Issues

- Closed: #1808
- Closed: #1805

### Testing

- Added a test verifying that the global `EventManager` remains active as long as at least one crawler is running, and is shut down only after the last crawler finishes.
- Added a test to verify that both `event_manager` instances are active while the crawler is running.

### DIFF
--- a/src/crawlee/events/_event_manager.py
+++ b/src/crawlee/events/_event_manager.py
@@ -129,6 +129,7 @@ class EventManager:
         if self._active_ref_count > 1:
             # Emit persist state event to ensure the latest state is saved before closing the context.
             await self._emit_persist_state_event()
+            await self.wait_for_all_listeners_to_complete(timeout=self._close_timeout)
             self._active_ref_count -= 1
             return
 
--- a/src/crawlee/events/_local_event_manager.py
+++ b/src/crawlee/events/_local_event_manager.py
@@ -75,7 +75,10 @@ class LocalEventManager(EventManager):
         It starts emitting system info events at regular intervals.
         """
         await super().__aenter__()
-        self._emit_system_info_event_rec_task.start()
+
+        if self._active_ref_count == 1:
+            self._emit_system_info_event_rec_task.start()
+
         return self
 
     async def __aexit__(
@@ -88,7 +91,9 @@ class LocalEventManager(EventManager):
 
         It stops emitting system info events and closes the event manager.
         """
-        await self._emit_system_info_event_rec_task.stop()
+        if self._active_ref_count == 1:
+            await self._emit_system_info_event_rec_task.stop()
+
         await super().__aexit__(exc_type, exc_value, exc_traceback)
 
     async def _emit_system_info_event(self) -> None:


src/crawlee/events/_event_manager.py
```diff
@@ -93,6 +93,9 @@ def __init__(
             delay=self._persist_state_interval,
         )
 
+        # Reference count for active contexts.
+        self._ref_count = 0
+
         # Flag to indicate the context state.
         self._active = False
```
Pijukatel: Do we need `self._active` anymore? 

```
    @property
    def active(self) -> bool:
        """Indicate whether the context is active."""
        return self._ref_count>0
```

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -768,22 +768,26 @@ def sigint_handler() -> None:
...
+                local_event_manager,
                 self._snapshotter,
                 self._statistics,
                 self._session_pool if self._use_session_pool else None,
                 self._http_client,
                 self._crawler_state_rec_task,
                 *self._additional_context_managers,
             )
-            if cm and getattr(cm, 'active', False) is False
+            if cm and (isinstance(cm, EventManager) or not getattr(cm, 'active', False))
```
Pijukatel: I think it would be more readable to just do it on a standalone line. Something like
`contexts_to_enter.append([local_event_manager, global_event_manager])`

src/crawlee/events/_event_manager.py
```diff
@@ -93,6 +93,9 @@ def __init__(
             delay=self._persist_state_interval,
         )
 
+        # Reference count for active contexts.
+        self._ref_count = 0
```
Pijukatel: Maybe `self._active_ref_count`

src/crawlee/events/_event_manager.py
```diff
@@ -130,9 +131,16 @@ async def __aexit__(
...
+        self._ref_count -= 1
+
+        # Emit persist state event to ensure the latest state is saved before closing the context.
+        await self._emit_persist_state_event()
+
+        if self._ref_count > 0:
+            return
+
         # Stop persist state event periodic emission and manually emit last one to ensure latest state is saved.
         await self._emit_persist_state_event_rec_task.stop()
```
Pijukatel, line 137: Having
```
self._emit_persist_state_event_rec_task.stop()
await self._emit_persist_state_event()
```
reduced the risk of a possible scenario of two persist state events very close to each other, I would prefer to keep it this way and just add 
```
 if self._ref_count > 0:
            await self._emit_persist_state_event()
            return
```

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -773,26 +773,31 @@ async def _run_crawler(self) -> None:
         if local_event_manager is global_event_manager:
             local_event_manager = None  # Avoid entering the same event manager context twice
 
+        # The event managers are always entered.
+        contexts_to_enter: list[Any] = (
+            [global_event_manager, local_event_manager] if local_event_manager else [global_event_manager]
+        )
```
janbuchar, line 779: I'm not so sure about this. Why should the crawler initialize a component that it doesn't use? And then tear it down? I think we should try to see a bigger picture first. 

The linked issues are caused by `Snapshotter` and `RecoverableState` subscribing to an event manager that doesn't emit anything - how does that happen?

Mantisus: I would prefer to remove `event_manager` from the `BasicCrawler` constructor entirely and keep only the global one, configurable via `service_locator.set_event_manager`. However, this requires discussion and should probably be part of v2.

> I'm not so sure about this. Why should the crawler initialize a component that it doesn't use? And then tear it down? I think we should try to see a bigger picture first.

The global `event_manager` being started inside the crawler, while only being used by the crawler's internal components, is an early architectural decision. I assume that it was done for a better user experience.

> The linked issues are caused by Snapshotter and RecoverableState subscribing to an event manager that doesn't emit anything - how does that happen?

I believe the root cause is bugs introduced during our work on supporting multiple parallel crawlers:
- An edge case we didn't account for, demonstrated by `test_multiple_crawlers_with_global_event_manager`. The first crawler to activate the `event_manager` is fully responsible for its lifecycle. When the crawler finishes, it tears down the `event_manager`, which clears all subscriptions, leaving any still-running crawlers without a functioning `event_manager`.
- The introduction of an internal `_service_locator` in `BasicCrawler`, which caused the user-provided `event_manager` to be used only for `Event.CRAWLER_STATUS` in `_crawler_state_task`, while internal components (`Snapshotter`, `RecoverableState`) continued subscribing to the global `service_locator`, which may never be started.

All that made the behavior of `service_locator.get_event_manager()` unstable and unpredictable.

This PR is an attempt to fix these issues within the current architecture without introducing breaking changes.

Pijukatel: I think that the fixes that solve 2 Crawlers using the same `EventManager` are good. Regarding the global event manager. I see this as a quick fix. 

> I would prefer to remove event_manager from the BasicCrawler constructor entirely...

I think this would lead to counterintuitive behavior due to `Configuration` having `EventManager` specific fields and they would be ignored in cases like this:
```
# Global event manager already exists
BasicCrawler(configuration=Configuration(persist_state_interval=...))
```

Maybe it would be possible for each subscriber to the event manager to have at least an optional init argument that would allow passing an explicit `event_manager` (if it is creating storages, then maybe `service_locator` argument would be even better) instead of relying on the global one.

janbuchar: When porting the service locator to crawlee-js (https://github.com/apify/crawlee/pull/3325), I actually made the global `serviceLocator` object into a proxy that resolves to the crawler-scoped ServiceLocator when called from a crawler instance and to the global one otherwise.

If we adopted that in Python as well, it would solve the inconsistency caused by `Snapshotter` and friends still using the global locator.

Mantisus: Thanks, I'll study that solution

janbuchar: Let me know if you get stuck on something, it's kinda magic. But that's the cost of making the library usable without factories, builders and explicit dependency containers.

src/crawlee/events/_event_manager.py
```diff
@@ -93,24 +93,20 @@ def __init__(
...
         """Indicate whether the context is active."""
-        return self._active
+        return self._active_ref_count > 0
 
     async def __aenter__(self) -> EventManager:
-        """Initialize the event manager upon entering the async context.
+        """Initialize the event manager upon entering the async context."""
+        self._active_ref_count += 1
+        if self._active_ref_count > 1:
+            return self
```
vdusek, line 108: Potential task leak for multi-crawler scenarios; from Claude:

### Bug: `LocalEventManager` not updated for ref-counting — causes task leak

The base class now correctly guards `_emit_persist_state_event_rec_task.start()` behind `ref_count == 1` (line 110). However, `LocalEventManager.__aenter__` (in `_local_event_manager.py:72-79`) **unconditionally** calls `self._emit_system_info_event_rec_task.start()` on every entry:

```python
async def __aenter__(self) -> LocalEventManager:
    await super().__aenter__()
    self._emit_system_info_event_rec_task.start()  # Called EVERY time!
    return self
```

Since `RecurringTask.start()` creates a new `asyncio.Task` and overwrites `self.task` without cancelling the previous one (`_utils/recurring_task.py:60-66`), the old task is **leaked** (still running, but no reference to cancel it).

In the multi-crawler scenario with the default shared global `LocalEventManager`:
1. Crawler 1 enters → `ref_count=1`, system_info **task A** created
2. Crawler 2 enters → `ref_count=2`, system_info **task B** created, **task A orphaned**
3. Crawler 1 exits → `stop()` cancels task B, `ref_count=1` — but **task A is still running** with no reference
4. Crawler 2 exits → `stop()` tries to cancel already-cancelled task B — **task A never stopped**

`LocalEventManager` needs the same ref-count-aware guards:

```python
async def __aenter__(self) -> LocalEventManager:
    await super().__aenter__()
    if self._active_ref_count == 1:  # Only start on first entry
        self._emit_system_info_event_rec_task.start()
    return self

async def __aexit__(self, ...) -> None:
    if self._active_ref_count == 1:  # Only stop on last exit
        await self._emit_system_info_event_rec_task.stop()
    await super().__aexit__(exc_type, exc_value, exc_traceback)
```

Note: checking `== 1` in `__aexit__` must happen **before** `super().__aexit__()` decrements the count.

src/crawlee/events/_event_manager.py
```diff
@@ -127,17 +123,23 @@ async def __aexit__(
...
         """
-        if not self._active:
+        if not self.active:
             raise RuntimeError(f'The {self.__class__.__name__} is not active.')
 
+        if self._active_ref_count > 1:
+            # Emit persist state event to ensure the latest state is saved before closing the context.
+            await self._emit_persist_state_event()
+            self._active_ref_count -= 1
+            return
```
vdusek, line 134: We don't call `wait_for_all_listeners_to_complete` in this branch.

From Claude:

### Consideration: Intermediate exit skips listener cleanup

When `ref_count > 1`, the exit emits a persist-state event but does **not** call `wait_for_all_listeners_to_complete()` or clean up listeners registered by the exiting crawler.

In practice this is likely fine because individual components (`Snapshotter`, `RecoverableState`) clean up their own listeners via `off()` in their `__aexit__`/`teardown()`. But any component that relies on the event manager's `remove_all_listeners()` as the cleanup mechanism would have stale listeners firing after its context is torn down.

Worth adding a comment here explaining this design choice — e.g. that individual components are responsible for their own listener cleanup, and `remove_all_listeners()` on final exit is just a safety net.

tests/unit/crawlers/_basic/test_basic_crawler.py
```diff
@@ -2047,3 +2046,104 @@ async def error_handler(context: BasicCrawlingContext, error: Exception) -> Requ
...
+
+    # After both crawlers are finished, event manager should be inactive.
+    assert event_manager.active is False
+
+    await rq1.drop()
+    await rq2.drop()
+
+
+async def test_global_and_local_event_manager_in_crawler_run() -> None:
+    """Test that both global and local event managers are used in crawler run"""
```
vdusek:
```suggestion
    """Test that both global and local event managers are used in crawler run."""
```

Mantisus:
I studied the implementation of crawlee-js (https://github.com/apify/crawlee/pull/3325) and discussed it with @janbuchar. It doesn't work here due to how `RecoverableState` is designed. It can be initialized outside the crawler's context and expects events from the global event manager rather than the one defined within the crawler's scope.

Therefore, we decided to keep this PR as a temporary solution and discussed a possible update in v2. I will create a follow-up issue after we agree on some details.

vdusek:
@janbuchar Do you wanna re-check this, or it can be merged?

janbuchar:
> @janbuchar Do you wanna re-check this, or it can be merged?

No, I think it's fine as a temporary fix. Go ahead and merge it.

PR #1813, opened 2026-03-26
fix(playwright): Validate browser context options before calling launch_persistent_context

ReinerBRO:
Fixes the persistent Playwright browser path so unsupported browser_new_context_options are rejected early with a clear error instead of failing later inside launch_persistent_context(). Adds a regression test covering storage_state.

### DIFF
--- a/src/crawlee/browsers/_playwright_browser.py
+++ b/src/crawlee/browsers/_playwright_browser.py
@@ -17,6 +17,8 @@ if TYPE_CHECKING:
 
 logger = getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


Pijukatel:
Hello, we already have  https://github.com/apify/crawlee-python/pull/1796 in progress, so I guess we should just finalize and merge that one.

vdusek:
Hi @ReinerBRO, thanks for the PR, however we already have https://github.com/apify/crawlee-python/pull/1796 opened, so closing in favour of that.

PR #1814, opened 2026-03-27
fix: Defer page object cleanup to make it accessible in error handlers

### LINKED ISSUE #1482: PlaywrightCrawler error_handler cannot access Page object
In the Javascript version, the error handler is able to access the `Page` object via `PlaywrightCrawlingContext.page`. I discovered that the Python version doesn't implement this when porting the `ContextPipeline` to Javascript.

### Test case

```py
async def test_error_handler_can_access_page(server_url: URL) -> None:
    crawler = PlaywrightCrawler(max_request_retries=2)

    request_handler = mock.AsyncMock(side_effect=RuntimeError('Intentional crash'))
    crawler.router.default_handler(request_handler)

    error_handler_calls: list[str | None] = []

    @crawler.error_handler
    async def error_handler(context: BasicCrawlingContext | PlaywrightCrawlingContext, _error: Exception) -> None:
        error_handler_calls.append(
            await context.page.content() if isinstance(context, PlaywrightCrawlingContext) else None
        )

    await crawler.run([str(server_url / 'hello-world')])

    assert error_handler_calls == [HELLO_WORLD, HELLO_WORLD, HELLO_WORLD]
```

### Possible solutions

1. Run the error handlers before the cleanup step of the context pipeline
    - this is a fairly big change and we probably want to do it after #1474
    - changing this in the adaptive playwright crawler will be especially tricky
2. Add some "deferred cleanup" step to the context pipeline and call _that_ after error handlers are done
    - it's unclear how this would fit in the current async generator based middleware model
    - considerable refactoring of the `_run_request_handler` and `__run_task_function` would still be necessary - error handlers are called by the latter and context pipeline is only handled in the former

janbuchar:
- closes #1482

### DIFF
--- a/src/crawlee/_types.py
+++ b/src/crawlee/_types.py
@@ -14,7 +14,7 @@ if TYPE_CHECKING:
     import json
     import logging
     import re
-    from collections.abc import Callable, Coroutine, Sequence
+    from collections.abc import Awaitable, Coroutine, Sequence
 
     from typing_extensions import NotRequired, Required, Self, Unpack
 
@@ -38,6 +38,9 @@ T = TypeVar('T')
 HttpMethod = Literal['GET', 'HEAD', 'POST', 'PUT', 'DELETE', 'CONNECT', 'OPTIONS', 'TRACE', 'PATCH']
 
 HttpPayload = bytes
+
+DeferredCleanupCallback = Callable[[], 'Awaitable[Any]']
+"""An async callback to be called after request processing completes (including error handlers)."""
 
 RequestTransformAction = Literal['skip', 'unchanged']
 
@@ -661,7 +664,7 @@ class BasicCrawlingContext:
     log: logging.Logger
     """Logger instance."""
 
-    register_deferred_cleanup: Callable[[Callable[[], Coroutine[None, None, None]]], None]
+    register_deferred_cleanup: Callable[[DeferredCleanupCallback], None]
     """Register an async callback to be called after request processing completes (including error handlers)."""
 
     async def get_snapshot(self) -> PageSnapshot:
--- a/src/crawlee/crawlers/_playwright/_playwright_crawler.py
+++ b/src/crawlee/crawlers/_playwright/_playwright_crawler.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import asyncio
-import contextlib
 import logging
 import warnings
 from datetime import timedelta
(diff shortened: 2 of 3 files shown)


src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1509,6 +1512,13 @@ async def __run_task_function(self) -> None:
             )
             raise
 
+        finally:
+            for cleanup in deferred_cleanup:
+                try:
+                    await cleanup()
+                except Exception:  # noqa: PERF203
```
vdusek, line 1519: Can't we put the entire loop into the try-except block as the lint rule suggests?

src/crawlee/crawlers/_playwright/_playwright_crawler.py
```diff
@@ -296,63 +298,73 @@ async def _navigate(
...
+        await context.page.__aenter__()
 
-                route_handler = self._prepare_request_interceptor(
-                    method=context.request.method,
-                    headers=context.request.headers,
-                    payload=context.request.payload,
-                )
+        async def _close_page() -> None:
+            with contextlib.suppress(Exception):
+                await context.page.__aexit__(None, None, None)
```
vdusek: Should we really suppress all exceptions completely? Maybe we could at least log them? In the `BasicCrawler` cleanup loop, we're already catching and logging exceptions, so we could even let them propagate to that level.

Just an idea, maybe there is a good reason to completely suppress all exceptions on this level.

src/crawlee/_types.py
```diff
@@ -661,6 +661,9 @@ class BasicCrawlingContext:
     log: logging.Logger
     """Logger instance."""
 
+    register_deferred_cleanup: Callable[[Callable[[], Coroutine[None, None, None]]], None]
```
vdusek: The type is huge, maybe we could utilize a type alias `DeferredCleanupCallback`?

src/crawlee/crawlers/_basic/_basic_crawler.py
janbuchar: That would mean that if one cleanup hook fails, all those that follow it will, too. I wouldn't worry about performance too much here, there won't be more than one for 99% of users.

src/crawlee/crawlers/_playwright/_playwright_crawler.py
Pijukatel: As far as I understand this change. This should now be redundant,

So all this can be deleted and also the whole `early` argument and all its usage (`_early_reported_errors`) . `ErrorTracker` is private, so such a change is fine.
```
        # Collect data in case of errors, before the page object is closed.
        if error:
            await self.statistics.error_tracker.add(error=error, context=context, early=True)
```

janbuchar: Thanks, implemented!

PR #1816, opened 2026-03-28
fix: remove double usage_count increment in Session.retire()

vdusek:
## Summary
- `Session.retire()` incremented `_usage_count`, but callers (`mark_good()`, `mark_bad()`) already increment it before calling `retire()`, causing a double-increment when a session is retired at the usage limit.
- Removed the redundant `_usage_count += 1` from `retire()`.
- Added tests covering the no-double-increment invariant for `mark_good()`, `mark_bad()`, and direct `retire()` calls.

## Test plan
- [x] Existing session tests pass
- [x] New tests verify `usage_count` is incremented exactly once per `mark_good()`/`mark_bad()` call, even when retirement is triggered

### DIFF
--- a/src/crawlee/sessions/_session.py
+++ b/src/crawlee/sessions/_session.py
@@ -217,8 +217,6 @@ class Session:
         to use `mark_bad` method.
         """
         self._error_score += self._max_error_score
-        self._usage_count += 1
-        # Note: We emit an event here because of the Puppeteer in TS implementation.
 
     def is_blocked_status_code(
         self,
--- a/tests/unit/sessions/test_session.py
+++ b/tests/unit/sessions/test_session.py
@@ -82,9 +82,27 @@ def test_multiple_marks(session: Session) -> None:
 
 def test_retire_method(session: Session) -> None:
     """Test that retire method properly sets the session as unusable."""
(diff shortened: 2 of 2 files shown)


Pijukatel (commented):
What if someone retires a session explicitly by directly calling `session.retire()`?


While ugly, I think it would be safer to avoid the increment in the branches that call the retire. Something like:

```
    def mark_bad(self) -> None:
        """Mark the session as bad after an unsuccessful session usage."""
        self._error_score += 1

        # Retire the session if it is not usable anymore
        if not self.is_usable:
            self.retire()
        else:
            self._usage_count += 1
```

Pijukatel:
> What if someone retires a session explicitly by directly calling `session.retire()`?
> 
> While ugly, I think it would be safer to avoid the increment in the branches that call the retire. Something like:
> 
> ```
>     def mark_bad(self) -> None:
>         """Mark the session as bad after an unsuccessful session usage."""
>         self._error_score += 1
> 
>         # Retire the session if it is not usable anymore
>         if not self.is_usable:
>             self.retire()
>         else:
>             self._usage_count += 1
> ```

Actually, it does not matter. You can ignore this comment

PR #1846, opened 2026-04-15
ci: Extract docs versioning and release into reusable workflows

vdusek:
## Summary

**Docs versioning** (`manual_version_docs.yaml`):
- Extract the inline `version_docs` job from `manual_release_stable.yaml` into a standalone reusable workflow
- Can be triggered manually from the GitHub UI (with an optional ref input) or called from the release pipeline
- Fix `api:version` ENOENT bug by running docusaurus commands through `uv run`
- Clean up all existing versions for the same major version, not just exact major.minor match
- Remove non-docs artifacts (pyproject.toml, .gitignore, caches) from versioned doc snapshots

**Docs release** (`manual_release_docs.yaml`):
- Rename `_release_docs.yaml` to `manual_release_docs.yaml` to match the naming convention
- Add optional `ref` input to `workflow_dispatch` so it can be triggered from the GitHub UI with a specific ref (falling back to the default branch)
- Replace `CHECKOUT_REF` env var approach with a `Determine checkout ref` step, consistent with `manual_version_docs.yaml`
- Update all references in `manual_release_stable.yaml` and `on_master.yaml`

### DIFF
--- a/.github/workflows/manual_release_docs.yaml
+++ b/.github/workflows/manual_release_docs.yaml
@@ -1,13 +1,20 @@
-name: Doc release
+name: Release docs
 
 on:
   # Runs when manually triggered from the GitHub UI.
   workflow_dispatch:
+    inputs:
+      ref:
+        description: Git ref to checkout (branch, tag, or SHA). Defaults to the default branch.
+        required: false
+        type: string
+        default: ""
 
   # Runs when invoked by another workflow.
   workflow_call:
     inputs:
       ref:
+        description: Git ref to checkout (branch, tag, or SHA)
         required: true
         type: string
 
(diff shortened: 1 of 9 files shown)


Pijukatel (approved):
I would prefer to have this defined once in the workflow repo. 

Similar:
https://github.com/apify/crawlee-python/pull/1846
https://github.com/apify/apify-client-python/pull/728
https://github.com/apify/apify-sdk-python/pull/857
https://github.com/apify/apify-client-js/pull/876

vdusek:
> I would prefer to have this defined once in the workflow repo.

Let's go step by step, merge it, migrate to pnpm, see whether everything works, and then try to extract the general part to the workflows.

PR #1855, opened 2026-04-23
fix: Fix StorageInstanceManager cache eviction

janbuchar:
- See the added test to understand the problematic scenario.
- There is a good chance that this wouldn't cause any problems in practice, but better safe than sorry.


vdusek (approved):
LGTM

Pijukatel (approved):

PR #1860, opened 2026-04-27
fix: Report integer count in 'Experiencing problems' status log

vdusek:
## Summary

The walrus operator at `_basic_crawler.py:1656` binds less tightly than `>`, so `failed_requests := X - Y > 0` was parsed as `failed_requests := ((X - Y) > 0)` per [PEP 572](https://peps.python.org/pep-0572/). As a result, `failed_requests` was assigned a `bool` instead of the integer difference, and the periodic status log emitted:

```
Experiencing problems, True failed requests since last status update.
```

instead of the actual count.

## Fix

Move the parentheses so `:=` captures the integer subtraction and `> 0` sits outside.

## Regression test

Added `test_status_message_reports_failed_request_count`, which directly invokes `_crawler_state_task` with a mocked `_statistics.state` and a separate `_previous_crawler_state`, asserting the rendered message contains the integer count `3`, not `True`.

### DIFF
--- a/src/crawlee/crawlers/_basic/_basic_crawler.py
+++ b/src/crawlee/crawlers/_basic/_basic_crawler.py
@@ -1654,11 +1654,9 @@ class BasicCrawler(Generic[TCrawlingContext, TStatisticsState]):
         current_state = self.statistics.state
 
         if (
-            failed_requests := (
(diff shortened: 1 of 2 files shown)


Mantisus (approved):
LGTM

PR #1861, opened 2026-04-27
fix: Preserve `forefront` flag on `RequestQueue` retry path

vdusek:
### Description

`RequestQueue._process_batch` accepts a `forefront` flag and forwards it on the first attempt to `add_batch_of_requests`. When the storage client returned unprocessed requests, the recursive retry call omitted `forefront`, so it fell back to the parameter default `False` regardless of what the original caller asked for.

A user calling `add_requests(..., forefront=True)` would silently see requests appended to the back of the queue whenever they happened to be in an `unprocessed_requests` response, breaking the priority-ordering guarantee.

### Fix

Forward `forefront=forefront` in the recursive `_process_batch` call.

### Test

Added a regression test (parametrized over `True` and `False`) that mocks `add_batch_of_requests` to return all requests as unprocessed on the first call and asserts the retry uses the same `forefront` value as the original call.


Mantisus (approved):
LGTM

PR #1864, opened 2026-04-30
fix: Filter sitemap-derived URLs by enqueue strategy

vdusek:
## Summary

`SitemapRequestLoader` accepted every `<loc>` it parsed (including nested sitemaps and `robots.txt` directives) without checking the host, so a sitemap could push arbitrary URLs into the queue. This wires the existing `EnqueueStrategy` mechanism into the sitemap loader and the `robots.txt` discovery path.

## What changed

- `SitemapRequestLoader.__init__` gains `enqueue_strategy: EnqueueStrategy = 'same-hostname'`, applied to nested `<sitemap><loc>` and `<urlset><url><loc>` entries against the parent sitemap URL. Pass `'all'` to opt out. The strategy is stamped on emitted `Request`s, so `BasicCrawler._check_url_after_redirects` continues policing across redirects.
- `RobotsTxtFile.get_sitemaps(*, enqueue_strategy)` filters cross-host `Sitemap:` directives (already disallowed by spec). The `discover_valid_sitemaps` path is closed as well.
- Hoisted the strategy matcher into `crawlee._utils.urls.matches_enqueue_strategy` so request loaders and `BasicCrawler` share one implementation. `BasicCrawler` now uses `yarl.URL` to match the rest of the codebase.
- Memoised the public-suffix lookup used by `same-domain` so it isn't re-run per URL on hot paths.

## Behavior change

The new default `'same-hostname'` matches `enqueue_links`. Callers depending on cross-host sitemap entries must pass `enqueue_strategy='all'` (or `'same-domain'`). `RobotsTxtFile.get_sitemaps()` no longer returns cross-host entries; its tests were rewritten to assert filtering, with a new `test_extract_same_host_sitemaps_urls` covering the legitimate path.

### DIFF
--- a/src/crawlee/_utils/urls.py
+++ b/src/crawlee/_utils/urls.py
@@ -16,6 +16,13 @@ if TYPE_CHECKING:
     from crawlee._types import EnqueueStrategy
 
 
+_ALLOWED_SCHEMES: frozenset[str] = frozenset({'http', 'https'})
+"""URL schemes Crawlee accepts for fetching and enqueuing."""
+
+_HTTP_URL_ADAPTER: TypeAdapter[AnyHttpUrl] = TypeAdapter(AnyHttpUrl)
+"""Pydantic validator for HTTP and HTTPS URLs."""
+
+
 def is_url_absolute(url: str) -> bool:
     """Check if a URL is absolute."""
     url_parsed = URL(url)
@@ -44,28 +51,19 @@ def to_absolute_url_iterator(base_url: str, urls: Iterator[str], logger: Logger 
             yield converted_url
 
 
-_http_url_adapter = TypeAdapter(AnyHttpUrl)
-
-
 def validate_http_url(value: str | None) -> str | None:
     """Validate the given HTTP URL.
 
+    Args:
+        value: The URL to validate, or `None` to skip validation.
+
     Raises:
-        pydantic.ValidationError: If the URL is not valid.
+        pydantic.ValidationError: If the URL is malformed or its scheme is not `http`/`https`.
     """
     if value is not None:
-        _http_url_adapter.validate_python(value)
+        _HTTP_URL_ADAPTER.validate_python(value)
 
     return value
-
-
-@lru_cache(maxsize=1)
-def _get_tld_extractor() -> TLDExtract:
-    """Return a lazily-initialized `TLDExtract` instance shared across the module."""
-    # `mkdtemp` (vs `TemporaryDirectory`) returns a path whose lifetime is tied to the process — `TemporaryDirectory`
-    # is collected immediately when its return value is discarded, which would race the directory out from under
-    # tldextract.
-    return TLDExtract(cache_dir=tempfile.mkdtemp())
 
 
 def matches_enqueue_strategy(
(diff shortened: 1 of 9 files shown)


src/crawlee/_utils/urls.py
```diff
@@ -51,3 +58,52 @@ def validate_http_url(value: str | None) -> str | None:
...
+    Args:
+        strategy: The enqueue strategy to apply.
+        target_url: The URL to be evaluated.
+        origin_url: The reference URL the target is compared against.
+
+    Returns:
+        `True` if `target_url` is allowed under `strategy` relative to `origin_url`, `False` otherwise.
+    """
+    target = urlparse(target_url) if isinstance(target_url, str) else target_url
+    origin = urlparse(origin_url) if isinstance(origin_url, str) else origin_url
```
Mantisus: Can we use `yarl.URL` instead of `urlparse`?

It performs better, and we're already using it in many places for the same purpose.

Mantisus (changes requested):
Just one note about parsing URLs. Everything else is LGTM.

Mantisus (approved):
LGTM

tests/unit/_utils/test_urls.py
```diff
@@ -55,3 +65,30 @@ def test_validate_http_url() -> None:
...
     with pytest.raises(ValidationError):
         validate_http_url(invalid_url)
+
+
+@pytest.mark.parametrize(
+    ('strategy', 'origin', 'target', 'expected'),
+    [
+        # 'all' lets everything through, even with empty/cross-host targets
+        ('all', 'https://example.com/', 'https://other.test/', True),
+        ('all', 'https://example.com/', 'gopher://internal:6379/_PING', True),
```
janbuchar: I'm pretty sure that even with `enqueue_strategy='all'`, this should still filter that url out, or throw an error

vdusek: maybe we could introduce `allowed_schemes` arg with http / https schemas as default

janbuchar: Let's be real, is there, like anywhere in the world, a legitimate sitemap that contains gopher urls? If someone, somehow, finds that and decides to scrape it, either they'll let us know or they'll just handle the URLs themselves, without `SitemapRequestLoader`.

vdusek: Alright, let's limit it strictly to HTTP / HTTPS

src/crawlee/_utils/urls.py
```diff
@@ -38,16 +51,73 @@ def to_absolute_url_iterator(base_url: str, urls: Iterator[str], logger: Logger
...
+    assert_never(strategy)
+
+
+@lru_cache(maxsize=1)
+def _get_tld_extractor() -> TLDExtract:
+    """Return a lazily-initialized `TLDExtract` instance shared across the module."""
+    # `mkdtemp` (vs `TemporaryDirectory`) returns a path whose lifetime is tied to the process — `TemporaryDirectory`
+    # is collected immediately when its return value is discarded, which would race the directory out from under
+    # tldextract.
+    return TLDExtract(cache_dir=tempfile.mkdtemp())
```
janbuchar, line 149: 👏 Nice that this is not a part of `BasicCrawler` anymore

janbuchar (approved):

PR #1865, opened 2026-04-30
feat: Add send_request_enqueue_strategy option to BasicCrawler

vdusek:
## Summary

Adds an opt-in `send_request_enqueue_strategy: EnqueueStrategy = 'all'` option to `BasicCrawler` so handlers can reject cross-host `send_request` calls. Default `'all'` preserves current behavior.

Follow-up to #1864, which applied the same `EnqueueStrategy` mechanism to `SitemapRequestLoader`.

## What changed

- New `send_request_enqueue_strategy` option on `BasicCrawler` / `_BasicCrawlerOptions`.
- When set to anything other than `'all'`, `_prepare_send_request_function` validates the target URL against the current request's `loaded_url` (falling back to `url`) via `_check_enqueue_strategy` and raises `ValueError` on mismatch.
- The closure captures the `Request`, so post-redirect `loaded_url` is the origin once set.

### DIFF
--- a/src/crawlee/crawlers/_basic/_basic_crawler.py
+++ b/src/crawlee/crawlers/_basic/_basic_crawler.py
@@ -204,6 +204,12 @@ class _BasicCrawlerOptions(TypedDict):
     """If set to `True`, the crawler will automatically try to fetch the robots.txt file for each domain,
     and skip those that are not allowed. This also prevents disallowed URLs to be added via `EnqueueLinksFunction`."""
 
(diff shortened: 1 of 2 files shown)


Mantisus (approved):
LGTM

B4nan:
Isn't this weird? If the URL is explicitly provided, we should just follow it. If the URL was added to the queue, it must have already respected the enqueue links strategy, or again, be explicitly provided.

PR #1866, opened 2026-05-04
chore: refresh uv.lock and resolve type errors after upgrading ty

vdusek:
## Summary

The `ty` type-checker upgrade to 0.0.34 (in `uv.lock`) surfaced several issues that were either silently passing or guarded by stale ignore directives.

### Changes

- **`Unpack[Union[...]]` is no longer accepted (per PEP 692)**: `BasicCrawler.export_data` now uses a new combined `ExportDataKwargs` TypedDict that inherits from both `ExportDataJsonKwargs` and `ExportDataCsvKwargs` instead of `Unpack[ExportDataJsonKwargs | ExportDataCsvKwargs]`.
- **`ExportDataCsvKwargs.quoting`** tightened from `int` to `Literal[0, 1, 2, 3]` — that's what `csv.writer` actually accepts (the `csv.QUOTE_*` constants).
- **`partial(block_requests, page=...)`** in `PlaywrightCrawler` produced a result whose typed signature (`functools.partial` collapses keyword-bound params into a fully keyword-only signature) didn't satisfy the `BlockRequestsFunction` protocol. Replaced with a `_make_block_requests` static helper that returns an explicit closure.
- **Removed two stale `# ty: ignore[invalid-argument-type]`** directives in `_adaptive_playwright_crawler.py` that ty now flags as unused.
- **Annotated the test `cycle` default** in `_SimpleRenderingTypePredictor` so it produces `Iterator[RenderingType]` instead of `Iterator[str]`.
- **Pre-existing runtime bug fix**: the CSV export example used `quoting='all'`, which raises `TypeError: "quoting" must be an integer, not str` at runtime. Replaced with `csv.QUOTE_ALL`.


src/crawlee/crawlers/_playwright/_playwright_crawler.py
```diff
@@ -558,6 +557,18 @@ def post_navigation_hook(self, hook: Callable[[PlaywrightPostNavCrawlingContext]
         """
         self._post_navigation_hooks.append(hook)
 
+    @staticmethod
```
janbuchar: I'd honestly prefer to see a cast rather than 10 lines of boilerplate with repeated function signatures. This issue feels like more of a `ty` or `typeshed` problem than something we should bend our code around...

tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
```diff
@@ -87,7 +87,8 @@ def __init__(
     ) -> None:
         super().__init__()
 
-        self._rendering_types = rendering_types or cycle(['static'])
+        default_rendering_types: list[RenderingType] = ['static']
+        self._rendering_types = rendering_types or cycle(default_rendering_types)
```
janbuchar, line 89: This also feels like a `ty` regression 🙂

janbuchar (approved):
LGTM, left two non-blocking comments

src/crawlee/crawlers/_playwright/_playwright_crawler.py
vdusek: Yeah, no strong preference here, so replacing it with cast.

PR #1874, opened 2026-05-06
ci: install Apify CLI with pnpm in scheduled e2e tests

vdusek:
### Description

The scheduled E2E tests workflow has been failing on every matrix job at the **Install dependencies** step (latest example: [run 25421420663](https://github.com/apify/crawlee-python/actions/runs/25421420663)).

`npm install -g apify-cli` now fails because the published `apify-cli` package ships a `preinstall` hook (`npx only-allow pnpm`) that refuses non-pnpm installs. On the GitHub runners that hook can't auto-fetch `only-allow` and exits before any test runs:

```
npm error code 127
npm error path /opt/hostedtoolcache/node/22.22.2/x64/lib/node_modules/apify-cli
npm error command sh -c npx only-allow pnpm
npm error sh: 1: only-allow: not found
```

### Changes

- Add a `Setup pnpm` step using `pnpm/action-setup@v4`.
- Install `apify-cli` globally via `pnpm add -g apify-cli` instead of `npm install -g`.

This mirrors the pnpm-based approach already adopted by the docs workflows in #1871.

### DIFF
--- a/.github/workflows/on_schedule_tests.yaml
+++ b/.github/workflows/on_schedule_tests.yaml
@@ -42,8 +42,13 @@ jobs:
         with:
           node-version: ${{ env.NODE_VERSION }}
 
-      - name: Install dependencies
-        run: npm install -g apify-cli
+      - name: Setup pnpm
+        uses: pnpm/action-setup@v4
+        with:
+          version: 10
+
+      - name: Install Apify CLI
+        run: pnpm add -g apify-cli
 
       - name: Set up Python ${{ env.PYTHON_VERSION }}
         uses: actions/setup-python@v6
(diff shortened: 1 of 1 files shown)


B4nan:
We need to fix this in the CLI, we can't enforce pnpm downstream, that would be a pretty crazy move for an OSS project.

B4nan:
On the other hand, we should use pnpm here, so let's merge this if it helps. Just saying that this happening is a bug in the CLI on its own.

janbuchar:
https://github.com/apify/setup-apify-cli-action

vdusek:
@janbuchar please recheck

janbuchar (approved):
LGTM

PR #1875, opened 2026-05-06
ci: inline beta release jobs to fix PyPI trusted publishing

vdusek:
## Summary

PyPI's Trusted Publishing rejects OIDC tokens issued from reusable workflows:

> The claims in this token suggest that the calling workflow is a reusable workflow. Reusable workflows are not currently supported by PyPI's Trusted Publishing.

`on_master.yaml` was invoking `manual_release_beta.yaml` via `uses:`, which made the OIDC token reflect a reusable workflow call. The same fix was applied in `apify/apify-shared-python#63`.

## Changes

- `on_master.yaml`: inline the four beta release jobs (`release_prepare`, `changelog_update`, `pypi_publish`, `doc_release_post_publish`) directly, instead of calling `manual_release_beta.yaml` as a reusable workflow.
- `manual_release_beta.yaml`: remove the `workflow_call` trigger (no longer invoked from another workflow) and add a comment explaining why the duplication exists.
- Drop the unused `tag_name` output from `release_prepare` in both files.

## Follow-up

The PyPI Trusted Publisher for `crawlee` is currently configured for `manual_release_beta.yaml`. After this is merged, an entry for `on_master.yaml` needs to be added on PyPI so the automatic beta release passes verification.

See: https://docs.pypi.org/trusted-publishers/troubleshooting/#reusable-workflows-on-github

### DIFF
--- a/.github/workflows/manual_release_beta.yaml
+++ b/.github/workflows/manual_release_beta.yaml
@@ -2,10 +2,11 @@ name: Beta release
 
 on:
   # Runs when manually triggered from the GitHub UI.
+  # Note: This workflow is intentionally NOT a reusable workflow (no `workflow_call`)
+  # because PyPI's Trusted Publishing does not currently support reusable workflows.
+  # The same jobs are duplicated in `on_master.yaml` for the automatic beta release on push to master.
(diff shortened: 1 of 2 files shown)


vdusek:
merging it, as I have a feat PR in the pipeline that I can use for testing it

PR #1878, opened 2026-05-06
fix(redis): Prevent counter corruption from concurrent mark handled in Redis RQ

### LINKED ISSUE #1873: Negative pending request count when using RedisRequestQueueClient
### Background

I am using a Sitemap crawler with the Redis storage client to manage the request queue. 

### Bug

During a long running job, the `pending_request_count` went negative, preventing the crawler from terminating.

As you can see from my Redis state:

```
LLEN  request_queues:<name>:queue           = 0
HLEN  request_queues:<name>:in_progress     = 0
SCARD request_queues:<name>:pending_set     = 0
JSON.GET ...:metadata $.pending_request_count   = -6
JSON.GET ...:metadata $.total_request_count     = 89582
JSON.GET ...:metadata $.handled_request_count   = 89588   # > total
```

The `pending_request_count = -6`, this causes the `RedisRequestQueueClient.is_empty()` to return `False`:

```python
return metadata.pending_request_count == 0
```

This results in crawlee continuing to endlessly query redis every 1-3ms (across each connection).

### Possible cause

`mark_request_as_handled` guards on `hexists(in_progress, unique_key)` in a separate round-trip from the pipeline that does `delta_handled_request_count=+1, delta_pending_request_count=-1`. If two coroutines race on the same `unique_key` (or a retry/error path calls it twice), both `hexists` checks can pass before either pipeline executes, and the counters get double-applied.

Mantisus:
### Description

- Fixes counter corruption on concurrent `mark_request_as_handled` calls for the same request. When two coroutines concurrently called `mark_request_as_handled` for the same request, both could pass the `hexists` check before either executed the pipeline, causing counters to be updated multiple times for the same request, breaking the queue operation.

### Issues

- Closes: #1873

### Testing

- Added a test to verify that counters are updated correctly during concurrent execution.
- Added a test to verify that the request is correctly restored to `in_progress` after failure.

### Checklist

- [ ] CI passed

### DIFF
--- a/src/crawlee/storage_clients/_redis/_request_queue_client.py
+++ b/src/crawlee/storage_clients/_redis/_request_queue_client.py
@@ -396,9 +396,12 @@ class RedisRequestQueueClient(RequestQueueClient, RedisClientMixin):
     @retry_on_error(RedisError)
     @override
     async def mark_request_as_handled(self, request: Request) -> ProcessedRequest | None:
-        # Check if the request is in progress.
-        check_in_progress = await await_redis_response(self._redis.hexists(self._in_progress_key, request.unique_key))
-        if not check_in_progress:
+        # `hdel` is an atomic operation, so we can be sure that if it returns 1, the request was in progress and is now
+        # removed from `in_progress`.
+        deleted = await await_redis_response(self._redis.hdel(self._in_progress_key, request.unique_key))
+
+        # If not deleted, the request was not in progress.
+        if not deleted:
             logger.warning(f'Marking request {request.unique_key} as handled that is not in progress.')
             return None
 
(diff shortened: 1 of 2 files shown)


src/crawlee/storage_clients/_redis/_request_queue_client.py
```diff
@@ -396,41 +396,56 @@ async def get_request(self, unique_key: str) -> Request | None:
...
-        )
+            return ProcessedRequest(
+                unique_key=request.unique_key,
+                was_already_present=True,
+                was_already_handled=True,
+            )
+        except Exception:
+            blocked_until = int(datetime.now(tz=timezone.utc).timestamp() * 1000) + self._BLOCK_REQUEST_TIME
+            # If we fail to mark the request as handled after removing it from in_progress, we restore request in
+            # `in_progress` hash.
```
janbuchar, line 440: And that will result in a retry, correct?

janbuchar (approved):

src/crawlee/storage_clients/_redis/_request_queue_client.py
Mantisus: Yes, if it's a Redis-related error, it will trigger a retry using `@retry_on_error(RedisError)`.
If something unexpected happens, the error will be propagated.

vdusek (approved):

PR #1883, opened 2026-05-07
fix: Fall back to drop+recreate when `RequestQueue.purge` is unsupported

vdusek:
## Summary

Some storage clients do not support purging the request queue — most notably the Apify platform client, which raises `NotImplementedError`. Since `BasicCrawler` started calling `request_manager.purge()` automatically on repeated runs (PR #1762), this surfaced as a regression on the Apify platform: a previously working pattern now blew up.

`RequestQueue.purge()` now catches `NotImplementedError`, logs a warning, drops the underlying queue, and re-opens it via `RequestQueue.open()`, swapping the new client/id into the existing instance so callers like `BasicCrawler` keep working transparently. This restores the pre-#1762 behavior on the platform (drop the default queue, let the platform recreate it) without exposing the fallback to callers.

The new `test_purge_falls_back_to_drop_when_not_implemented` covers the path against all four storage backends (memory, file_system, sql, redis).

### DIFF
--- a/src/crawlee/storages/_request_queue.py
+++ b/src/crawlee/storages/_request_queue.py
@@ -153,12 +153,23 @@ class RequestQueue(Storage, RequestManager):
         try:
             await self._client.purge()
         except NotImplementedError:
+            client_name = type(self._client).__name__
+            if self._name is not None:
+                logger.warning(
+                    f'Storage client "{client_name}" does not support purging the request queue. '
+                    f'Skipping purge for named queue "{self._name}" to avoid destroying persistent data; '
+                    'the queue contents are left intact.'
+                )
+                return
             logger.warning(
-                f'Storage client "{type(self._client).__name__}" does not support purging the request queue. '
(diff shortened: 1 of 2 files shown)


src/crawlee/storages/_request_queue.py
```diff
@@ -150,7 +150,17 @@ async def drop(self) -> None:
...
-        await self._client.purge()
+        try:
+            await self._client.purge()
+        except NotImplementedError:
+            logger.warning(
+                f'Storage client "{type(self._client).__name__}" does not support purging the request queue. '
+                'Falling back to dropping and recreating it; the request queue ID may change.'
+            )
+            await self.drop()
+            new_rq = await RequestQueue.open(name=self._name)
```
janbuchar: Doing this to a named queue on Apify actually scares me a fair bit. Maybe we should only do this for unnamed queues (`self._name is None`).

janbuchar (approved):

PR #1888, opened 2026-05-09
feat(cli): Add Adaptive and Stagehand crawler templates

vdusek:
Adds three crawler types to the cookiecutter project template and the scheduled E2E templates tests:

- `adaptive-beautifulsoup` — uses `AdaptivePlaywrightCrawler.with_beautifulsoup_static_parser()`.
- `adaptive-parsel` — uses `AdaptivePlaywrightCrawler.with_parsel_static_parser()`.
- `stagehand` — uses `StagehandCrawler`. The template ships a placeholder API key; the E2E test only validates the build and skips the run.


src/crawlee/project_template/templates/main_stagehand.py
```diff
@@ -0,0 +1,17 @@
...
+from crawlee.crawlers import StagehandCrawler
+# % endblock
+
+# % block instantiation
+crawler = StagehandCrawler(
+    request_handler=router,
+    headless=True,
+    max_requests_per_crawl=10,
+    stagehand_options=StagehandOptions(
+        model_api_key='<YOUR_OPENAI_API_KEY>',
```
janbuchar: It's fairly reasonable to try an environment variable here (`os.environ['OPENAI_API_KEY']`). An added benefit of that is that it will pick up an API key in the CI. If we add one 😁

janbuchar (approved):

PR #1889, opened 2026-05-11
fix: Update `push_data` and `user_data` annotation  with `JsonSerializable` instead of `Any`

### LINKED ISSUE #1191: Update `push_data` annotations to use `JsonSerializable` type

Mantisus:
### Description

- Improved annotation for arguments that accept JSON data by replacing implicit `Any` with explicit `JsonSerializable` type for `push_data` and `user_data` parameters.

### Issues

- Closes: #1191

### DIFF
--- a/src/crawlee/_types.py
+++ b/src/crawlee/_types.py
@@ -198,7 +198,7 @@ class PushDataKwargs(TypedDict):
 
 
 class PushDataFunctionCall(PushDataKwargs):
-    data: list[dict[str, Any]] | dict[str, Any]
+    data: Sequence[Mapping[str, JsonSerializable]] | Mapping[str, JsonSerializable]
     dataset_id: str | None
     dataset_name: str | None
     dataset_alias: str | None
@@ -300,7 +300,7 @@ class RequestHandlerRunResult:
 
     async def push_data(
         self,
-        data: list[dict[str, Any]] | dict[str, Any],
+        data: Sequence[Mapping[str, JsonSerializable]] | Mapping[str, JsonSerializable],
         dataset_id: str | None = None,
         dataset_name: str | None = None,
         dataset_alias: str | None = None,
@@ -392,7 +392,7 @@ class EnqueueLinksFunction(Protocol):
         selector: str | None = None,
         attribute: str | None = None,
         label: str | None = None,
-        user_data: dict[str, Any] | None = None,
+        user_data: dict[str, JsonSerializable] | None = None,
         transform_request_function: Callable[[RequestOptions], RequestOptions | RequestTransformAction] | None = None,
         rq_id: str | None = None,
         rq_name: str | None = None,
@@ -417,7 +417,7 @@ class EnqueueLinksFunction(Protocol):
         selector: str | None = None,
         attribute: str | None = None,
         label: str | None = None,
-        user_data: dict[str, Any] | None = None,
+        user_data: dict[str, JsonSerializable] | None = None,
         transform_request_function: Callable[[RequestOptions], RequestOptions | RequestTransformAction] | None = None,
         requests: Sequence[str | Request] | None = None,
         rq_id: str | None = None,
@@ -465,7 +465,7 @@ class ExtractLinksFunction(Protocol):
         selector: str = 'a',
         attribute: str = 'href',
         label: str | None = None,
-        user_data: dict[str, Any] | None = None,
+        user_data: dict[str, JsonSerializable] | None = None,
         transform_request_function: Callable[[RequestOptions], RequestOptions | RequestTransformAction] | None = None,
         **kwargs: Unpack[EnqueueLinksKwargs],
     ) -> Coroutine[None, None, list[Request]]:
@@ -543,7 +543,7 @@ class PushDataFunction(Protocol):
 
     def __call__(
         self,
-        data: list[dict[str, Any]] | dict[str, Any],
+        data: Sequence[Mapping[str, JsonSerializable]] | Mapping[str, JsonSerializable],
         dataset_id: str | None = None,
         dataset_name: str | None = None,
         dataset_alias: str | None = None,
--- a/src/crawlee/_types.py
+++ b/src/crawlee/_types.py
@@ -27,9 +27,7 @@ if TYPE_CHECKING:
     from crawlee.storage_clients import StorageClient
     from crawlee.storages import KeyValueStore
 
-    # Workaround for https://github.com/pydantic/pydantic/issues/9445
-    J = TypeVar('J', bound='JsonSerializable')
-    JsonSerializable = list[J] | dict[str, J] | str | bool | int | float | None
+    JsonSerializable = dict[str, 'JsonSerializable'] | list['JsonSerializable'] | str | int | float | bool | None
 else:
     from pydantic import JsonValue as JsonSerializable
 
--- a/src/crawlee/_types.py
+++ b/src/crawlee/_types.py
@@ -14,7 +14,7 @@ if TYPE_CHECKING:
     import json
     import logging
     import re
-    from collections.abc import Awaitable, Coroutine, Sequence
+    from collections.abc import Awaitable, Coroutine, MutableMapping, Sequence
 
     from typing_extensions import NotRequired, Required, Self, Unpack
 
@@ -390,7 +390,7 @@ class EnqueueLinksFunction(Protocol):
         selector: str | None = None,
         attribute: str | None = None,
         label: str | None = None,
-        user_data: dict[str, JsonSerializable] | None = None,
+        user_data: Mapping[str, JsonSerializable] | None = None,
         transform_request_function: Callable[[RequestOptions], RequestOptions | RequestTransformAction] | None = None,
         rq_id: str | None = None,
         rq_name: str | None = None,
@@ -415,7 +415,7 @@ class EnqueueLinksFunction(Protocol):
         selector: str | None = None,
         attribute: str | None = None,
         label: str | None = None,
-        user_data: dict[str, JsonSerializable] | None = None,
+        user_data: Mapping[str, JsonSerializable] | None = None,
         transform_request_function: Callable[[RequestOptions], RequestOptions | RequestTransformAction] | None = None,
         requests: Sequence[str | Request] | None = None,
         rq_id: str | None = None,
@@ -463,7 +463,7 @@ class ExtractLinksFunction(Protocol):
         selector: str = 'a',
         attribute: str = 'href',
         label: str | None = None,
-        user_data: dict[str, JsonSerializable] | None = None,
+        user_data: Mapping[str, JsonSerializable] | None = None,
         transform_request_function: Callable[[RequestOptions], RequestOptions | RequestTransformAction] | None = None,
         **kwargs: Unpack[EnqueueLinksKwargs],
     ) -> Coroutine[None, None, list[Request]]:
@@ -614,8 +614,8 @@ class UseStateFunction(Protocol):
 
     def __call__(
         self,
-        default_value: dict[str, JsonSerializable] | None = None,
-    ) -> Coroutine[None, None, dict[str, JsonSerializable]]:
+        default_value: MutableMapping[str, JsonSerializable] | None = None,
+    ) -> Coroutine[None, None, MutableMapping[str, JsonSerializable]]:
         """Call dunder method.
 
         Args:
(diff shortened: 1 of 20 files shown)


vdusek (changes requested):
Could you please try using this type:

```python
JsonSerializable = dict[str, 'JsonSerializable'] | list['JsonSerializable'] | str | int | float | bool | None
"""Recursive type for JSON-serializable values - primitives plus objects and arrays with JSON-serializable contents.

Based on the definition discussed in https://github.com/python/typing/issues/182.
"""
```

All major type checkers support recursive types now, so we can finally type this correctly. I recently made the same change in the API client as well - https://github.com/apify/apify-client-python/blob/master/src/apify_client/_types.py#L30C1-L34C4.

src/crawlee/_utils/file.py
```diff
@@ -150,7 +150,7 @@ async def atomic_write(
 
 
 async def export_json_to_stream(
-    iterator: AsyncIterator[dict[str, Any]],
+    iterator: AsyncIterator[Mapping[str, JsonSerializable]],
```
vdusek, line 153:
```suggestion
    iterator: AsyncIterator[JsonSerializable],
```

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
```diff
@@ -200,7 +200,7 @@ async def extract_links(
             selector: str = 'a',
             attribute: str = 'href',
             label: str | None = None,
-            user_data: dict[str, Any] | None = None,
+            user_data: dict[str, JsonSerializable] | None = None,
```
vdusek:
```suggestion
            user_data: JsonSerializable | None = None,
```

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -941,7 +941,7 @@ async def export_data(
 
     async def _push_data(
         self,
-        data: list[dict[str, Any]] | dict[str, Any],
+        data: Sequence[Mapping[str, JsonSerializable]] | Mapping[str, JsonSerializable],
```
vdusek, line 944:
```suggestion
        data: JsonSerializable,
```

vdusek (commented):
The point of a better `JsonSerializable` type should be that it lets us type its various usages more accurately and with less complexity. Here are a few examples:

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
Mantisus: Thank you for this example.

However, this type is not valid for Crawlee's logic; `user_data` cannot be a list. `dict[str, JsonSerializable]` is the correct type in this case.

src/crawlee/_utils/file.py
Mantisus: Thank you for this example.

Each item in the dataset is always a dictionary-like element, so the signature `Mapping[str, JsonSerializable]` is correct in this case.

src/crawlee/crawlers/_basic/_basic_crawler.py
Mantisus: Thank you for the example and explanation. I will update the signatures where it will be correct.

Mantisus: After reviewing the codebase, I don’t think using `JsonSerializable` would be accurate here. 
The method specifically expects a dataset element to be a `dict[str, JsonSerializable]` or a `list` of them. Using only `JsonSerializable` would also allow `str`, `int`, `float`, or `None`, which is incorrect in this context. 

Overall, I don't see any more places in the codebase where we could further improve or simplify the typing with `JsonSerializable` beyond what has already been done in this PR.

vdusek (commented):
Return value of `iterate_items` is still `AsyncIterator[dict[str, Any]]` in the `FileSystemDatasetClient`, `SqlDatasetClient`, and `RedisDatasetClient`. Is that intended?

In Redis, there is also a relevant cast.

vdusek (commented):
Also see `data` field in the `DatasetItemDb`:

```python
data: Mapped[list[dict[str, Any]] | dict[str, Any]] = mapped_column(JsonField, nullable=False)
```

=>

```python
data: Mapped[dict[str, JsonSerializable]] = mapped_column(JsonField, nullable=False)
```

src/crawlee/_types.py
```diff
@@ -27,9 +27,7 @@
     from crawlee.storage_clients import StorageClient
     from crawlee.storages import KeyValueStore
 
-    # Workaround for https://github.com/pydantic/pydantic/issues/9445
-    J = TypeVar('J', bound='JsonSerializable')
-    JsonSerializable = list[J] | dict[str, J] | str | bool | int | float | None
+    JsonSerializable = dict[str, 'JsonSerializable'] | list['JsonSerializable'] | str | int | float | bool | None
 else:
     from pydantic import JsonValue as JsonSerializable
```
vdusek, line 32: Just wondering whether now we can't just use Pydantic's type for type checking as well, have you tried it? 🙂

src/crawlee/storage_clients/_base/_dataset_client.py
```diff
@@ -91,3 +93,9 @@ async def iterate_items(
         raise NotImplementedError
         if False:
             yield {}
+
+    @staticmethod
+    def _is_list_of_items(
```
vdusek: nit: maybe we can name it "is sequence"? instead of "is list" 🙂

src/crawlee/sessions/_session.py
```diff
@@ -63,7 +65,7 @@ def __init__(
         """
         self._id = id or crypto_random_object_id(length=10)
         self._max_age = max_age
-        self._user_data = user_data or {}
+        self._user_data: dict[str, JsonSerializable] = dict(user_data) if user_data is not None else {}
```
vdusek: no?

```suggestion
        self._user_data: MutableMapping[str, JsonSerializable] = dict(user_data) if user_data is not None else {}
```

src/crawlee/sessions/_models.py
```diff
@@ -24,7 +26,7 @@ class SessionModel(BaseModel):
 
     id: Annotated[str, Field(alias='id')]
     max_age: Annotated[timedelta, Field(alias='maxAge')]
-    user_data: Annotated[dict, Field(alias='userData')]
+    user_data: Annotated[dict[str, JsonSerializable], Field(alias='userData')]
```
vdusek: `Mapping` / `MutableMapping` no?

src/crawlee/crawlers/_playwright/_playwright_crawler.py
```diff
@@ -448,7 +385,7 @@ async def extract_links(
             selector: str = 'a',
             attribute: str = 'href',
             label: str | None = None,
-            user_data: dict | None = None,
+            user_data: dict[str, JsonSerializable] | None = None,
```
vdusek: `Mapping` / `MutableMapping` no?

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1015,7 +1015,7 @@ async def enqueue_links(
             selector: str | None = None,
             attribute: str | None = None,
             label: str | None = None,
-            user_data: dict[str, Any] | None = None,
+            user_data: dict[str, JsonSerializable] | None = None,
```
vdusek: `Mapping` / `MutableMapping` no?

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
vdusek: `Mapping` / `MutableMapping` no?

src/crawlee/_types.py
Mantisus: Yeah, we can. I kept the current solution because I would prefer to have an explicit definition of such a complex type, which is used everywhere. But I won't insist on it. 🙂

Mantisus:
> Also see `data` field in the `DatasetItemDb`:
> 
> ```python
> data: Mapped[list[dict[str, Any]] | dict[str, Any]] = mapped_column(JsonField, nullable=False)
> ```
> 
> =>
> 
> ```python
> data: Mapped[dict[str, JsonSerializable]] = mapped_column(JsonField, nullable=False)
> ```

`SQLAlchemy` uses its own mechanism for handling types and their mapping. This prevents the use of `JsonSerializable`.

src/crawlee/storage_clients/_base/_dataset_client.py
Mantisus: Updated

src/crawlee/sessions/_session.py
Mantisus: Updated

src/crawlee/sessions/_models.py
Mantisus: Updated

src/crawlee/crawlers/_playwright/_playwright_crawler.py
Mantisus: Updated

src/crawlee/crawlers/_basic/_basic_crawler.py
Mantisus: Updated

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
Mantisus: Updated

src/crawlee/_types.py
vdusek: Let's keep it as it is for now.

vdusek (approved):
LGTM, just before merging, could you please prepare a draft PR to the SDK with

```toml
dependencies = [
    # ...
    "crawlee @ git+https://github.com/apify/crawlee-python.git@master",
    # ...
]
```

so we can make sure the new typing doesn't cause any issues there?

janbuchar (approved):

PR #1891, opened 2026-05-11
test: Fix `test_import_error_handled` and add import error tests for all optional extras

### LINKED ISSUE #838: Add tests for import errors where there is missing extra
### Problem

- The [test_import_error_handled](https://github.com/apify/crawlee-python/blob/c58e9730d8d09a75923e8aaa90429f802d81a170/tests/unit/parsel_crawler/test_parsel_crawler.py#L233) is currently not functional with the new `crawlee._utils.try_import.ImportWrapper`. 
- As a result, the test is marked as skipped.

### Solution

- Come up with a way to ensure that `ImportError` is properly raised when required extras are missing.
- Extend this testing approach to cover all extras -  `parsel`, `beautifulsoup4, `playwright`, and `curl-cffi`.

Mantisus:
- Closes: #838

### DIFF
--- a/tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
+++ b/tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
@@ -915,7 +915,7 @@ async def test_adaptive_playwright_crawler_with_sql_storage(test_urls: list[str]
 
 
 @pytest.mark.parametrize(
-    ('optional_module_name'),
+    'optional_module_name',
     [
         pytest.param('playwright', id='playwright'),
         pytest.param('jaro', id='jaro'),
--- a/tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
+++ b/tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import sys
 from dataclasses import dataclass
 from datetime import timedelta
 from itertools import cycle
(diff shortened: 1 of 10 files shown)


tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
```diff
@@ -911,3 +912,29 @@ async def request_handler(_context: AdaptivePlaywrightCrawlingContext) -> None:
         await crawler.run(test_urls[:1])
 
         mocked_handler.assert_called()
+
+
+@pytest.mark.parametrize(
+    ('optional_module_name'),
```
vdusek: `('optional_module_name')` is not a tuple

If it is intended to be a tuple, use `('optional_module_name',)`, if it is intended to be a string, use just `'optional_module_name'` 🙂 .

tests/unit/test_cli.py
```diff
@@ -251,3 +252,28 @@ def test_create_existing_folder_interactive_multiple_attempts(
             'install_project': True,
         },
     )
+
+
+@pytest.mark.parametrize(
+    ('optional_module_name'),
```
vdusek: `('optional_module_name')` is not a tuple

If it is intended to be a tuple, use `('optional_module_name',)`, if it is intended to be a string, use just `'optional_module_name'` 🙂 .

Mantisus: Updated

tests/unit/crawlers/_adaptive_playwright/test_adaptive_playwright_crawler.py
Mantisus: Updated

vdusek (approved):

PR #1893, opened 2026-05-12
ci: Add package pre-publish check

vdusek:
## Summary

Adds an end-to-end check that the built sdist and wheel install cleanly, expose the expected sources and data files (`py.typed`, `project_template/**`, `_redis/lua_scripts/*.lua`), and pass an import + `crawlee create` smoke test — guarding against the silent failure mode behind #1890.

Wired into every PR via a new reusable `_check_package.yaml`, and into both stable and beta release workflows so the exact artifact about to hit PyPI is verified first.

### DIFF
--- a/.github/workflows/_check_package.yaml
+++ b/.github/workflows/_check_package.yaml
@@ -27,7 +27,7 @@ jobs:
         run: uv run poe build
 
       - name: Verify built package
-        uses: apify/actions/python-package-check@main
+        uses: apify/actions/python-package-check@v1.1.0
         with:
           package_name: crawlee
           dist_dir: dist
--- a/.github/workflows/_check_package.yaml
+++ b/.github/workflows/_check_package.yaml
@@ -27,10 +27,9 @@ jobs:
         run: uv run poe build
 
       - name: Verify built package
-        uses: apify/workflows/python-package-check@main
+        uses: apify/actions/python-package-check@feat/python-package-check-action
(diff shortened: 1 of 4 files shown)


.github/workflows/_check_package.yaml
```diff
@@ -0,0 +1,47 @@
...
+            from crawlee.crawlers import (
+                HttpCrawler, BeautifulSoupCrawler, ParselCrawler,
+                PlaywrightCrawler, AdaptivePlaywrightCrawler,
+            )
+            from crawlee.storages import Dataset, KeyValueStore, RequestQueue
+            from crawlee.http_clients import HttpxHttpClient, ImpitHttpClient
+            from crawlee import Request
+            HttpCrawler()
+            BeautifulSoupCrawler()
+            ParselCrawler()
```
janbuchar, line 46: I see this is repeated three times in the diff - is that necessary? Or a lesser evil than setting up a reusable workflow/action?

vdusek: At first, I tried to implement it as a dedicated reusable workflow, but I didn't see much benefit there. This is just a configuration, and each repo has its own smoke test, can have a different rollout timing for new Python versions, and so on. When I prepared it, it felt like an unnecessary middle layer with very little added value. So I removed it and kept only the action itself there.

janbuchar (approved):

PR #1895, opened 2026-05-13
fix: include `sql_mysql` in the `all` extra

vdusek:
## Summary

The `all` meta-extra was missing `sql_mysql`, so installing `crawlee[all]` did not pull in the MySQL storage client dependencies (`aiomysql`, `cryptography`). This adds it alongside the other SQL extras.


Mantisus (approved):
Thanks! LGTM.

PR #1896, opened 2026-05-14
ci: Use custom action to commit changes in CI instead of `git commit`

fnesveda:
We want to enforce commit signing for all commits in our repositories.
To do that, we need to make sure even commits created by CI workflows are signed.

It would be possible to sign using GPG keys, but that would require a lot of maintenance.

Instead, we can commit using the GitHub GraphQL API, which automatically signs commits.

This PR replaces direct `git commit` / `git push` usage (and third-party commit actions like `EndBug/add-and-commit`)
with the `apify/actions/signed-commit` action, which uses the GraphQL API under the hood.

As agreed on Slack, I'm also simplifying the workflows to always work on top of the default branch, rather than on a ref passed down through the input, as it should effectively be the same.


vdusek:
Swapping Pepa with Honza, as Pepa is on the US mission.

vdusek (approved):
LGTM

janbuchar (approved):

PR #1897, opened 2026-05-15
test: bound `serve_in_thread` startup to prevent CI hangs

vdusek:
### Description

The session-scoped `http_server` fixture uses `unused_tcp_port_factory()` to pick a port, then starts uvicorn in a background thread. When another xdist worker grabs that port before this fixture's `bind()` runs, the server thread fails with `[Errno 98] address already in use` and exits without ever setting `server.started = True`. The startup poll loop in `serve_in_thread` had no timeout, so it spun forever and pytest-timeout killed the whole worker 30 minutes later — marking every queued test on that worker as a timeout error.

This is a companion to https://github.com/apify/crawlee-python/pull/1892, which bounded the **teardown**; the same class of bug lived in the **startup**.

### Fix

In `serve_in_thread`, bound the wait with:

1. A 30s deadline (`time.monotonic()` based).
2. A `thread.is_alive()` check that detects when uvicorn has exited (bind failure being the realistic cause).

Either condition raises a `RuntimeError` with a descriptive message, so the fixture fails fast and the failing test surfaces a clear error instead of a 30-minute hang.

### Observed failure

https://github.com/apify/crawlee-python/actions/runs/25895087554/job/76106390899

### DIFF
--- a/tests/unit/server.py
+++ b/tests/unit/server.py
@@ -506,8 +506,8 @@ class TestServer(Server):
     async def restart(self) -> None:
         """Request server restart and wait for it to complete.
 
-        This method can be called from a different thread than the one the server
-        is running on, and from a different async environment.
+        This method can be called from a different thread than the one the server is running on,
+        and from a different async environment.
         """
         self.started = False
         self.restart_requested.set()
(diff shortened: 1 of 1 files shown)


Mantisus (approved):
LGTM

PR #1906, opened 2026-05-20
fix(stagehand): Inject `--no-sandbox` into Stagehand's Chromium launch when sandbox is disabled

vdusek:
## Summary

The scheduled e2e tests for the Stagehand template have failed for several days. After PR #1900 fixed the `OPENAI_API_KEY` propagation, the next layer of failure surfaced — every Stagehand variant in [today's run](https://github.com/apify/crawlee-python/actions/runs/26136278735) failed identically on the Apify platform with:

```
stagehand.InternalServerError: Error code: 500 - {'success': False, 'message': 'connect ECONNREFUSED 127.0.0.1:<port>'}
```

## Root cause

Reproduced locally inside the `apify/actor-python-playwright:3.13` image: Stagehand's `BrowserLaunchOptions` accepts a `chromium_sandbox` field, but the field is **not propagated** to the underlying Chromium launch. When the actor runs as root (which it does on Apify), Chromium silently refuses to start because the setuid sandbox can't initialise — and Stagehand's local SEA server then hits `ECONNREFUSED` when it tries to connect to the missing Chromium CDP port. The regular `PlaywrightCrawler` is unaffected because Playwright's Python binding *does* translate `chromium_sandbox=False` into `--no-sandbox`.

## Fix

In `StagehandBrowserPlugin`, when the sandbox is disabled (`config.disable_browser_sandbox=True`, which Apify auto-sets), append `'--no-sandbox'` to `browser_launch_options['args']` as an explicit workaround. Verified end-to-end in the same actor base image — the crawler now completes requests successfully.

The `apify push` hang from the same scheduled run is a separate flake and is addressed in #1905.

### DIFF
--- a/src/crawlee/browsers/_stagehand_browser_plugin.py
+++ b/src/crawlee/browsers/_stagehand_browser_plugin.py
@@ -88,6 +88,15 @@ class StagehandBrowserPlugin(BrowserPlugin):
         if user_data_dir is not None:
             self._browser_launch_options['user_data_dir'] = str(user_data_dir)
 
+        # Stagehand's BrowserLaunchOptions accept `chromium_sandbox` but the field is not propagated
+        # to the underlying Chromium launch, so containers running as root hit ECONNREFUSED on the CDP port.
+        # Inject `--no-sandbox` explicitly when the sandbox is disabled.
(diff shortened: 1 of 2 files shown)


Mantisus (approved):
LGTM

PR #1912, opened 2026-05-25
ci: Alert Slack on failure across CI workflows

vdusek:
Let's make "Notify Slack on failure" reusable, and reuse it across other places where it makes sense. Apply it to other Python repos once this is merged.

### DIFF
--- a/.github/workflows/on_issue.yaml
+++ b/.github/workflows/on_issue.yaml
@@ -27,17 +27,3 @@ jobs:
               repo: context.repo.repo,
               labels: ["t-tooling"]
             })
-
-  # Send a Slack notification to the team alerting channel when there is a failure.
-  notify_on_failure:
-    name: Notify Slack on failure
-    needs: label_issues
-    if: failure()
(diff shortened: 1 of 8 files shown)


.github/workflows/on_issue.yaml
janbuchar, line 1: Is this really worth a Slack notification? If so, maybe we should just set up an org-level hook to monitor all failed workflow runs...

vdusek: You mean just `on_issue`, right? And the rest is ok?
 
Agreed, it's not really needed there.
 
For the other workflows, I'd like to have them - mostly for myself, so I have a single place that notifies me about failing runs, instead of having to check each repo manually.

janbuchar: As long as it saves work, it's a good thing to have. But if you want to monitor more repositories than just this one, I still think that an organization-level webhook would be less intrusive 🙂

janbuchar (approved):

PR #1794, opened 2026-03-15
docs: add CloakBrowser stealth browser example

Cloak-HQ:
### Description

Add CloakBrowser integration example, following the same pattern as the existing Camoufox example.

[CloakBrowser](https://github.com/CloakHQ/CloakBrowser) is a patched Chromium binary with source-level fingerprint modifications that passes Cloudflare Turnstile, reCAPTCHA v3 (0.9 score), and common bot detection checks where stock Chromium and playwright-stealth get blocked.

The integration uses a custom `PlaywrightBrowserPlugin` that swaps the default Chromium for CloakBrowser's binary. Same plugin pattern as Camoufox — override `new_browser()`, return a `PlaywrightBrowserController`.

### Files

- `docs/examples/code_examples/playwright_crawler_with_cloakbrowser.py` — example code
- `docs/examples/playwright_crawler_with_cloakbrowser.mdx` — docs page

### Testing

Tested on Linux x64 with Crawlee 1.5.0. The plugin launches CloakBrowser's binary, all stealth checks pass (webdriver=false, spoofed GPU/platform/plugins/hardwareConcurrency). Zero failed requests.

### Checklist

- [x] CI passed

PR #1921, opened 2026-05-27
docs: add Invisible Firefox stealth browser example

feder-cr:
opening as draft to check interest first.

this mirrors the CloakBrowser example added in PR #1794 (merged 2026-03-31) for the Firefox side. PR #1794 covers the Chromium-side path; some target sites flag Chromium UA by default and the same approach makes sense on the Firefox side.

invisible_playwright (feder-cr/invisible_playwright) drives a patched Firefox 150 binary (feder-cr/invisible_firefox, MPL-2, same license as Firefox upstream, fingerprint patches at the C++ source code level so no JS shims to detect). it follows the same `pip install` + `ensure_binary()` + `PlaywrightBrowserPlugin` subclass pattern as CloakBrowser, so the integration shape is identical.

file structure exactly mirrors PR #1794:
- one new example file under `docs/guides/code_examples/avoid_blocking/`
- one new section in `docs/guides/avoid_blocking.mdx`
- the same versioned mirror under `website/versioned_docs/version-1.7/`

issues against the backend route to feder-cr/invisible_playwright, not here. tracking discussion: #1920

if not in scope i'll close without noise.


vdusek:
Hi @feder-cr, thanks, we'll check it out.

feder-cr:
@vdusek my bad, forgot to tag you on the original. in case you missed the notif.

vdusek:
Hi @feder-cr, thanks for the PR, but we're going to pass on this one for now. We only want to showcase projects with an established track record and real-world adoption in our docs, and your project is still very new. However, we plan to add a new doc page on building custom extensions (#1936), which third-party integrations like yours could then link to. Closing.

PR #1922, opened 2026-05-27
fix(templates): Install pinned playwright into system env in uv Dockerfile

vdusek:
Since #1904 the `uv` Dockerfile branch uses a standalone uv binary and pins playwright via `uv pip install`. But `UV_PROJECT_ENVIRONMENT` is ignored by `uv pip`, so with no venv and no `--system` it fails with `No virtual environment found` — breaking all Playwright-based `uv` e2e jobs ([run 26485871617](https://github.com/apify/crawlee-python/actions/runs/26485871617)).

Adding `--system` installs the pin into `/usr/local`, the same env `uv sync` populates. Verified with a real `docker build`.


vdusek:
Merging since E2E tests are failing

PR #1927, opened 2026-05-31
feat: provide Request instances in skipped request callbacks

lorenz-lb:
### Description

This PR changes the skipped request handler to receive a `Request` object instead of only the URL `str`.

Skipped URLs are now converted into `Request` objects before the callback is invoked. This ensures that request metadata remains available for skipped requests, including `request.user_data` and other request attributes.

### Issues

- N/A

### Testing

- Added/updated unit tests for skipped request handlers.
- Verified that skipped request handlers receive the original Request object.
- Ran uv run poe check-code.

### Checklist

- [x] CI passed

### Future improvements

While working on this change, I noticed that many parts of the internal processing pipeline operate on `str | Request` unions.

A possible future improvement could be to standardize on `Request` objects internally and only accept `str | Request` at the public API boundary. URLs could then be wrapped into `Request` instances immediately, allowing the rest of the codebase to operate exclusively on `Request` objects.

This could simplify typing, reduce repeated union handling, and make request metadata consistently available throughout the pipeline. However, such a change would require broader refactoring and additional testing, so it is outside the scope of this PR.

### DIFF
--- a/src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
+++ b/src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
@@ -206,6 +206,18 @@ class AbstractHttpCrawler(
             **kwargs: Unpack[EnqueueLinksKwargs],
         ) -> list[Request]:
             requests = list[Request]()
+            skipped = list[Request]()
+
+            def create_request(request_options: RequestOptions) -> Request | None:
+                try:
+                    return Request.from_url(**request_options)
+                except ValidationError as exc:
+                    context.log.debug(
+                        f'Skipping URL "{request_options["url"]}" due to invalid format: {exc}. '
+                        'This may be caused by a malformed URL or unsupported URL scheme. '
+                        'Please ensure the URL is correct and retry.'
+                    )
+                    return None
 
             base_user_data = user_data or {}
 
@@ -226,11 +238,19 @@ class AbstractHttpCrawler(
                 else context.request.loaded_url or context.request.url
             )
             links_iterator = to_absolute_url_iterator(base_url, links_iterator, logger=context.log)
+            skipped_iterator = iter([])
 
             if robots_txt_file:
-                skipped, links_iterator = partition(robots_txt_file.is_allowed, links_iterator)
-            else:
-                skipped = iter([])
+                skipped_iterator, links_iterator = partition(robots_txt_file.is_allowed, links_iterator)
+
+            for url in skipped_iterator:
+                request_options = RequestOptions(
+                    url=url, user_data={**base_user_data}, label=label, enqueue_strategy=strategy
+                )
+                request = create_request(request_options)
+
+                if request is not None:
+                    skipped.append(request)
 
             for url in self._enqueue_links_filter_iterator(links_iterator, context.request.url, **kwargs):
                 request_options = RequestOptions(
@@ -244,17 +264,10 @@ class AbstractHttpCrawler(
                     if transform_request_options != 'unchanged':
                         request_options = transform_request_options
 
-                try:
-                    request = Request.from_url(**request_options)
-                except ValidationError as exc:
-                    context.log.debug(
-                        f'Skipping URL "{url}" due to invalid format: {exc}. '
-                        'This may be caused by a malformed URL or unsupported URL scheme. '
-                        'Please ensure the URL is correct and retry.'
-                    )
-                    continue
+                request = create_request(request_options)
 
-                requests.append(request)
+                if request is not None:
+                    requests.append(request)
 
             skipped_tasks = [
                 asyncio.create_task(self._handle_skipped_request(request, 'robots_txt')) for request in skipped
--- a/src/crawlee/crawlers/_basic/_basic_crawler.py
+++ b/src/crawlee/crawlers/_basic/_basic_crawler.py
@@ -110,7 +110,7 @@ T = TypeVar('T')
 
 ErrorHandler = Callable[[TCrawlingContext, Exception], Awaitable[Request | None]]
 FailedRequestHandler = Callable[[TCrawlingContext, Exception], Awaitable[None]]
-SkippedRequestCallback = Callable[[str, SkippedReason], Awaitable[None]]
+SkippedRequestCallback = Callable[[Request, SkippedReason], Awaitable[None]]
 
 
 class _BasicCrawlerOptions(TypedDict):
@@ -1210,17 +1210,15 @@ class BasicCrawler(Generic[TCrawlingContext, TStatisticsState]):
                 raise UserDefinedErrorHandlerError('Exception thrown in user-defined failed request handler') from e
 
     async def _handle_skipped_request(
-        self, request: Request | str, reason: SkippedReason, *, need_mark: bool = False
+        self, request: Request, reason: SkippedReason, *, need_mark: bool = False
     ) -> None:
         if need_mark and isinstance(request, Request):
             request.state = RequestState.SKIPPED
             await self._mark_request_as_handled(request)
 
-        url = request.url if isinstance(request, Request) else request
-
         if self._on_skipped_request:
             try:
-                await self._on_skipped_request(url, reason)
+                await self._on_skipped_request(request, reason)
             except Exception as e:
                 raise UserDefinedErrorHandlerError('Exception thrown in user-defined skipped request callback') from e
 
--- a/src/crawlee/crawlers/_playwright/_playwright_crawler.py
+++ b/src/crawlee/crawlers/_playwright/_playwright_crawler.py
@@ -459,6 +459,18 @@ class PlaywrightCrawler(
             The `PlaywrightCrawler` implementation of the `ExtractLinksFunction` function.
             """
             requests = list[Request]()
+            skipped = list[Request]()
+
+            def create_request(request_options: RequestOptions) -> Request | None:
+                try:
+                    return Request.from_url(**request_options)
+                except ValidationError as exc:
+                    context.log.debug(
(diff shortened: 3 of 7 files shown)


src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -110,7 +110,7 @@
 
 ErrorHandler = Callable[[TCrawlingContext, Exception], Awaitable[Request | None]]
 FailedRequestHandler = Callable[[TCrawlingContext, Exception], Awaitable[None]]
-SkippedRequestCallback = Callable[[str, SkippedReason], Awaitable[None]]
+SkippedRequestCallback = Callable[[Request, SkippedReason], Awaitable[None]]
```
vdusek, line 113: This is a breaking change to a public API. Existing callbacks like `async def cb(url: str, reason)` will now receive a `Request`; string usage like `url.startswith(...)` will crash. It should work with both `str` and `Request`.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1210,17 +1210,15 @@ async def _handle_failed_request(self, context: TCrawlingContext | BasicCrawling
...
         if need_mark and isinstance(request, Request):
             request.state = RequestState.SKIPPED
             await self._mark_request_as_handled(request)
 
-        url = request.url if isinstance(request, Request) else request
-
         if self._on_skipped_request:
             try:
-                await self._on_skipped_request(url, reason)
+                await self._on_skipped_request(request, reason)
```
vdusek, line 1221: This now forwards `request` straight to the callback, but `add_requests()` (just below, ~line 841) wasn't updated — it still builds `skipped` from the original `Sequence[str | Request]` and passes a possibly-`str` item here:

```python
for request in requests:
    check_url = request.url if isinstance(request, Request) else request
    if await self._is_allowed_based_on_robots_txt_file(check_url):
        allowed_requests.append(request)
    else:
        skipped.append(request)  # <- can be a plain str
```

So with `respect_robots_txt_file=True`, `await crawler.add_requests(['https://disallowed/...'])` on a robots-disallowed URL delivers a `str` to the callback, and `request.url` (as in the docs example) raises `AttributeError` → `UserDefinedErrorHandlerError`.

Suggest normalizing `str → Request` once at the choke point (here, or in `add_requests`) instead of only in the two `extract_links` impls. That also makes the `isinstance(request, Request)` guard a few lines up dead code (it now contradicts the `request: Request` annotation). This path also has no test coverage.

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
```diff
@@ -206,6 +206,18 @@ async def extract_links(
             **kwargs: Unpack[EnqueueLinksKwargs],
         ) -> list[Request]:
             requests = list[Request]()
+            skipped = list[Request]()
+
+            def create_request(request_options: RequestOptions) -> Request | None:
```
vdusek, line 211: `create_request` is now duplicated verbatim (including the multi-line debug message) with `PlaywrightCrawler.extract_links` (`_playwright_crawler.py:464`). Both crawlers extend `BasicCrawler`, so this looks like a good candidate for a single shared helper in `crawlee/_utils` (taking a `logger`) to keep the two copies from drifting.

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
```diff
@@ -206,6 +206,18 @@ async def extract_links(
...
+                try:
+                    return Request.from_url(**request_options)
+                except ValidationError as exc:
+                    context.log.debug(
+                        f'Skipping URL "{request_options["url"]}" due to invalid format: {exc}. '
+                        'This may be caused by a malformed URL or unsupported URL scheme. '
+                        'Please ensure the URL is correct and retry.'
```
vdusek, line 218: Now that this helper is also used for robots-skipped, auto-discovered links (not just user enqueues), the "Please ensure the URL is correct and retry." wording is misleading — the operator never submitted this URL and there's nothing to retry. Consider a neutral message, or a separate one for the skip path.

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
```diff
@@ -226,11 +238,19 @@ async def extract_links(
                 else context.request.loaded_url or context.request.url
             )
             links_iterator = to_absolute_url_iterator(base_url, links_iterator, logger=context.log)
+            skipped_iterator = iter([])
```
vdusek, line 241: Minor: `skipped_iterator = iter([])` followed by a conditional reassignment inside `if robots_txt_file:` reads a little awkwardly (and `iter([])` infers `Iterator[Never]`). The previous `if/else`, or guarding the skipped-building loop under the same `if robots_txt_file:`, is clearer.

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
```diff
@@ -226,11 +238,19 @@ async def extract_links(
...
+                skipped_iterator, links_iterator = partition(robots_txt_file.is_allowed, links_iterator)
+
+            for url in skipped_iterator:
+                request_options = RequestOptions(
+                    url=url, user_data={**base_user_data}, label=label, enqueue_strategy=strategy
+                )
+                request = create_request(request_options)
+
+                if request is not None:
+                    skipped.append(request)
```
vdusek, line 253: Two things about this skipped-building loop:

1. **Behavior change / silent drop:** previously every robots-disallowed URL reached the callback as a raw string; now they go through `Request.from_url`, which enforces http/https (`validate_http_url`). A disallowed-but-non-http(s) or malformed-but-absolute URL (one that passes `to_absolute_url_iterator` but fails `from_url`) is now silently dropped and the skip callback never fires for it. Intended?

2. **Consistency:** these skipped requests are built directly and don't pass through `transform_request_function`, while the enqueued ones do. So a skipped `Request` carries the base `label`/`user_data`/`enqueue_strategy` but not the user's per-request transform — inconsistent with enqueued requests from the same links.

src/crawlee/crawlers/_playwright/_playwright_crawler.py
```diff
@@ -459,6 +459,18 @@ async def extract_links(
             The `PlaywrightCrawler` implementation of the `ExtractLinksFunction` function.
             """
             requests = list[Request]()
+            skipped = list[Request]()
+
+            def create_request(request_options: RequestOptions) -> Request | None:
```
vdusek, line 464: Same as the HTTP crawler: this `create_request` is a verbatim duplicate (candidate for a shared helper), and the skipped-building loop below has the same silent-drop and `transform_request_function`-not-applied behavior. See the comments on `_abstract_http_crawler.py`.

tests/unit/crawlers/_beautifulsoup/test_beautifulsoup_crawler.py
```diff
@@ -246,18 +246,22 @@ async def request_handler(context: BeautifulSoupCrawlingContext) -> None:
...
+    expected_skip_urls = {
+        str(server_url / 'page_1'),
+        str(server_url / 'page_2'),
+        str(server_url / 'page_3'),
+        str(server_url / 'page_4'),
+    }
+
+    requests = [call.args[0] for call in skip.call_args_list]
+
+    all(isinstance(request, Request) for request in requests)
```
vdusek, line 263: No assert?

tests/unit/crawlers/_parsel/test_parsel_crawler.py
```diff
@@ -330,18 +330,22 @@ async def request_handler(context: ParselCrawlingContext) -> None:
...
+    expected_skip_urls = {
+        str(server_url / 'page_1'),
+        str(server_url / 'page_2'),
+        str(server_url / 'page_3'),
+        str(server_url / 'page_4'),
+    }
+
+    requests = [call.args[0] for call in skip.call_args_list]
+
+    all(isinstance(request, Request) for request in requests)
```
vdusek, line 347: No assert?

tests/unit/crawlers/_playwright/test_playwright_crawler.py
```diff
@@ -765,18 +765,22 @@ async def request_handler(context: PlaywrightCrawlingContext) -> None:
...
+    expected_skip_urls = {
+        str(server_url / 'page_1'),
+        str(server_url / 'page_2'),
+        str(server_url / 'page_3'),
+        str(server_url / 'page_4'),
+    }
+
+    requests = [call.args[0] for call in skip.call_args_list]
+
+    all(isinstance(request, Request) for request in requests)
```
vdusek, line 782: No assert?

vdusek (changes requested):

lorenz-lb:
Thanks for the review! Sorry for some obvious mistakes, I started to use crawlee only recently, clearly skill issue on my side. I'll incorporate your feedback very very soon!

src/crawlee/crawlers/_basic/_basic_crawler.py
janbuchar: I don't think you can start sending `Request` instances into existing callbacks that expect `str` without breaking BC without runtime type inspection — do we really want to go that way?

vdusek: > I don't think you can start sending Request instances into existing callbacks that expect str without breaking BC without runtime type inspection — do we really want to go that way?

You're right. Well, this may be a v2.0 material.

vdusek:
No response from the author for more than 3 weeks, closing.

PR #1938, opened 2026-06-03
fix: Reset private state correctly in sitemap parsers

vdusek:
## Description

Fixes two attribute-name typos in `src/crawlee/_utils/sitemap.py` where the underscore prefix was missing on assignment, so each statement silently created a new public attribute instead of resetting the intended private state:

- `_XMLSaxSitemapHandler.endElement` set `self.current_tag = None` instead of `self._current_tag = None`. The handler therefore never left the "inside a tracked tag" state, so stray text between elements kept being appended to the buffer and a duplicate close tag could re-process stale buffer contents.
- `_TxtSitemapParser.flush` set `self.buffer = ''` instead of `self._buffer = ''`. Reusing the parser after `flush()` concatenated the leftover URL with the next chunk, yielding corrupted URLs like `https://b.com/https://c.com/`.

Both fixes add the missing underscore. Regression tests cover the state reset in the XML handler and the buffer corruption in the TXT parser.

### DIFF
--- a/src/crawlee/_utils/sitemap.py
+++ b/src/crawlee/_utils/sitemap.py
@@ -122,7 +122,7 @@ class _XMLSaxSitemapHandler(ContentHandler):
             elif name == 'changefreq' and text in VALID_CHANGE_FREQS:
                 self._current_url['changefreq'] = text
 
-            self.current_tag = None
+            self._current_tag = None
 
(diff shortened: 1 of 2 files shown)


Mantisus (approved):
Good catch!

vdusek:
@copilot resolve the merge conflicts in this pull request

PR #1939, opened 2026-06-03
fix(redis): Return all items from Redis dataset `get_data` with `desc=True` and `limit=None`

vdusek:
## Description

`RedisDatasetClient.get_data` called with `desc=True, offset=0, limit=None` fell through to the generic descending match arm and built the Redis JSON path `$[:-0]`. Slice semantics treat `-0` as `0`, so the path evaluated to an empty array and a non-empty dataset silently returned zero items, while the symmetric ascending call correctly returned everything.

This adds an explicit `case (True, 0, None)` arm producing `$[:]` (all items, reversed afterwards by the existing `desc` handling), plus a regression test.

### DIFF
--- a/src/crawlee/storage_clients/_redis/_dataset_client.py
+++ b/src/crawlee/storage_clients/_redis/_dataset_client.py
@@ -185,6 +185,8 @@ class RedisDatasetClient(DatasetClient, RedisClientMixin):
         match (desc, offset, limit):
             case (True, 0, int()):
                 json_path += f'[-{limit}:]'
+            case (True, 0, None):
+                json_path += '[:]'
             case (True, int(), None):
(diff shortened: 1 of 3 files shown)


Mantisus (approved):
One suggestion.

vdusek:
> One suggestion.

@Mantisus Which one? :slightly_smiling_face:

Mantisus:
> @Mantisus Which one? 🙂

And it got lost somewhere 🙂

Maybe we should move the test to `storages/test_dataset.py`. I think it would be useful for all storage clients.

Mantisus (approved):
LGTM

janbuchar (approved):

PR #1940, opened 2026-06-03
fix: Do not raise KeyError in parse_sitemap when partial options are provided

vdusek:
## Description

- `ParseSitemapOptions` is declared as `TypedDict(total=False)`, so every key is optional, but `parse_sitemap` read `emit_nested_sitemaps`, `max_depth`, and `sitemap_retries` with direct subscripting. Any caller passing a partial options dict (e.g. just `timeout`) via `Sitemap.load()`, `Sitemap.parse()`, or `parse_sitemap()` hit a `KeyError`.
- Read all options uniformly with `.get()` and the existing defaults, collapsing the duplicated `if/else` branches.
- Add a regression test covering a partial options dict.

### DIFF
--- a/src/crawlee/_utils/sitemap.py
+++ b/src/crawlee/_utils/sitemap.py
@@ -437,17 +437,11 @@ async def parse_sitemap(
     up to the specified maximum depth.
     """
     # Set default options
-    default_timeout = timedelta(seconds=30)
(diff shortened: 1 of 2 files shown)


Mantisus (approved):
LGTM

PR #1941, opened 2026-06-03
fix(memory-storage): Avoid duplicate processed requests in memory request queue client

vdusek:
## Description

- `MemoryRequestQueueClient.add_batch_of_requests` appended a `ProcessedRequest` twice for requests that were already present in the queue but not yet handled — once in the `was_already_present` branch and again at the unconditional append after the if/else block, which that branch did not skip.
- This violated the contract that `AddRequestsResponse.processed_requests` contains exactly one entry per input request and inflated results for any caller iterating the list.
- Removed the inner append so the shared append at the end handles both branches, matching the structure of the file system client. Added a regression test.


Mantisus (approved):
LGTM

janbuchar (approved):

PR #1942, opened 2026-06-03
fix(redis): Preserve shared Redis index hashes when dropping a storage

vdusek:
### Description

`RedisClientMixin._drop` called `pipe.delete(f'{MAIN_KEY}:id_to_name', storage_id)` (and the same for `name_to_id`), but Redis `DEL` deletes whole top-level keys — it does not remove a field from a hash. Since `id_to_name` and `name_to_id` are shared hashes holding the mappings for every storage of a given type, dropping any single Dataset/KVS/RequestQueue wiped the entire index, breaking subsequent `open(id=...)` calls for all sibling storages (open-by-name kept working, since it reads metadata directly).

This replaces the two `DEL` calls with `HDEL`, removing only the dropped storage's fields, and adds a regression test verifying that a sibling storage's index entries survive a drop and that it can still be opened by ID.


Mantisus (approved):
Good catch!

PR #1943, opened 2026-06-03
fix: retry sitemap fetching on error and raise when retries are exhausted

vdusek:
## Description

The `try/except` in `_fetch_and_process_sitemap` wrapped the entire `while retries_left > 0` loop instead of the loop body. Any exception raised while streaming a sitemap therefore exited the loop after a single attempt, logged a warning, and silently ended the async generator — no retry ever happened, `sitemap_retries` was effectively dead code, and callers (`Sitemap.load`, `parse_sitemap`, `SitemapRequestLoader`) saw empty or partial results with no error.

This PR moves the `try/except` inside the loop body so each attempt is independently retried, and raises the error once all retries are exhausted instead of swallowing it.

### Behavior notes

- A sitemap that keeps failing after all retries now raises to the caller instead of silently producing empty results. `SitemapRequestLoader` already logs and propagates this via its own error handling.
- If a stream fails after yielding some items, the retry re-fetches the sitemap from scratch, so items yielded before the failure may be yielded again. This is pre-existing streaming-generator semantics, now just reachable via the (working) retry path.

### DIFF
--- a/src/crawlee/_utils/sitemap.py
+++ b/src/crawlee/_utils/sitemap.py
@@ -321,9 +321,9 @@ async def _fetch_and_process_sitemap(
 
     sitemap_url = source['url']
 
-    try:
-        while retries_left > 0:
-            retries_left -= 1
+    while retries_left > 0:
+        retries_left -= 1
+        try:
             async with http_client.stream(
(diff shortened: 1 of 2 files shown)


Mantisus (approved):
LGTM

janbuchar (approved):

PR #1950, opened 2026-06-05
fix: validate storage name and alias values

vdusek:
### Summary

Storage aliases were not validated the same way storage names are. This change:

- Adds validation for storage **aliases** (previously unvalidated), allowing the same simple identifier characters used for names plus the reserved `__default__` alias.
- Ensures a storage name or alias always maps to a single subdirectory inside the configured storage directory, both via the high-level API and when the file-system storage clients are used directly.

Includes unit tests for the new validation and for the file-system clients.

### DIFF
--- a/src/crawlee/_utils/file.py
+++ b/src/crawlee/_utils/file.py
@@ -63,9 +63,9 @@ else:
 def validate_subdirectory(base_dir: Path, subdirectory: str) -> Path:
     """Resolve a storage subdirectory inside a base directory.
 
-    Joins `subdirectory` onto `base_dir` and verifies that the result stays inside `base_dir`, so a storage
-    name or alias always maps to a directory under the storage directory rather than somewhere else (e.g. a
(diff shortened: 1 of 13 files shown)


Pijukatel (commented):
One thing confuses me.

This alias: `with/slash`

- It is not a valid alias at the high-level API. (rejected by `validate_storage_alias`)
- It is a valid alias when using the storage client directly. (approved by `validate_subdirectory`)

Is there a specific reason for it?

vdusek:
> One thing confuses me.
>
> This alias: `with/slash`
>
> - It is not a valid alias at the high-level API. (rejected by `validate_storage_alias`)
> - It is a valid alias when using the storage client directly. (approved by `validate_subdirectory`)
>
> Is there a specific reason for it?

Oh, good catch, it should be rejected by both of course, I'll fix it.

Pijukatel (approved):

PR #1955, opened 2026-06-08
fix: derive Pydantic camelCase aliases via to_camel generator

### LINKED ISSUE #1954: Use `alias_generator=to_camel` on Pydantic models instead of manual per-field aliases
### Background

We're on Pydantic ≥2.11. Many models map snake_case Python fields to camelCase external keys (storage/state/API JSON), and today every field spells its alias out by hand. For example in `src/crawlee/statistics/_models.py`:

```python
stats_id: Annotated[int | None, Field(alias='statsId')] = None
requests_finished: Annotated[int, Field(alias='requestsFinished')] = 0
crawler_started_at: Annotated[datetime | None, Field(alias='crawlerStartedAt')] = None
```

### Proposal

Set an alias generator once in the model config so the camelCase aliases are derived automatically:

```python
from pydantic.alias_generators import to_camel

model_config = ConfigDict(validate_by_name=True, validate_by_alias=True, alias_generator=to_camel)
```

Fields then drop the manual alias (`stats_id: int | None = None`) and still (de)serialize as `statsId`. We already standardize on `validate_by_name=True, validate_by_alias=True` (the 2.11 successors of `populate_by_name`), so the only new piece is `alias_generator`. Best placed on a shared base model, since subclasses currently redeclare the config.

### Where it makes sense

- Apply to models that already use camelCase aliases: `Request`, `SessionModel`, `StatisticsState`, the `storage_clients/models.py` set, event data classes, etc.
- Keep an explicit `Field(alias=...)` wherever `to_camel` produces the wrong key (acronyms / irregular casing). Each field's generated alias must equal the current one.
- Skip models whose serialized form is snake_case, e.g. `RequestQueueState` in `_file_system/_request_queue_client.py`, and `Configuration` (`BaseSettings` with env-var `AliasChoices`).

### Scope

Refactor only. The wire format must not change; generated aliases have to match the existing ones, guarded by the current (de)serialization tests.

vdusek:
Sets `alias_generator=to_camel` in each Pydantic model's `ConfigDict` so snake_case fields (de)serialize to camelCase keys automatically, instead of spelling out every `Field(alias=...)`. Applied inline per model (no shared base class).

Explicit `Field(alias=...)` is kept only where `to_camel` yields the wrong key: the `*Millis` durations and `crawlerLastStartTimestamp` in `StatisticsState`, `ProcessedRequest.id` → `requestId`, `EventMigratingData.time_remaining` → `timeRemainingSecs`, and `SitemapRequestLoaderState.completed` → `sitemapCompleted`.

A few fields were previously serialized with snake_case keys; they now use camelCase like everything else: the three `AdaptivePlaywrightCrawlerStatisticState` counters, `EventCrawlerStatusData.crawler_id` → `crawlerId`, and `CrawleeRequestData.session_id` → `sessionId`. Old persisted snake_case data still loads via `validate_by_name`.

Skipped: `Configuration` (env-var `AliasChoices`), `RequestQueueState` (snake_case form), and `UserData` (arbitrary user keys).

Closes #1954

### DIFF
--- a/src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler_statistics.py
+++ b/src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler_statistics.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from typing import Annotated
 
 from pydantic import ConfigDict, Field
+from pydantic.alias_generators import to_camel
 
 from crawlee._utils.docs import docs_group
 from crawlee.statistics import StatisticsState
@@ -12,8 +13,12 @@ from crawlee.statistics import StatisticsState
 class AdaptivePlaywrightCrawlerStatisticState(StatisticsState):
     """Statistic data about a crawler run with additional information related to adaptive crawling."""
 
-    model_config = ConfigDict(validate_by_name=True, validate_by_alias=True, ser_json_inf_nan='constants')
+    model_config = ConfigDict(
+        validate_by_name=True, validate_by_alias=True, alias_generator=to_camel, ser_json_inf_nan='constants'
+    )
 
+    # These fields are serialized with snake_case keys, so they keep explicit aliases that override the
+    # camelCase alias generator.
     http_only_request_handler_runs: Annotated[int, Field(alias='http_only_request_handler_runs')] = 0
     """Number representing how many times static http based crawling was used."""
 
--- a/src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler_statistics.py
+++ b/src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler_statistics.py
@@ -1,8 +1,6 @@
 from __future__ import annotations
 
-from typing import Annotated
-
-from pydantic import ConfigDict, Field
+from pydantic import ConfigDict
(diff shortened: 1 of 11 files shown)


src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler_statistics.py
```diff
@@ -12,8 +13,12 @@
 class AdaptivePlaywrightCrawlerStatisticState(StatisticsState):
     """Statistic data about a crawler run with additional information related to adaptive crawling."""
 
-    model_config = ConfigDict(validate_by_name=True, validate_by_alias=True, ser_json_inf_nan='constants')
+    model_config = ConfigDict(
+        validate_by_name=True, validate_by_alias=True, alias_generator=to_camel, ser_json_inf_nan='constants'
+    )
 
+    # These fields are serialized with snake_case keys, so they keep explicit aliases that override the
```
vdusek: @Pijukatel the `snake_case` alias is a mistake, right?

vdusek:
Well, there might be some compatibility issues; maybe we should resolve this as part of [v2.0](https://github.com/apify/crawlee-python/milestone/271)

PR #1956, opened 2026-06-08
fix: Constrain default sitemap loading

Pijukatel:
### Description

Constrain default sitemap loading.
Allow previous behavior by passing a more relaxed enqueue strategy.

### Testing

- Unit tests

### Checklist

- [ ] CI passed

### DIFF
--- a/src/crawlee/_utils/sitemap.py
+++ b/src/crawlee/_utils/sitemap.py
@@ -246,19 +246,19 @@ async def _process_sitemap_item(
     # Handle sitemap URL references (nested sitemaps)
     if item_type == 'sitemap_url' and 'url' in item_copy:
         sitemap_url = item_copy['url']
-        parent_url = source.get('url')
-        if sitemap_url and sitemap_url not in visited_sitemap_urls and parent_url:
-            ok, reason = filter_url(target=sitemap_url, strategy=enqueue_strategy, origin=parent_url)
-            if not ok:
-                logger.warning(f'Skipping nested sitemap {sitemap_url!r} (parent {parent_url!r}): {reason}.')
-                return
+        if sitemap_url and sitemap_url not in visited_sitemap_urls:
+            if parent_url := source.get('url'):
+                ok, reason = filter_url(target=sitemap_url, strategy=enqueue_strategy, origin=parent_url)
+                if not ok:
+                    logger.warning(f'Skipping nested sitemap {sitemap_url!r} (parent {parent_url!r}): {reason}.')
+                    return
 
             # Add to processing queue
             sources.append(SitemapSource(type='url', url=sitemap_url, depth=depth + 1))
 
             # Output the nested sitemap reference if requested
             if emit_nested_sitemaps:
-                yield NestedSitemap(loc=sitemap_url, origin_sitemap_url=None)
+                yield NestedSitemap(loc=sitemap_url, origin_sitemap_url=parent_url)
 
     # Handle individual URL entries
     elif item_type == 'url' and 'loc' in item_copy:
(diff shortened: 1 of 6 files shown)


src/crawlee/_utils/sitemap.py
```diff
@@ -242,7 +246,13 @@ async def _process_sitemap_item(
     # Handle sitemap URL references (nested sitemaps)
     if item_type == 'sitemap_url' and 'url' in item_copy:
         sitemap_url = item_copy['url']
-        if sitemap_url and sitemap_url not in visited_sitemap_urls:
+        parent_url = source.get('url')
+        if sitemap_url and sitemap_url not in visited_sitemap_urls and parent_url:
```
vdusek: Well, I guess we have a bug here. With the addition of `and parent_url` this whole block becomes a no-op for raw sources (where `source.get('url')` is `None`). As a result, nested `<sitemap>` refs are no longer processed. Am I correct?

tests/unit/_utils/test_sitemap.py
```diff
@@ -115,24 +78,49 @@ def encode_base64(data: bytes) -> str:
...
+        base64=encode_base64(get_basic_sitemap(url=different_url).encode()), c_type='application/xml; charset=utf-8'
+    )
+    sitemap = await Sitemap.load(str(sitemap_url), http_client=http_client)
+
+    assert len(sitemap.urls) == 0
+
+
+async def test_sitemap_different_url_allowed(server_url: URL, http_client: HttpClient) -> None:
+    """Test loading a basic sitemap when sitemap contains links to different url, and it is explicitly allowed."""
+    different_url = 'https://other.com'
```
vdusek: URL `'https://other.com'` has no trailing slash, so the helper builds `https://other.comcatalog?...` (host `other.comcatalog`) instead of a real path. The test still passes (both sides use the same helper), but I guess this is not what we want, right 🙂? 

The same applies to L91.

vdusek (commented):
Two comments, and these 3 bullet points from AI:

- **The default behavior change isn't documented on the public API.** `Sitemap.load` / `Sitemap.parse` / `Sitemap.from_xml_string` now silently drop cross-host URLs, and the only opt-out is the (undocumented) `parse_sitemap_options={'enqueue_strategy': 'all'}`. Worth a line in those docstrings and on the new `ParseSitemapOptions.enqueue_strategy` field.
- **`RobotsTxtFile.parse_sitemaps` / `parse_urls_from_sitemaps` (outside the diff) don't forward `enqueue_strategy`.** They pass it to `get_sitemaps` (which filters the sitemap *list*) but call `Sitemap.load(sitemaps, ...)` with no options, so each sitemap's *contents* now get `same-hostname`-filtered regardless. As a result `parse_urls_from_sitemaps(enqueue_strategy='all')` no longer returns cross-host URLs. Consider threading the strategy through.
- **Minor cleanup:** every loader URL is filtered by `parse_sitemap` *and* again by `_passes_filters` with the same origin (`max_depth=0` ⇒ same parent), so `filter_url` runs twice and can warn twice; and the `same-hostname` default is hardcoded in both `parse_sitemap` and `SitemapRequestLoader.__init__`.

tests/unit/_utils/test_sitemap.py
Pijukatel: Fixed

src/crawlee/_utils/sitemap.py
Pijukatel: Fixed

vdusek (approved):
LGTM

PR #1957, opened 2026-06-09
docs: Add `HTTP headers` guide

### LINKED ISSUE #1923: Expose a browser-impersonation toggle directly on HTTP crawlers
Solutioning follow-up from #1911.

**What happened:** a plain `GET` worked with `requests` but 404'd through `HttpCrawler`. Not a Crawlee bug — all HTTP clients inject an `Accept-Language` header by default (browser impersonation), and this particular API 404s whenever that header is present.

**The friction:** turning impersonation off today means importing a client, disabling it there, and passing it back to the crawler — and the opt-out is named differently per client:

```python
from crawlee.http_clients import ImpitHttpClient

crawler = HttpCrawler(http_client=ImpitHttpClient(browser=None))
# or HttpxHttpClient(header_generator=None)
# or CurlImpersonateHttpClient(impersonate=None)
```

That's a lot of ceremony for a common need.

**Proposal — expose it on the HTTP crawlers.** Add a simple flag so it's a one-liner with no client import:

```python
crawler = HttpCrawler(impersonate=False)
```

Applies to `HttpCrawler`, `BeautifulSoupCrawler`, `ParselCrawler` (the `AbstractHttpCrawler` family). It would configure the default HTTP client under the hood. If the user passes their own `http_client`, the flag doesn't apply — they configure impersonation on that client directly (we should document this clearly, or raise on conflict).

**Also:** a short docs guide on default browser-like headers (esp. `Accept-Language`) — when impersonation helps vs. hurts.

Related: #1683, #1685, #1911

Mantisus:
### Description

- Add a guide on working with HTTP headers in web scraping (`docs/guides/http_headers.mdx`) with a runnable example.

### Issues

- Closes: #1923


Pijukatel (commented):
Hi, I think it would be better to do just a documentation change and keep the current implementation.

I wrote the reasons into the issue, as the current wording of the issue is asking for a code change.

https://github.com/apify/crawlee-python/issues/1923#issuecomment-4666858773

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
```diff
@@ -72,8 +73,20 @@ def __init__(
         *,
         parser: AbstractHttpParser[TParseResult, TSelectResult],
         navigation_timeout: timedelta | None = None,
+        impersonate: bool = True,
```
vdusek: I guess it should also be added to `HttpCrawlerOptions` TypedDict.

src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
```diff
@@ -86,6 +99,9 @@ def __init__(
                 'AbstractHttpCrawler._create_static_content_crawler_pipeline() method to initialize it.'
             )
 
+        if impersonate is False and 'http_client' not in kwargs:
```
vdusek: We should add a warning log if incompatible arguments are used simultaneously, and clearly communicate which one takes precedence.

vdusek (commented):
Two comments from my side.

---

And regarding this:

> Hi, I think it would be better to do just a documentation change and keep the current implementation.
>
> I wrote the reasons into the issue, as the current wording of the issue is asking for a code change.

Exposing high-level convenience arguments on crawlers, which configure the underlying components, is a Crawlee design choice. And we follow this all over the place - crawlers (and Actor in SDK) already act as partial facades over the components they compose. A few examples:

* `PlaywrightCrawler` - `headless`, `browser_type`, `browser_launch_options`, `use_incognito_pages`, `user_data_dir`, and similar BrowserPool/plugin internals directly on the crawler. With your approach, there should be only `browser_pool`.
* `StagehandCrawler` follows the same pattern.
* `BasicCrawler` has `use_session_pool: bool` next to the `session_pool` object, which is the same shape as `impersonate`.

So `impersonate` is not introducing a new pattern. It follows the same approach we already use for browsers, sessions, proxy configuration, concurrency, and more.

This design decision was made a long time ago, and we should be consistent and follow it, rather than diverging from it. And AFAIK @B4nan is a strong proponent of this approach.

> This creates the ugly edge case `HttpCrawler(impersonate=False, http_client=...)`.

We should simply validate the argument combination and raise an error when it is invalid, like in other places.

...

TLDR; A convenience flag (with guard and documentation) is consistent with the rest of Crawlee. It also keeps the simple case simple: users can turn off impersonation without needing to know what an HTTP client is.

Pijukatel:
> ...
Exposing high-level convenience arguments on crawlers, which configure the underlying components, is a Crawlee design choice.
...

Those internal components usually have more arguments than what is exposed on the Crawler level, and many internal component arguments remain unexposed (which is fine). I do not think we have sufficient evidence to say that this specific internal component argument is so useful for the general user base that it deserves to be exposed on the Crawler level. JS tooling has been around for a while. Was anyone missing such an argument?

We should exercise restraint when exposing those convenient arguments. The more we have, the harder it is to understand the code.

Mantisus:
Regarding the `impersonate` flag. To me, this situation is similar to #1487. Both cases require only a minor configuration change on the user’s part, and if it were entirely up to me, I would limit myself to providing documentation.

But since we’re already taking the approach of "giving the user a simple configuration option", this PR is fully consistent with that approach.

vdusek (approved):
LGTM from my side, but since we don't have consensus here yet, let's gather more input - @janbuchar, what do you think?

If we decide not to include this flag at the `HttpCrawler` level, what about adding it at the HTTP client level at least? It could just be a simple flag to turn this behavior on or off completely. Right now, different HTTP clients handle this differently (have different interfaces (arguments)).

janbuchar:
See [crawlee v4's session management guide](https://github.com/apify/crawlee/blob/v4/docs/guides/session_management.mdx). The way to fine-tune how fingerprint impersonation (is that a good way to call it? :thinking:) is already there and it's different and arguably more capable.

Then again, achieving full parity will be a major undertaking. But from my side, adding a single flag that stops working once you set up a custom http client is not optimal and we'd have to deprecate it once we have the new session management in place.

Mantisus:
> See [crawlee v4's session management guide](https://github.com/apify/crawlee/blob/v4/docs/guides/session_management.mdx). The way to fine-tune how fingerprint impersonation (is that a good way to call it? 🤔) is already there and it's different and arguably more capable.

If we plan to implement this in v2, then I would limit this PR to documentation only.

@vdusek, what do you think about that?

vdusek:
@janbuchar Thanks for your input.

@Mantisus Yeah, let's update the docs only. Thanks.

docs/guides/code_examples/http_headers/set_headers.py
```diff
@@ -0,0 +1,33 @@
...
+from crawlee import HttpHeaders, Request
+from crawlee.crawlers import HttpCrawler, HttpCrawlingContext
+from crawlee.http_clients import ImpitHttpClient
+
+
+async def main() -> None:
+    # Set default headers on the client. They are sent on every request.
+    http_client = ImpitHttpClient(headers={'X-Api-Key': 'secret'})
+
+    crawler = HttpCrawler(http_client=http_client)
```
vdusek: All crawler examples set `max_requests_per_crawl`.

```suggestion
    crawler = HttpCrawler(http_client=http_client, max_requests_per_crawl=10)
```

docs/guides/code_examples/http_headers/set_headers.py
```diff
@@ -0,0 +1,33 @@
...
+
+    @crawler.router.default_handler
+    async def request_handler(context: HttpCrawlingContext) -> None:
+        # `httpbin.org/headers` echoes the received request headers back.
+        response = (await context.http_response.read()).decode()
+        context.log.info(response)
```
vdusek: Both requests hit the same URL, so the two log lines are indistinguishable. Maybe we can add `unique_key` prefix so it's clear which response carried the per-request `Accept`?

```suggestion
        context.log.info(f'{context.request.unique_key}: {response}')
```

docs/guides/http_headers.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+Every request a crawler sends includes [HTTP headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers). These headers tell the server who is making the request, what content is acceptable, and in what language. The server reads them and decides what to return. The same URL can return different content, a different status code, or a blocked page depending on the headers it sees. This guide covers the headers that shape a scraping request, like `User-Agent`, `Accept-Language`, and `Content-Type`, what Crawlee sends by default, and how to change them.
+
+## What headers do
+
+Headers are key-value metadata attached to a request. Some of them shape what you get back. Others identify you or carry state. The ones that matter most for scraping are below.
+
+### Identity headers
+
+`User-Agent` identifies the client. Many sites serve different markup to a browser than to a crawler. Some reject requests whose `User-Agent` doesn't look like a real browser. It is one of the basic headers a server uses to identify the client, though not the only one.
```
vdusek: Doc writing style: Prefer a contraction.

```suggestion
`User-Agent` identifies the client. Many sites serve different markup to a browser than to a crawler. Some reject requests whose `User-Agent` doesn't look like a real browser. It's one of the basic headers a server uses to identify the client, though not the only one.
```

docs/guides/http_headers.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+`Referer` says which page the request came from. Some sites gate content, images, or API responses behind an expected `Referer`. A direct request with no `Referer`, or the wrong one, gets a different answer than a click from inside the site.
+
+### Content negotiation
+
+These headers tell the server what the client can handle. The server uses them to pick what to send.
+
+`Accept` lists the formats the client wants. The same endpoint can return HTML to one `Accept` and JSON to another. If you need data from an API, try setting it to `application/json` to get JSON instead of a rendered page.
+
+`Accept-Language` lists the languages the client prefers, in priority order. It is a preference, not a switch. A server honors it only for content it actually serves in more than one language, and ignores it otherwise. Where it applies, it changes translated text, date and number formats, and sometimes currency. Set it to match the locale you expect, then confirm from the response that the server applied it.
```
vdusek: Doc writing style: Prefer a contraction.

```suggestion
`Accept-Language` lists the languages the client prefers, in priority order. It's a preference, not a switch. A server honors it only for content it actually serves in more than one language, and ignores it otherwise. Where it applies, it changes translated text, date and number formats, and sometimes currency. Set it to match the locale you expect, then confirm from the response that the server applied it.
```

docs/guides/http_headers.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+### Headers don't guarantee a result
+
+A header is a request, not a command. The server decides what to do with it. It can ignore a header entirely, so a value you set has no effect on the response. It can reject a value it doesn't accept and answer with an error. And some headers only take effect in combination with others. Setting a header is the first step. Confirm from the response that it did what you expected.
+
+## Default headers in Crawlee
+
+All built-in HTTP clients impersonate a browser by default. Instead of a bare library `User-Agent` like `python-httpx/0.27`, they send a realistic set of browser-like headers: a browser `User-Agent`, an `Accept`, an `Accept-Language`, and client hints where the client supports them. This makes a crawl look like normal browser traffic and avoids the simplest forms of blocking.
```
vdusek: Doc writing style: The sentence opens with a bare "This" as the subject — give it a noun (the verb then agrees as "avoid").

```suggestion
All built-in HTTP clients impersonate a browser by default. Instead of a bare library `User-Agent` like `python-httpx/0.27`, they send a realistic set of browser-like headers: a browser `User-Agent`, an `Accept`, an `Accept-Language`, and client hints where the client supports them. Such headers make a crawl look like normal browser traffic and avoid the simplest forms of blocking.
```

docs/guides/http_headers.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+<ApiLink to="class/HttpxHttpClient">`HttpxHttpClient`</ApiLink> and <ApiLink to="class/CurlImpersonateHttpClient">`CurlImpersonateHttpClient`</ApiLink> take the same `headers` argument.
+
+Header names are case insensitive, and <ApiLink to="class/HttpHeaders">`HttpHeaders`</ApiLink> normalizes the casing for you, so `user-agent` and `User-Agent` refer to the same header.
+
+## Header order and fingerprinting
+
+Anti-bot systems look at more than header values. They look at which headers are present, their casing, and the order they arrive in. Real browsers send a consistent, recognizable set. A request that has a browser `User-Agent` but the wrong header order, or missing client hints, still looks automated.
+
+This is why `ImpitHttpClient` and `CurlImpersonateHttpClient` replicate the browser at the transport layer rather than just attaching headers. Setting a browser `User-Agent` on a plain client is not enough to pass these checks. If a target uses fingerprinting, prefer an impersonating client over hand-set headers.
```
vdusek: Doc writing style: Replace the bare "This" opener with a noun, and contract "is not" → "isn't".

```suggestion
This fingerprinting is why `ImpitHttpClient` and `CurlImpersonateHttpClient` replicate the browser at the transport layer rather than just attaching headers. Setting a browser `User-Agent` on a plain client isn't enough to pass these checks. If a target uses fingerprinting, prefer an impersonating client over hand-set headers.
```

vdusek (commented):
A few comments

vdusek:
Hi @szaganek, could we ask you for a final doc style review? Thank you.

vdusek (approved):
LGTM

docs/guides/http_headers.mdx
```diff
@@ -0,0 +1,122 @@
...
+import ApiLink from '@site/src/components/ApiLink';
+import RunnableCodeBlock from '@site/src/components/RunnableCodeBlock';
+
+import SetHeadersExample from '!!raw-loader!roa-loader!./code_examples/http_headers/set_headers.py';
+
+Every request a crawler sends includes [HTTP headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers). These headers tell the server who is making the request, what content is acceptable, and in what language. The server reads them and decides what to return. The same URL can return different content, a different status code, or a blocked page depending on the headers it sees. This guide covers the headers that shape a scraping request, like `User-Agent`, `Accept-Language`, and `Content-Type`, what Crawlee sends by default, and how to change them.
+
+## What headers do
+
+Headers are key-value metadata attached to a request. Some of them shape what you get back. Others identify you or carry state. The ones that matter most for scraping are below.
```
szaganek:
```suggestion
Headers are key-value metadata attached to a request. Some of them shape what you get back. Others identify you or carry state.
```
Fluff :D

docs/guides/http_headers.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+### Identity headers
+
+`User-Agent` identifies the client. Many sites serve different markup to a browser than to a crawler. Some reject requests whose `User-Agent` doesn't look like a real browser. It's one of the first signals a server reads, though not the only one.
```
szaganek:
```suggestion
`User-Agent` identifies the client. Many sites serve different markup to a browser than to a crawler. Some reject requests whose `User-Agent` doesn't look like a real browser. It's one of the first signals a server reads.
```

docs/guides/http_headers.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+`Referer` says which page the request came from. Some sites gate content, images, or API responses behind an expected `Referer`. A direct request with no `Referer`, or the wrong one, gets a different answer than a click from inside the site.
+
+### Content negotiation
+
+These headers tell the server what the client can handle. The server uses them to pick what to send.
```
szaganek:
```suggestion
These headers tell the server what the client can handle and the server uses them to pick what to send:
```
I'd also turn what follows into bullets.

szaganek (approved):

vdusek:
@Pijukatel Can we merge this?

Pijukatel (approved):

PR #1960, opened 2026-06-11
test: Fix DeprecationWarning and RuntimeWarning emitted by unit tests

vdusek:
### Log from tests:

```
=============================== warnings summary ===============================
tests/unit/_utils/test_system.py::test_memory_estimation_does_not_overestimate_due_to_shared_memory
  /opt/hostedtoolcache/Python/3.14.5/x64/lib/python3.14/multiprocessing/popen_fork.py:76: DeprecationWarning: This process (pid=3748) is multi-threaded, use of fork() may lead to deadlocks in the child.
    self.pid = os.fork()

tests/unit/crawlers/_basic/test_basic_crawler.py::test_global_and_local_event_manager_in_crawler_run
  /home/runner/work/crawlee-python/crawlee-python/tests/unit/crawlers/_basic/test_basic_crawler.py:2151: RuntimeWarning: coroutine 'AsyncMockMixin._execute_mock_call' was never awaited
    handler_call(local_event_manager.active, global_event_manager.active)
  Enable tracemalloc to get traceback where the object was allocated.
  See https://docs.pytest.org/en/stable/how-to/capture-warnings.html#resource-warnings for more info.

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

### Description

The unit-test jobs in CI emit two kinds of warnings:

- `DeprecationWarning: This process is multi-threaded, use of fork() may lead to deadlocks in the child` from `multiprocessing` on Linux with Python 3.12+. It is triggered by `test_system.py` (explicit `fork` context), the `ProcessPoolExecutor` tests in `test_basic_crawler.py`, and the `proxy.py` acceptor processes used by the proxy fixtures, because the default start method on Linux up to Python 3.13 is `fork` and the pytest process is multi-threaded (uvicorn test server thread, xdist).
- `RuntimeWarning: coroutine 'AsyncMockMixin._execute_mock_call' was never awaited` in `test_global_and_local_event_manager_in_crawler_run`, which called an `AsyncMock` without awaiting it.

Changes:

- Set the `spawn` start method in `tests/unit/conftest.py` for all subprocesses created in unit tests. This also aligns Python <= 3.13 runs with Python 3.14, which no longer defaults to `fork`. `proxy.py` has no API for choosing a start method, so the default has to be set globally.
- Refactor the shared-memory estimation test in `test_system.py` to module-level (picklable) child functions running under an explicit `spawn` context instead of `fork`.
- Replace the unawaited `AsyncMock` with a plain `Mock`, since the handler only records call arguments synchronously.

The full unit suite passes with an empty warnings summary.

### DIFF
--- a/tests/unit/_utils/test_system.py
+++ b/tests/unit/_utils/test_system.py
@@ -12,6 +12,14 @@ from crawlee._utils.system import get_cpu_info, get_memory_info
 
 if TYPE_CHECKING:
     from collections.abc import Callable
+    from multiprocessing.sharedctypes import Synchronized
+
+# The `spawn` start method is used for all processes in the shared-memory estimation test. Unlike `fork`, it is safe
+# to use from the multi-threaded pytest process (forking a multi-threaded process is deprecated since Python 3.12),
+# but it requires all process targets to be picklable, module-level functions.
+_ctx = get_context('spawn')
+
+_EXTRA_MEMORY_SIZE = 1024 * 1024 * 100  # 100 MB
 
 
 def test_get_memory_info_returns_valid_values() -> None:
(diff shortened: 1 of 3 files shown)


vdusek:
Closing in favor of #1968. This deserves more time and effort.

PR #1964, opened 2026-06-14
feat: Add PydanticAiCrawler with AI-powered HTML extraction

### LINKED ISSUE #1593: Add support for AI/LLM-based HTML parsing (selectors)
At recent events I attended, I was asked about AI/LLM-based HTML parsing. I also found a few dedicated AI-based scraping frameworks, such as [ScrapeGraphAI](https://github.com/ScrapeGraphAI/Scrapegraph-ai) and [Parsera](https://github.com/raznem/parsera), that appear to be gaining traction.

Right now, we provide an AI-selector workflow only through the `PlaywrightCrawler` via [Stagehand guide](https://crawlee.dev/python/docs/guides/playwright-crawler-stagehand).

This means:
- AI-based selectors are supported only for Playwright, not for HTTP-based crawlers.
- Even for `PlaywrightCrawler`, the integration is not very smooth compared to the tools mentioned above.

Example from the `ScrapeGraphAI`:

```python
# Create the SmartScraperGraph instance
smart_scraper_graph = SmartScraperGraph(
    prompt="Extract useful information from the webpage, including a description of what the company does, founders and social media links",
    source="https://scrapegraphai.com/",
    config=graph_config
)

# Run the pipeline
result = smart_scraper_graph.run()
```

It might be worth exploring a more native solution:
- Better Stagehand integration so that AI-based selectors in Playwright crawlers are as straightforward as in the dedicated AI-scraping libraries.
- Introduce an AI/LLM-powered crawler built on top of `AbstractHttpCrawler`, enabling AI/LLM selectors for HTTP-based scraping as well.

This could make Crawlee more usable for AI/LLM-based extractions, and/or for faster prototype scrapers without manual CSS/XPath selectors.

Mantisus:
### Description

- Adds `PydanticAiCrawler` - a new HTTP crawler that parses pages with [`parsel`](https://parsel.readthedocs.io/en/latest/) and uses [`pydantic-ai`](https://github.com/pydantic/pydantic-ai) as the layer for LLM interaction.
- `PydanticAiHtmlDistiller` is a protocol for distillers that clean HTML and convert it to a compact format (e.g., cleaned HTML, Markdown) for an LLM.
  - `PydanticAiCleanHtmlDistiller` removes comments, noisy attributes, and scripts, returning a compact HTML version.
  - `PydanticAiSkeletonDistiller` extends `PydanticAiCleanHtmlDistiller` by truncating text and collapsing repeated siblings.
- `PydanticAiHtmlExtractor` is a protocol for extractors that turn a page into structured data using a distiller and an LLM.
  - `PydanticAiDirectExtractor` sends the distilled page to an LLM together with a Pydantic schema describing the target data and returns the validated result.
  - `PydanticAiSelectorExtractor` asks the LLM for CSS selectors once and caches them in a `KeyValueStore`, so later pages are extracted without an LLM call.

### Issues

- Closes: #1593

### Testing

- Added new unit tests for `PydanticAiCrawler`, `PydanticAiCleanHtmlDistiller`, `PydanticAiSelectorExtractor`, `PydanticAiDirectExtractor`, and `PydanticAiSelectorExtractor`.


src/crawlee/crawlers/_pydantic_ai/_skeleton_distiller.py
```diff
@@ -0,0 +1,216 @@
...
+
+                saved, dropped = group[: self._keep_siblings], group[self._keep_siblings :]
+
+                for elem in dropped:
+                    parent.remove(elem)
+
+                classes = '.'.join(sig[1])
+                label = f'{sig[0]}.{classes}' if classes else sig[0]
+
+                saved[-1].addnext(etree.Comment(f' ...{len(dropped)} more <{label}> siblings omitted '))
```
vdusek, line 187: Possible IndexError found by Claude:

> `AiSkeletonDistiller(keep_siblings=0)` raises `IndexError` here. The guard on line 169 only skips when `len(group) <= keep_siblings`, so with `keep_siblings=0` any repeated run still proceeds, `saved` becomes `group[:0]` (empty), and `saved[-1].addnext(...)` blows up (verified empirically on a page with a repeated-sibling run). `keep_siblings=0` (collapse all, keep none) is a reasonable request, so either validate it in `__init__` or place the marker without `saved[-1]` when `saved` is empty.

src/crawlee/crawlers/_pydantic_ai/_selector_extractor.py
```diff
@@ -0,0 +1,636 @@
...
+    async def _delegate_to_fallback(
+        self,
+        selector: Selector,
+        schema: type[TSchema],
+        additional_instructions: str | None,
+    ) -> TSchema:
+        if self._fallback is None:
+            raise RuntimeError('Cannot delegate to a fallback extractor because none is configured.')
+        # The scope was already applied to `selector`, so the fallback gets the subtree and no scope.
+        return await self._fallback.extract(
```
vdusek, line 484: From Claude:

> Suggestion: `_delegate_to_fallback` forwards `additional_instructions` but drops `cache_tag` (and `scope`, which the comment explains is intentional). Harmless for the default `AiDirectExtractor` fallback since it has no cache, but a fallback that caches per tag (e.g. another `AiSelectorExtractor`) would bucket under the empty tag instead of the caller's. Consider forwarding `cache_tag` for consistency.

B4nan:
I kinda hate the name `AiCrawler` 🙃

vdusek:
> I kinda hate the name `AiCrawler` 🙃 

Why? Is it too generic / "buzzword-y"?

If so, maybe we could consider something like `PydanticAiCrawler` + `PydanticAiCrawlingContext`. Since this is built on top of Pydantic and PydanticAI. Which might make sense, because it also highlights that selectors can be defined using Pydantic models (similar to `ParselCrawler`, `BeautifulSoupCrawler`, ...).

docs/guides/ai_crawler.mdx
```diff
@@ -0,0 +1,150 @@
...
+
+<ApiLink to="class/AiDirectExtractor">`AiDirectExtractor`</ApiLink> sends the distilled page to the model in one call. The schema is the model's output type. Pydantic AI validates the result; on a mismatch, it sends the error back to the model to fix, bounded by `retries`.
+
+It reads each page on its own, so extraction is accurate per page. It accepts schemas of any shape: nested models, lists, dictionaries, unions, and deep nesting. The cost is one model call per page, which scales poorly on a large site.
+
+Use `additional_instructions` to focus the model on the data you want:
+
+<CodeBlock className="language-python">
+    {AdditionalInstructionsExample}
+</CodeBlock>
```
vdusek: We should make it clear that `AiDirectExtractor` is the default, or explicitly include it in the code example. While reading, I wasn't sure whether it was intentional or if something was missing.

docs/guides/code_examples/ai_crawler/selector_extractor_example.py
```diff
@@ -0,0 +1,56 @@
...
+async def main() -> None:
+    model = OpenAIChatModel(
+        'gpt-5.4-nano',
+        provider=OpenAIProvider(api_key='your-openai-api-key'),
+    )
+    crawler = AiCrawler(
+        extractor=AiSelectorExtractor(
+            model=model,
+            # Pages the cached selectors cannot handle fall back to direct extraction.
+            fallback=AiDirectExtractor(model=model),
```
vdusek: What is the default behaviour? How can I provide a custom fallback? What can it be?

I have these questions after reading the doc guide.

If there are no other options for fallbacks, it can be just a boolean flag.

B4nan:
Yes, exactly, I would rather name it this way so it's clear what it does. AiCrawler sounds like a product name, not a library feature to me.

docs/guides/ai_crawler.mdx
```diff
@@ -0,0 +1,150 @@
...
+
+Both extractors share two more knobs. `retries` caps how many times the model may fix output that fails schema validation (default 1 for <ApiLink to="class/AiDirectExtractor">`AiDirectExtractor`</ApiLink>, 3 for <ApiLink to="class/AiSelectorExtractor">`AiSelectorExtractor`</ApiLink>). `instructions` replaces the base task instructions entirely.
+
+## Distillers
+
+A distiller reduces raw HTML to a compact representation the model reads cheaply. Each extractor uses one. Replace it with the extractor's `distiller` argument (the crawler itself has no `distiller` argument).
+
+<ApiLink to="class/AiDirectExtractor">`AiDirectExtractor`</ApiLink> defaults to an <ApiLink to="class/AiCleanHtmlDistiller">`AiCleanHtmlDistiller`</ApiLink>: cleaned, structure-preserving HTML that keeps the full page text. <ApiLink to="class/AiSelectorExtractor">`AiSelectorExtractor`</ApiLink> uses an <ApiLink to="class/AiSkeletonDistiller">`AiSkeletonDistiller`</ApiLink> internally to ask the model for selectors; you rarely set it yourself.
```
vdusek: What happens if I use `AiCleanHtmlDistiller` in `AiSelectorExtractor`? And/or `AiSkeletonDistiller` in `AiDirectExtractor`? I also don't know what the difference between them is. I would expect this section to follow this structure:

```
## Distillers

...

### AiCleanHtmlDistiller

...

### AiSkeletonDistiller

...

### Custom distiller

...
```

Mantisus:
> I kinda hate the name AiCrawler 🙃

Yeah, me too 😄. I chose it just as a "hyped" marketing name

> If so, maybe we could consider something like PydanticAiCrawler + PydanticAiCrawlingContext.

I agree.

vdusek (commented):
Looks good! A few comments...

src/crawlee/crawlers/_ai/_selector_extractor.py
```diff
@@ -0,0 +1,636 @@
...
+
+    async def __aenter__(self) -> AiSelectorExtractor:
+        """Initialize the selector cache eagerly."""
+        if self._active:
+            raise RuntimeError(f'The {type(self).__name__} is already active.')
+
+        self._active = True
+        if not self._selector_cache.is_initialized:
+            await self._selector_cache.initialize()
+        return self
```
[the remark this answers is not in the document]
Mantisus: Updated

src/crawlee/crawlers/_pydantic_ai/_selector_extractor.py
```diff
@@ -0,0 +1,636 @@
...
+    def _check_fields_covered(plan: SelectorMap, schema: type[BaseModel]) -> None:
+        """Raise `ModelRetry` when the plan misses required schema fields."""
+        missing = [name for name in schema.model_fields if name not in plan.selectors]
+        if missing:
+            raise ModelRetry(f'No selector provided for fields: {missing}')
+
+    @staticmethod
+    def _is_leaf_selector_form(selector: str) -> bool:
+        """Whether a selector targets a value (ends with `::text` or `::attr(...)`)."""
+        return selector.endswith('::text') or '::attr(' in selector
```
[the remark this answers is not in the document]
Mantisus: Updated

src/crawlee/crawlers/_pydantic_ai/_selector_extractor.py
```diff
@@ -0,0 +1,636 @@
...
+
+        Scope and tag are part of the identity. The same schema extracted from a different region or page kind gets
+        its own selector bucket.
+        """
+        return hashlib.sha256(
+            json.dumps(schema.model_json_schema(), sort_keys=True).encode()
+            + (scope or '').encode()
+            + b'\x00'
+            + (cache_tag or '').encode()
+        ).hexdigest()[:16]
```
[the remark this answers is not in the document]
Mantisus: Updated

src/crawlee/crawlers/_pydantic_ai/_skeleton_distiller.py
Mantisus: Fixed

src/crawlee/crawlers/_pydantic_ai/_selector_extractor.py
Mantisus: Fixed

docs/guides/code_examples/ai_crawler/selector_extractor_example.py
Mantisus: The default behaviour without a fallback raises `UnexpectedModelBehavior` if it fails to generate selectors or `ValueError` if a complex, unsupported schema is passed.

A fallback could be another `AiSelectorExtractor` with a different model, distiller, or instructions, or a `AiDirectExtractor` with custom settings. We can't reduce it to a boolean flag.

docs/guides/ai_crawler.mdx
Mantisus: > What happens if I use `AiCleanHtmlDistiller` in `AiSelectorExtractor`?

Using `AiCleanHtmlDistiller` with `AiSelectorExtractor` can improve extraction because the page keeps more data. That likely makes selector generation easier for the model. The price is more tokens that are wasted if the generation still fails.

> And/or `AiSkeletonDistiller` in `AiDirectExtractor`?

`AiSkeletonDistiller` with `AiDirectExtractor` is trickier. If the data is in the text, the model can't read it correctly because the text is truncated. But if the data is in attributes, it works and saves tokens.

The best distiller depends on the site and the task. For example, for news sites where the goal is to extract and organize the visible text, a `MarkdownDistiller` (the custom distiller example) with `PydanticAiDirectExtractor` will probably work best.

Mixing different extractors and distillers lets users find the optimal balance between extraction quality and the resources it costs.

Pijukatel:
Hi, I started reviewing by using the Crawler. It worked fine on the doc examples, so I tried something slightly harder, and I got stuck on something trivial, but hard to debug.

The default model settings ran out of tokens, but PydanticAI does not expose this at all. It hides it behind a model validation error, which is only the consequence of running out of tokens. I had to go deep inside in debug mode to even figure this out.

Example settings that should produce this error even on example code:
```
    model = AnthropicModel(
        'claude-opus-4-8',
        provider=AnthropicProvider(
            api_key='...'),
        settings=ModelSettings(max_tokens=100) # Some obviously small number
    )
```

Having this in the example Crawler, it logs the following error:

> [AiCrawler] WARN  Retrying request to https://crawlee.dev/ due to: Exceeded maximum output retries (1). File "/home/pijukatel/repos/crawlee-python/ai/.venv/lib/python3.14/site-packages/pydantic_ai/_output.py", line 947, in validate,     return self.validator.validate_python(,            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^,         data or {}, allow_partial=pyd_allow_partial, context=validation_context,         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^,     )

This error does not have the root cause anywhere in the exception chain, so it is very misleading. I see this is more of a PydanticAI problem, but still, I think we should solve this as this kind of error will probably be common, and users will have no clue why it happened based on the useless Pydantic validation error.

Here is the debug that uncovers where the root cause is hiding, but PydanticAI never exposes this:
image

Maybe there could be a debug section in the guide that focuses on model responses like that. Also adding some handling around PydanticAI to surface these errors would be nice.

Mantisus:
> This error does not have the root cause anywhere in the exception chain, so it is very misleading. I see this is more of a PydanticAI problem, but still, I think we should solve this as this kind of error will probably be common, and users will have no clue why it happened based on the useless Pydantic validation error.
> Maybe there could be a debug section in the guide that focuses on model responses like that. Also adding some handling around PydanticAI to surface these errors would be nice.

Great idea. I'll add a section on debugging.

If you find this helpful for using the package, use `capture_run_messages`  in situations like this one

```python
from pydantic_ai import capture_run_messages

with capture_run_messages() as messages:
    try:
        article = await context.extract(Article)
    except:
        raise
    finally:
        print('Messages from the model run:')
        for message in messages:
            print(message)
```

This allows you to view interactions with the model.

Pijukatel:
> If you find this helpful for using the package, use `capture_run_messages` in situations like this one

Why not enhance some errors out of the box? Maybe something like:

```
class ExtractorModelError(UnexpectedModelBehavior):
    def __init__(self, *args, finish_reasons, **kwargs) -> None:
        super().__init__(*args, **kwargs)
        self.finish_reasons = finish_reasons

    def __str__(self):
        return f"{super().__str__()}, Model finish reasons: {self.finish_reasons}"
```
Wrapping this either in  extractors or in `extract` that is defined in AiCrawler
```
        with capture_run_messages() as model_messages:
            try:
                ... depeneds on where we plug it...
            except UnexpectedModelBehavior as e:
                raise ExtractorModelError(
                    finish_reasons={
                        model_message.finish_reason for model_message in model_messages if isinstance(model_message, ModelResponse)}, message=e.message
                ) from e
```

Then the error would contain an actually useful hint:

> [AiCrawler] WARN  Retrying request to https://crawlee.dev/ due to: Exceeded maximum output retries (1), Model finish reasons: {'length'}. File "... )

Mantisus:
> Why not enhance some errors out of the box? Maybe something like

I'd prefer to avoid that. Error `UnexpectedModelBehavior` will also occur if the model generates data that fails validation and the number of retries has already been exhausted. There are likely other situations as well that we haven't encountered yet.

```bash
pydantic_ai.exceptions.UnexpectedModelBehavior: Exceeded maximum output retries (1)
```

I also believe that, depending on the model provider, the information provided by `ModelResponse` may vary and further confuse the user.

src/crawlee/crawlers/_pydantic_ai/_clean_html_distiller.py
```diff
@@ -0,0 +1,263 @@
...
+
+if TYPE_CHECKING:
+    from lxml.html import HtmlElement
+    from lxml_html_clean import Cleaner
+
+
+# Attributes that carry selector targets or semantic meaning for an LLM.
+_SEMANTIC_ATTRS = frozenset(
+    {
+        'class',
```
Pijukatel, line 25: Tiny detail just for readability. Please sort it alphabetically.

src/crawlee/crawlers/_pydantic_ai/_pydantic_ai_crawler.py
```diff
@@ -0,0 +1,173 @@
...
+    from typing_extensions import Unpack
+
+    from crawlee import Request
+    from crawlee.crawlers._abstract_http import ParsedHttpCrawlingContext
+
+    from ._types import ExtractFunction, PydanticAiHtmlExtractor, PydanticAiUsageStats, TSchema
+
+
+# Default model
+_DEFAULT_AI_MODEL = 'openai:gpt-5.4-nano'
```
Pijukatel, line 31: Maybe it would be better to have no vendor-specific default model at all. The user has to set up the token in the environment accordingly anyway, so the user already has to be aware of the model, or at least the provider.

src/crawlee/crawlers/_pydantic_ai/_utils.py
```diff
@@ -0,0 +1,28 @@
+from lxml_html_clean import Cleaner
+
+
+def get_basic_http_cleaner(**kwargs: object) -> Cleaner:
+    """Build the default `lxml_html_clean.Cleaner` used by built-in distillers.
+
+    Args:
+        kwargs: Overrides for individual `Cleaner` options.
+    """
+    options: dict = {
```
Pijukatel, line 10: Tiny detail just for readability. Please sort it alphabetically.

Pijukatel (approved):
Nice. Just some minor suggestions.

src/crawlee/crawlers/_pydantic_ai/_selector_extractor.py
```diff
@@ -0,0 +1,661 @@
...
+        logger.debug(
+            f'Selector generation for {schema.__name__} used {run_usage.requests} request(s), '
+            f'{run_usage.input_tokens} input + {run_usage.output_tokens} output tokens.'
+        )
+        return result.output
+
+    @staticmethod
+    def _check_fields_covered(plan: SelectorMap, schema: type[BaseModel]) -> None:
+        """Raise `ModelRetry` when the plan misses required schema fields."""
+        missing = [name for name in schema.model_fields if name not in plan.selectors]
```
vdusek: Potential bug

> `_check_fields_covered` requires a selector for every `model_fields` entry, and `_check_selectors_compile_and_match` (line 466) rejects any selector that matches nothing. So an optional field that's genuinely absent on the page used for generation makes every plan fail → retries exhaust → `UnexpectedModelBehavior` (or silent per-page fallback, defeating the cache). Optional vs. required isn't distinguished at generation time.

src/crawlee/crawlers/_pydantic_ai/_selector_extractor.py
```diff
@@ -0,0 +1,661 @@
...
+            ) from exc
+
+    def _apply_selectors(
+        self,
+        plan: SelectorMap,
+        selector: Selector,
+        schema: type[TSchema],
+    ) -> TSchema:
+        """Run `plan` against `selector` and build a validated `schema` instance."""
+        return schema.model_validate(self._apply_fields(plan.selectors, selector, schema))
```
vdusek: We probably want to use model validation by name

> `_apply_fields` builds the dict keyed by `model_fields` names, but `model_validate` validates by alias when a field defines one. A schema using `Field(alias=...)` without `populate_by_name` fails validation on every page (silently skipped on the cache path, retry-exhausted on generation). The direct extractor is unaffected since pydantic-ai handles aliases there. Consider `model_validate(raw, by_name=True)`.

src/crawlee/crawlers/_pydantic_ai/_pydantic_ai_crawler.py
```diff
@@ -0,0 +1,173 @@
...
+                extract=self._create_extract_function(parsel_context.selector, parsel_context.request),
+                ai_usage=self._ai_usage,
+            )
+
+        kwargs.setdefault('_logger', getLogger(__name__))
+        kwargs['_context_pipeline'] = self._create_static_content_crawler_pipeline().compose(final_step)
+
+        # If the extractor is an async context manager, add it to the crawler's additional context managers so it's
+        # properly entered and exited around the crawl.
+        if isinstance(extractor, AbstractAsyncContextManager):
```
vdusek, line 128: Consider this please:

> Only the top-level extractor is added to `_additional_context_managers`. A `PydanticAiSelectorExtractor` used as a `fallback` (or any extractor used standalone outside `async with`) is reached only via lazy `initialize()`, which registers a `PERSIST_STATE` listener that's never `teardown()`'d — so the listener leaks and the fallback's learned selectors are never persisted on shutdown.

src/crawlee/crawlers/_pydantic_ai/_clean_html_distiller.py
```diff
@@ -0,0 +1,263 @@
...
+
+        return result
+
+    def _parse_and_clean(self, html: str) -> HtmlElement:
+        """Parse raw HTML and run the cleaning stage in place.
+
+        Args:
+            html: The raw HTML markup.
+        """
+        tree = lxml.html.fromstring(html)
```
vdusek, line 131: I would say we should handle this.

> Note: `lxml.html.fromstring` raises `ValueError: Unicode strings with encoding declaration are not supported` for an (X)HTML string beginning with `<?xml ... encoding=...?>`, and `ParserError` on some malformed input. The `not html.strip()` guard at line 112 doesn't catch these, so they propagate and crash the handler instead of degrading.

src/crawlee/crawlers/_pydantic_ai/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+with _try_import(__name__, 'PydanticAiSelectorExtractor'):
+    from ._selector_extractor import PydanticAiSelectorExtractor
+with _try_import(__name__, 'BasePydanticAiHtmlDistiller'):
+    from ._base_distiller import BasePydanticAiHtmlDistiller
+with _try_import(__name__, 'PydanticAiCleanHtmlDistiller'):
+    from ._clean_html_distiller import PydanticAiCleanHtmlDistiller
+with _try_import(__name__, 'PydanticAiSkeletonDistiller'):
+    from ._skeleton_distiller import PydanticAiSkeletonDistiller
+with _try_import(__name__, 'PydanticAiHtmlDistiller', 'PydanticAiHtmlExtractor', 'PydanticAiUsageStats'):
+    from ._types import PydanticAiHtmlDistiller, PydanticAiHtmlExtractor, PydanticAiUsageStats
```
vdusek, line 25: `ExtractFunction` is part of the public docs via `@docs_group('Functions')`, but is not exposed here. Is that intentional?

src/crawlee/crawlers/_pydantic_ai/_pydantic_ai_crawler.py
```diff
@@ -0,0 +1,173 @@
...
+            extractor = PydanticAiDirectExtractor(model if model is not None else _DEFAULT_AI_MODEL)
+
+        # Call the notification only once.
+        warnings.warn(
+            'The PydanticAiCrawler is experimental and its public API may change in future releases.',
+            category=UserWarning,
+            stacklevel=2,
+        )
+
+        self._ai_usage = extractor.ai_usage
```
vdusek: Consider this please

> This snapshots `extractor.ai_usage` once and hands that object to `crawler.ai_usage` and every context. `set_ai_usage()` (public, used to re-share with fallbacks) rebinds the extractor's accumulator, after which `crawler.ai_usage`/`context.ai_usage` point at a stale object and a token-budget guard reads frozen usage. Prefer having the `ai_usage` property return `self._extractor.ai_usage` live.

src/crawlee/crawlers/_pydantic_ai/_pydantic_ai_crawler.py
```diff
@@ -0,0 +1,173 @@
...
+            cache_tag: str | None = None,
+            additional_instructions: str | None = None,
+        ) -> TSchema:
+            # `PydanticAiHtmlExtractor.extract` accepts a Selector directly, so the already-parsed tree is handed over
+            # without a serialize round trip.
+            return await self._extractor.extract(
+                selector,
+                schema,
+                scope=scope,
+                cache_tag=cache_tag if cache_tag is not None else request.label,
```
vdusek, line 168: Consider this

> `cache_tag` defaults to `request.label`, which is `None` for unlabeled requests; `_build_cache_digest` folds that to `''`, so different page kinds crawled without labels collapse into one bucket and churn against `max_variants`. Worth a doc note or a warning.

src/crawlee/crawlers/_pydantic_ai/_base_extractor.py
```diff
@@ -0,0 +1,113 @@
...
+
+        Args:
+            model: A provider-prefixed name (e.g. `'openai:gpt-5.4-nano'`) or a pydantic-ai `Model`. Credentials are
+                read from the provider's environment variable (e.g. `OPENAI_API_KEY`) or passed explicitly through a
+                `Model` instance.
+            distiller: The HTML distiller shaping the LLM input.
+            instructions: Base task instructions. The distiller's prompt notes are appended automatically.
+            usage_limits: Optional pydantic-ai `UsageLimits` applied to every single run.
+        """
+        self._model = infer_model(model)
```
vdusek, line 54: I would say we should handle this

> Note: `infer_model(model)` runs at construction, so `PydanticAiCrawler()` with the default model and no `OPENAI_API_KEY` raises a low-level provider error at instantiation — before the experimental warning a couple lines later. Consider deferring resolution or wrapping it in a Crawlee-level message.

src/crawlee/crawlers/_pydantic_ai/_skeleton_distiller.py
```diff
@@ -0,0 +1,223 @@
...
+
+        Args:
+            tree: The lxml tree.
+        """
+
+        def signature(el: HtmlElement) -> tuple:
+            return (
+                el.tag,
+                tuple(sorted((el.get('class') or '').split())),
+                tuple(el.get(attr) for attr in _IDENTITY_ATTRS),
```
vdusek, line 161: Worth consideration

> The collapse `signature` omits `id` (`_IDENTITY_ATTRS` has none), so consecutive `<section id="pricing">` / `<section id="faq">` with the same class collapse to the first + a marker. A selector the LLM later derives for `#faq` then targets content dropped from the skeleton. Adding `'id'` to `_IDENTITY_ATTRS` would prevent it.

src/crawlee/crawlers/_pydantic_ai/_selector_extractor.py
```diff
@@ -0,0 +1,661 @@
...
+        return raw
+
+    @staticmethod
+    def _build_cache_digest(schema: type[BaseModel], scope: str | None, cache_tag: str | None) -> str:
+        """Build the digest identifying a `(schema, scope, cache_tag)` bucket.
+
+        Scope and tag are part of the identity. The same schema extracted from a different region or page kind gets
+        its own selector bucket.
+        """
+        return hashlib.sha256(
```
vdusek: Worth consideration

> The cache digest hashes `schema.model_json_schema()`, which embeds field `description`/defaults. A cosmetic description edit changes the digest → a fresh bucket (extra paid generation) and invalidates the persisted cache across deploys. Hashing only selector-relevant shape (field names + kinds) would be more stable.

src/crawlee/crawlers/_pydantic_ai/_selector_extractor.py
```diff
@@ -0,0 +1,661 @@
...
+            kind, inner = self._classify_field(info.annotation)
+
+            if kind is _FieldKind.LIST_MODEL and inner is not None:
+                raw[name] = [
+                    self._apply_fields(field_selector.fields or {}, element, inner)
+                    for element in scope.css(field_selector.selector)
+                ]
+            elif kind is _FieldKind.NESTED_MODEL and inner is not None:
+                matched = scope.css(field_selector.selector)
+                raw[name] = self._apply_fields(field_selector.fields or {}, matched[0], inner) if matched else None
```
vdusek: Worth consideration

> A required (non-Optional) `NESTED_MODEL` whose container matches nothing becomes `None` → `ValidationError`, whereas `LIST_MODEL` yields `[]`. On the cache path this silently forces an LLM regeneration for a present-but-empty region. The asymmetry vs. the list case is surprising.

src/crawlee/crawlers/_pydantic_ai/_base_extractor.py
```diff
@@ -0,0 +1,113 @@
...
+            selector: The Parsel selector to query.
+            scope: A CSS selector identifying the region of interest.
+
+        Raises:
+            ValueError: When the scope matches nothing on the page.
+        """
+        scoped = selector.css(scope)
+        if not scoped:
+            raise ValueError(f'Extraction scope {scope!r} matched nothing on the page.')
+        return scoped[0]
```
vdusek, line 116: I would say we should handle the case when scoped's length is more than 1. It can be just a warning log.

src/crawlee/crawlers/_pydantic_ai/_pydantic_ai_crawler.py
```diff
@@ -0,0 +1,173 @@
...
+                cached-selector extraction. Provide at most one of `model` or `extractor`.
+            kwargs: Additional keyword arguments to pass to the underlying `AbstractHttpCrawler`.
+        """
+        if model is not None and extractor is not None:
+            raise ValueError('Provide at most one of `model` or `extractor`.')
+
+        if extractor is None:
+            extractor = PydanticAiDirectExtractor(model if model is not None else _DEFAULT_AI_MODEL)
+
+        # Call the notification only once.
```
vdusek, line 97: Shouldn't there be a guard? Log `warnings.warn` runs on every construction, doesn't it?

docs/guides/pydantic_ai_crawler.mdx
```diff
@@ -0,0 +1,180 @@
...
+crawler = PydanticAiCrawler(model=model)
+```
+
+## Extractors
+
+An extractor turns a page into your schema. Extractors implement different strategies for working with the LLM, and each one uses a <ApiLink to="class/PydanticAiHtmlDistiller">`PydanticAiHtmlDistiller`</ApiLink> to shape the model's input. Crawlee ships two.
+
+### PydanticAiDirectExtractor
+
+<ApiLink to="class/PydanticAiDirectExtractor">`PydanticAiDirectExtractor`</ApiLink> sends the distilled page to the model in one call. The schema is the model's output type. Pydantic AI validates the result; on a mismatch, it sends the error back to the model to fix, bounded by `retries`.
```
vdusek: Doc style thing

```suggestion
<ApiLink to="class/PydanticAiDirectExtractor">`PydanticAiDirectExtractor`</ApiLink> sends the distilled page to the model in one call. The schema is the model's output type. Pydantic AI validates the result. On a mismatch, it sends the error back to the model to fix, bounded by `retries`.
```

docs/guides/pydantic_ai_crawler.mdx
```diff
@@ -0,0 +1,180 @@
...
+
+## Distillers
+
+A distiller reduces raw HTML to a compact representation the model reads cheaply. Each extractor uses one, set through its `distiller` argument (the crawler has no `distiller` argument).
+
+Any distiller pairs with any extractor, that lets you balance extraction quality against token cost. A distiller that keeps more of the page helps the model read values or generate selectors, but costs more tokens. One that keeps only structure and attributes is cheaper, but drops the page text. Pick by your task and the strategy your extractor implements.
+
+### PydanticAiCleanHtmlDistiller
+
+<ApiLink to="class/PydanticAiCleanHtmlDistiller">`PydanticAiCleanHtmlDistiller`</ApiLink> produces cleaned, structure-preserving HTML and keeps the full page text. Scripts and styling are removed; tags, nesting, and data-bearing attributes (`href`, `class`, `datetime`, ...) stay. It is the default for <ApiLink to="class/PydanticAiDirectExtractor">`PydanticAiDirectExtractor`</ApiLink>, that reads the values straight from the document, so it fits pages where the data lives in the text.
```
vdusek: Doc style thing

```suggestion
<ApiLink to="class/PydanticAiCleanHtmlDistiller">`PydanticAiCleanHtmlDistiller`</ApiLink> produces cleaned, structure-preserving HTML and keeps the full page text. Scripts and styling are removed. Tags, nesting, and data-bearing attributes (`href`, `class`, `datetime`, ...) stay. It is the default for <ApiLink to="class/PydanticAiDirectExtractor">`PydanticAiDirectExtractor`</ApiLink>, that reads the values straight from the document, so it fits pages where the data lives in the text.
```

src/crawlee/crawlers/_pydantic_ai/_pydantic_ai_crawler.py
Mantisus: This is for parity with `StagehandCrawler`, the same model and the same provider.

docs/guides/pydantic_ai_crawler.mdx
```diff
@@ -0,0 +1,180 @@
+---
+id: pydantic-ai-crawler
+title: AI crawler
```
vdusek: Please use full name as a title - Pydantic AI crawler

vdusek (commented):
A few more comments

vdusek:
Hi @szaganek, could we ask you for a final doc style review? Thank you.

docs/guides/pydantic_ai_crawler.mdx
```diff
@@ -0,0 +1,180 @@
...
+import UsageLimitExample from '!!raw-loader!./code_examples/pydantic_ai_crawler/usage_limit_example.py';
+
+A <ApiLink to="class/PydanticAiCrawler">`PydanticAiCrawler`</ApiLink> extracts structured data from a page with an LLM. It fetches each page over plain HTTP and parses it with Parsel, then exposes an <ApiLink to="class/ExtractFunction">`extract`</ApiLink> helper: pass a Pydantic model and get a validated instance back. Instead of writing CSS selectors for every field, you describe the data with a schema and the model fills it in.
+
+The model layer is [Pydantic AI](https://pydantic.dev/docs/ai/overview/), so any provider it supports (OpenAI, Anthropic, Gemini, Ollama, ...) works through the `model` argument. The context is a <ApiLink to="class/PydanticAiCrawlingContext">`PydanticAiCrawlingContext`</ApiLink>, that extends the <ApiLink to="class/ParselCrawlingContext">`ParselCrawlingContext`</ApiLink>, so the manual <ApiLink to="class/ParselCrawlingContext#selector">`selector`</ApiLink> and <ApiLink to="class/EnqueueLinksFunction">`enqueue_links`</ApiLink> stay available next to <ApiLink to="class/ExtractFunction">`extract`</ApiLink>.
```
szaganek:
```suggestion
The model layer is [Pydantic AI](https://pydantic.dev/docs/ai/overview/), so any provider it supports (OpenAI, Anthropic, Gemini, Ollama, and more) works through the `model` argument. The context is a <ApiLink to="class/PydanticAiCrawlingContext">`PydanticAiCrawlingContext`</ApiLink>, that extends the <ApiLink to="class/ParselCrawlingContext">`ParselCrawlingContext`</ApiLink>, so the manual <ApiLink to="class/ParselCrawlingContext#selector">`selector`</ApiLink> and <ApiLink to="class/EnqueueLinksFunction">`enqueue_links`</ApiLink> stay available next to <ApiLink to="class/ExtractFunction">`extract`</ApiLink>.
```

docs/guides/pydantic_ai_crawler.mdx
```diff
@@ -0,0 +1,180 @@
...
+
+The `pydantic-ai` extra installs the OpenAI integration by default. To use another provider, add the matching [pydantic-ai-slim](https://pydantic.dev/docs/ai/overview/install/#slim-install) extra. For example, for Anthropic:
+
+```bash
+pip install 'crawlee[pydantic-ai]' 'pydantic-ai-slim[anthropic]'
+```
+
+## Basic usage
+
+Provide a `model` and call <ApiLink to="class/PydanticAiCrawlingContext#extract">`context.extract`</ApiLink> with a Pydantic model inside the handler. With only a `model`, the crawler wraps it in a <ApiLink to="class/PydanticAiDirectExtractor">`PydanticAiDirectExtractor`</ApiLink>, the default extractor, that sends each distilled page to the model in one call. The example below extracts an article and pushes it to the dataset.
```
szaganek:
```suggestion
Provide a `model` and call <ApiLink to="class/PydanticAiCrawlingContext#extract">`context.extract`</ApiLink> with a Pydantic model inside the handler. With only a `model`, the crawler wraps it in a <ApiLink to="class/PydanticAiDirectExtractor">`PydanticAiDirectExtractor`</ApiLink>, the default extractor, that sends each distilled page to the model in one call. The following example extracts an article and pushes it to the dataset.
```

docs/guides/pydantic_ai_crawler.mdx
```diff
@@ -0,0 +1,180 @@
...
+
+An extractor turns a page into your schema. Extractors implement different strategies for working with the LLM, and each one uses a <ApiLink to="class/PydanticAiHtmlDistiller">`PydanticAiHtmlDistiller`</ApiLink> to shape the model's input. Crawlee ships two.
+
+### PydanticAiDirectExtractor
+
+<ApiLink to="class/PydanticAiDirectExtractor">`PydanticAiDirectExtractor`</ApiLink> sends the distilled page to the model in one call. The schema is the model's output type. Pydantic AI validates the result. On a mismatch, it sends the error back to the model to fix, bounded by `retries`.
+
+It reads each page on its own, so extraction is accurate per page. It accepts schemas of any shape: nested models, lists, dictionaries, unions, and deep nesting. The cost is one model call per page, that scales poorly on a large site.
+
+Use `additional_instructions` to focus the model on the data you want:
```
szaganek:
```suggestion
To focus the model on the data you want, use `additional_instructions`:
```

docs/guides/pydantic_ai_crawler.mdx
```diff
@@ -0,0 +1,180 @@
...
+
+### PydanticAiSkeletonDistiller
+
+<ApiLink to="class/PydanticAiSkeletonDistiller">`PydanticAiSkeletonDistiller`</ApiLink> builds on the <ApiLink to="class/PydanticAiCleanHtmlDistiller">`PydanticAiCleanHtmlDistiller`</ApiLink>, then truncates text nodes to short samples and collapses runs of repeated siblings. The output shows the page structure rather than its content, that is what the model needs to write CSS selectors, and its smaller size costs fewer tokens. It is the default for <ApiLink to="class/PydanticAiSelectorExtractor">`PydanticAiSelectorExtractor`</ApiLink>.
+
+### Custom distiller
+
+Subclass <ApiLink to="class/BasePydanticAiHtmlDistiller">`BasePydanticAiHtmlDistiller`</ApiLink> and implement <ApiLink to="class/BasePydanticAiHtmlDistiller#distill">`distill`</ApiLink> to send a different representation. Set `prompt_notes` so the model knows the input format. The extractor appends the notes to its instructions.
+
+The example below converts the cleaned page to Markdown with [html-to-markdown](https://pypi.org/project/html-to-markdown/), an extra dependency:
```
szaganek:
```suggestion
The following example converts the cleaned page to Markdown with [html-to-markdown](https://pypi.org/project/html-to-markdown/), an extra dependency:
```

docs/guides/pydantic_ai_crawler.mdx
```diff
@@ -0,0 +1,180 @@
...
+</CodeBlock>
+
+The errors point at the cause:
+
+- `UnexpectedModelBehavior` - the model couldn't produce a valid result within `retries`, for example its output kept failing validation, a content filter fired, or the token limit cut off a tool call. See [Pydantic AI model errors](https://pydantic.dev/docs/ai/core-concepts/agent/#model-errors).
+- `ModelHTTPError` - the provider returned a 4xx or 5xx response: a missing or wrong API key, a rate limit, or a provider outage.
+- `ValueError` - the schema shape is unsupported by <ApiLink to="class/PydanticAiSelectorExtractor">`PydanticAiSelectorExtractor`</ApiLink>, or a `scope` matched nothing.
+- `UsageLimitExceeded` - a run hit the configured `usage_limits`.
+
+Of these, `UnexpectedModelBehavior`, `ModelHTTPError`, and `UsageLimitExceeded` come from the model run and share the base `AgentRunError`, so catch it to handle any model failure in one place. `ValueError` is raised by the extractor itself, around the schema or `scope`, not by the model.
```
szaganek:
```suggestion
`UnexpectedModelBehavior`, `ModelHTTPError`, and `UsageLimitExceeded` come from the model run and share the base `AgentRunError`, so catch it to handle any model failure in one place. `ValueError` is raised by the extractor itself, around the schema or `scope`, not by the model.
```

szaganek (approved):

janbuchar (approved):
I gave this just a brief read, but I like it. LGTM!

src/crawlee/crawlers/_pydantic_ai/_clean_html_distiller.py
Mantisus: When called from the extractor or within the flow crawler, these errors will not occur, since the HTML has already been normalized in Parsel and passed using `selector.get()`.

src/crawlee/crawlers/_pydantic_ai/__init__.py
Mantisus: Yeah. I'm following the general pattern. We don't expose helper protocols, such as `EnqueueLinksFunction` and `AddRequestsFunction`.

src/crawlee/crawlers/_pydantic_ai/_base_extractor.py
Mantisus: I've updated the order so that `infer_model(model)` is now called after the warning. But I don't think we need to change the error handling, since it depends on the model provider and `pydantic-ai` provides fairly informative errors about missing credentials.

src/crawlee/crawlers/_pydantic_ai/_pydantic_ai_crawler.py
Mantisus: I used an incorrect `stacklevel`, it should be 1. Currently, the warning is logged correctly only once.

src/crawlee/crawlers/_pydantic_ai/_pydantic_ai_crawler.py
Mantisus: Updated docstring.

src/crawlee/crawlers/_pydantic_ai/_skeleton_distiller.py
Mantisus: Added `id` to `_IDENTITY_ATTRS`

src/crawlee/crawlers/_pydantic_ai/_clean_html_distiller.py
vdusek: ah, okay, great

src/crawlee/crawlers/_pydantic_ai/__init__.py
vdusek: Yeah, got it, because there should be no reason for people to import it, if I am not mistaken.

src/crawlee/crawlers/_pydantic_ai/_base_extractor.py
vdusek: Fair enough, thanks

vdusek (approved):
LGTM, thanks 🚀

PR #1978, opened 2026-06-18
chore: switch renovate lock file maintenance to a reviewed PR

vdusek:
Lock file maintenance was effectively stuck. Renovate cannot yet pass an age constraint to `uv lock` (renovatebot/renovate#41654), so `minimumReleaseAge` only added a `renovate/stability-days` check that blocked the merge while `uv lock --upgrade` still resolved to the newest versions. Combined with branch automerge and strict internal checks, the branch was rebased and regenerated on every `master` change, so the check never cleared and the lockfile refresh never landed, leaving transitive dependencies frozen.

This drops `automerge`/`automergeType` and sets `minimumReleaseAge: "0 days"` on `lockFileMaintenance`, so the refresh opens a regular PR for review instead: the PR reliably appears (no stuck stability check) and the transitive bumps get a human/CI check before merge. The top-level `minimumReleaseAge` still gates direct dependency updates.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -9,8 +9,7 @@
     "internalChecksFilter": "strict",
     "lockFileMaintenance": {
         "enabled": true,
-        "automerge": true,
-        "automergeType": "branch"
+        "minimumReleaseAge": "0 days"
     },
     "packageRules": [
         {


B4nan:
> This drops automerge/automergeType and sets minimumReleaseAge: "0 days" on lockFileMaintenance, so the refresh opens a regular PR for review instead: the PR reliably appears (no stuck stability check) and the transitive bumps get a human/CI check before merge. The top-level minimumReleaseAge still gates direct dependency updates.

This is a bad idea, the whole point of the min release age dance (or at least the main issue we are trying to solve) is that we don't want to run fresh deps through our CI (so we can't expose our secrets that are available in the workflows). So unless `uv` would fail to install the deps, we can't go this way. 

Lockfile maintenance itself is pretty simple task (drop old lockfile, generate new one, open a PR), I am sure we could craft our own scheduled workflow to do this, if renovate doesn't support `uv` properly just yet.

PR #1979, opened 2026-06-18
fix: Gracefully close sitemap stream on `SitemapRequestLoader` abort

Mantisus:
### Description

Use `contextlib.aclosing` to let the
stream request finishes gracefully and avoids errors when closing generators.

Example error:
```bash
an error occurred during closing of asynchronous generator <async_generator object HttpxHttpClient.stream at 0x7bad2c3fbe10>
asyncgen: <async_generator object HttpxHttpClient.stream at 0x7bad2c3fbe10>
RuntimeError: aclose(): asynchronous generator is already running
```

### DIFF
--- a/src/crawlee/request_loaders/_sitemap_request_loader.py
+++ b/src/crawlee/request_loaders/_sitemap_request_loader.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import asyncio
 from collections import deque
-from contextlib import suppress
+from contextlib import aclosing, suppress
 from logging import getLogger
 from typing import TYPE_CHECKING, Annotated, Any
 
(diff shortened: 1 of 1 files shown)


vdusek (commented):
Isn't this just a non-optimal workaround? 🙂 

CC suggests utilizing this pattern:

```python
  from contextlib import aclosing, suppress  # already importing suppress here

  # in _load_sitemaps, replace the bare `async for item in parse_sitemap(...)`:
  async with aclosing(
      parse_sitemap(
          [SitemapSource(type='url', url=sitemap_url)],
          self._http_client,
          proxy_info=self._proxy_info,
          options=parse_options,
      )
  ) as sitemap_items:
      async for item in sitemap_items:
          ...
```

Mantisus:
Well, `asyncio.sleep(0)` is pretty common when working with HTTP clients. But yeah, `contextlib.aclosing` is better in this case.

vdusek (approved):
LGTM

Pijukatel (approved):

PR #1982, opened 2026-06-21
fix: Add `is_finished` and correct `is_empty` semantics in `RequestQueueClient`

### LINKED ISSUE #1598: Crawler can get softlocked when consuming the RequestQueue
Example scenario:

There is only 1 request in RQ
The request is processed for a long time - it fails twice, and it will pass only on the third attempt.
After each failure, the request is reclaimed.

The problem is that `AutoscalePool` can go completely rouge and can start burning resources up to the point where it blocks itself due to the system being overloaded.

The root cause is that `__is_task_ready_function` (which decides whether a new task should be scheduled) is checking if the `not await request_manager.is_empty()`. Since there is one request in progress and RQ is not empty, this will return `True` and the `AutoscalePool` will schedule a new task. The problem is that RQ has no available requests as the only one is already in progress. This will create a situation where new tasks are spawned as fast as possible, only to get no request from RQ. 

Preferably, in such scenario `__is_task_ready_function` should return `False`, because the RQ should be aware of the fact that the only request available is already being handled. So `__is_task_ready_function` should return True if there are some available requests and return False if the only requests available are already being processed - but there is no such interface on RQ for this. 

**Bottom lime `not RequestManager.is_empty` does not imply `__is_task_ready`, but our code has this expectation.**

### LINKED ISSUE #1966: Add an `is_finished` method to the `StorageClient` interface
- Reasoning here
    - https://github.com/apify/crawlee/pull/3729#issuecomment-4672024151
    - https://github.com/apify/crawlee/pull/3729#issuecomment-4683157033
- This can be made non-breaking by making the method optional and assuming `is_empty` <=> `is_finished` if `is_finished` is not provided

Mantisus:
### Description

- Adds an `is_finished` method to the `RequestQueueClient` interface. It's not a breaking change. The base class defaults `is_finished` to `is_empty` when it isn't overridden. Splits the queue predicates. `is_empty` means there are no requests available to fetch. `is_finished` means all requests are fully processed (nothing to fetch and nothing in progress).
- Implements `is_finished` in all built-in `RequestQueueClient`.
- Updates `is_empty` in all built-in `RequestQueueClient` to match the new meaning.
- Affects the `AutoscaledPool`. It now skips scheduling new tasks when no requests are available to fetch. For example, while the only request is being processed.

### Issues

- Closes: #1966
- Closes: #1598 

### Testing

- Added a test for `RequestQueue` that checks the `is_empty` and `is_finished` values through the queue lifecycle. It covers all built-in clients.
- Added a regression test that checks the `AutoscaledPool` doesn't schedule new tasks while the only request is being processed.

### DIFF
--- a/src/crawlee/storage_clients/_base/_request_queue_client.py
+++ b/src/crawlee/storage_clients/_base/_request_queue_client.py
@@ -131,9 +131,19 @@ class RequestQueueClient(ABC):
         """
 
     async def is_finished(self) -> bool:
-        """Check if the request queue is finished. That means the queue is empty and no requests are being processed.
+        """Check if the request queue is finished.
+
+        A finished queue is empty and has no requests currently being processed.
+
+        Warning:
+            This default only checks `is_empty`, which reports whether requests are available to fetch, not
+            whether requests are still being processed. It can therefore return `True` while requests are in
+            progress. Subclasses that track in-progress requests should override this method; all built-in
+            clients already do.
 
         Returns:
             True if the request queue is finished, False otherwise.
         """
+        # TODO: Make this method abstract.
+        # https://github.com/apify/crawlee-python/issues/1985
         return await self.is_empty()
--- a/src/crawlee/storage_clients/_base/_request_queue_client.py
+++ b/src/crawlee/storage_clients/_base/_request_queue_client.py
@@ -124,8 +124,16 @@ class RequestQueueClient(ABC):
 
     @abstractmethod
     async def is_empty(self) -> bool:
-        """Check if the request queue is empty.
+        """Check if the request queue is empty. That means there are no requests available to fetch.
 
         Returns:
             True if the request queue is empty, False otherwise.
         """
+
+    async def is_finished(self) -> bool:
+        """Check if the request queue is finished. That means the queue is empty and no requests are being processed.
+
+        Returns:
+            True if the request queue is finished, False otherwise.
+        """
+        return await self.is_empty()
--- a/src/crawlee/storage_clients/_file_system/_request_queue_client.py
+++ b/src/crawlee/storage_clients/_file_system/_request_queue_client.py
@@ -114,6 +114,9 @@ class FileSystemRequestQueueClient(RequestQueueClient):
         self._request_cache_needs_refresh = True
         """Flag indicating whether the cache needs to be refreshed from filesystem."""
 
+        self._is_empty_cache: bool | None = None
+        """Cache for is_empty result: None means unknown, True/False is cached state."""
+
         self._state = recoverable_state
         """Recoverable state to maintain request ordering, in-progress status, and handled status."""
 
@@ -288,6 +291,9 @@ class FileSystemRequestQueueClient(RequestQueueClient):
             self._request_cache.clear()
             self._request_cache_needs_refresh = True
 
+            # Invalidate is_empty cache.
+            self._is_empty_cache = None
+
     @override
     async def purge(self) -> None:
         async with self._lock:
@@ -309,6 +315,9 @@ class FileSystemRequestQueueClient(RequestQueueClient):
                 new_total_request_count=0,
             )
 
+            # Invalidate is_empty cache.
+            self._is_empty_cache = None
+
     @override
     async def add_batch_of_requests(
         self,
@@ -317,6 +326,7 @@ class FileSystemRequestQueueClient(RequestQueueClient):
         forefront: bool = False,
     ) -> AddRequestsResponse:
         async with self._lock:
+            self._is_empty_cache = None
             new_total_request_count = self._metadata.total_request_count
             new_pending_request_count = self._metadata.pending_request_count
             processed_requests = list[ProcessedRequest]()
@@ -425,6 +435,9 @@ class FileSystemRequestQueueClient(RequestQueueClient):
             if forefront:
                 self._request_cache_needs_refresh = True
 
+            # Invalidate is_empty cache.
+            self._is_empty_cache = None
+
             return AddRequestsResponse(
                 processed_requests=processed_requests,
                 unprocessed_requests=unprocessed_requests,
@@ -463,12 +476,15 @@ class FileSystemRequestQueueClient(RequestQueueClient):
 
             if next_request is not None:
                 state.in_progress_requests.add(next_request.unique_key)
+                # `in_progress_requests` is updated, so we need to invalidate the `is_empty` cache.
+                self._is_empty_cache = None
 
             return next_request
 
     @override
     async def mark_request_as_handled(self, request: Request) -> ProcessedRequest | None:
         async with self._lock:
+            self._is_empty_cache = None
             state = self._state.current_value
 
             # Check if the request is in progress.
@@ -516,6 +532,7 @@ class FileSystemRequestQueueClient(RequestQueueClient):
         forefront: bool = False,
     ) -> ProcessedRequest | None:
         async with self._lock:
+            self._is_empty_cache = None
             state = self._state.current_value
 
             # Check if the request is in progress.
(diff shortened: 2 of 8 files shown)


src/crawlee/storage_clients/_file_system/_request_queue_client.py
```diff
@@ -586,39 +571,23 @@ async def reclaim_request(
...
-
-            # If we have a cached requests, check them first (fast path).
-            if self._request_cache:
-                for req in self._request_cache:
-                    if req.unique_key not in state.handled_requests:
-                        self._is_empty_cache = False
-                        return False
-                self._is_empty_cache = True
-                return len(state.in_progress_requests) == 0
-
```
janbuchar, line 608: Are you sure it's safe to remove all this?

janbuchar (commented):
Looks sound, the only thing I'm mildly concerned about is the removal of the is_empty return value caching — I'd feel better if I knew some historical context and why it's safe to do this.

vdusek (commented):
Thanks! Could you please implement https://github.com/apify/apify-sdk-python/issues/987 as well?

That way, we can verify that it works on the platform too.

I'd like this to be part of SDK v4, built on top of Crawlee v1.8 with this included.

Mantisus:
> Thanks! Could you please implement https://github.com/apify/apify-sdk-python/issues/987 as well?

Sure.

src/crawlee/storage_clients/_file_system/_request_queue_client.py
Mantisus: My bad. I was sure that, in the current implementation of `atomic_write`, it guarantees that the metadata file is always valid. However, I just double-checked this, and on Windows, there is still a risk of creating an invalid metadata file that would cause the queue to fail.

Mantisus: Updated with cache logic

src/crawlee/storage_clients/_base/_request_queue_client.py
```diff
@@ -124,8 +124,16 @@ async def reclaim_request(
...
             True if the request queue is empty, False otherwise.
         """
+
+    async def is_finished(self) -> bool:
+        """Check if the request queue is finished. That means the queue is empty and no requests are being processed.
+
+        Returns:
+            True if the request queue is finished, False otherwise.
+        """
+        return await self.is_empty()
```
vdusek, line 149: Hmm, this default implementation equates "finished" with "empty", but the new `is_empty` can return `True` while requests are still (or could still be) in progress.

This means that a potential `RequestQueueClient` implementation that overrides only the abstract `is_empty` would inherit an `is_finished` that can report finished mid-processing, which contradicts this method's docstring and could allow a crawl to stop with work still while running.

I would rather see this as an abstract method as well. But that would be breaking for the users who wrote their own RQ client already. Or maybe we could come up with a better (safer) default? Or at least document it, and open an issue for v2 to make this abstract as well.

src/crawlee/storage_clients/_redis/_request_queue_client.py
```diff
@@ -509,9 +505,18 @@ async def is_empty(self) -> bool:
...
+
+    @retry_on_error(RedisError)
+    @override
+    async def is_finished(self) -> bool:
+        is_empty = await self.is_empty()
+
         metadata = await self.get_metadata()
 
-        return metadata.pending_request_count == 0
+        return is_empty and metadata.pending_request_count == 0
```
vdusek: IMO, a worthwhile optimization suggestion from Claude:

> Suggestion: short-circuit when the queue isn't empty so a non-empty queue skips the extra `get_metadata()` round-trip (the SQL client already does this).

```suggestion
        if not await self.is_empty():
            return False

        metadata = await self.get_metadata()
        return metadata.pending_request_count == 0
```

src/crawlee/storage_clients/_file_system/_request_queue_client.py
```diff
@@ -592,34 +594,46 @@ async def is_empty(self) -> bool:
...
-                return len(state.in_progress_requests) == 0
+                return True
 
             # Fallback: check state for unhandled requests.
             await self._update_metadata(update_accessed_at=True)
 
-            # Check if there are any requests that are not handled
-            all_requests = set(state.forefront_requests.keys()) | set(state.regular_requests.keys())
-            unhandled_requests = all_requests - state.handled_requests
+            # Check pending requests is state.
```
vdusek: typo?

```suggestion
            # Check pending requests in state.
```

src/crawlee/storage_clients/_memory/_request_queue_client.py
```diff
@@ -314,8 +314,18 @@ async def is_empty(self) -> bool:
...
         await self._update_metadata(update_accessed_at=True)
 
-        # Queue is empty if there are no pending requests and no requests in progress.
-        return len(self._pending_requests) == 0 and len(self._in_progress_requests) == 0
+        # Queue is empty if there are no pending requests.
+        return len(self._pending_requests) == 0
+
+    @override
+    async def is_finished(self) -> bool:
+        """Check if the queue is finished.
```
vdusek: This PR removes the docstrings from the `is_empty` overrides in the Redis/SQL clients and from the file-system `is_finished` (replacing them with inline comments), while the memory client keeps the full "returns" docstrings on both `is_empty` and `is_finished`.

Could we align all four implementations? I would prefer keeping the full docstrings.

tests/unit/crawlers/_basic/test_basic_crawler.py
```diff
@@ -142,6 +142,29 @@ async def handler(context: BasicCrawlingContext) -> None:
...
+
+    with patch.object(
+        request_manager,
+        'fetch_next_request',
+        wraps=request_manager.fetch_next_request,
+    ) as fetch_counter:
+        await crawler.run(['https://a.placeholder.com'])
+
+        # `concurrency` tasks can be scheduled if control was never yielded with `await` during task scheduling
+        assert fetch_counter.call_count <= 4
```
vdusek: One comment from Claude, please consider it:

> Suggestion: this `call_count` is deterministically `== concurrency` (the bug pushed it to ~660), so `<= 4` works but under-specifies. Consider asserting `== concurrency` plus a positive lower bound, since 0 fetches (e.g. if storage isolation ever regressed and the request were dedup'd away) would also pass `<= 4` as a silent no-op.

janbuchar (approved):
Looks about right to me. @vdusek please review it too 🙂

vdusek (changes requested):
A few comments

tests/unit/crawlers/_basic/test_basic_crawler.py
Mantisus: `fetch_counter.call_count == concurrency` is a rather flaky condition, since this behavior can vary depending on the storage client being used.

But I made the condition stricter.

src/crawlee/storage_clients/_memory/_request_queue_client.py
Mantisus: I removed the docstrings for all implementations. This is consistent with other methods that use `override`, and the docstring is inherited from the base class.

src/crawlee/storage_clients/_sql/_request_queue_client.py
```diff
@@ -595,10 +595,34 @@ async def reclaim_request(
...
+
+            await self._add_buffer_record(session)
+
+            return not result.scalar()
+
+    @retry_on_error(SQLAlchemyError)
+    @override
+    async def is_finished(self) -> bool:
+        # If the queue is not empty, it is not finished
+        if not await self.is_empty():
```
vdusek, line 623: `is_finished` calls `await self.is_empty()` here, which opens a DB session, and then opens a second session below for `get_metadata()` and the remaining queries. Since the autoscaled pool polls `is_finished`/`is_empty` while scheduling requests, this could have a performance impact.

Maybe we could handle the empty check directly inside `is_finished`, using only a single session?

src/crawlee/storage_clients/_redis/_request_queue_client.py
```diff
@@ -509,8 +505,18 @@ async def is_empty(self) -> bool:
...
             self._next_reclaim_stale = datetime.now(tz=timezone.utc) + self._RECLAIM_INTERVAL
 
-        metadata = await self.get_metadata()
+        # Check if there are any requests in the queue.
+        requests_in_queue = await await_redis_response(self._redis.llen(self._queue_key))
+        return requests_in_queue == 0
+
+    @retry_on_error(RedisError)
+    @override
+    async def is_finished(self) -> bool:
```
vdusek: Nit: unnecessary empty blank line at the top of the method body

tests/unit/crawlers/_basic/test_basic_crawler.py
```diff
@@ -142,6 +142,29 @@ async def handler(context: BasicCrawlingContext) -> None:
     ]
 
 
+async def test_no_new_tasks_while_only_request_in_progress() -> None:
```
vdusek, line 145: Could we add a one-line docstring? (like other tests have it)

```suggestion
async def test_no_new_tasks_while_only_request_in_progress() -> None:
    """No new tasks are scheduled while the only request is still being processed."""
```

tests/unit/crawlers/_basic/test_basic_crawler.py
```diff
@@ -142,6 +142,29 @@ async def handler(context: BasicCrawlingContext) -> None:
...
+
+    with patch.object(
+        request_manager,
+        'fetch_next_request',
+        wraps=request_manager.fetch_next_request,
+    ) as fetch_counter:
+        await crawler.run(['https://a.placeholder.com'])
+
+        # `concurrency` tasks can be scheduled if control was never yielded with `await` during task scheduling
+        assert 1 <= fetch_counter.call_count <= 4
```
vdusek, line 166: If I am not mistaken, we can do it like this

```suggestion
        assert fetch_counter.call_count <= concurrency
```

vdusek (commented):
A few more comments (three nits and one thing that might have a performance impact)

tests/unit/crawlers/_basic/test_basic_crawler.py
Mantisus: That was the initial implementation, but I agree with Claude that we should explicitly exclude 0 from the condition.

https://github.com/apify/crawlee-python/pull/1982#discussion_r3457582543

vdusek: oh sorry, up to you then

vdusek (commented):
I'd probably use a `fix` prefix for the PR title. It does add a new method, true, but only because its absence is the root cause of the bug. Something like this:

> fix: add is_finished and correct is_empty semantics in RequestQueueClient

would be better in my opinion.

src/crawlee/storage_clients/_sql/_request_queue_client.py
Mantisus: Getting a session from the pool is quite cheap, but it can actually impact performance if the pool is small. However, this will only become apparent when the queue is empty but not yet finished.

But merging `is_empty` and `get_metadata` into a single session in `is_finished` will make the method harder to read.

If we need to optimize these methods, I would consider removing the `_add_buffer_record` calls in `is_empty` and `is_finished`, just to update `accessed_at` in the metadata. This will have a greater impact on performance.

tests/unit/crawlers/_basic/test_basic_crawler.py
Mantisus: Sure. added

src/crawlee/storage_clients/_sql/_request_queue_client.py
vdusek: Okay, could you please write these optimization ideas down in an issue?

vdusek (approved):
LGTM

PR #1993, opened 2026-06-28
build: Bump ty to v0.0.55 and fix type issues

Mantisus:


uv.lock
```diff
@@ -1237,7 +1237,7 @@ name = "exceptiongroup"
 version = "1.3.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "typing-extensions", marker = "python_full_version < '3.11'" },
```
vdusek, line 1240: Interesting, you must have something different in your `uv` setup.

vdusek (approved):
LGTM. I just limited the `uv.lock` diff to `ty` only, and avoided one type checker suppression.

PR #1996, opened 2026-06-30
fix(templates): Pin Playwright to base image version for templates

Pijukatel:
### Description

- Use fixed `playwright_version` to determine the version of the base image and also the version of installed Playwright

### DIFF
--- a/src/crawlee/project_template/{{cookiecutter.project_name}}/Dockerfile
+++ b/src/crawlee/project_template/{{cookiecutter.project_name}}/Dockerfile
@@ -1,16 +1,22 @@
 # First, specify the base Docker image.
 # You can see the Docker images from Apify at https://hub.docker.com/r/apify/.
 # You can also use any other image from Docker Hub.
+
+# The Playwright version baked into the Apify Playwright base images. Keep this in sync with the
+# `playwright` version resolved by the lockfile so the installed package matches the browser
+# binaries shipped in the base image.
+# % set playwright_version = '1.60.0'
+
 # % if cookiecutter.crawler_type == 'playwright' or cookiecutter.crawler_type.startswith('adaptive-') or cookiecutter.crawler_type == 'stagehand'
-# % set base_image = 'apify/actor-python-playwright:3.13-1.60.0'
+# % set base_image = 'apify/actor-python-playwright:3.13-' ~ playwright_version
 # % elif cookiecutter.crawler_type == 'playwright-camoufox'
-# % set base_image = 'apify/actor-python-playwright-camoufox:3.13-1.60.0'
+# % set base_image = 'apify/actor-python-playwright-camoufox:3.13-' ~ playwright_version
 # % elif cookiecutter.crawler_type == 'playwright-chrome'
-# % set base_image = 'apify/actor-python-playwright-chrome:3.13-1.60.0'
+# % set base_image = 'apify/actor-python-playwright-chrome:3.13-' ~ playwright_version
 # % elif cookiecutter.crawler_type == 'playwright-firefox'
-# % set base_image = 'apify/actor-python-playwright-firefox:3.13-1.60.0'
+# % set base_image = 'apify/actor-python-playwright-firefox:3.13-' ~ playwright_version
 # % elif cookiecutter.crawler_type == 'playwright-webkit'
-# % set base_image = 'apify/actor-python-playwright-webkit:3.13-1.60.0'
+# % set base_image = 'apify/actor-python-playwright-webkit:3.13-' ~ playwright_version
 # % else
 # % set base_image = 'apify/actor-python:3.13'
 # % endif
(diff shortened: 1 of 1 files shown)


Pijukatel:
E2E tests: https://github.com/apify/crawlee-python/actions/runs/28428916800

B4nan:
I see we still have this python specific setup in the docker images. We should align this, in JS we have docker image tags for playwright versions, and we pin those in the templates (together with the dependency version, to ensure they are aligned, plus we have a warning in the docker build if those versions are not the same).

cc @vladfrangu 

> it is pulled in transitively via crawlee[playwright]

This is also weird, in JS, we don't dictate the playwright version, its a peer dependency, users need to install it explicitly. Can we do the same in python?

vdusek (commented):
Maybe we should merge https://github.com/apify/crawlee-python/pull/1997 first. After that, couldn't we use a build ARG or ENV  instead? I don't think we need to derive the Playwright version this way - we should already know it.

Pijukatel:
@vdusek We can fix the version in the template, for example, like this https://github.com/apify/crawlee-python/pull/1996/changes/c25bd1437aa8acb7baca9d7f869b5aae45abbd52

There are some edge cases about users who want a specific Playwright version or specific browser versions, but for that they have to do some custom Dockerfile changes.

E2E tests green: https://github.com/apify/crawlee-python/actions/runs/28436119038

src/crawlee/project_template/{{cookiecutter.project_name}}/Dockerfile
```diff
@@ -1,16 +1,22 @@
 # First, specify the base Docker image.
 # You can see the Docker images from Apify at https://hub.docker.com/r/apify/.
 # You can also use any other image from Docker Hub.
+
+# The Playwright version baked into the Apify Playwright base images. Keep this in sync with the
```
vdusek: Dockerfile comment vs Jinja comment

> This is a plain `#` comment, so it renders into the generated project's Dockerfile, where the template-only `playwright_version` variable no longer exists, leaving guidance the end user can't act on. It's also aimed at the wrong invariant: the `--force-reinstall` below overrides the lockfile, so the value that must stay in sync is the base-image tag's Playwright version (`3.13-X`), not the lockfile. Consider a Jinja comment (`{# ... #}`) so it stays template-only, reworded around matching the base image. (Both reviewers.)

src/crawlee/project_template/{{cookiecutter.project_name}}/Dockerfile
```diff
@@ -32,21 +38,18 @@ COPY pyproject.toml poetry.lock ./
...
- # Replace playwright version so that it matches whatever is pre-installed in the image (the `hash` checks if playwright is installed)
-    sed "s/^playwright==\(.*\)/playwright==$(hash playwright 2>/dev/null && (playwright --version | cut -d ' ' -f 2) || echo '\1')/" | \
- # Install everything using pip (ignore dependency checks - the lockfile is correct, period)
-    pip install -r /dev/stdin --no-dependencies \
+ # Export packages from poetry.lock and install everything using pip
+ # (ignore dependency checks - the lockfile is correct, period)
+ && poetry export -f requirements.txt --without-hashes | pip install -r /dev/stdin --no-dependencies \
  && echo "All installed Python packages:" \
  && pip freeze
+# % if cookiecutter.crawler_type.startswith('playwright') or cookiecutter.crawler_type.startswith('adaptive-') or cookiecutter.crawler_type == 'stagehand'
```
vdusek: We can do the base image selection only once.

> The 'uses a Playwright base image' predicate is now spelled two different ways: the base-image selector above (`crawler_type == 'playwright' or startswith('adaptive-') or == 'stagehand'` plus the elif chain) vs. this `startswith('playwright') or ...`, which is also repeated verbatim in the uv and pip blocks. Two spellings can drift (adding a new `playwright-*` type, or a non-playwright type starting with `playwright`) and yield a base image with no matching pin, or a pin with no matching image. Consider computing it once near the top (`# % set uses_playwright_base = ...`) and reusing it for both the `FROM` selection and all three pin blocks.

src/crawlee/project_template/{{cookiecutter.project_name}}/Dockerfile
```diff
@@ -32,21 +38,18 @@ COPY pyproject.toml poetry.lock ./
...
+# Pin playwright to the version baked into the base image so it matches the browser binaries.
+RUN pip install --no-deps --force-reinstall "playwright=={{ playwright_version }}"
```
vdusek, line 49: The playwright versions should match now, right? So why do we need to force-reinstall it?

Or do we reinstall the version the user provided (python package)?

vdusek (commented):
A few questions / suggestions

src/crawlee/project_template/{{cookiecutter.project_name}}/Dockerfile
Pijukatel: Ok

src/crawlee/project_template/{{cookiecutter.project_name}}/Dockerfile
Pijukatel: That depends on when it is being installed.

It can happen that our base image is on `1.60.0`, while the newest version is already `1.61.0`. So the Playwright might be updated during the project installation and this is the situation where we want to downgrade back to the base image version.

src/crawlee/project_template/{{cookiecutter.project_name}}/Dockerfile
Pijukatel: Ok

Pijukatel:
New run of E2E tests with the latest commit:
https://github.com/apify/crawlee-python/actions/runs/28504970808

src/crawlee/project_template/{{cookiecutter.project_name}}/Dockerfile
vdusek: thanks

vdusek (approved):

PR #2003, opened 2026-07-01
ci: Add workflow to bump the project template's Playwright version

Pijukatel:
### Description

The cookiecutter project template's Dockerfile
(`src/crawlee/project_template/{{cookiecutter.project_name}}/Dockerfile`) pins a single
Playwright version. That value selects the Apify base image tag
(`apify/actor-python-playwright*:3.13-<version>`) and the in-image `playwright==<version>`
pin, and it was bumped by hand. This adds automation that keeps it current, in the same
spirit as the template docker-image bumping in `apify/actor-templates`.


.github/workflows/update_playwright_version.yaml
```diff
@@ -0,0 +1,106 @@
...
+          PR_NUMBER=$(gh pr list --head "$BRANCH_NAME" --base master --json number --jq '.[0].number // empty')
+          if [ -n "$PR_NUMBER" ]; then
+            gh pr merge "$PR_NUMBER" --disable-auto || true
+          fi
+
+          # Delete the remote branch if it exists. The signed-commit step below recreates it from
+          # the current HEAD (origin/master).
+          if git ls-remote --heads origin "$BRANCH_NAME" | grep -q "$BRANCH_NAME"; then
+            echo "Deleting existing remote branch $BRANCH_NAME"
+            git push origin --delete "$BRANCH_NAME"
```
vdusek: Maybe we should force-update the branch instead of deleting it, since deleting the branch closes the existing PR and causes each daily run to open a new PR. Which means losing prior comments and approvals. WDYT?

> Deleting the branch here closes the existing open PR — GitHub auto-closes a PR when its head branch is deleted — so the "PR #N already exists … updating it" path in the create step below is unreachable. While a bump sits unmerged, every daily run rewrites master's still-old pin, sees `has-changes=true`, deletes this branch (closing the current PR), and opens a brand-new one, discarding the previous PR's comments/approvals. Consider force-updating the branch instead of deleting it so the open PR is preserved. (Both reviewers flagged the reset logic.)

scripts/update_playwright_version.py
```diff
@@ -0,0 +1,71 @@
...
+
+from __future__ import annotations
+
+import json
+import re
+import urllib.request
+from pathlib import Path
+
+DOCKERFILE = Path(__file__).resolve().parent / '{{cookiecutter.project_name}}/Dockerfile'
+TAGS_URL = 'https://hub.docker.com/v2/repositories/apify/actor-python-playwright/tags?page_size=100'
```
vdusek, line 24: A note about the Playwright image variants with potential missing tags and broken builds.

> This queries only the `apify/actor-python-playwright` repo, but the bumped `playwright_version` also selects the `-camoufox`/`-chrome`/`-firefox`/`-webkit` base images. If the plain image publishes a new `3.13-<v>` tag before a variant does, the template will pin e.g. `apify/actor-python-playwright-firefox:3.13-<v>` to a tag that doesn't exist yet, breaking `docker build` for those crawler types. Worth confirming the variants publish in lockstep, or intersecting the available tags across all five repos.

scripts/update_playwright_version.py
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env python3
```
vdusek, line 1: We should move this script to the `scripts/`.

> `src/crawlee/project_template/` is excluded from both ruff (`extend-exclude`) and ty, and is shipped in the published wheel (`packages = ["src/crawlee"]`). So this maintenance script's style/types are never enforced and it's distributed to end users. A `scripts/`-dir or repo-root location would be linted, type-checked, and kept out of the wheel — at the cost of the co-location with the Dockerfile the docstring calls out.

src/crawlee/project_template/update_playwright_version.py
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env python3
+"""Bump the Playwright version pinned by the project template's Dockerfile.
+
+The template Dockerfile pins a single Playwright version (a Jinja ``# % set playwright_version
```
vdusek: Docstrings convention

> RST-style double backticks in the docstring (here on lines 4–5, and again on lines 5 and 8). Repo/user convention is single backticks for code references in docstrings and comments.

vdusek (commented):
A few comments

scripts/update_playwright_version.py
Pijukatel: This change is gated by E2E tests on those variant Docker images. In the scenario where those variants are missing, the E2E tests will fail, so it is safe in this context.

.github/workflows/update_playwright_version.yaml
Pijukatel: Good catch — fixed. Rather than delete/recreate the branch, the workflow now checks out the existing `ci/update-playwright-version` branch when one is open (`git ls-remote` + `git fetch` + `git checkout -B … FETCH_HEAD`) and runs the bump there, so the update lands on the same PR and its history/approvals are preserved. If the branch already holds the latest version, the run is a no-op (`git diff --quiet` → no commit); a newer release just appends a commit to the same PR. A new branch/PR is only created when none exists. Since `apify/actions/signed-commit` pushes onto the existing branch without force, this avoids the delete entirely.

---
_Generated by [Claude Code](https://claude.ai/code)_

scripts/update_playwright_version.py
vdusek: But this also means updating the targets of static analysis tools.

vdusek (changes requested):

vdusek (approved):
LGTM

PR #2010, opened 2026-07-02
fix(logging): Preserve host application logging configuration

vdusek:
## Problem

`BasicCrawler.__init__` (with the default `configure_logging=True`) called `configure_logger(root_logger, remove_old_handlers=True)`, wiping the host application's root logging handlers. Embedding Crawlee in a larger app silently removed the host's own logging configuration.

## Changes

- The root logger is now configured only when it has no handlers yet, mirroring `logging.basicConfig` semantics. A host that has already configured logging keeps its configuration.
- The `httpx` logger silencing was moved out of `BasicCrawler` into `HttpxHttpClient`. It used to live unconditionally in the base crawler, even though the default HTTP client (`ImpitHttpClient`) doesn't use the `httpx` logger at all, and it reached into a third-party logger straight from the base class. The silencing now applies only when `HttpxHttpClient` is actually instantiated, which is where the noisy per-request `httpx` INFO logs originate.

## Verification

Regression tests that fail without the fix:
- host-installed root handlers survive crawler construction,
- Crawlee still configures an unconfigured root logger,
- `HttpxHttpClient` silences the `httpx` logger to `WARNING` by default.

## Behavior change

This intentionally changes observable behavior, since the old behavior was the bug: host apps keep their logging config.

For `HttpxHttpClient` users, the `httpx` logger is now silenced whenever the client is instantiated. Previously the silencing happened only during `BasicCrawler` construction, with `configure_logging=True` and an unconfigured root logger. As a result it now also applies with `configure_logging=False` and to standalone `send_request` usage.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,7 +52,6 @@ dependencies = [
 [project.optional-dependencies]
 all = ["crawlee[adaptive-crawler,pydantic-ai,beautifulsoup,cli,curl-impersonate,httpx,parsel,playwright,otel,sql_sqlite,sql_postgres,sql_mysql,stagehand,redis]"]
 adaptive-crawler = [
-    "crawlee[beautifulsoup,parsel]",
     "jaro-winkler>=2.0.3",
     "playwright>=1.27.0",
     "scikit-learn>=1.6.0",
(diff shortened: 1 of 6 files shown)


Mantisus (approved):
LGTM

Pijukatel (approved):

PR #2011, opened 2026-07-02
perf(file-system): Fix quadratic file rescans in the FS request queue

vdusek:
## Problem

The file system request queue client was O(n²):

- Every cache refresh globbed and parsed every request file ever written (handled files are never deleted from disk).
- `forefront=True` invalidated the whole cache, and `RequestManagerTandem` adds every loader request with `forefront=True`, which triggered a full rescan per processed request.
- The blocking `json.load` ran on the event loop.

Processing a 10k `RequestList` through the tandem meant roughly 50M file parses.

## Changes

- `_refresh_cache` now derives pending requests from the recoverable state's key→sequence maps and reads only those files, instead of globbing and parsing the whole directory.
- New forefront requests go straight to the front of the in-memory cache. The cache is invalidated only when an existing request is repositioned, or when it's at capacity.
- Handled requests are pruned from the pending state maps (they stay in `handled_requests` for deduplication), so cache refreshes, `is_empty`, and the persisted state blob scale with the backlog rather than with every request ever processed.
- `_parse_request_file` reads the file in a single `asyncio.to_thread` call instead of running `json.load` on the event loop.

## Results

Processing 200 requests through the tandem pattern went from 20,100 file parses (2.8 s) to 1 (0.3 s), and scaling is now linear (2,000 requests in about 2.6 s). Ordering semantics (forefront LIFO, regular FIFO), deduplication, and crash-recovery behavior are preserved and covered by new regression tests.

## Logging changes

The file system request queue no longer emits the `Request <unique_key> not found in state, skipping.` warning. The cache refresh no longer scans the queue directory, so orphaned request files (e.g. written by another process) are silently ignored.

### DIFF
--- a/src/crawlee/storage_clients/_file_system/_request_queue_client.py
+++ b/src/crawlee/storage_clients/_file_system/_request_queue_client.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import asyncio
-import functools
 import json
 import shutil
 from collections import deque
@@ -333,7 +332,8 @@ class FileSystemRequestQueueClient(RequestQueueClient):
             unprocessed_requests = list[UnprocessedRequest]()
             state = self._state.current_value
 
-            all_requests = state.forefront_requests | state.regular_requests
+            def is_in_queue(unique_key: str) -> bool:
+                return unique_key in state.forefront_requests or unique_key in state.regular_requests
 
             requests_to_enqueue = {}
 
@@ -352,7 +352,7 @@ class FileSystemRequestQueueClient(RequestQueueClient):
                 # Or if the request is already in the queue and the `forefront` flag is not used, we do not change the
                 # position of the request.
                 elif (request.unique_key in state.in_progress_requests) or (
-                    request.unique_key in all_requests and not forefront
+                    is_in_queue(request.unique_key) and not forefront
                 ):
                     processed_requests.append(
                         ProcessedRequest(
@@ -368,7 +368,7 @@ class FileSystemRequestQueueClient(RequestQueueClient):
             # Process each request in the batch.
             for request in requests_to_enqueue.values():
                 # If the request is not already in the RQ, this is a new request.
-                if request.unique_key not in all_requests:
+                if not is_in_queue(request.unique_key):
                     request_path = self._get_request_path(request.unique_key)
                     # Add sequence number to ensure FIFO ordering using state.
                     if forefront:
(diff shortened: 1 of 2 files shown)


src/crawlee/storage_clients/_file_system/_request_queue_client.py
```diff
@@ -696,63 +709,36 @@ async def _update_metadata(
...
-            self._request_cache.appendleft(request)
 
-        # Add regular requests to the end of the cache (right side).
-        for request, _ in regular_requests:
-            if len(self._request_cache) >= self._MAX_REQUESTS_IN_CACHE:
-                break
-            self._request_cache.append(request)
+            request = await self._parse_request_file(self._get_request_path(unique_key))
+            if request is not None:
+                self._request_cache.append(request)
```
Mantisus: Let's add a log warning when there is no file for the `unique_key`.

src/crawlee/storage_clients/_file_system/_request_queue_client.py
```diff
@@ -384,6 +384,16 @@ async def add_batch_of_requests(
                     request_data = await json_dumps(request.model_dump())
                     await atomic_write(request_path, request_data)
 
+                    # A new forefront request belongs to the very front of the queue, so it can go straight
+                    # to the cache without a full refresh, unless the cache is at capacity. New regular
+                    # requests must not be appended to the cache: it may be truncated, so its tail is not
+                    # necessarily the end of the queue; they are picked up by a refresh once the cache drains.
+                    if forefront:
+                        if len(self._request_cache) < self._MAX_REQUESTS_IN_CACHE:
+                            self._request_cache.appendleft(request)
```
Mantisus, line 394: We can probably skip this if `self._request_cache_needs_refresh` is already `True`.

tests/unit/storage_clients/_file_system/test_fs_rq_client.py
```diff
@@ -259,3 +259,99 @@ async def test_get_request_does_not_mark_in_progress(rq_client: FileSystemReques
...
+    for i in range(n):
+        await rq_client.add_batch_of_requests([Request.from_url(f'https://example.com/{i}')], forefront=True)
+        request = await rq_client.fetch_next_request()
+        assert request is not None
+        await rq_client.mark_request_as_handled(request)
+
+    # Before the fix, each forefront add invalidated the cache and each fetch re-parsed every request file
+    # ever written, giving n * (n + 1) / 2 parses in total. With the fix, new forefront requests go straight
+    # to the cache, so only the initial refresh parses anything.
+    assert parse_count <= 5
```
Mantisus: Maybe we could use `unittest.mock.patch` for this?

For example

```python
from unittest.mock import patch

async def test_forefront_add_fetch_handle_parses_linear_number_of_files(
    rq_client: FileSystemRequestQueueClient,
) -> None:
    with patch.object(
        rq_client,
        '_parse_request_file',
        wraps=rq_client._parse_request_file,
    ) as call_counter:

        n = 25
        for i in range(n):
            await rq_client.add_batch_of_requests([Request.from_url(f'https://example.com/{i}')], forefront=True)
            request = await rq_client.fetch_next_request()
            assert request is not None
            await rq_client.mark_request_as_handled(request)

    assert call_counter.call_count <= 5
```

Mantisus (approved):
LGTM. Only few suggestions.

src/crawlee/storage_clients/_file_system/_request_queue_client.py
```diff
@@ -352,7 +352,7 @@ async def add_batch_of_requests(
                 # Or if the request is already in the queue and the `forefront` flag is not used, we do not change the
                 # position of the request.
                 elif (request.unique_key in state.in_progress_requests) or (
-                    request.unique_key in all_requests and not forefront
+                    is_in_queue(request.unique_key) and not forefront
```
Pijukatel: Just a nit, maybe put `not forefront` first so that we get a fast fail through without any other longer membership tests in some cases

Pijukatel (approved):

PR #2012, opened 2026-07-02
fix(sitemap): bound sitemap decompression size and nesting depth to prevent DoS

vdusek:
## Problem

The sitemap parsing pipeline processes content from arbitrary crawled hosts, which are untrusted. Two vectors let a malicious host exhaust crawler memory or trap it in unbounded work:

- **Gzip decompression bomb.** `parse_sitemap` decompressed gzipped sitemaps with no output cap, so a ~50 KB payload could inflate to gigabytes.
- **Unbounded nested-sitemap recursion.** `parse_sitemap` defaulted `max_depth` to `float('inf')` with only exact-URL dedup, so a chain of unique nested-sitemap URLs recursed forever.

## Changes

- **Gzip decompression bomb.** Decompression and parsing now run in bounded pieces capped at `MAX_SITEMAP_SIZE` (50 MB, the sitemap-protocol limit). The read loop also stops once the gzip stream ends (otherwise trailing bytes accumulate unbounded in the decompressor's `unused_data`) and once too many raw bytes are read (guarding a stream that produces little or no output). On a repro, peak memory dropped from ~680 MB to ~15 MB.
- **Unbounded nested-sitemap recursion.** The default is now `DEFAULT_MAX_DEPTH = 10`. `SitemapRequestLoader` — the main consumer — drives nesting itself (it passes `max_depth=0`), so it now tracks per-sitemap depth in its persisted state and stops following chains past the same limit.

## Out of scope

A third listed vector, unbounded `robots.txt` reads, is intentionally not addressed here. Capping it requires switching `RobotsTxtFile.load` from `send_request` to streaming, which regresses `CurlImpersonateHttpClient` (a multi-second stall in the crawl pipeline that fails `test_respect_robots_txt_with_problematic_links[curl]`). Since `send_request` reads the whole body eagerly and there is no size-limited read primitive in the HTTP client layer, there is no safe non-streaming cap. It is the lowest-severity vector (no compression amplification — the attacker must transmit the full bytes), so it is left as a follow-up: add a size-limited read to the `HttpClient` streaming layer, then cap `robots.txt` there.

## Tests

Adds regression tests for gzip-bomb truncation, stop-after-gzip-member, `parse_sitemap` depth, and loader nested-chain depth.

### DIFF
--- a/src/crawlee/_utils/sitemap.py
+++ b/src/crawlee/_utils/sitemap.py
@@ -219,7 +219,10 @@ class _XmlSitemapParser:
 
 def _get_parser(content_type: str = '', url: str | None = None) -> _XmlSitemapParser | _TxtSitemapParser:
     """Create appropriate parser based on content type and URL."""
-    if 'text/plain' in content_type.lower() or (url and URL(url).path.endswith('.txt')):
+    # Strip a trailing `.gz` so gzipped sitemaps are classified by their real extension (e.g. `sitemap.txt.gz`
+    # is a text sitemap, not XML) even when the compressed response is served as `application/gzip`.
+    path = URL(url).path.removesuffix('.gz') if url else ''
+    if 'text/plain' in content_type.lower() or path.endswith('.txt'):
         return _TxtSitemapParser()
     # Default to XML parser for most cases
     return _XmlSitemapParser()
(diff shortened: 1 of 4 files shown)


Pijukatel (approved):

Mantisus (approved):
LGTM

PR #2017, opened 2026-07-03
fix: preserve case-sensitive URL parts

### LINKED ISSUE #2008: `normalize_url` lowercases the entire URL, silently deduplicating case-distinct pages
### Description

### Impact

### Proposed fix

roian6:
### Description

- Preserve path, query-name, query-value, and fragment casing in `normalize_url` while still relying on `yarl` to canonicalize scheme/host casing.
- Update the normalization docstring so it no longer implies all URL casing is trivial.
- Add regressions showing default unique keys no longer collide for case-distinct paths or query components.

### Issues

- Closes: #2008

### Testing

- `uv run pytest tests/unit/_utils/test_requests.py -q`
- `uv run poe lint`
- `uv run poe type-check`
- `git diff --check`

I also attempted `uv run poe unit-tests`. After installing Playwright browsers, the suite reached `2152 passed, 6 skipped` but still failed in `tests/unit/browsers/test_stagehand_browser_plugin.py` because the local environment lacks the Stagehand SEA binary (`stagehand-linux-arm64`). That failure appears unrelated to this URL normalization change.

### Checklist

- [x] CI passed

### DIFF
--- a/tests/unit/_utils/test_requests.py
+++ b/tests/unit/_utils/test_requests.py
@@ -15,7 +15,9 @@ from crawlee._utils.requests import compute_unique_key, normalize_url
             'http://example.com/?another_key=another_value&key=value',
             False,
         ),
-        ('HTTPS://EXAMPLE.COM/?KEY=VALUE', 'https://example.com/?key=value', False),
+        ('HTTPS://EXAMPLE.COM/?KEY=VALUE', 'https://example.com/?KEY=VALUE', False),
+        ('HTTPS://EXAMPLE.COM/Product/ABC?token=SeCrEt', 'https://example.com/Product/ABC?token=SeCrEt', False),
+        ('HTTP://EXAMPLE.COM/Path#Frag', 'http://example.com/Path#Frag', True),
         ('', '', False),
         ('http://example.com/#fragment', 'http://example.com/#fragment', True),
         ('http://example.com/#fragment', 'http://example.com', False),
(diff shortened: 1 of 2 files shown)


tests/unit/_utils/test_requests.py
```diff
@@ -38,6 +42,19 @@ def test_normalize_url(url: str, expected_output: str, *, keep_url_fragment: boo
     assert output == expected_output
 
 
+def test_compute_unique_key_preserves_case_sensitive_path_and_query() -> None:
```
Pijukatel: Could you please write it as a parametrized test with 1 assert and 3 sets of parameters and ids explaining a specific test case (path_key, query_value_key, query_name_key)

vdusek:
Hi @roian6, thanks for the PR. However, this issue is currently milestoned for v2 because the fix would introduce breaking changes to how requests are represented in storages.

Because of that, this needs to be handled carefully as part of a major version update, along with an appropriate upgrade guide. I'm going to close this for now.

tests/unit/_utils/test_requests.py
roian6: Thanks! Updated the regression to a parametrized test with one assert and the requested ids: `path_key`, `query_value_key`, and `query_name_key`.

I also rebased onto the current `origin/master` and re-ran:

- `uv run pytest tests/unit/_utils/test_requests.py -q` (`20 passed`)
- `uv run poe lint`
- `uv run poe type-check`
- `git diff origin/master --check`

The Semgrep PR check is passing on the updated head `7108d4ba45093e60a1166676ec84e65a1cc654cb` as well.

PR #2018, opened 2026-07-03
test: Deflake test_error_handler_can_access_page on Windows CI

vdusek:
`test_error_handler_can_access_page` fails intermittently on the Windows CI runner ([example run](https://github.com/apify/crawlee-python/actions/runs/28652457948/job/84973524931)). Under `xdist` load, Playwright's `Page.goto` occasionally fails with `net::ERR_NO_BUFFER_SPACE` before the page is navigated. That attempt's error handler then receives a `BasicCrawlingContext` (recorded as `None`) instead of a `PlaywrightCrawlingContext`, so the exact-list assertion `[HELLO_WORLD, HELLO_WORLD]` breaks with `[None, HELLO_WORLD]`.

The crawler behaves correctly here (it retries, and a non-navigated attempt legitimately yields a basic context), so this is a test flake, not a product bug. The same `net::ERR_NO_BUFFER_SPACE` failure is already documented for three sibling tests (`test_resource_management`, `test_adaptive_crawling_pre_nav_change_to_context`, `test_stagehand_browser_controller`), which handle it with `@pytest.mark.flaky(reruns=3)`.

Rather than retry, this makes the assertion robust to spurious pre-navigation failures: it ignores attempts that never reached the page (`None`) and asserts that at least one attempt did, and that every attempt that did exposes the page HTML. Removes the test's sensitivity to the flake deterministically while preserving what it verifies (that the error handler can read the page). No product code changes.

### DIFF
--- a/tests/unit/crawlers/_playwright/test_playwright_crawler.py
+++ b/tests/unit/crawlers/_playwright/test_playwright_crawler.py
@@ -1238,8 +1238,18 @@ async def test_error_handler_can_access_page(server_url: URL) -> None:
 
     await crawler.run([str(server_url / 'hello-world')])
 
-    assert error_handler_calls == [HELLO_WORLD.decode(), HELLO_WORLD.decode()]
-    assert failed_handler_calls == [HELLO_WORLD.decode()]
+    # The error handler runs on each retry and the failed-request handler on the final failure, each recording the
+    # page content when it received a `PlaywrightCrawlingContext` or `None` otherwise. On CI (notably Windows under
+    # `xdist` load) navigation can spuriously fail with `net::ERR_NO_BUFFER_SPACE` before the page is created, so that
+    # attempt surfaces a `BasicCrawlingContext` recorded as `None`. Such attempts are environmental noise rather than
+    # the behavior under test, so assert on the attempts that actually reached the page: at least one must have, and
+    # every one that did must expose the page HTML.
+    page_error_calls = [content for content in error_handler_calls if content is not None]
+    page_failed_calls = [content for content in failed_handler_calls if content is not None]
+
+    assert page_error_calls, 'the error handler never received a PlaywrightCrawlingContext'
+    assert all(content == HELLO_WORLD.decode() for content in page_error_calls)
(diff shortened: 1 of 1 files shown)


tests/unit/crawlers/_playwright/test_playwright_crawler.py
Pijukatel: What is the point of keeping the `[None]` calls and then filtering them out later?

vdusek: fixed

tests/unit/crawlers/_playwright/test_playwright_crawler.py
```diff
@@ -1220,32 +1220,30 @@ async def test_error_handler_can_access_page(server_url: URL) -> None:
...
-    assert all(content == HELLO_WORLD.decode() for content in page_error_calls)
-    assert all(content == HELLO_WORLD.decode() for content in page_failed_calls)
+    # The error handler runs on each retry and the failed-request handler on the final failure. Each records the page
+    # content only when it received a `PlaywrightCrawlingContext`. On CI (notably Windows under `xdist` load) navigation
+    # can spuriously fail with `net::ERR_NO_BUFFER_SPACE` before the page is created, yielding a `BasicCrawlingContext`
+    # that never reached the page. Such attempts are environmental noise rather than the behavior under test, so assert
+    # only on the attempts that actually reached the page: at least one must have, and every one that did exposes it.
+    assert error_handler_calls, 'the error handler never received a PlaywrightCrawlingContext'
+    assert all(content == HELLO_WORLD.decode() for content in error_handler_calls)
+    assert all(content == HELLO_WORLD.decode() for content in failed_handler_calls)
```
Pijukatel: This looks more like a set, but it does not raise when `failed_handler_calls= []`, is that intentional or not?
assert all(content == HELLO_WORLD.decode() for content in []) -> OK?

Otherwise, I would suggest turning `error_handler_calls` and `failed_handler_calls` into sets
    assert {HELLO_WORLD.decode()} == error_handler_calls
    assert {HELLO_WORLD.decode()} == failed_handler_calls

Pijukatel (approved):

PR #2030, opened 2026-07-07
chore(ci): Update renovate schedules

Pijukatel:
- Replace deprecated `breejslater` syntax with `cron` syntax
https://docs.renovatebot.com/key-concepts/scheduling/#deprecated-breejslater-syntax
- Use these triggers, limit the window to avoid spam:
  - Every **Tuesday - Friday** between **0:00 and 6:00** to update **dependencies**
  - Every **Monday** between **0:00 and 6:00** to update **lockfiles** 
- Do not run both jobs on the same day.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -10,7 +10,8 @@
     "lockFileMaintenance": {
         "enabled": true,
         "automerge": true,
-        "automergeType": "branch"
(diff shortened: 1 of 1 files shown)


vdusek (approved):
great, thanks

PR #2033, opened 2026-07-07
chore(ci): modernize Renovate config

anxkhn:
`renovate.json` uses three Renovate options that were renamed or removed in Renovate v36. Renovate still accepts them by running an internal config migration on every run, but because this config sets `dependencyDashboard: false` and does not enable `configMigration`, Renovate never opens a migration PR or surfaces a dashboard checkbox, so the legacy option names otherwise persist.

This updates them to the current names:

- `extends`: `config:base` -> `config:recommended` (the preset was renamed in v36).
- `pinVersions: false` -> `rangeStrategy: "replace"`. Renovate's config migration converts `pinVersions: false` into `rangeStrategy: "replace"`, so making it explicit matches the config Renovate already computes internally.
- `packageRules[].matchPaths` -> `matchFileNames` (`matchPaths` and `matchFiles` were merged into `matchFileNames` in v36).

These edits match what `renovate-config-validator` emits when it migrates the current config, so there is no behavior change, only the removal of deprecated option names.

### Issues

- N/A (config housekeeping; no related issue).

### Testing

Validated with Renovate's official config validator:

```bash
npx --package renovate -- renovate-config-validator --strict renovate.json
```

- Before: exit 1, `WARN: Config migration necessary`.
- After: exit 0, `INFO: Config validated successfully`.

No unit test applies because this is CI/tooling config; no in-repo workflow, pre-commit hook, or formatter touches `renovate.json`.

### Checklist

- [ ] CI passed

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,7 +1,7 @@
 {
-    "extends": ["config:base", ":semanticCommitTypeAll(chore)"],
+    "extends": ["config:recommended", ":semanticCommitTypeAll(chore)"],
     "ignorePaths": ["docs/**", "src/crawlee/project_template/**", "website/versioned_docs/**"],
-    "pinVersions": false,
+    "rangeStrategy": "replace",
     "separateMajorMinor": false,
     "dependencyDashboard": false,
     "semanticCommits": "enabled",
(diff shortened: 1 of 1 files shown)


vdusek (approved):
LGTM

PR #456, opened 2024-08-22
refactor!: declare private and public interface

vdusek:
### Description

- declare private and public interface

### Issues

- N/A

### Testing

- All existing tests are passing.

### Checklist

- [x] CI passed

PR #2036, opened 2026-07-07
chore: remove redundant desired concurrency defaults from docs

anxkhn:
`ConcurrencySettings.desired_concurrency` defaults to `10`
(`src/crawlee/_types.py`, `desired_concurrency: int = 10`), and that value is used
verbatim by the autoscaled pool. The docstring and the scaling guide, however,
still say it defaults to `min_concurrency`, which is stale and misleading for
anyone tuning concurrency.

This corrects both documentation sites to state the real default. Docs-only, no
behavior change.

- `src/crawlee/_types.py`: docstring for `desired_concurrency` now reads "By
  default, it is `10`." (was "By default, it is `min_concurrency`.")
- `docs/guides/scaling_crawlers.mdx`: the "Desired concurrency" section now reads
  "It defaults to `10`." (was "It defaults to the same value as
  `min_concurrency`.")

Background: the default was changed from `None` (resolved to `min_concurrency`) to
a hardcoded `10` in #1419, but the docstring (originally from #456) and the guide
were not updated at the time.

### Issues

- No related issue. Self-identified documentation drift; this repo has no
  `good first issue` label and issues are maintainer-assigned. Happy to open an
  issue first if preferred.

### Testing

- `uv run poe lint` passes clean (ruff format check + ruff check).
- `uv run poe type-check` shows the same 30 pre-existing diagnostics as `master`
  (the optional `pydantic_ai` unresolved import); zero introduced by this change.
- Runtime sanity: `ConcurrencySettings().desired_concurrency == 10` while
  `min_concurrency == 1`, confirming the old wording was wrong.
- No unit test added: this is a pure docs correction with no behavior change, and
  every existing test that builds `ConcurrencySettings` passes `desired_concurrency`
  explicitly, so nothing depends on the default. (Consistent with the docstring-fix
  precedent in #742.)

### Checklist

- [ ] CI passed

### DIFF
--- a/docs/guides/scaling_crawlers.mdx
+++ b/docs/guides/scaling_crawlers.mdx
@@ -38,7 +38,7 @@ If you set `min_concurrency` too high compared to the available system resources
 
 ## Desired concurrency
 
-The `desired_concurrency` option in the <ApiLink to="class/ConcurrencySettings">`ConcurrencySettings`</ApiLink> specifies the initial number of parallel tasks to start with, assuming sufficient resources are available. It defaults to `10`.
+The `desired_concurrency` option in the <ApiLink to="class/ConcurrencySettings">`ConcurrencySettings`</ApiLink> specifies the initial number of parallel tasks to start with, assuming sufficient resources are available.
 
 <RunnableCodeBlock className="language-python" language="python">
     {MinAndMaxConcurrencyExample}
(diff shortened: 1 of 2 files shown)


vdusek (approved):
rather remove the defaults from the docs (they are redundant)

PR #2042, opened 2026-07-07
fix(curl-impersonate): Avoid false already consumed error for buffered streams in `CurlImpersonateHttpClient`

Mantisus:
### Description

- `curl_cffi` buffers a streamed response in a queue, so a small response body could be fully buffered before reading started. This caused a false `RuntimeError: Cannot read stream, it was already consumed.` from `read_stream`.

### Testing

- Added new regression tests for HTTP clients

### DIFF
--- a/tests/unit/http_clients/test_http_clients.py
+++ b/tests/unit/http_clients/test_http_clients.py
@@ -15,6 +15,7 @@ from crawlee import Request
 from crawlee.errors import ProxyError
 from crawlee.http_clients import CurlImpersonateHttpClient, HttpClient, HttpxHttpClient, ImpitHttpClient
 from crawlee.statistics import Statistics
+from tests.unit.server import generate_file_content
 from tests.unit.server_endpoints import HELLO_WORLD
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


tests/unit/http_clients/test_http_clients.py
```diff
@@ -177,6 +178,31 @@ async def test_stream(http_client: HttpClient, server_url: URL) -> None:
     assert content_body == HELLO_WORLD
 
 
+async def test_stream_read_after_transfer_finished(http_client: HttpClient, server_url: URL) -> None:
+    """A small body can be fully buffered before `read_stream` is called. It must still be readable."""
+    content_body: bytes = b''
+
+    async with http_client.stream(str(server_url)) as response:
+        # Give the event loop time to finish the transfer before the body is consumed.
+        await asyncio.sleep(0.1)
```
vdusek: Couldn't we write this test in a deterministic way instead of using wait?

Mantisus: Updated. The test now uses a response with a small body that can be buffered by the client. `asyncio.sleep(0)` is needed to ensure that the context switches to the buffering task.

vdusek (approved):
LGTM

Issue #874, opened 2025-01-06
Reconsider HttpClient interface


janbuchar:
As of now, the python `BaseHttpClient` look like this:

https://github.com/apify/crawlee-python/blob/beac9fa0eb415caafc04cdaef2888e77fad915e0/src/crawlee/http_clients/_base.py#L55

It has two methods, `send_request` and `crawl`. This is the first iteration of decoupled HTTP clients.

Later on, we refactored the JS version to use this one:

https://github.com/apify/crawlee/blob/f912b8b06da2bc4f3f3db508cc39c936a5c87f23/packages/core/src/http_clients/base-http-client.ts#L179

It also has two methods, `sendRequest` and `stream`. Unlike the python version, the signatures of those methods match quite well. It is worth noting that the two serious attempts at implementing this interface (so far) both couldn't manage to implement `stream` correctly. Although we could probably live without it in the most common case, streaming is paramount for downloading files (potentially large ones), which is a use case that we want to support.

We should simplify this interface and make it look the same in both versions. Any thoughts on how to achieve that? @vdusek @B4nan @barjin... and whoever else wants to chat :slightly_smiling_face:

B4nan:
I like the JS version, but given many HTTP clients won't support streaming or it will be challenging to implement, I think we could have a default implementation of the `stream` method as part of the base class. So in JS instead of having an interface, it would be an abstract class with the stream method implemented with a fallback like we do with the curl-impersonate client.

I am not sure why there is that `crawl` method in python, if we don't support streaming there at all, ideally we would drop it (and implement a proxy at the call site if we need a different signature). Feels weird that `crawler.run` and `send_request` would use something else. And we should surely think of how to support streaming there too, but that's another story.

vdusek:
> I am not sure why there is that crawl method in python, if we don't support streaming there at all, ideally we would drop it (and implement a proxy at the call site if we need a different signature). Feels weird that crawler.run and send_request would use something else. And we should surely think of how to support streaming there too, but that's another story.

Mostly historical reasons - the first attempt of implementation of HTTP clients. The `crawl` method is being called from the `crawler.run`, and `send_request` is being called from the `send_request` context helper. They have a slightly different APIs, but I believe they could be unified.

[...]

PR #2043, opened 2026-07-08
feat: add FileDownloadCrawler for downloading and streaming files

### LINKED ISSUE #875: Implement a `FileDownload` crawler
https://github.com/apify/crawlee/blob/master/packages/http-crawler/src/internals/file-download.ts may serve as inspiration

Mantisus:
### Description

- Adds `FileDownloadCrawler` with support for streaming the response body in the request handler.

### Issues

- Closes: #875

### Testing

- Added tests for `FileDownloadCrawler`

### DIFF
--- a/docs/examples/code_examples/file_download_stream.py
+++ b/docs/examples/code_examples/file_download_stream.py
@@ -14,9 +14,9 @@ async def main() -> None:
     # been read yet.
     crawler = FileDownloadCrawler(
         stream=True,
-        # The streamed request runs under this timeout, so raise it for large downloads.
+        # Bounds establishing the connection and receiving the response headers.
         navigation_timeout=timedelta(minutes=5),
-        # The handler stays active while streaming the body to disk.
+        # The body is downloaded inside the handler, so this bounds the transfer itself.
         request_handler_timeout=timedelta(minutes=5),
         # Limit the crawl to max requests. Remove or increase it for crawling all links.
         max_requests_per_crawl=10,
--- a/docs/examples/code_examples/file_download_stream.py
+++ b/docs/examples/code_examples/file_download_stream.py
@@ -1,0 +1,49 @@
+import asyncio
+from datetime import timedelta
+from pathlib import Path
+
+from yarl import URL
+
+from crawlee.crawlers import FileDownloadCrawler, FileDownloadCrawlingContext
+
+DOWNLOAD_DIR = Path('downloads')
+
+
+async def main() -> None:
+    # With stream=True, the request handler receives a response whose body has not
+    # been read yet.
+    crawler = FileDownloadCrawler(
+        stream=True,
+        # The streamed request runs under this timeout, so raise it for large downloads.
+        navigation_timeout=timedelta(minutes=5),
+        # The handler stays active while streaming the body to disk.
+        request_handler_timeout=timedelta(minutes=5),
+        # Limit the crawl to max requests. Remove or increase it for crawling all links.
+        max_requests_per_crawl=10,
+    )
+
+    # Define the default request handler, which will be called for every request.
+    @crawler.router.default_handler
+    async def request_handler(context: FileDownloadCrawlingContext) -> None:
+        context.log.info(f'Downloading {context.request.url} ...')
+
+        file_name = URL(context.request.url).name
+
+        # Write each chunk to disk as it arrives, without buffering the whole file.
+        with (DOWNLOAD_DIR / file_name).open('wb') as file:
+            async for chunk in context.http_response.read_stream():
+                file.write(chunk)
+
+        context.log.info(f'Saved {file_name}')
+
+    # Run the crawler with the list of files to download.
+    await crawler.run(
+        [
+            'https://samplelib.com/mp4/sample-15s-720p.mp4',
+        ]
+    )
+
+
+if __name__ == '__main__':
+    DOWNLOAD_DIR.mkdir(exist_ok=True)
+    asyncio.run(main())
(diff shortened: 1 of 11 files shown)


src/crawlee/crawlers/_file_download/_file_download_crawler.py
```diff
@@ -0,0 +1,154 @@
...
+                        headers=context.request.headers,
+                        payload=context.request.payload,
+                        session=context.session,
+                        proxy_info=context.proxy_info,
+                        timeout=remaining_timeout,
+                    )
+                )
+
+            self._statistics.register_status_code(response.status_code)
+            context.request.state = RequestState.AFTER_NAV
```
vdusek, line 150: From Claude:

> The streamed path never sets `context.request.loaded_url`, whereas the buffered `crawl()` path does (`request.loaded_url = str(response.url)`). As a result, in stream mode `_check_url_after_redirects` is effectively skipped (it's guarded by `if loaded_url is not None`), so off-domain redirects aren't filtered by the enqueue strategy, and `context.request.loaded_url` stays `None` for user code. It's a silent divergence between the two modes. The `stream()` API doesn't expose the final URL, so this may be hard to fix cleanly, but worth a note/follow-up.

docs/examples/code_examples/file_download_stream.py
```diff
@@ -0,0 +1,49 @@
...
+
+DOWNLOAD_DIR = Path('downloads')
+
+
+async def main() -> None:
+    # With stream=True, the request handler receives a response whose body has not
+    # been read yet.
+    crawler = FileDownloadCrawler(
+        stream=True,
+        # The streamed request runs under this timeout, so raise it for large downloads.
```
vdusek: Claude's nit:

> In stream mode `navigation_timeout` only bounds connection establishment + headers, not the body download (that runs under `request_handler_timeout`, as the next comment correctly says). This wording reads as if the whole download is bounded by `navigation_timeout`, so a reader tuning only this value for a large file would still hit the handler timeout.

src/crawlee/crawlers/_file_download/_file_download_crawler.py
```diff
@@ -0,0 +1,154 @@
...
+    # Define the default request handler, which will be called for every request.
+    @crawler.router.default_handler
+    async def request_handler(context: FileDownloadCrawlingContext) -> None:
+        context.log.info(f'Downloading {context.request.url} ...')
+
+        # Read the whole file and save it to the key-value store.
+        content = await context.http_response.read()
+        kvs = await context.get_key_value_store()
+        await kvs.set_value(
+            key=context.request.url.split('/')[-1],
```
vdusek: Claude's nit:

> `context.request.url.split('/')[-1]` yields `''` for a trailing-slash URL, and `kvs.set_value(key='')` would raise. The runnable example (`file_download.py`) uses the safer `URL(context.request.url).name` — worth matching here (and at line 73) so the docstring doesn't teach the more fragile pattern.

docs/guides/http_crawlers.mdx
```diff
@@ -131,6 +134,10 @@ The following examples demonstrate how to integrate with several popular parsing
     </TabItem>
 </Tabs>
 
+## FileDownloadCrawler
+
+The <ApiLink to="class/FileDownloadCrawler">`FileDownloadCrawler`</ApiLink> downloads files instead of scraping pages. It accepts any content type without parsing and gives the request handler direct access to the response body. By default the whole file is buffered in memory. For large files, construct the crawler with `stream=True` and consume the body in chunks via <ApiLink to="class/HttpResponse#read_stream">`read_stream()`</ApiLink>. For usage, see the [Download files](../examples/file-download) example.
```
Pijukatel, line 139: For large files, I was wondering if we could benefit from https://github.com/apify/apify-client-python/pull/927

Can you explore a little bit how to utilize this from the Crawlee level?

Pijukatel (approved):
For large files, it would be great to have something like https://pypi.org/project/smart-open/ , but our storages lacks interface to support such chunked uploads

docs/guides/http_crawlers.mdx
Mantisus: > For large files, I was wondering if we could benefit from https://github.com/apify/apify-client-python/pull/927

The compression added in the PR (the `Content-Encoding` header) is a transport-level optimization that only affects how data moves over the network between the client and the server. It's fully transparent to the user, who never sees the compressed bytes directly. Similarly, when the server sends compressed data based on the client's `Accept-Encoding` header, decompression occurs automatically at the HTTP client level (`impit`, `curl_cffi`, `httpx`).

So we already get all the benefits from this.

> Can you explore a little bit how to utilize this from the Crawlee level?

If we wanted to add compression at the Crawlee level, it would need to be a storage-level feature instead. It would compress data before it's written to the storage backend (files, Redis, SQL, etc.). Since that changes the stored format, it should be explicit and opt-in rather than automatic. For example, as a literal argument to `KeyValueStore.set_value`.

vdusek (commented):
Otherwise it looks great

src/crawlee/crawlers/_file_download/_file_download_crawler.py
Mantisus: That's true. But we can't do that without updating the HTTP clients API. This will be fixed along with #874 when the clients are synchronized with Crawlee JS v4.

vdusek (changes requested):
LGTM. However, let's wait with merging for now so we can release one more patch version before we start merging changes for 1.9.0.

vdusek (approved):

Issue #2063, opened 2026-07-13
Migrate CLI to typer >=0.26 and drop the `typer<0.26` pin


vdusek:
## Summary

`crawlee[cli]` currently pins `typer>=0.12.0,<0.26` in `pyproject.toml`. This issue tracks migrating
`src/crawlee/_cli.py` to typer 0.26's choice handling so the pin can be dropped.

## Background

typer 0.26 vendored click into a private `typer._click` module. `_cli.py` builds its
`--crawler-type` / `--http-client` / `--package-manager` options by passing `click.Choice(...)`
(from the external `click`) into `typer.Option(click_type=...)`:

```python
from click import Choice

crawler_type: str | None = typer.Option(
    ...,
    click_type=Choice(crawler_choices),
    ...
)
```

Under typer 0.26 that `click.Choice` is a *foreign* `ParamType` (typer expects its own
`typer._click.types.ParamType`), which breaks in two ways:

1. **Type checking** — `ty` reports `invalid-argument-type`: `Expected ParamType | None, found
   Choice[...]`. Parametrizing as `Choice[str]` does not help; the two `ParamType` classes are
   simply distinct.
2. **Runtime regression** — an invalid choice raises an uncaught `click.BadParameter` (full
   traceback, exit code 1) instead of a clean usage error (`Error: Invalid value for
   '--crawler-type': ...`, exit code 2), because typer's error handling only catches its own
   vendored `ClickException`. typer wraps the foreign `Choice` as a `FuncParamType`, so the real
   click exception escapes.

typer 0.26 exposes no public `Choice`, so there is no drop-in replacement.

## Task

- Migrate the three choice-based options in `_cli.py` to typer 0.26's choice handling. Options:
  - Validate the values manually against the choice lists (`crawler_choices`, `http_client_choices`,
    `package_manager_choices`) and raise `typer.BadParameter` on mismatch, or
  - Use an `Enum`-based approach.
- Verify `crawlee create --help` still lists the choices and shell completion still works.
- Verify an invalid choice produces a clean usage error (exit code 2), not a traceback.
- Drop the `typer<0.26` pin in `pyproject.toml` (the `cli` extra) and re-lock.

## References

- Pin introduced in the lock-file-maintenance PR #2062.

Pijukatel:
This should also be confirmed by running E2E tests on any change that touches the typer version.

PR #2064, opened 2026-07-13
fix(cli): pin typer<0.26 to keep CLI choice options working

vdusek:
typer 0.26 vendored click into a private `typer._click` module, so the `click.Choice` we pass to `typer.Option(click_type=...)` in `_cli.py` becomes a foreign `ParamType`. That fails `ty` and, worse, makes an invalid CLI choice raise an uncaught `click.BadParameter` (full traceback, exit code 1) instead of a clean usage error (exit code 2).

This pins `typer<0.26` in the `cli` extra so the CLI keeps working, and removes the now-unnecessary `# ty: ignore` comments that #2062 had added. The full migration to typer 0.26's choice handling is tracked in #2063.


Pijukatel:
Running the E2E tests https://github.com/apify/crawlee-python/actions/runs/29239052291

vdusek:
@Pijukatel working

Pijukatel (approved):

PR #2065, opened 2026-07-13
fix(robots): disallow crawling when robots.txt returns a 5xx

anxkhn:
`RobotsTxtFile.load` only special-cased 4xx responses (file unavailable -> allow
all) and parsed the raw response body as robots rules for every other status. The
HTTP clients do not raise on 5xx, so a server-error response reached that branch and
its body was handed to Protego. Protego reads a non-robots body (for example an HTML
`500` error page, or an empty body) as allow-all, so a transient 5xx on `robots.txt`
silently permitted crawling every URL, the opposite of what the site expects while it
is failing.

Per RFC 9309 section 2.3.1.4, a `robots.txt` that is unreachable due to server errors
(5xx) is undefined and a crawler "MUST assume complete disallow". This PR handles 5xx
explicitly with a disallow-all directive, using the existing
`is_status_code_server_error` helper. The 4xx (allow-all, section 2.3.1.3), 2xx/3xx
(parse the body), and network-error paths are left unchanged.

This only affects users who opt in with `respect_robots_txt_file=True`, and only when
`robots.txt` returns a 5xx: previously such a response allowed crawling everything, now
it correctly disallows until the file is reachable again. It also matches crawlee-JS,
which never parses a non-2xx `robots.txt` body as rules.

### Issues

- No open issue tracks this specific behavior; it was found by code review. Happy to
  open one first if the maintainers prefer to track it separately.

### Testing

- Added `tests/unit/_utils/test_robots.py::test_server_error_disallows_all`: loads
  `robots.txt` from the test server's `status/500` route and asserts the resulting
  rules disallow arbitrary paths. It runs against all three HTTP clients
  (httpx / impit / curl) via the existing `http_client` fixture.
- Verified red -> green: on the pre-fix code the new test fails on all three clients
  (a 5xx is allow-all); with the fix it passes.
- `uv run pytest tests/unit/_utils/test_robots.py` -> 17 passed.
- `uv run pytest tests/unit/crawlers/{_basic,_beautifulsoup,_parsel} -k robots` ->
  13 passed (no regression in `respect_robots_txt_file`).
- `uv run poe lint` and `uv run poe type-check`: clean (no new diagnostics).

### Checklist

- [ ] CI passed

### DIFF
--- a/src/crawlee/_utils/robots.py
+++ b/src/crawlee/_utils/robots.py
@@ -8,7 +8,7 @@ from yarl import URL
 
 from crawlee._utils.sitemap import Sitemap
 from crawlee._utils.urls import filter_url
-from crawlee._utils.web import is_status_code_client_error
+from crawlee._utils.web import is_status_code_client_error, is_status_code_server_error
 
 if TYPE_CHECKING:
     from typing_extensions import Self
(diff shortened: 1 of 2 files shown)


vdusek (approved):
LGTM

Mantisus (approved):
LGTM

PR #2066, opened 2026-07-13
fix(fingerprint): keep the original error as the cause when generation fails

anxkhn:
`BrowserforgeFingerprintGenerator.generate` retries `browserforge` fingerprint
generation (which is known to be flaky and raises `ValueError` when the screen
or header constraints cannot be satisfied) inside `for attempt in range(max_attempts)`
with an `if attempt == max_attempts: raise` guard:

```python
max_attempts = 10
for attempt in range(max_attempts):
    try:
        return self._generator.generate(**self._options)
    except ValueError:  # noqa:PERF203
        if attempt == max_attempts:
            raise
raise RuntimeError('Failed to generate fingerprint.')
```

`range(10)` yields `0..9`, so `attempt == max_attempts` (`== 10`) is never true
and the inner `raise` is dead code. On persistent failure the loop falls through
to a bare `RuntimeError('Failed to generate fingerprint.')` whose `__cause__` and
`__context__` are both `None`, so the original `ValueError` (which carries the
real reason the constraints could not be satisfied) is discarded and the failure
is hard to diagnose.

This drops the dead guard and chains the last `ValueError` through the final
`RuntimeError` via `raise ... from`, so the diagnostic context is preserved. The
retry count (10) and the raised `RuntimeError` type and message are unchanged;
the only behavior change is that the exception now has a `__cause__`.

This also makes the method consistent with its sibling
`PatchedHeaderGenerator.generate` in the same module, which already loops
`for _attempt in range(max_attempts)` and raises `RuntimeError` only after the
loop, with no in-loop guard.

### Issues

- No existing issue; this was found while reading the fingerprint suite. The
  behavior was introduced in #829.

### Testing

- Added `test_fingerprint_generator_preserves_error_on_persistent_failure` in
  `tests/unit/fingerprint_suite/test_adapters.py`: it monkeypatches the
  underlying generator to always raise `ValueError`, then asserts that
  `generate()` raises `RuntimeError` after exactly 10 attempts and that
  `__cause__` is the original `ValueError`. It fails on the current code (the
  `RuntimeError`'s `__cause__` is `None`) and passes with this change.
- `uv run pytest tests/unit/fingerprint_suite/ -q` -> 17 passed, no regressions.
- `uv run poe lint` and `uv run poe type-check` are clean for the changed files.

### Checklist

- [ ] CI passed

### DIFF
--- a/src/crawlee/fingerprint_suite/_browserforge_adapter.py
+++ b/src/crawlee/fingerprint_suite/_browserforge_adapter.py
@@ -230,13 +230,13 @@ class BrowserforgeFingerprintGenerator(FingerprintGenerator):
         # During test runs around 10 % flakiness was detected.
         # Max attempt set to 10 as (0.1)^10 is considered sufficiently low probability.
         max_attempts = 10
-        for attempt in range(max_attempts):
+        last_error: ValueError | None = None
+        for _attempt in range(max_attempts):
             try:
                 return self._generator.generate(**self._options)
-            except ValueError:  # noqa:PERF203
-                if attempt == max_attempts:
-                    raise
-        raise RuntimeError('Failed to generate fingerprint.')
+            except ValueError as exc:  # noqa:PERF203
+                last_error = exc
+        raise RuntimeError('Failed to generate fingerprint.') from last_error
 
 
 class BrowserforgeHeaderGenerator:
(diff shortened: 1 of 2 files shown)


vdusek (approved):
LGTM

Pijukatel (approved):

PR #2074, opened 2026-07-14
fix(memory-storage): stop re-adds from duplicating and resetting pending requests

anxkhn:
- Store pending requests in an `OrderedDict` keyed by unique key instead of a `deque`. Repositioning a request to the forefront is now O(1) instead of an O(n) `deque.remove()` scan, so re-adding K already-pending requests is O(K) instead of O(K*N) — measured 22.4s → 0.004s for 2000 forefront re-adds into a 20k-request queue. The scan only triggered for requests already waiting in the queue, which is what `add_requests(..., forefront=True)` hits when a handler re-discovers URLs that are still pending.
- Fix the same URL being handed out repeatedly. The old scan silently failed once an earlier regular re-add had replaced the registered request object with a differently-valued one, leaving a stale entry behind and appending a second one for the same unique key. Five URLs re-added over three rounds yielded 20 fetches instead of 5 and left `pending_request_count` at -15. Duplicates are now impossible by construction.
- Keep the originally enqueued request on a re-add. The old code replaced the registered request with the incoming duplicate, which on a forefront re-add also replaced the queued one — resetting `retry_count` (a request could retry forever), `label` (dispatch to a different handler) and `user_data`. On a regular re-add it left the two out of sync, so `get_request` and `fetch_next_request` disagreed. This restores the documented contract ("Duplicates will be identified but not re-added to the queue") and matches the file-system, SQL and Redis clients, as well as `@crawlee/memory-storage` in Crawlee for JS.
- Make `get_request` consistent after a reclaim: it now returns the reclaimed object, the same one `fetch_next_request` hands back.
- Write the queue metadata once per batch instead of once per new request, saving N-1 `datetime.now()` calls on a batch add.

*✍️ Drafted by Claude Code*


vdusek (commented):
.

src/crawlee/storage_clients/_memory/_request_queue_client.py
```diff
@@ -175,19 +174,15 @@ async def add_batch_of_requests(
                 )
                 continue
 
-            # If the request is already in the queue but not handled, update it.
+            # If the request is already in the queue but not handled, we only reposition it when `forefront`
+            # is set; a regular re-add leaves the already-pending entry untouched.
```
vdusek: > This drops the pre-existing behavior where a non-forefront re-add re-stored the request object (`_requests_by_unique_key[key] = request`), i.e. picked up the new object's `user_data` / headers. Now that data is silently discarded on a regular re-add. Intentional? It does match the file-system client (which also doesn't re-store on a non-forefront re-add), so likely yes — but since it's a behavior change for the memory client, a one-line code comment saying the drop is deliberate parity would help.

src/crawlee/storage_clients/_memory/_request_queue_client.py
```diff
@@ -175,19 +174,15 @@ async def add_batch_of_requests(
...
                 if forefront:
-                    # Update the existing request with any new data and
-                    # remove old request from pending queue if it's there.
-                    with suppress(ValueError):
-                        self._pending_requests.remove(existing_request)
-
-                    # Add updated request back to queue.
+                    # Move the request to the front. The old entry is left in the deque instead of being
+                    # located and removed (`deque.remove` is O(n), which makes a batch of forefront re-adds
+                    # O(n^2)); registering the new object here supersedes the old one, and the stale entry is
```
vdusek: > Leaving the old entry in the deque trades the O(n) `remove` for potential unbounded growth — repeated forefront re-adds of still-pending requests pile up stale entries that are reclaimed only when `fetch_next_request` pops them (or `is_empty` front-prunes). Trailing stale entries behind a live front persist until fetched, so a forefront-heavy pattern that outpaces fetching can grow the deque well beyond the live request count. Fine for normal fetch cycles, but worth acknowledging as a trade-off.

src/crawlee/storage_clients/_memory/_request_queue_client.py
```diff
@@ -309,6 +309,13 @@ async def reclaim_request(
     async def is_empty(self) -> bool:
         await self._update_metadata(update_accessed_at=True)
 
+        # Discard stale entries left at the front by forefront repositioning; a live request is always
+        # enqueued ahead of the stale entry it supersedes, so pruning stops at the first live request.
+        while self._pending_requests and (
```
vdusek: > `is_empty` is a predicate that now mutates `_pending_requests` (and it's also reached via `is_finished` and the higher-level `RequestQueue.is_empty`). It's safe here — no `await` inside the loop, so the prune is atomic w.r.t. other tasks — but a query with a side effect is surprising; consider a short comment noting the pruning is intentional.

src/crawlee/storage_clients/_memory/_request_queue_client.py
```diff
@@ -309,6 +309,13 @@ async def reclaim_request(
     async def is_empty(self) -> bool:
         await self._update_metadata(update_accessed_at=True)
 
+        # Discard stale entries left at the front by forefront repositioning; a live request is always
+        # enqueued ahead of the stale entry it supersedes, so pruning stops at the first live request.
```
vdusek: > This invariant isn't strictly guaranteed — a `reclaim_request(forefront=False)` after a forefront reposition re-appends the live copy *behind* the stale one. The code stays correct because the loop prunes *all* leading stale entries (not just one), but a future "prune a single entry" tweak would silently break `is_empty`. Maybe reword to lean on "prune all leading stale entries" rather than the ordering claim.

tests/unit/storage_clients/_memory/test_memory_rq_client.py
```diff
@@ -93,3 +94,105 @@ async def test_memory_metadata_updates(rq_client: MemoryRequestQueueClient) -> N
...
+    await rq_client.add_batch_of_requests(requests)
+
+    # Re-add the same, still-pending requests with `forefront=True`. Previously each re-add scanned the
+    # whole pending deque with `deque.remove` to reposition the existing entry, which is O(n) per request.
+    await rq_client.add_batch_of_requests(requests, forefront=True)
+
+    assert CountingDeque.remove_calls == 0
+
+
+async def test_forefront_readd_preserves_order_and_dedup(rq_client: MemoryRequestQueueClient) -> None:
```
vdusek, line 131: > Consider adding a test that reclaims a request (`reclaim_request`, non-forefront) after it was repositioned to the forefront. That's the path where the live copy ends up behind its stale tombstone and it exercises the full front-prune in `is_empty` — the one lifecycle the current tests don't cover.

tests/unit/storage_clients/_memory/test_memory_rq_client.py
```diff
@@ -93,3 +94,105 @@ async def test_memory_metadata_updates(rq_client: MemoryRequestQueueClient) -> N
...
+        'https://example.com/2',
+    ]
+
+    # No stale duplicates should linger after all live requests are drained.
+    assert await rq_client.is_empty() is True
+    assert await rq_client.is_finished() is True
+
+
+async def test_regular_readd_of_pending_request_is_not_dropped(rq_client: MemoryRequestQueueClient) -> None:
+    """Test that a regular (non-forefront) re-add of a still-pending request keeps it fetchable.
```
vdusek: One line docstrings for tests please.

vdusek (changes requested):

src/crawlee/storage_clients/_memory/_request_queue_client.py
```diff
@@ -175,27 +174,21 @@ async def add_batch_of_requests(
...
 
             # Add the new request to the queue.
             else:
                 if forefront:
-                    self._pending_requests.appendleft(request)
+                    self._pending_requests[request.unique_key] = request
+                    self._pending_requests.move_to_end(request.unique_key, last=False)
                 else:
-                    self._pending_requests.append(request)
+                    self._pending_requests[request.unique_key] = request
```
vdusek: Identical assignment in both arms — hoist it and keep only the `move_to_end` conditional.

```suggestion
                self._pending_requests[request.unique_key] = request
                if forefront:
                    self._pending_requests.move_to_end(request.unique_key, last=False)
```

src/crawlee/storage_clients/_memory/_request_queue_client.py
```diff
@@ -175,27 +174,21 @@ async def add_batch_of_requests(
                 )
                 continue
 
-            # If the request is already in the queue but not handled, update it.
+            # If the request is already in the queue but not handled, update it without changing its position.
```
vdusek: "without changing its position" is inaccurate — forefront re-adds move it to the front.

```suggestion
            # If the request is already in the queue but not handled, update it; forefront moves it to the front.
```

anxkhn:

vdusek:
@anxkhn Thank you. For future reviews, I'd really appreciate it if you could address all the feedback in a single follow-up commit, rather than splitting the changes across multiple commits, merging them into earlier commits, and force-pushing it. Your current approach makes the changes really hard to review.

anxkhn:
understood, i’ll keep future review fixes in a single follow-up commit without rewriting earlier commits.

tests/unit/storage_clients/_memory/test_memory_rq_client.py
```diff
@@ -93,3 +93,115 @@ async def test_memory_metadata_updates(rq_client: MemoryRequestQueueClient) -> N
...
+
+    for _ in range(10):
+        await rq_client.add_batch_of_requests(requests, forefront=True)
+
+    fetched = []
+    while (request := await rq_client.fetch_next_request()) is not None:
+        fetched.append(request)
+        await rq_client.mark_request_as_handled(request)
+
+    assert len(fetched) == len(requests)
```
vdusek: Note: this doesn't guard the perf fix anymore. Running the new tests against `master` shows only `test_regular_readd_of_pending_request_is_not_dropped` failing - this one passes on the old O(n) `deque.remove` implementation too, since that path also produced no duplicates. With `OrderedDict` the O(1) is structural, so rather than trying to assert it, it may be better to say so in the docstring.

*✍️ Drafted by Claude Code*

src/crawlee/storage_clients/_memory/_request_queue_client.py
```diff
@@ -175,27 +174,19 @@ async def add_batch_of_requests(
                 )
                 continue
 
-            # If the request is already in the queue but not handled, update it.
+            # If the request is already in the queue but not handled, update it; forefront moves it to the front.
             if was_already_present and existing_request:
```
vdusek: Nit: the two arms are now nearly identical - both write the same two indexes and run the same `move_to_end`. Only the metadata bump is unique to the new-request path, so this could collapse into a single write path with `if not was_already_present:` guarding the `_update_metadata` call.

*✍️ Drafted by Claude Code*

vdusek (commented):
Thanks. It's much cleaner after the rewrite. Two things left.

tests/unit/storage_clients/_memory/test_memory_rq_client.py
anxkhn: updated the docstring to state that the o(1) guarantee is structural, while keeping the public-api duplicate check.

src/crawlee/storage_clients/_memory/_request_queue_client.py
anxkhn: collapsed both branches into one index update path, with only the metadata update guarded for new requests.

anxkhn:
@vdusek addressed both new threads, rebased, and re-requesting review. thank you.

vdusek:
Summary:

- Store pending requests in an `OrderedDict` keyed by unique key instead of a `deque`. Repositioning a request to the forefront is now O(1) instead of an O(n) `deque.remove()` scan, so re-adding K already-pending requests is O(K) instead of O(K*N) — measured 22.4s → 0.004s for 2000 forefront re-adds into a 20k-request queue. The scan only triggered for requests already waiting in the queue, which is what `add_requests(..., forefront=True)` hits when a handler re-discovers URLs that are still pending.
- Fix the same URL being handed out repeatedly. The old scan silently failed once an earlier regular re-add had replaced the registered request object with a differently-valued one, leaving a stale entry behind and appending a second one for the same unique key. Five URLs re-added over three rounds yielded 20 fetches instead of 5 and left `pending_request_count` at -15. Duplicates are now impossible by construction.
- Keep the originally enqueued request on a re-add. The old code replaced the registered request with the incoming duplicate, which on a forefront re-add also replaced the queued one — resetting `retry_count` (a request could retry forever), `label` (dispatch to a different handler) and `user_data`. On a regular re-add it left the two out of sync, so `get_request` and `fetch_next_request` disagreed. This restores the documented contract ("Duplicates will be identified but not re-added to the queue") and matches the file-system, SQL and Redis clients, as well as `@crawlee/memory-storage` in Crawlee for JS.
- Make `get_request` consistent after a reclaim: it now returns the reclaimed object, the same one `fetch_next_request` hands back.
- Write the queue metadata once per batch instead of once per new request, saving N-1 `datetime.now()` calls on a batch add.

*✍️ Drafted by Claude Code*

vdusek:
@Mantisus could you please check this as well?

vdusek (approved):
LGTM

src/crawlee/storage_clients/_memory/_request_queue_client.py
Mantisus: I believe we can drop it and check this directly in `self._pending_requests`, `self._handled_requests` and `elf._in_progress_requests`

tests/unit/storage_clients/_memory/test_memory_rq_client.py
```diff
@@ -93,3 +93,261 @@ async def test_memory_metadata_updates(rq_client: MemoryRequestQueueClient) -> N
...
+
+    def counting_eq(self: Request, other: object) -> bool:
+        nonlocal comparison_count
+        comparison_count += 1
+        return original_eq(self, other)
+
+    monkeypatch.setattr(Request, '__eq__', counting_eq)
+    await rq_client.add_batch_of_requests(requests, forefront=True)
+
+    assert comparison_count == 0
```
Mantisus: Let's try doing this using a mock. Don't forget to add the import `from unittest.mock import patch`

```suggestion
    with patch.object(Request, '__eq__', autospec=True, side_effect=Request.__eq__) as eq_counter:
        await rq_client.add_batch_of_requests(requests, forefront=True)

    eq_counter.assert_not_called()
```

Mantisus (changes requested):
That's a great PR, thanks for that. A couple of points from me.

PR #2075, opened 2026-07-14
fix(key-value store): Avoid "dictionary changed size" error in KVS autosave

anxkhn:
`KeyValueStore.get_auto_saved_value` mutates the shared per-store autosave cache
(`self._autosaved_values[self.id]`) while holding `self._autosave_lock`. But the
two consumers of that cache iterated it and awaited I/O inside the loop *without*
taking the lock:

- `persist_autosaved_values` iterated `cache.values()` and `await`ed
  `value.persist_state()` per entry.
- `_clear_cache` (called from `drop()`) iterated `cache.values()` and `await`ed
  `value.teardown()` per entry, then cleared the cache.

Because both loops suspend at an `await`, a request handler that calls
`get_auto_saved_value` with a new key while a persist or drop is mid-iteration
inserts into the same dict and grows it during iteration, raising
`RuntimeError: dictionary changed size during iteration` and aborting the
persist/drop.

This wraps both loops in `async with self._autosave_lock:`, the same lock
`get_auto_saved_value` already uses, so the iteration and any concurrent insert
are serialized. No public API changes; `_clear_cache` still calls
`cache.clear()` (now inside the lock, which is strictly safer).

The awaited bodies (`RecoverableState.persist_state()` / `.teardown()`) bottom
out in `set_value` and never re-enter `get_auto_saved_value`,
`persist_autosaved_values`, or `_clear_cache`, so the non-reentrant
`asyncio.Lock` is never re-acquired on the same instance (no deadlock).

### Issues

- No existing issue; found while reading the autosave/`use_state` code path.

### Testing

Added two regression tests in `tests/unit/storages/test_key_value_store.py`
(both run across all four storage backends via the `storage_client` fixture):

- `test_persist_autosaved_values_with_concurrent_new_key`
- `test_drop_with_concurrent_new_autosaved_key`

Each parks the persist/clear loop mid-iteration (by blocking the underlying
`set_value`, which `persist_state`/`teardown` call) and then inserts a new
autosaved key concurrently. Without the fix they fail with `RuntimeError:
dictionary changed size during iteration`; with the fix they pass.

- `uv run pytest tests/unit/storages/test_key_value_store.py` -> 296 passed, 2 skipped
- `uv run pytest tests/unit/storages/` -> 887 passed, 6 skipped
- `uv run poe lint` -> clean
- `uv run poe type-check` -> no new diagnostics vs `master`

### Checklist

- [ ] CI passed

### DIFF
--- a/tests/unit/storages/test_key_value_store.py
+++ b/tests/unit/storages/test_key_value_store.py
@@ -1,7 +1,8 @@
 from __future__ import annotations
 
+import asyncio
 import json
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


tests/unit/storages/test_key_value_store.py
```diff
@@ -1180,3 +1181,74 @@ async def test_record_with_noascii_chars(kvs: KeyValueStore) -> None:
...
+
+    drop_task = asyncio.ensure_future(kvs.drop())
+    await clear_entered.wait()
+
+    insert_task = asyncio.ensure_future(kvs.get_auto_saved_value('added-during-drop'))
+    await asyncio.sleep(0)  # Give the insert a chance to reach the shared cache.
+    release_clear.set()
+
+    # Should not raise `RuntimeError: dictionary changed size during iteration`.
+    await asyncio.gather(drop_task, insert_task)
```
vdusek: Potential test flakiness:

> After `_clear_cache` releases the autosave lock, `drop()` continues into `await self._client.drop()` concurrently with the now-unblocked insert's `get_auto_saved_value` -> `get_value` on the same store. It passes today, but the ordering isn't guaranteed, so on a backend that rejects reads of a dropped store (e.g. SQL) this could flake in CI. Minor: that concurrent insert also re-registers a `PERSIST_STATE` listener on the dropped store that is never torn down.

tests/unit/storages/test_key_value_store.py
```diff
@@ -1180,3 +1181,74 @@ async def test_record_with_noascii_chars(kvs: KeyValueStore) -> None:
     value = await kvs.get_value(key)
     assert value is not None
     assert value == init_value
+
+
+async def test_persist_autosaved_values_with_concurrent_new_key(
+    kvs: KeyValueStore,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Test that persisting autosaved values tolerates a concurrent new autosaved key.
```
vdusek: Single line docstrings for tests please.

vdusek (changes requested):

vdusek (approved):
LGTM

PR #2077, opened 2026-07-15
fix: preserve case-sensitive URL parts in normalize_url

### LINKED ISSUE #2008: `normalize_url` lowercases the entire URL, silently deduplicating case-distinct pages

JD114514-CPU:
## Summary
- stop lowercasing the full normalized URL so path and query casing stay intact
- keep the existing scheme/host normalization behavior while aligning the docstring with the implementation
- add regression coverage for normalize_url and compute_unique_key

Closes #2008.

## Testing
- uv run python -m pytest tests/unit/_utils/test_requests.py -q
- uv run ruff check src/crawlee/_utils/requests.py tests/unit/_utils/test_requests.py


vdusek:
This is planned for v2 because it introduces a breaking change.

PR #2078, opened 2026-07-16
fix(system): do not fail memory snapshots on inaccessible processes

anxkhn:
- `get_memory_info()` no longer raises when a process cannot be inspected. In restricted environments (`hidepid`, hardened containers) it failed on every sample, which meant a traceback in the log every second and an autoscaler running on a stale memory history. Processes that deny inspection, exit mid-measurement, or lose their `/proc` entry are now skipped instead.
- Failure to list the child processes no longer aborts the snapshot either - the parent's own usage is still reported.
- A kernel that exposes no `smaps` at all makes psutil alias `memory_full_info` to `memory_info`, whose result has no `pss` field, so reading it raised `AttributeError` on every sample. The metric is now read defensively, and the verdict is latched: such a machine is detected once instead of being re-probed for every process on every sample.
- A `smaps` file can be empty, which psutil parses to a PSS of zero. That was reported as zero bytes used, which the autoscaler reads as free memory. A PSS of zero now falls back to the RSS of the same process, which `memory_full_info()` has already read anyway.
- A single process denying PSS falls back to RSS just for itself - a denial says nothing about the other processes, so it is not latched.
- Every degraded path warns once, so an estimate that misses a subprocess or falls back from PSS to RSS shows up in the log instead of being silently wrong.
- `LoggerOnce` is now thread-safe, since memory metrics are sampled in a worker thread - this PR adds its first caller that runs off the event loop.
- Nothing changes when PSS is readable: on a normal Linux process tree the reported `current_size` is byte-for-byte what `master` reports, with no warnings and no RSS fallbacks.
- Tests cover the denied, exited, zombie, and missing-`/proc` cases, both PSS fallbacks and their warnings, the latch, and a vanished process not being misreported as a denial.

*✍️ Drafted by Claude Code*


src/crawlee/_utils/system.py
```diff
@@ -122,14 +122,19 @@ def get_memory_info() -> MemoryInfo:
...
 
-    # Retrieve estimated memory usage of the current process.
-    current_size_bytes = _get_used_memory(current_process)
+    # Retrieve estimated memory usage of the current process. On Linux `_get_used_memory` reads PSS via
+    # `memory_full_info`, which can raise `AccessDenied` in restricted environments (e.g. hardened containers);
+    # fall back to RSS, which a process can always read for itself.
+    try:
+        current_size_bytes = _get_used_memory(current_process)
+    except psutil.AccessDenied:
+        current_size_bytes = int(current_process.memory_info().rss)
```
vdusek: Silent fallback - a `logger.debug` (PSS denied, using RSS) would explain why `current_size` reads higher than expected in restricted environments.

src/crawlee/_utils/system.py
```diff
@@ -122,14 +122,19 @@ def get_memory_info() -> MemoryInfo:
...
 
     # Sum memory usage by all children processes, try to exclude shared memory from the sum if allowed by OS.
     for child in current_process.children(recursive=True):
-        # Ignore any NoSuchProcess exception that might occur if a child process ends before we retrieve
-        # its memory usage.
-        with suppress(psutil.NoSuchProcess):
+        # Ignore a child that ends before we retrieve its memory usage (`NoSuchProcess`) or that we are not
+        # allowed to inspect (`AccessDenied`, e.g. an unreadable subprocess in a restricted environment).
+        with suppress(psutil.NoSuchProcess, psutil.AccessDenied):
```
vdusek: The current process falls back to RSS, but an unreadable child is dropped entirely, so `current_size` silently undercounts.

anxkhn:

vdusek:
Same as here https://github.com/apify/crawlee-python/pull/2074#issuecomment-5091741396

> @anxkhn Thank you. For future reviews, I'd really appreciate it if you could address all the feedback in a single follow-up commit, rather than splitting the changes across multiple commits, merging them into earlier commits, and force-pushing it. Your current approach makes the changes really hard to review.

src/crawlee/_utils/system.py
```diff
@@ -122,15 +122,24 @@ def get_memory_info() -> MemoryInfo:
...
-        with suppress(psutil.NoSuchProcess):
-            current_size_bytes += _get_used_memory(child)
+    with suppress(psutil.NoSuchProcess, psutil.AccessDenied):
+        for child in current_process.children(recursive=True):
+            try:
+                current_size_bytes += _get_used_memory(child)
+            except psutil.AccessDenied:  # noqa: PERF203
+                logger.debug('PSS access denied for child process %s, falling back to RSS.', child.pid)
+                with suppress(psutil.NoSuchProcess, psutil.AccessDenied):
+                    current_size_bytes += int(child.memory_info().rss)
```
vdusek: Important: moving `suppress` outside the loop regresses existing behavior. The inner `try` only catches `AccessDenied`, so a child that exits mid-iteration raises `NoSuchProcess` from `_get_used_memory`, which escapes to the outer `suppress` and aborts the whole loop - silently dropping every remaining child. On `master` the per-child `suppress` kept iterating, so this breaks exactly the case the original comment documented.

Guard only the `children()` call and keep per-child suppression inside the loop:

```suggestion
    children: list[psutil.Process] = []
    with suppress(psutil.NoSuchProcess, psutil.AccessDenied):
        children = current_process.children(recursive=True)

    for child in children:
        # Ignore a child that ends before we retrieve its memory usage (`NoSuchProcess`) or that we are not
        # allowed to inspect at all (`AccessDenied`).
        with suppress(psutil.NoSuchProcess, psutil.AccessDenied):
            try:
                current_size_bytes += _get_used_memory(child)
            except psutil.AccessDenied:
                logger.debug('PSS access denied for child process %s, falling back to RSS.', child.pid)
                current_size_bytes += int(child.memory_info().rss)
```

*✍️ Drafted by Claude Code*

src/crawlee/_utils/system.py
```diff
@@ -122,15 +122,24 @@ def get_memory_info() -> MemoryInfo:
...
-    # Retrieve estimated memory usage of the current process.
-    current_size_bytes = _get_used_memory(current_process)
+    # Retrieve estimated memory usage of the current process. On Linux `_get_used_memory` reads PSS via
+    # `memory_full_info`, which can raise `AccessDenied` in restricted environments (e.g. hardened containers);
+    # fall back to RSS, which a process can always read for itself.
+    try:
+        current_size_bytes = _get_used_memory(current_process)
+    except psutil.AccessDenied:
+        logger.debug('PSS access denied for the current process, falling back to RSS.')
+        current_size_bytes = int(current_process.memory_info().rss)
```
vdusek: Suggestion: this PSS->RSS fallback is now duplicated for children below. Extracting it into a module-level helper dedupes it and lets the `noqa: PERF203` go away:

```python
def _get_used_memory_with_rss_fallback(process: psutil.Process) -> int:
    try:
        return _get_used_memory(process)
    except psutil.AccessDenied:
        logger.debug('PSS access denied for process %s, falling back to RSS.', process.pid)
        return int(process.memory_info().rss)
```

*✍️ Drafted by Claude Code*

tests/unit/_utils/test_system.py
```diff
@@ -21,6 +24,54 @@ def test_get_memory_info_returns_valid_values() -> None:
     assert memory_info.current_size < memory_info.total_size
 
 
+def test_get_memory_info_falls_back_to_rss_for_children_with_access_denied(monkeypatch: pytest.MonkeyPatch) -> None:
```
vdusek: Add a single-line docstring, please.

*✍️ Drafted by Claude Code*

tests/unit/_utils/test_system.py
```diff
@@ -21,6 +24,54 @@ def test_get_memory_info_returns_valid_values() -> None:
...
+    child = psutil.Process()  # any process object works; only `_get_used_memory` behavior matters here
```
vdusek: Nit: "any process object works" is misleading - this is the current process, and the test only works because the `process is child` identity check below distinguishes it from the separate `psutil.Process(os.getpid())` object built inside `get_memory_info`.

*✍️ Drafted by Claude Code*

tests/unit/_utils/test_system.py
```diff
@@ -21,6 +24,54 @@ def test_get_memory_info_returns_valid_values() -> None:
...
+    def fake_get_used_memory(process: psutil.Process) -> int:
+        if process is child:
+            raise psutil.AccessDenied(pid=child.pid)
+        return 100
+
+    monkeypatch.setattr(system, '_get_used_memory', fake_get_used_memory)
+
+    memory_info = get_memory_info()
+
+    assert memory_info.current_size == ByteSize(150)
```
vdusek: Note: nothing covers a child raising `NoSuchProcess` mid-loop, which is why CI stays green despite the loop regression flagged in `system.py`. Worth adding next to this test: two children, the first raising `NoSuchProcess`, asserting the second is still counted.

*✍️ Drafted by Claude Code*

tests/unit/_utils/test_system.py
```diff
@@ -21,6 +24,54 @@ def test_get_memory_info_returns_valid_values() -> None:
...
+
+
+@pytest.mark.parametrize('error', [psutil.AccessDenied(), psutil.NoSuchProcess(pid=1)])
```
vdusek: Suggestion: bare exception instances produce opaque generated ids; `pytest.param` with explicit ids is used elsewhere in the suite (e.g. `test_file.py`).

```suggestion
@pytest.mark.parametrize(
    'error',
    [
        pytest.param(psutil.AccessDenied(), id='access denied'),
        pytest.param(psutil.NoSuchProcess(pid=1), id='no such process'),
    ],
)
```

*✍️ Drafted by Claude Code*

vdusek (commented):
Thanks for the revision. Both of my earlier comments are correctly addressed - the `logger.debug` for the PSS->RSS fallback, and the child RSS fallback so an unreadable child is no longer dropped.

However, the change made for the Copilot comment about guarding `children()` introduces a regression: wrapping the entire loop in `suppress` means one child dying mid-iteration now aborts the loop and drops all remaining children. I verified this locally - a repro with a dead first child and a live second child worth 40 B yields 140 B on `master` and 100 B on this branch. CI does not catch it because that path is untested.

Details inline.

*✍️ Drafted by Claude Code*

src/crawlee/_utils/system.py
anxkhn: kept the children() guard separate and restored per-child suppression. a new two-child regression test verifies that a dead first child does not hide the live second child.

src/crawlee/_utils/system.py
anxkhn: extracted the shared pss-to-rss fallback helper and removed the per-loop exception handler.

tests/unit/_utils/test_system.py
anxkhn: added the single-line docstring.

tests/unit/_utils/test_system.py
anxkhn: removed the misleading comment.

tests/unit/_utils/test_system.py
anxkhn: added a regression test with a dead first child and live second child, asserting the second is counted.

tests/unit/_utils/test_system.py
anxkhn: added explicit parameter ids.

anxkhn:
@vdusek addressed all new threads, including the child-exit regression test, rebased, and re-requesting review. thank you.

vdusek:
Summary:

- `get_memory_info()` no longer raises when a process cannot be inspected. In restricted environments (`hidepid`, hardened containers) it failed on every sample, which meant a traceback in the log every second and an autoscaler running on a stale memory history. Processes that deny inspection, exit mid-measurement, or lose their `/proc` entry are now skipped instead.
- Failure to list the child processes no longer aborts the snapshot either - the parent's own usage is still reported.
- A kernel that exposes no `smaps` at all makes psutil alias `memory_full_info` to `memory_info`, whose result has no `pss` field, so reading it raised `AttributeError` on every sample. The metric is now read defensively, and the verdict is latched: such a machine is detected once instead of being re-probed for every process on every sample.
- A `smaps` file can be empty, which psutil parses to a PSS of zero. That was reported as zero bytes used, which the autoscaler reads as free memory. A PSS of zero now falls back to the RSS of the same process, which `memory_full_info()` has already read anyway.
- A single process denying PSS falls back to RSS just for itself - a denial says nothing about the other processes, so it is not latched.
- Every degraded path warns once, so an estimate that misses a subprocess or falls back from PSS to RSS shows up in the log instead of being silently wrong.
- `LoggerOnce` is now thread-safe, since memory metrics are sampled in a worker thread - this PR adds its first caller that runs off the event loop.
- Nothing changes when PSS is readable: on a normal Linux process tree the reported `current_size` is byte-for-byte what `master` reports, with no warnings and no RSS fallbacks.
- Tests cover the denied, exited, zombie, and missing-`/proc` cases, both PSS fallbacks and their warnings, the latch, and a vanished process not being misreported as a denial.

*✍️ Drafted by Claude Code*

vdusek:
@Mantisus could you please check this as well?

vdusek (approved):
LGTM

src/crawlee/_utils/system.py
```diff
@@ -2,37 +2,124 @@
...
+        Raises:
+            psutil.Error: If the process refuses inspection or is gone.
+            OSError: If a `/proc` entry of the process is missing.
+        """
+        if _PssAvailability.is_available:
+            try:
+                # A system that does not expose `smaps` at all makes psutil alias `memory_full_info` to
+                # `memory_info`, whose result has no `pss` field.
+                memory = process.memory_full_info()
+            except (psutil.NoSuchProcess, psutil.ZombieProcess):
```
Mantisus: Nit. The `psutil.ZombieProcess` is a child of the `psutil.NoSuchProcess`

tests/unit/_utils/test_system.py
```diff
@@ -1,31 +1,224 @@
...
+        call_count += 1
+        return SimpleNamespace(rss=1024)
+
+    monkeypatch.setattr(psutil.Process, 'memory_full_info', memory_full_info)
+    monkeypatch.setattr(psutil.Process, 'memory_info', lambda _process: SimpleNamespace(rss=1024))
+
+    for _ in range(3):
+        assert system._get_used_memory(psutil.Process()) == 1024
+
+    assert call_count == 1
```
Mantisus: Please use `Mock` with `from unittest.mock import Mock`. 

```suggestion
    memory_full_info = Mock(return_value=SimpleNamespace(rss=1024))
    monkeypatch.setattr(psutil.Process, 'memory_full_info', memory_full_info)
    monkeypatch.setattr(psutil.Process, 'memory_info', lambda _process: SimpleNamespace(rss=1024))

    for _ in range(3):
        assert system._get_used_memory(psutil.Process()) == 1024

    memory_full_info.assert_called_once()
```

Mantisus (approved):
LGTM. Just a few nits.

Also, could you update the PR description?

PR #2079, opened 2026-07-16
feat(dataset): add `collect_all_keys` option and keep CSV export columns aligned

anxkhn:
- Fixes CSV export writing values under the wrong column. Rows were written positionally with
  `csv.writer`, so any item whose key set or key order differed from the first item's had its values
  shifted or placed under the wrong header. Rows are now written by key with `csv.DictWriter`.
- Columns are taken from the first non-empty item. Keys that appear only in later items are dropped
  instead of being appended out of position, and an item missing a column now gets an empty cell
  instead of a short row. Exports of datasets with mixed keys therefore change content.
- Dropped keys are reported once per export in a warning that names them. Crawlee for JS drops them
  silently, so this is a Python-only addition.
- Adds a `collect_all_keys` parameter to `Dataset.export_to` and `BasicCrawler.export_data`. When
  True, the columns are the union of all item keys in first-seen order, so nothing is dropped. It
  reads the whole dataset before writing the header, unlike the default. Named and behaving like
  `collectAllKeys` in Crawlee for JS.
- Adds `restval`, the value written for columns an item has no key for. It was rejected before,
  because `csv.writer` has no such argument.
- Documents `ExportDataCsvKwargs` as mirroring `csv.DictWriter` and `ExportDataJsonKwargs` as
  mirroring `json.dump`, so Crawlee-specific options stay out of them.
- Documents the column behavior in the dataset export example.
- Alignment with Crawlee for JS: the by-key row mapping, deriving columns from the first item, and
`collectAllKeys` all match `Dataset.exportTo` in `packages/core/src/storages/dataset.ts`.

*✍️ Drafted by Claude Code*


src/crawlee/_utils/file.py
```diff
@@ -203,16 +203,18 @@ async def export_csv_to_stream(
...
+    # the behavior of Crawlee for JavaScript.
     async for item in iterator:
         if not item:
             continue
 
-        if write_header:
-            writer.writerow(item.keys())
-            write_header = False
+        if writer is None:
+            writer = csv.DictWriter(dst, fieldnames=list(item.keys()), extrasaction='ignore', **kwargs)
```
Mantisus: We can't just ignore extra columns.

tests/unit/_utils/test_file.py
```diff
@@ -19,6 +23,48 @@ async def test_json_dumps() -> None:
...
+                {'age': 40, 'name': 'Carol'},
+            ]
+        ),
+        dst,
+        lineterminator='\n',
+    )
+
+    # The header comes from the first item; later items are mapped by key (extra keys such as `city` are dropped),
+    # so `Carol`/`40` stay aligned with the `name`/`age` columns instead of being written positionally.
+    assert dst.getvalue() == 'name,age\nAlice,30\nBob,25\nCarol,40\n'
```
Mantisus, line 48: The correct output should include a `city` column

Mantisus (changes requested):
Thank you for finding this bug. However, we cannot use `extrasaction=‘ignore’`

src/crawlee/_utils/file.py
```diff
@@ -203,16 +203,12 @@ async def export_csv_to_stream(
...
-            write_header = False
-
-        writer.writerow(item.values())
+    items = [item async for item in iterator if item]
+    if not items:
+        return
+
+    fieldnames = list(dict.fromkeys(key for item in items for key in item))
+    writer = csv.DictWriter(dst, fieldnames=fieldnames, **kwargs)
+    writer.writeheader()
```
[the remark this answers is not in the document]
Mantisus: This conflicts with my review. I believe we shouldn't just skip the additional fields that aren't present in the first element.

@vdusek, @Pijukatel. What do you think?

src/crawlee/_utils/file.py
```diff
@@ -203,16 +203,12 @@ async def export_csv_to_stream(
...
-    async for item in iterator:
-        if not item:
-            continue
-
-        if write_header:
-            writer.writerow(item.keys())
-            write_header = False
-
-        writer.writerow(item.values())
+    items = [item async for item in iterator if item]
```
vdusek: This buffers the whole dataset in memory instead of streaming (!). I don't think we want to do that.

src/crawlee/_utils/file.py
```diff
@@ -203,16 +203,12 @@ async def export_csv_to_stream(
...
+    if not items:
```
vdusek: The `if not items: return` branch isn't covered by a test.

vdusek (changes requested):

src/crawlee/_utils/file.py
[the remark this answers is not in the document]
vdusek: Yeah, probably not. This should also be checked against the JS implementation.

anxkhn:

vdusek:
Same as here https://github.com/apify/crawlee-python/pull/2074#issuecomment-5091741396

> @anxkhn Thank you. For future reviews, I'd really appreciate it if you could address all the feedback in a single follow-up commit, rather than splitting the changes across multiple commits, merging them into earlier commits, and force-pushing it. Your current approach makes the changes really hard to review.

src/crawlee/_utils/file.py
```diff
@@ -203,16 +203,21 @@ async def export_csv_to_stream(
...
-    async for item in iterator:
-        if not item:
-            continue
-
-        if write_header:
-            writer.writerow(item.keys())
-            write_header = False
-
-        writer.writerow(item.values())
+    fieldnames = dict[str, None]()
```
vdusek: Important: I asked to check this against the JS implementation, and JS does the opposite by default. It takes the first item's keys and exposes the union only behind an opt-in flag (`packages/core/src/storages/dataset.ts:366`, same code in `basic-crawler.ts:1329`):

```js
const keys = options?.collectAllKeys
    ? Array.from(new Set(items.flatMap(Object.keys)))   // opt-in: union of all keys
    : Object.keys(items[0]);                            // default: first item's keys
```

This PR hardcodes the opt-in branch, so the JS default is not reachable at all. For the input from the new test:

```text
{'name': 'Alice', 'age': 30}
{'name': 'Bob', 'city': 'NYC', 'age': 25}
{'age': 40, 'name': 'Carol'}
```

```text
master (the bug)        this PR              JS default           JS collectAllKeys: true
name,age                name,age,city        name,age             name,age,city
Alice,30                Alice,30,            Alice,30             Alice,30,
Bob,NYC,25  <- shifted   Bob,25,NYC           Bob,25               Bob,25,NYC
40,Carol    <- swapped   Carol,40,            Carol,40             Carol,40,
```

(the two JS columns are actual output from `csv-stringify` running that snippet, not from reading the code)

The union is also what forces the buffering discussed below: the header line can't be written until the very last item has been seen, because any item may still introduce a new column - so everything has to be collected first. With the first item's header you know it after one item, write it, and stream the rest. No temp file, no JSON round trip, and the alignment bug this PR is about is still fixed:

```python
writer = None
async for item in iterator:
    if not item:
        continue
    if writer is None:
        writer = csv.DictWriter(dst, fieldnames=list(item), extrasaction='ignore', **kwargs)
        writer.writeheader()
    writer.writerow(item)
```

Let's mirror JS here: make that the default, and add a `collect_all_keys: bool = False` parameter for the union, so only that branch pays for the buffering. That keeps the two implementations consistent and makes dropping columns an explicit user choice rather than a silent default.

*✍️ Drafted by Claude Code*

src/crawlee/_utils/file.py
```diff
@@ -203,16 +203,21 @@ async def export_csv_to_stream(
...
+    with tempfile.TemporaryFile(mode='w+', encoding='utf-8') as items:
```
vdusek: Important: this doesn't bound memory end to end. Both callers pass a `StringIO` and then take `.getvalue()` (`_dataset.py:357`, `_basic_crawler.py:942`), so the complete CSV sits in RAM no matter what the exporter does internally - the temp file only adds a second full copy of the data alongside it.

And the copy is not necessarily on disk: with systemd defaults `/tmp` is a tmpfs (Fedora, Ubuntu 24.04+, Arch), so `TemporaryFile()` is RAM-backed there and a large export hits `ENOSPC` at ~50% of RAM instead of `MemoryError`. It also writes outside Crawlee's configured storage directory.

Measured through the current call path, 300k items with 4 keys each (17.6 MB of CSV):

| | peak RSS | peak tmpfs | total |
| --- | --- | --- | --- |
| master | 57.6 MB | 0 MB | **57.6 MB** |
| this PR | 58.5 MB | 29.3 MB | **87.8 MB** |

So on a tmpfs host this is roughly **52% more** peak memory than master, not less. The JSONL spill is even larger than the CSV itself (29.3 vs 17.6 MB) because every row repeats the keys.

*✍️ Drafted by Claude Code*

src/crawlee/_utils/file.py
```diff
@@ -203,16 +203,21 @@ async def export_csv_to_stream(
...
+        async for item in iterator:
+            if not item:
+                continue
+
+            fieldnames.update(dict.fromkeys(item))
+            json.dump(item, items)
```
vdusek: Important: the JSON round trip regresses values that the old `csv.writer` rendered with `str()`. `MemoryStorageClient` stores items verbatim, so `Dataset.export_to(content_type='csv')` on a dataset holding a `datetime` (or `bytes`, `set`) now raises `TypeError: Object of type datetime is not JSON serializable`, where master exported `2020-01-01 00:00:00+00:00`. Other values change silently: `(1, 2)` -> `[1, 2]`, `{1: 'a'}` -> `{'1': 'a'}`.

```suggestion
            json.dump(item, items, default=str)
```

*✍️ Drafted by Claude Code*

src/crawlee/_utils/file.py
```diff
@@ -203,16 +203,21 @@ async def export_csv_to_stream(
...
+            items.write('\n')
+
+        if not fieldnames:
+            return
+
+        writer = csv.DictWriter(dst, fieldnames=fieldnames, **kwargs)
```
vdusek: Nit: `fieldnames` is a `dict` here - `list(fieldnames)` reads better as a `DictWriter` fieldnames argument.

*✍️ Drafted by Claude Code*

src/crawlee/_utils/file.py
```diff
@@ -203,16 +203,21 @@ async def export_csv_to_stream(
...
+        writer.writeheader()
+        items.seek(0)
+        for item in items:
```
vdusek, line 224: Note: the spill and read-back are blocking sync I/O inside an `async def`. This module already offloads file work with `asyncio.to_thread` (see `json_dumps` and `atomic_write`).

*✍️ Drafted by Claude Code*

tests/unit/_utils/test_file.py
```diff
@@ -19,6 +23,53 @@ async def test_json_dumps() -> None:
     assert await json_dumps(datetime(2022, 1, 1, tzinfo=timezone.utc)) == '"2022-01-01 00:00:00+00:00"'
 
 
+# Tests for export_csv_to_stream (dataset CSV export).
+
+
+async def _async_iter(
```
vdusek: Nit: test-local helpers don't need the underscore prefix - nothing imports from a test module. `async_iter` is enough.

*✍️ Drafted by Claude Code*

tests/unit/_utils/test_file.py
```diff
@@ -19,6 +23,53 @@ async def test_json_dumps() -> None:
...
+    await export_csv_to_stream(
+        _async_iter([{}, {'id': 1, 'name': 'Item 1'}, {}, {'id': 2, 'name': 'Item 2'}]),
+        dst,
+        lineterminator='\n',
+    )
+
+    assert dst.getvalue() == 'id,name\n1,Item 1\n2,Item 2\n'
+
+
+async def test_export_csv_to_stream_handles_empty_iterator() -> None:
```
vdusek, line 84: Add a single-line docstring, please.

vdusek (commented):
A few important things and then a few nits.

src/crawlee/_utils/file.py
anxkhn: changed the default to the first non-empty item keys, matching js, and added collect_all_keys as an opt-in union mode.

src/crawlee/_utils/file.py
anxkhn: removed the temporary file and jsonl spill. the default now streams directly to the provided destination; only collect_all_keys buffers because its header depends on the full input.

src/crawlee/_utils/file.py
anxkhn: removed the json round trip and added a datetime regression test, preserving csv writer string conversion.

src/crawlee/_utils/file.py
anxkhn: the temporary fieldnames mapping is gone; both modes now pass lists.

src/crawlee/_utils/file.py
anxkhn: removed all temporary-file i/o. default export is streaming, and collect_all_keys buffers objects in memory without blocking file operations.

tests/unit/_utils/test_file.py
anxkhn: renamed the helper to async_iter.

tests/unit/_utils/test_file.py
anxkhn: added the single-line docstring.

anxkhn:
@vdusek addressed all new threads, including js-default parity, streaming behavior, bounded default memory, and value preservation tests. rebased and re-requesting review. thank you.

vdusek:
Summary:

- Fixes CSV export writing values under the wrong column. Rows were written positionally with
  `csv.writer`, so any item whose key set or key order differed from the first item's had its values
  shifted or placed under the wrong header. Rows are now written by key with `csv.DictWriter`.
- Columns are taken from the first non-empty item. Keys that appear only in later items are dropped
  instead of being appended out of position, and an item missing a column now gets an empty cell
  instead of a short row. Exports of datasets with mixed keys therefore change content.
- Dropped keys are reported once per export in a warning that names them. Crawlee for JS drops them
  silently, so this is a Python-only addition.
- Adds a `collect_all_keys` parameter to `Dataset.export_to` and `BasicCrawler.export_data`. When
  True, the columns are the union of all item keys in first-seen order, so nothing is dropped. It
  reads the whole dataset before writing the header, unlike the default. Named and behaving like
  `collectAllKeys` in Crawlee for JS.
- Adds `restval`, the value written for columns an item has no key for. It was rejected before,
  because `csv.writer` has no such argument.
- Documents `ExportDataCsvKwargs` as mirroring `csv.DictWriter` and `ExportDataJsonKwargs` as
  mirroring `json.dump`, so Crawlee-specific options stay out of them.
- Documents the column behavior in the dataset export example.
- Alignment with Crawlee for JS: the by-key row mapping, deriving columns from the first item, and
`collectAllKeys` all match `Dataset.exportTo` in `packages/core/src/storages/dataset.ts`.

*✍️ Drafted by Claude Code*

Mantisus (approved):
LGTM

vdusek (approved):
LGTM

PR #2080, opened 2026-07-19
fix(cli): Migrate CLI to `typer` >=0.26

### LINKED ISSUE #2063: Migrate CLI to typer >=0.26 and drop the `typer<0.26` pin

Mantisus:
### Description

- Cli has been updated for better compatibility with `typer` and no longer uses direct calls to `click`

### Issues

- Closes: #2063

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,9 +65,7 @@ cli = [
     "cookiecutter>=2.6.0",
     "inquirer>=3.3.0",
     "rich>=13.9.0",
-    # Pinned below 0.26, which breaks our `click.Choice` CLI options.
-    # See https://github.com/apify/crawlee-python/issues/2063.
-    "typer>=0.12.0,<0.26",
+    "typer>=0.12.0",
 ]
 curl-impersonate = ["curl-cffi>=0.9.0"]
 httpx = ["httpx[brotli,http2,zstd]>=0.27.0", "apify_fingerprint_datapoints>=0.0.2", "browserforge>=1.2.3"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,7 +65,7 @@ cli = [
     "cookiecutter>=2.6.0",
     "inquirer>=3.3.0",
     "rich>=13.9.0",
-    "typer>=0.12.0",
+    "typer>=0.26.0",
 ]
 curl-impersonate = ["curl-cffi>=0.9.0"]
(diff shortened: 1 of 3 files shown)


Mantisus:
After the update, `crawlee create --help` displays the correct information

image

An error caused by an incorrect input is displayed correctly

image

vdusek (approved):
LGTM

Pijukatel:
Looks good, lets just verify that E2E tests are ok: https://github.com/apify/crawlee-python/actions/runs/29748094500

Pijukatel (approved):

PR #2088, opened 2026-07-21
fix(events): prevent deadlock when closing or waiting for listeners from within a listener

vdusek:
Calling `EventManager.wait_for_all_listeners_to_complete()` - or closing the manager via `__aexit__` - from *within* an event listener deadlocked. The listener runs in a task that is itself registered in `_listener_tasks`, so the wait ends up awaiting the very task that is awaiting it. Under a close timeout this cycle degrades further into a `RecursionError`.

Changes in `EventManager`:

- Listener tasks currently blocked in `wait_for_all_listeners_to_complete()` are tracked in `_waiting_listener_tasks` and excluded from the wait, so listener waiters never await themselves or each other. A caller that is not a listener is outside the cycle and still awaits every listener, waiting ones included.
- The wait no longer wraps the gather in an inner task. The one-tick defer that task provided is now explicit: `emit` only schedules the listener wrappers, and each registers its listener task once it starts running, so the wait yields before snapshotting `_listener_tasks`. This also drops the `Event listener raised an exception.` log line, which duplicated the ERROR the listener wrapper already logs with the traceback.
- The listener wrapper's `finally` uses `set.discard()` instead of `set.remove()`, since `__aexit__` may have already cleared the task set while the listener was mid-flight (avoids a spurious `KeyError`).

Regression tests cover waiting from within a listener, several listeners waiting at once, closing the manager from within a listener, and waiting from outside while a listener is itself waiting.

This unblocks apify/apify-sdk-python#1061, where `Actor.exit()` is called from inside an event listener (e.g. an `ABORTING` handler) - with this fix the SDK can drop its `_detach_current_listener_task` workaround.

*✍️ Drafted by Claude Code*


src/crawlee/events/_event_manager.py
```diff
@@ -256,17 +260,27 @@ async def wait_for_all_listeners_to_complete(self, *, timeout: timedelta | None
...
+            results = await asyncio.gather(*listener_tasks, return_exceptions=True)
             for result in results:
                 if isinstance(result, Exception):
                     logger.exception('Event listener raised an exception.', exc_info=result)
 
         tasks = [asyncio.create_task(wait_for_listeners(), name=f'Task-{wait_for_listeners.__name__}')]
 
-        await wait_for_all_tasks_for_finish(tasks=tasks, logger=logger, timeout=timeout)
+        try:
+            await wait_for_all_tasks_for_finish(tasks=tasks, logger=logger, timeout=timeout)
```
Pijukatel: Why do we have `wait_for_listeners` and also `wait_for_all_tasks_for_finish`? There is probably some reason, but it is not obvious to me. So the natural question that comes to my mind: 

Why don't we drop `wait_for_listeners` and call the other one directly?
```
listener_tasks = [task for task in self._listener_tasks if task not in self._waiting_listener_tasks]
await wait_for_all_tasks_for_finish(tasks=tasks, logger=logger, timeout=timeout)
```

vdusek: you're right, removing it

tests/unit/events/test_event_manager.py
```diff
@@ -206,6 +209,144 @@ async def test_methods_raise_error_when_not_active(event_system_info_data: Event
...
+    event_manager.on(event=Event.SYSTEM_INFO, listener=other_listener)
+    event_manager.on(event=Event.SYSTEM_INFO, listener=closing_listener)
+
+    tasks_before = asyncio.all_tasks()
+    event_manager.emit(event=Event.SYSTEM_INFO, event_data=event_system_info_data)
+
+    try:
+        await asyncio.wait_for(closed.wait(), timeout=5)
+        # Drain the wrapper tasks so their `finally` blocks run before we assert - no arbitrary sleep.
+        spawned = asyncio.all_tasks() - tasks_before - {asyncio.current_task()}
```
Mantisus: While the `await` runs, tasks unrelated to this test can enter the event loop, and they won't be in the `tasks_before` snapshot.

tests/unit/events/test_event_manager.py
```diff
@@ -206,6 +209,144 @@ async def test_methods_raise_error_when_not_active(event_system_info_data: Event
...
+    """Closing the event manager from within a listener (as `Actor.exit()` does) must not deadlock or raise."""
+    event_manager = EventManager()
+    await event_manager.__aenter__()
+
+    # A wrapper finalizing after close raises onto the loop (its `error` listener is gone by then), so watch
+    # both channels for a stray exception.
+    emitter_errors: list[BaseException] = []
+    event_manager._event_emitter.add_listener('error', emitter_errors.append)
+    loop_errors: list[dict[str, Any]] = []
+    asyncio.get_running_loop().set_exception_handler(lambda _loop, context: loop_errors.append(context))
```
Mantisus: Let's make sure this doesn't leak into other tests.

Mantisus (approved):
LGTM, just a few nits

B4nan (approved):

PR #2092, opened 2026-07-27
docs: align maxRequestsPerCrawl intro example with the JS SDK

### LINKED ISSUE #2091: docs: `max_requests_per_crawl` intro example differs from the JS SDK (10 vs 20)
While going through the introduction, I noticed the **"Limit your crawls"** section uses a different `max_requests_per_crawl` value than the equivalent section in the JavaScript SDK docs.

- **Python** — [Introduction → Adding more URLs](https://crawlee.dev/python/docs/introduction/adding-more-urls): the example uses `max_requests_per_crawl=10`, and the text reads *"no new requests will be started after the 10th request is finished."*
- **JavaScript** — [Introduction → Adding more URLs](https://crawlee.dev/js/docs/introduction/adding-urls): the same section uses `maxRequestsPerCrawl: 20` and *"after the 20th request is finished."*

The two introductions are otherwise near-identical ports of each other, so the differing number looks incidental rather than deliberate. It's harmless, but a little confusing when cross-referencing the two SDKs while learning.

For what it's worth, on `main` the Python section is already internally consistent (both the code and the prose say 10) — so this isn't an internal mismatch, just a cross-SDK one.

I'm happy to open a small PR aligning the Python example to `20` to match JS. If you'd rather standardize on `10` and have the JS side change instead, just say so and I'll close this.

govarthenan:
### Description

The **"Limit your crawls"** section of the Python introduction uses `max_requests_per_crawl=10`, whereas the [equivalent section in the JavaScript SDK docs](https://crawlee.dev/js/docs/introduction/adding-urls) uses `20`. The two introductions are otherwise near-identical ports of each other, so this aligns the Python example to `20` for cross-SDK consistency.

Changes:

- `docs/introduction/03_adding_more_urls.mdx` — the inline snippet and the explanatory prose in the "Limit your crawls" section (`10` → `20`, and "10th" → "20th").
- `docs/introduction/code_examples/03_finding_new_links.py` — the runnable example rendered immediately below on the same page, so the page stays consistent with the section above it.

I deliberately scoped this to the "Limit your crawls" section and its adjacent runnable example. A few other intro examples (e.g. `06_scraping.py`, `07_final_code.py`) also use `max_requests_per_crawl=10` as an incidental testing limit — I left those untouched to keep the diff minimal, but I'm happy to broaden it to the whole introduction if page-wide consistency is preferred.

### Issues

- Closes: #2091

### Testing

Documentation-only change — the edits swap the integer value in an existing, already-runnable example and its accompanying prose; no code paths or behavior are affected.

### Checklist

- [ ] CI passed

### DIFF
--- a/docs/introduction/03_adding_more_urls.mdx
+++ b/docs/introduction/03_adding_more_urls.mdx
@@ -43,10 +43,10 @@ The <ApiLink to="class/EnqueueLinksFunction">`enqueue_links`</ApiLink> function 
 When you're just testing your code or when your crawler could potentially find millions of links, it's very useful to set a maximum limit of crawled pages. The option is called <ApiLink to="class/BasicCrawlerOptions#max_requests_per_crawl">`max_requests_per_crawl`</ApiLink>, is available in all crawlers, and you can set it like this:
 
 ```python
-crawler = BeautifulSoupCrawler(max_requests_per_crawl=10)
+crawler = BeautifulSoupCrawler(max_requests_per_crawl=20)
 ```
 
-This means that no new requests will be started after the 10th request is finished. The actual number of processed requests might be a little higher thanks to parallelization, because the running requests won't be forcefully aborted. It's not even possible in most cases.
+This means that no new requests will be started after the 20th request is finished. The actual number of processed requests might be a little higher thanks to parallelization, because the running requests won't be forcefully aborted. It's not even possible in most cases.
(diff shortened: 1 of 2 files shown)


vdusek:
`max_requests_per_crawl=10` example is used throughout the codebase (210 occurrences across 208 files).

PR #2097, opened 2026-07-28
docs: Add `cookies management` guide

### LINKED ISSUE #2060: Docs: explain how sessions interact with retries (keeping cookies/session across a retry)
## Summary

The [Session management guide](https://crawlee.dev/python/docs/guides/session-management) doesn't explain how sessions interact with **retries**, nor how to keep a session (and its cookies) across a retry. This is a recurring source of confusion.

## Motivation

See discussion [#2028](https://github.com/apify/crawlee-python/discussions/2028): a user sets cookies via `send_request` (or establishes login state), raises to trigger a retry, and finds the cookies "gone" on the retry.

The cookies aren't lost, they're stored on the session. The catch is that an **unpinned request draws a new random session on every attempt**. When the handler raises and the request is retried, the crawler picks a fresh session from the pool (default `max_pool_size=1000`), so the retry runs on a different, cookie-less session. Nothing in the docs sets this expectation, and the natural mental model ("same request → same session") is wrong by default.

## Proposal

Add a short section to the Session management guide, e.g. **"Reusing a session across retries"**, covering the default behavior and the ways to keep a session across retries.

### Option 1 — single-session pool

The simplest fix, keeps a raise-to-retry flow unchanged. One session for everything, so cookies set on the first attempt are still there on the retry. Trade-off: no session rotation.

```python
from crawlee.crawlers import HttpCrawler
from crawlee.sessions import SessionPool


async def main() -> None:
    crawler = HttpCrawler(session_pool=SessionPool(max_pool_size=1))
    # ...
```

### Option 2 — pin the request to its session

Keeps rotation for everything else. Instead of raising, re-enqueue the same URL bound to the session that now has the cookies. The `user_data` flag guards against re-enqueuing in a loop, and `always_enqueue=True` bypasses deduplication so the same URL runs again.

```python
from crawlee import Request
from crawlee.crawlers import HttpCrawler, HttpCrawlingContext

crawler = HttpCrawler()


async def handler(context: HttpCrawlingContext) -> None:
    if context.request.user_data.get('prepared'):
        # Cookies from the setup step are on this pinned session.
        # ... do the real work ...
        return

    # First pass: establish cookies on the current session.
    await context.send_request('https://example.com/set-something')

    # Re-enqueue the same URL, pinned to the same session.
    await context.add_requests(
        [
            Request.from_url(
                context.request.url,
                session_id=context.session.id,  # Pins the retry to this session.
                user_data={'prepared': True},
                always_enqueue=True,  # Bypasses deduplication so the same URL runs again.
            )
        ]
    )
```

### Option 3 — re-apply cookies in a pre-navigation hook

Keeps the raise-to-retry flow and rotation, at the cost of some manual plumbing. Snapshot the cookies after the setup step into an external store keyed by the request's `unique_key`, then re-apply them onto whichever session the retry picks in a `pre_navigation_hook` (which runs before navigation and receives `context.session`).

```python
from crawlee.crawlers import HttpCrawler, HttpCrawlingContext

crawler = HttpCrawler()
pending_cookies: dict[str, list] = {}


async def handler(context: HttpCrawlingContext) -> None:
    if not has_what_i_need(context):  # your existing check, e.g. cookies present
        # First pass: establish cookies, then snapshot them for the retry.
        await context.send_request('https://example.com/set-something')
        pending_cookies[context.request.unique_key] = context.session.cookies.get_cookies_as_dicts()
        raise RuntimeError('retry with cookies')

    pending_cookies.pop(context.request.unique_key, None)
    # ... do the real work; cookies are on context.session ...


@crawler.pre_navigation_hook
async def restore_cookies(context: HttpCrawlingContext) -> None:
    if saved := pending_cookies.get(context.request.unique_key):
        context.session.cookies.set_cookies(saved)
```

> **Note:** use an external store (as above), not `context.request.user_data`, to carry the snapshot across the retry. During a handler run `context.request` is an isolated copy, so in-handler mutations to it (including `user_data`) don't survive the retry (see [#2061](https://github.com/apify/crawlee-python/issues/2061)).

The guide should also note that cookies are scoped by domain (the `send_request` target must set them for the same host you're crawling), and that a custom HTTP client needs to attach the session's cookies to outgoing `send_request` calls.

## Notes for a future redesign (v2)

Reusing a session across retries isn't ergonomic today, and we expect to rethink this in Crawlee for Python v2 / Crawlee for JS v4. Worth capturing here as prior art / parity references:

- Crawlee for JS [Session management](https://crawlee.dev/js/docs/guides/session-management) and its `SessionPool` / `sessionPoolOptions.maxPoolSize`.
- Related parity gap: in Crawlee for JS, `request.userData` mutations persist across retries (same `Request` instance), whereas in Crawlee for Python `context.request` is an isolated `deepcopy` during handler execution, so in-handler mutations to `context.request` (including `user_data`) are silently dropped on retry. That difference makes some of the natural workarounds (stash state on `request.user_data`, then read it on the retry) fail. This is worth deciding on explicitly as part of the redesign.

Mantisus:
### Description

- Add a guide describing how to manage cookies

### Issues

- Closes: #2060

### DIFF
--- a/docs/guides/code_examples/cookie_management/persist_cookies.py
+++ b/docs/guides/code_examples/cookie_management/persist_cookies.py
@@ -1,0 +1,41 @@
+import asyncio
+from datetime import timedelta
+
+from crawlee.crawlers import HttpCrawler, HttpCrawlingContext
+from crawlee.sessions import SessionPool
+
+
+async def main() -> None:
+    # Define the pool outside the crawler so its state can be read before the run.
+    # `persist_state_kvs_name` is required: an unnamed store is purged on start,
+    # so persistence would silently not survive a restart.
+    session_pool = SessionPool(
+        max_pool_size=1,
+        create_session_settings={
+            # Keep the single session alive across runs.
+            'max_usage_count': 999_999,
+            'max_age': timedelta(hours=999_999),
+            'max_error_score': 100,
+        },
+        persistence_enabled=True,
+        persist_state_kvs_name='my-cookie-store',
+    )
+
+    async with session_pool:
+        # Read a session before crawling. On the first run its jar is empty. On
+        # later runs the cookies from the previous run are already restored.
+        session = await session_pool.get_session()
+        print(f'Cookies before run: {session.cookies.get_cookies_as_dicts()}')
+
+        # With a single session, don't burn retries trying to rotate to another one.
+        crawler = HttpCrawler(max_session_rotations=0, session_pool=session_pool)
+
+        @crawler.router.default_handler
+        async def handler(context: HttpCrawlingContext) -> None:
+            context.log.info(f'Processing {context.request.url}')
+
+        await crawler.run(['https://httpbingo.org/cookies/set?visited=1'])
+
+
+if __name__ == '__main__':
+    asyncio.run(main())
--- a/docs/guides/code_examples/cookie_management/read_write_cookies.py
+++ b/docs/guides/code_examples/cookie_management/read_write_cookies.py
@@ -28,7 +28,7 @@ async def main() -> None:
         token = context.session.cookies['csrf_token']
         context.log.info(f'CSRF token: {token}')
 
-    # The server sets a `session_id` cookie and echoes the request cookies back.
+    # The server sets a `session_id` cookie on the response.
     await crawler.run(['https://httpbingo.org/cookies/set?session_id=42'])
 
 
--- a/docs/guides/code_examples/cookie_management/read_write_cookies.py
+++ b/docs/guides/code_examples/cookie_management/read_write_cookies.py
@@ -1,0 +1,36 @@
+import asyncio
+
+from crawlee.crawlers import HttpCrawler, HttpCrawlingContext
+
+
+async def main() -> None:
+    crawler = HttpCrawler()
+
+    @crawler.router.default_handler
+    async def handler(context: HttpCrawlingContext) -> None:
+        if context.session is None:
+            return
+
+        # Set a cookie on the session. It's sent with every following
+        # request that runs on this session.
+        context.session.cookies.set(
+            name='csrf_token',
+            value='abc123',
+            domain='httpbingo.org',
+        )
+
+        # Iterate over the cookies stored on the session, including ones the
+        # server set via `Set-Cookie` on the current or an earlier response.
+        for cookie in context.session.cookies:
+            context.log.info(f'{cookie["name"]}={cookie["value"]}')
+
+        # Read a single cookie value by name.
+        token = context.session.cookies['csrf_token']
+        context.log.info(f'CSRF token: {token}')
+
+    # The server sets a `session_id` cookie and echoes the request cookies back.
+    await crawler.run(['https://httpbingo.org/cookies/set?session_id=42'])
+
+
+if __name__ == '__main__':
+    asyncio.run(main())
--- a/docs/guides/code_examples/cookie_management/retry_single_session.py
+++ b/docs/guides/code_examples/cookie_management/retry_single_session.py
@@ -1,0 +1,41 @@
+import asyncio
+from datetime import timedelta
+
+from crawlee.crawlers import HttpCrawler, HttpCrawlingContext
+from crawlee.sessions import SessionPool
+
+
+async def main() -> None:
+    crawler = HttpCrawler(
+        # With a single session, don't burn retries trying to rotate to another one.
+        max_session_rotations=0,
+        session_pool=SessionPool(
+            # A single session, so every attempt runs on the same one.
+            max_pool_size=1,
+            create_session_settings={
+                # Keep the single session alive for the whole crawl.
+                'max_usage_count': 999_999,
+                'max_age': timedelta(hours=999_999),
+                'max_error_score': 100,
+            },
+        ),
+    )
+
+    @crawler.router.default_handler
+    async def handler(context: HttpCrawlingContext) -> None:
+        if context.session is None:
+            return
+
+        # First attempt: establish cookies, then raise to trigger a retry.
+        if not context.session.cookies.get_cookies_as_dicts():
+            await context.send_request('https://httpbingo.org/cookies/set?logged_in=1')
+            raise RuntimeError('retry with cookies')
+
+        # The retry runs on the same session, so the cookies are still here.
+        context.log.info(f'Cookies: {context.session.cookies.get_cookies_as_dicts()}')
+
+    await crawler.run(['https://httpbingo.org/cookies'])
+
+
+if __name__ == '__main__':
+    asyncio.run(main())
--- a/docs/guides/code_examples/cookie_management/retry_single_session.py
+++ b/docs/guides/code_examples/cookie_management/retry_single_session.py
@@ -27,7 +27,7 @@ async def main() -> None:
             return
 
         # First attempt: establish cookies, then raise to trigger a retry.
-        if not context.session.cookies.get_cookies_as_dicts():
+        if not context.session.cookies:
             await context.send_request('https://httpbingo.org/cookies/set?logged_in=1')
             raise RuntimeError('retry with cookies')
 
--- a/docs/guides/cookie_management.mdx
+++ b/docs/guides/cookie_management.mdx
@@ -1,0 +1,106 @@
+---
+id: cookie-management
+title: Cookie management
+description: How to read, set, and persist cookies across requests, retries, and runs in Crawlee.
+---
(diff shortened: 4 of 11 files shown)


docs/guides/cookie_management.mdx
```diff
@@ -0,0 +1,106 @@
...
+    {ReadWriteCookies}
+</RunnableCodeBlock>
+
+:::note
+
+`context.session` is `None` only when the crawler runs without a session pool (<ApiLink to="class/BasicCrawler#__init__">`use_session_pool`</ApiLink> set to `False`). With the default pool it's always set. The examples guard with `if context.session is None: return` to cover that case.
+
+:::
+
+A cookie has a `name`, a `value`, and optional parameters such as `domain`. If a cookie has no `domain`, it applies to any domain. For the full set of parameters, see <ApiLink to="class/CookieParam">`CookieParam`</ApiLink>. To dump the whole jar, use <ApiLink to="class/SessionCookies#get_cookies_as_dicts">`get_cookies_as_dicts()`</ApiLink>, and to load several cookies at once, use <ApiLink to="class/SessionCookies#set_cookies">`set_cookies()`</ApiLink>.
```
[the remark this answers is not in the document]
Mantisus: Your statement is incorrect. `http.cookiejar` applies domain-less cookies to every domain.

docs/guides/cookie_management.mdx
```diff
@@ -0,0 +1,106 @@
...
+
+In Crawlee, cookies are stored on the <ApiLink to="class/Session">`Session`</ApiLink> that handles a request, in a <ApiLink to="class/SessionCookies">`SessionCookies`</ApiLink> jar. When a request runs on a session, the HTTP client sends the session's cookies with the request and writes any `Set-Cookie` from the response back onto the same jar. The next request on that session carries the updated cookies.
+
+Cookies follow the session, not the URL. Two requests share cookies when they run on the same session. For how sessions are created and rotated, see the [Session management](./session-management) guide.
+
+This guide covers how to read and set cookies, seed them on new sessions, handle them with <ApiLink to="class/PlaywrightCrawler">`PlaywrightCrawler`</ApiLink>, and keep them across retries and runs.
+
+## Reading and setting cookies
+
+Access the jar through <ApiLink to="class/Session#cookies">`context.session.cookies`</ApiLink>. Use <ApiLink to="class/SessionCookies#set">`set`</ApiLink> to add a cookie, iterate the jar to read all cookies, and index by name to read one value.
```
vdusek: **Note:** `SessionCookies.__getitem__` raises `KeyError` for a missing cookie rather than returning `None` (the `str | None` return type is about `Cookie.value`, not absence). Since readers will naturally index cookies the server may or may not have set, one clause noting that would help.

docs/guides/cookie_management.mdx
vdusek: **Suggestion:** the claim is accurate, but worth stating the consequence in a guide that opens by naming login state and CSRF tokens. I confirmed against `http.cookiejar` that a domain-less cookie goes to every host:

```
https://httpbingo.org/x    -> nodomain=1; scoped=2
https://evil.example.com/x -> nodomain=1
```

The examples already do the right thing by passing an explicit `domain`. The prose just never says why that matters.

docs/guides/code_examples/cookie_management/read_write_cookies.py
```diff
@@ -0,0 +1,36 @@
...
+        # Iterate over the cookies stored on the session, including ones the
+        # server set via `Set-Cookie` on the current or an earlier response.
+        for cookie in context.session.cookies:
+            context.log.info(f'{cookie["name"]}={cookie["value"]}')
+
+        # Read a single cookie value by name.
+        token = context.session.cookies['csrf_token']
+        context.log.info(f'CSRF token: {token}')
+
+    # The server sets a `session_id` cookie and echoes the request cookies back.
```
vdusek: **Nit:** the handler never reads the response body, so the "echoes the request cookies back" half describes something the reader can't observe here.

```suggestion
    # The server sets a `session_id` cookie on the response.
```

docs/guides/code_examples/cookie_management/retry_single_session.py
```diff
@@ -0,0 +1,41 @@
...
+        ),
+    )
+
+    @crawler.router.default_handler
+    async def handler(context: HttpCrawlingContext) -> None:
+        if context.session is None:
+            return
+
+        # First attempt: establish cookies, then raise to trigger a retry.
+        if not context.session.cookies.get_cookies_as_dicts():
```
vdusek: **Suggestion:** `SessionCookies` implements `__bool__`, so the jar can be tested directly instead of building the whole list of dicts just to check emptiness.

```suggestion
        if not context.session.cookies:
```

docs/guides/code_examples/cookie_management/persist_cookies.py
```diff
@@ -0,0 +1,41 @@
...
+        },
+        persistence_enabled=True,
+        persist_state_kvs_name='my-cookie-store',
+    )
+
+    async with session_pool:
+        # Read a session before crawling. On the first run its jar is empty. On
+        # later runs the cookies from the previous run are already restored.
+        session = await session_pool.get_session()
+        print(f'Cookies before run: {session.cookies.get_cookies_as_dicts()}')
```
vdusek, line 28: **Nit:** since this is a `RunnableCodeBlock`, a single "Run on Apify" click only ever prints an empty jar. The point lands on the second run. The prose does explain it, so fine as-is, just noting the button is a little anticlimactic here.

docs/guides/cookie_management.mdx
```diff
@@ -0,0 +1,106 @@
...
+
+Keeps rotation for the rest of the crawl. Instead of raising, create a new request for the same URL with its session bound through the `session_id` parameter of <ApiLink to="class/Request#from_url">`Request.from_url`</ApiLink>. Setting `always_enqueue=True` re-queues the URL despite deduplication.
+
+<RunnableCodeBlock className="language-python" language="python">
+    {RetryPinnedSession}
+</RunnableCodeBlock>
+
+### Re-apply cookies in a pre-navigation hook
+
+Keeps rotation and the raise-to-retry flow, and shares the cookies across every session. Snapshot the cookies after the setup step into the crawler-wide <ApiLink to="class/UseStateFunction">`use_state`</ApiLink> store. Then re-apply them onto whichever session handles the request in a <ApiLink to="class/AbstractHttpCrawler#pre_navigation_hook">`pre_navigation_hook`</ApiLink>, which runs before navigation and receives `context.session`.
```
vdusek: **Note:** `use_state` is marked experimental. `UseStateFunction`'s docstring warns that "the behavior and interface may change in future versions" (`src/crawlee/_types.py:594`). This guide is the only page in `docs/` that mentions it, and it presents the approach as an equal peer to two stable ones. A half-sentence or a `:::note` would set expectations.

docs/guides/cookie_management.mdx
```diff
@@ -0,0 +1,106 @@
...
+
+To start every session with a known set of cookies, for example a consent flag or a pre-obtained token, pass them through <ApiLink to="class/SessionPool#__init__">`create_session_settings`</ApiLink> on the <ApiLink to="class/SessionPool">`SessionPool`</ApiLink>. Each new session in the pool is created with these cookies already in its jar.
+
+<RunnableCodeBlock className="language-python" language="python">
+    {InitialCookies}
+</RunnableCodeBlock>
+
+## Cookies with PlaywrightCrawler
+
+<ApiLink to="class/PlaywrightCrawler">`PlaywrightCrawler`</ApiLink> keeps the session jar and the browser context in sync. Before navigation, the session's cookies are loaded into the browser context. After the handler returns, cookies from the browser context are merged back onto the session. The merge captures whatever the page picked up through `Set-Cookie`, JavaScript, or redirects.
```
vdusek, line 52: **Note:** the merge is broader than "cookies from the browser context". `_get_cookies` calls `page.context.cookies()` with no URL filter (`_playwright_crawler.py:610`), so with the default shared context it pulls in every cookie in that context, including ones other sessions' pages set for unrelated domains. The `use_incognito_pages` paragraph covers the outbound direction; this inbound leak onto the session jar is the more surprising half and isn't mentioned.

vdusek (commented):
LGTM, the text and examples are great, thank you.

Leaving here a few comments from Claude to consider, once they are resolved, I'll merge it (you can either address them or leave it as it is and just mark as resolved, up to you, they are mostly nits).

vdusek (approved):

PR #2100, opened 2026-07-29
fix(events): correct the listener invocation, removal, and context release

vdusek:
Get rid of `pyee` as we really don't need it and modernize the `EventManager` and `LocalEventManager`. No public API changes.

### Fixes

- A listener whose `__call__` is async (a class instance, not a function) was treated as sync and handed to `asyncio.to_thread`, which returned its coroutine without ever awaiting it - the listener silently never ran.
- `off()` decided between "remove this listener" and "remove all listeners of the event" by the truthiness of `listener`, so a listener object that is falsy in a boolean context wiped all listeners of the event.
- `off()` left an empty entry behind in `_listeners_to_wrappers`, keeping a reference to a listener that is no longer registered, and created entries for events and listeners that were never registered at all.
- A listener that fits neither call shape (it takes two parameters, say) raised out of its own task instead of being logged like any other failing listener. Whether the listener takes the event data is now resolved once at registration, so the invocation itself is fully covered by the logging.
- `__aexit__` released the context only on its happy path. A cancellation or a failing emission left the manager active for good, and since re-entering an active manager is a no-op, `PersistState` was never emitted again.

### pyee is gone

`emit` creates the listener task itself instead of going through `pyee.asyncio.AsyncIOEventEmitter`, which:

- halves the tasks per invocation - the emitter used to schedule a wrapper task that spawned and awaited a second, inner listener task,
- registers every listener task synchronously in `emit`, so `wait_for_all_listeners_to_complete()` can no longer miss the listeners of a just-emitted event - the one-tick defer that #2088 had to make explicit is not needed anymore,
- drops a runtime dependency (`pyee` is still pulled in by `playwright` for the browser extras).

The `_listeners_to_wrappers` layout is deliberately kept as it is - apify-sdk-python reaches into it in `Actor.reboot()`. Its full unit test suite passes against this branch.

### The rest

- Everything that depends only on the listener (whether it takes the event data, sync/async, its name) is resolved once at registration instead of on every invocation, and the wrapping moved out of `on()` into `_wrap_listener`.
- The two `__aexit__` branches were collapsed into a single `last_exit` flag, `__aenter__` mirrors it, and the manual "not active" check was replaced by the `ensure_context` decorator, so the message matches the one every other method raises.
- The three `EventManager.on.listener_wrapper(): ...` DEBUG lines are gone - they reported that a task is awaited, that it completed, and that it was discarded.
- `LocalEventManager` reads the CPU and the memory info concurrently - `get_cpu_info` alone blocks its thread for 100 ms while sampling the CPU.
- Typing: `__aenter__ -> Self` and `set[asyncio.Task[None]]`.
- `wait_for_all_tasks_for_finish` renamed to `wait_for_all_tasks_to_finish`.
- The listener helpers are private static methods of `EventManager`, and the attributes of both managers are documented with docstrings instead of comments.

New tests cover every fix above, the synchronous task registration in `emit`, the nested context teardown, and the concurrent system info readings.

*✍️ Drafted by Claude Code*


src/crawlee/events/_local_event_manager.py
Mantisus: Should probably use try/finally as well

src/crawlee/events/_event_manager.py
```diff
@@ -73,77 +68,80 @@ def __init__(
...
             await self._emit_persist_state_event()
             await self.wait_for_all_listeners_to_complete(timeout=self._close_timeout)
-            self._active_ref_count -= 1
-            return
+        finally:
+            # The context has to be left even if closing fails, be it a cancellation or a failed emission. Staying
+            # active would mean never emitting `PersistState` again, as re-entering the context would be a no-op.
+            if last_exit:
+                self._listener_tasks.clear()
+                self._listeners_to_wrappers.clear()
```
Mantisus, line 142:
```suggestion
                self._listener_tasks.clear()
                self._listeners_to_wrappers.clear()
                self._waiting_listener_tasks.clear()
```

tests/unit/events/test_event_manager.py
```diff
@@ -266,82 +477,93 @@ async def second_waiting_listener(_: Any) -> None:
...
-    # A wrapper finalizing after close raises onto the loop (its `error` listener is gone by then), so watch
-    # both channels for a stray exception.
-    emitter_errors: list[BaseException] = []
-    event_manager._event_emitter.add_listener('error', emitter_errors.append)
+    # A listener task finalizing after the close cleared the task set must not report a stray exception.
     loop_errors: list[dict[str, Any]] = []
-    loop = asyncio.get_running_loop()
-    original_exception_handler = loop.get_exception_handler()
-    loop.set_exception_handler(lambda _loop, context: loop_errors.append(context))
+    asyncio.get_running_loop().set_exception_handler(lambda _loop, context: loop_errors.append(context))
```
Mantisus: Let's make sure this doesn't leak into other tests

Mantisus (approved):
let's also update `__aexit__` in `LocalEventManager`. Otherwise LGTM.

B4nan (approved):

PR #2104, opened 2026-07-30
fix(httpx): honor session cookies across HTTP client request paths

Ayush7614:
## Summary
- **Httpx**: `send_request` / `stream` / `crawl` share `_build_request` for outbound session cookies. Clients stay shared **per proxy** (not per session); cookies are re-applied in `_HttpxTransport` on every hop so they survive redirects without pinning a client to each jar.
- **Impit** (default client): honors `persist_cookies_per_session`. When persistence is enabled, the session jar is attached to the client; when disabled, existing cookies are sent via a `Cookie` header so the shared client stays cached. Clients are cached by `(proxy, CookieJar)` identity.
- **Httpx fingerprints**: `Accept` / `Accept-Language` / `User-Agent` come from a single fingerprint profile. `HeaderGenerator.get_common_headers` / `get_random_user_agent_header` are deprecated in favor of `get_specific_headers`.

## Why
`context.send_request()` silently dropped session cookies under Httpx, breaking auth that worked for navigation (including after redirects). Impit's documented `persist_cookies_per_session=False` never took effect because the session jar was always attached and mutated in place.

## Test plan
- [x] `tests/unit/http_clients/test_http_clients.py` — shared cookie send/persist/redirect coverage across clients
- [x] `tests/unit/http_clients/test_httpx.py` — single-`generate()` fingerprint headers; proxy-only client cache across sessions
- [x] `tests/unit/http_clients/test_impit.py` — Cookie-header path when persist is disabled

### DIFF
--- a/src/crawlee/http_clients/_httpx.py
+++ b/src/crawlee/http_clients/_httpx.py
@@ -284,6 +284,7 @@ class HttpxHttpClient(HttpClient):
             method=method,
             headers=dict(headers) if headers else None,
             content=payload,
+            cookies=session.cookies.jar if session else None,
             extensions={'crawlee_session': session if self._persist_cookies_per_session else None},
             timeout=timeout or httpx.USE_CLIENT_DEFAULT,
         )
@@ -333,15 +334,19 @@ class HttpxHttpClient(HttpClient):
     def _combine_headers(self, explicit_headers: HttpHeaders | None) -> HttpHeaders | None:
         """Merge default headers with explicit headers for an HTTP request.
 
-        Generate a final set of request headers by combining default headers, a random User-Agent header,
-        and any explicitly provided headers.
+        Generate a final set of request headers by combining default headers from a single fingerprint
+        (Accept, Accept-Language, User-Agent) and any explicitly provided headers. Using one fingerprint
+        avoids mixing Accept headers from one browser profile with a User-Agent from another.
         """
-        common_headers = self._header_generator.get_common_headers() if self._header_generator else HttpHeaders()
-        user_agent_header = (
-            self._header_generator.get_random_user_agent_header() if self._header_generator else HttpHeaders()
-        )
+        if self._header_generator:
+            generated_headers = self._header_generator.get_specific_headers(
+                header_names={'Accept', 'Accept-Language', 'User-Agent'},
+            )
+        else:
+            generated_headers = HttpHeaders()
+
         explicit_headers = explicit_headers or HttpHeaders()
-        headers = common_headers | user_agent_header | explicit_headers
+        headers = generated_headers | explicit_headers
         return headers or None
 
     @staticmethod
--- a/src/crawlee/http_clients/_httpx.py
+++ b/src/crawlee/http_clients/_httpx.py
@@ -20,6 +20,7 @@ from crawlee.http_clients import HttpClient, HttpCrawlingResult, HttpResponse
 if TYPE_CHECKING:
     from collections.abc import AsyncGenerator, AsyncIterator
     from datetime import timedelta
+    from http.cookiejar import CookieJar
     from ssl import SSLContext
 
     from crawlee import Request
@@ -145,7 +146,7 @@ class HttpxHttpClient(HttpClient):
 
         self._transport: _HttpxTransport | None = None
 
-        self._client_by_proxy_url = dict[str | None, httpx.AsyncClient]()
+        self._client_cache = dict[tuple[str | None, int | None], httpx.AsyncClient]()
 
     @override
     async def crawl(
@@ -157,17 +158,17 @@ class HttpxHttpClient(HttpClient):
         statistics: Statistics | None = None,
         timeout: timedelta | None = None,
     ) -> HttpCrawlingResult:
-        client = self._get_client(proxy_info.url if proxy_info else None)
-        headers = self._combine_headers(request.headers)
+        cookie_jar = session.cookies.jar if session else None
+        client = self._get_client(proxy_info.url if proxy_info else None, cookie_jar)
 
-        http_request = client.build_request(
+        http_request = self._build_request(
+            client=client,
             url=request.url,
             method=request.method,
-            headers=headers,
-            content=request.payload,
-            cookies=session.cookies.jar if session else None,
-            extensions={'crawlee_session': session if self._persist_cookies_per_session else None},
-            timeout=timeout.total_seconds() if timeout is not None else httpx.USE_CLIENT_DEFAULT,
+            headers=request.headers,
+            payload=request.payload,
+            session=session,
+            timeout=httpx.Timeout(timeout.total_seconds()) if timeout is not None else None,
         )
 
         try:
@@ -202,7 +203,8 @@ class HttpxHttpClient(HttpClient):
     ) -> HttpResponse:
         validate_http_url(url)
 
-        client = self._get_client(proxy_info.url if proxy_info else None)
+        cookie_jar = session.cookies.jar if session else None
+        client = self._get_client(proxy_info.url if proxy_info else None, cookie_jar)
 
         http_request = self._build_request(
             client=client,
@@ -240,7 +242,8 @@ class HttpxHttpClient(HttpClient):
     ) -> AsyncGenerator[HttpResponse]:
         validate_http_url(url)
 
-        client = self._get_client(proxy_info.url if proxy_info else None)
+        cookie_jar = session.cookies.jar if session else None
+        client = self._get_client(proxy_info.url if proxy_info else None, cookie_jar)
 
         http_request = self._build_request(
             client=client,
@@ -289,13 +292,13 @@ class HttpxHttpClient(HttpClient):
             timeout=timeout or httpx.USE_CLIENT_DEFAULT,
         )
 
-    def _get_client(self, proxy_url: str | None) -> httpx.AsyncClient:
-        """Retrieve or create an HTTP client for the given proxy URL.
+    def _get_client(self, proxy_url: str | None, cookie_jar: CookieJar | None = None) -> httpx.AsyncClient:
+        """Retrieve or create an HTTP client for the given proxy URL and cookie jar.
 
-        If a client for the specified proxy URL does not exist, create and store a new one.
+        Clients are cached by `(proxy_url, id(cookie_jar))` so that redirects use the session cookie jar
+        attached to the client rather than per-request `cookies=` which httpx strips on redirect.
         """
         if not self._transport:
-            # Configure connection pool limits and keep-alive connections for transport
             limits = self._async_client_kwargs.get(
                 'limits', httpx.Limits(max_connections=1000, max_keepalive_connections=200)
             )
@@ -307,8 +310,9 @@ class HttpxHttpClient(HttpClient):
                 limits=limits,
             )
 
-        if proxy_url not in self._client_by_proxy_url:
-            # Prepare a default kwargs for the new client.
+        cache_key = (proxy_url, id(cookie_jar) if cookie_jar is not None else None)
+
+        if cache_key not in self._client_cache:
             kwargs: dict[str, Any] = {
                 'proxy': proxy_url,
                 'http1': self._http1,
@@ -316,7 +320,6 @@ class HttpxHttpClient(HttpClient):
                 'follow_redirects': True,
             }
 
-            # Update the default kwargs with any additional user-provided kwargs.
             kwargs.update(self._async_client_kwargs)
 
             kwargs.update(
@@ -326,10 +329,13 @@ class HttpxHttpClient(HttpClient):
                 }
             )
 
-            client = httpx.AsyncClient(**kwargs)
-            self._client_by_proxy_url[proxy_url] = client
+            if cookie_jar is not None:
+                kwargs['cookies'] = cookie_jar
 
-        return self._client_by_proxy_url[proxy_url]
+            client = httpx.AsyncClient(**kwargs)
+            self._client_cache[cache_key] = client
+
+        return self._client_cache[cache_key]
 
     def _combine_headers(self, explicit_headers: HttpHeaders | None) -> HttpHeaders | None:
         """Merge default headers with explicit headers for an HTTP request.
@@ -365,9 +371,9 @@ class HttpxHttpClient(HttpClient):
         return False
 
     async def cleanup(self) -> None:
-        for client in self._client_by_proxy_url.values():
+        for client in self._client_cache.values():
             await client.aclose()
-        self._client_by_proxy_url.clear()
+        self._client_cache.clear()
         if self._transport:
             await self._transport.aclose()
             self._transport = None
--- a/src/crawlee/http_clients/_httpx.py
+++ b/src/crawlee/http_clients/_httpx.py
@@ -4,6 +4,7 @@ import asyncio
 from contextlib import asynccontextmanager
 from logging import DEBUG, WARNING, getLogger
 from typing import TYPE_CHECKING, Any, cast
+from urllib.request import Request as UrllibRequest
 
 import httpx
 from typing_extensions import override
@@ -64,25 +65,40 @@ class _HttpxResponse:
 
 
 class _HttpxTransport(httpx.AsyncHTTPTransport):
-    """HTTP transport adapter that stores response cookies in a `Session`.
+    """HTTP transport adapter that keeps session cookies off the shared `httpx` client.
 
-    This transport adapter modifies the handling of HTTP requests to update the session cookies
-    based on the response cookies, ensuring that the cookies are stored in the session object
-    rather than the `HTTPX` client itself.
+    Outbound cookies are applied per hop (including redirects) from the jar in request extensions,
+    because httpx strips the `Cookie` header on redirect and rebuilds it from the client jar. Response
+    `Set-Cookie` values are stored on the session and removed from the response so the shared client
+    jar stays empty and reusable across sessions.
     """
 
     @override
     async def handle_async_request(self, request: httpx.Request) -> httpx.Response:
+        if cookie_jar := cast('CookieJar | None', request.extensions.get('crawlee_cookie_jar')):
+            self._apply_cookie_header(request, cookie_jar)
+
         response = await super().handle_async_request(request)
         response.request = request
 
-        if session := cast('Session', request.extensions.get('crawlee_session')):
+        if session := cast('Session | None', request.extensions.get('crawlee_session')):
             session.cookies.store_cookies(list(response.cookies.jar))
 
         if 'Set-Cookie' in response.headers:
             del response.headers['Set-Cookie']
 
         return response
+
+    @staticmethod
+    def _apply_cookie_header(request: httpx.Request, jar: CookieJar) -> None:
+        """Set the Cookie header from a jar for the current request URL."""
+        urllib_request = UrllibRequest(str(request.url), headers=dict(request.headers))  # noqa: S310
+        jar.add_cookie_header(urllib_request)
+        cookie_header = urllib_request.get_header('Cookie')
+        if cookie_header:
+            request.headers['cookie'] = cookie_header
+        else:
+            request.headers.pop('cookie', None)
 
 
 @docs_group('HTTP clients')
@@ -146,7 +162,7 @@ class HttpxHttpClient(HttpClient):
 
         self._transport: _HttpxTransport | None = None
 
-        self._client_cache = dict[tuple[str | None, int | None], httpx.AsyncClient]()
+        self._client_by_proxy_url = dict[str | None, httpx.AsyncClient]()
 
     @override
     async def crawl(
@@ -158,8 +174,7 @@ class HttpxHttpClient(HttpClient):
         statistics: Statistics | None = None,
         timeout: timedelta | None = None,
     ) -> HttpCrawlingResult:
-        cookie_jar = session.cookies.jar if session else None
-        client = self._get_client(proxy_info.url if proxy_info else None, cookie_jar)
+        client = self._get_client(proxy_info.url if proxy_info else None)
 
         http_request = self._build_request(
             client=client,
@@ -203,8 +218,7 @@ class HttpxHttpClient(HttpClient):
     ) -> HttpResponse:
         validate_http_url(url)
 
-        cookie_jar = session.cookies.jar if session else None
-        client = self._get_client(proxy_info.url if proxy_info else None, cookie_jar)
+        client = self._get_client(proxy_info.url if proxy_info else None)
 
         http_request = self._build_request(
             client=client,
@@ -242,8 +256,7 @@ class HttpxHttpClient(HttpClient):
     ) -> AsyncGenerator[HttpResponse]:
         validate_http_url(url)
 
-        cookie_jar = session.cookies.jar if session else None
-        client = self._get_client(proxy_info.url if proxy_info else None, cookie_jar)
+        client = self._get_client(proxy_info.url if proxy_info else None)
 
         http_request = self._build_request(
             client=client,
@@ -288,15 +301,19 @@ class HttpxHttpClient(HttpClient):
             headers=dict(headers) if headers else None,
             content=payload,
             cookies=session.cookies.jar if session else None,
-            extensions={'crawlee_session': session if self._persist_cookies_per_session else None},
+            extensions={
+                # Used by the transport to re-apply cookies on every hop (httpx strips Cookie on redirect).
+                'crawlee_cookie_jar': session.cookies.jar if session else None,
+                'crawlee_session': session if self._persist_cookies_per_session else None,
+            },
             timeout=timeout or httpx.USE_CLIENT_DEFAULT,
         )
 
-    def _get_client(self, proxy_url: str | None, cookie_jar: CookieJar | None = None) -> httpx.AsyncClient:
-        """Retrieve or create an HTTP client for the given proxy URL and cookie jar.
+    def _get_client(self, proxy_url: str | None) -> httpx.AsyncClient:
+        """Retrieve or create an HTTP client for the given proxy URL.
 
-        Clients are cached by `(proxy_url, id(cookie_jar))` so that redirects use the session cookie jar
-        attached to the client rather than per-request `cookies=` which httpx strips on redirect.
+        Clients are shared per proxy (not per session). Session cookies stay on the request /
+        transport path so concurrent sessions can reuse one client and its connection pool.
         """
         if not self._transport:
             limits = self._async_client_kwargs.get(
@@ -310,9 +327,7 @@ class HttpxHttpClient(HttpClient):
                 limits=limits,
             )
 
-        cache_key = (proxy_url, id(cookie_jar) if cookie_jar is not None else None)
-
-        if cache_key not in self._client_cache:
+        if proxy_url not in self._client_by_proxy_url:
             kwargs: dict[str, Any] = {
                 'proxy': proxy_url,
                 'http1': self._http1,
@@ -329,13 +344,10 @@ class HttpxHttpClient(HttpClient):
                 }
             )
 
-            if cookie_jar is not None:
-                kwargs['cookies'] = cookie_jar
-
             client = httpx.AsyncClient(**kwargs)
-            self._client_cache[cache_key] = client
+            self._client_by_proxy_url[proxy_url] = client
 
-        return self._client_cache[cache_key]
+        return self._client_by_proxy_url[proxy_url]
 
     def _combine_headers(self, explicit_headers: HttpHeaders | None) -> HttpHeaders | None:
         """Merge default headers with explicit headers for an HTTP request.
@@ -371,9 +383,9 @@ class HttpxHttpClient(HttpClient):
         return False
 
     async def cleanup(self) -> None:
-        for client in self._client_cache.values():
+        for client in self._client_by_proxy_url.values():
             await client.aclose()
-        self._client_cache.clear()
+        self._client_by_proxy_url.clear()
         if self._transport:
             await self._transport.aclose()
             self._transport = None
--- a/src/crawlee/http_clients/_impit.py
+++ b/src/crawlee/http_clients/_impit.py
@@ -2,10 +2,9 @@ from __future__ import annotations
 
 import asyncio
 from contextlib import asynccontextmanager
-from copy import deepcopy
-from http.cookiejar import CookieJar
 from logging import getLogger
 from typing import TYPE_CHECKING, Any, TypedDict
+from urllib.request import Request as UrllibRequest
 
 from cachetools import LRUCache
 from impit import AsyncClient, Browser, HTTPError, Response, TimeoutException, TransportError
@@ -22,6 +21,7 @@ from crawlee.http_clients import HttpClient, HttpCrawlingResult, HttpResponse
 if TYPE_CHECKING:
     from collections.abc import AsyncGenerator, AsyncIterator
     from datetime import timedelta
+    from http.cookiejar import CookieJar
 
     from crawlee import Request
     from crawlee._types import HttpMethod, HttpPayload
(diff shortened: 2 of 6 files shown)


src/crawlee/http_clients/_impit.py
```diff
@@ -116,7 +120,26 @@ def __init__(
...
+            return None
+
+        if self._persist_cookies_per_session:
+            return session.cookies.jar
+
+        # Copy cookies so Impit can attach a jar for outbound Cookie headers without mutating the session.
+        jar = CookieJar()
+        for cookie in session.cookies.jar:
+            jar.set_cookie(deepcopy(cookie))
+        return jar
```
Mantisus: `SessionCookies` already does exactly this, so we can use it here.

```python
from crawlee.sessions import SessionCookies

return SessionCookies(session.cookies).jar
```

The `deepcopy` isn't needed. `CookieJar.set_cookie` replaces the entry in its own dict rather than mutating the stored `Cookie`, and `extract_cookies` always builds new objects. A fresh jar is enough to isolate the session.

src/crawlee/http_clients/_impit.py
```diff
@@ -249,7 +285,13 @@ def _get_client(self, proxy_url: str | None, cookie_jar: CookieJar | None) -> As
 
         client = AsyncClient(**kwargs, cookie_jar=cookie_jar)
 
-        self._client_by_proxy_url[proxy_url] = _ClientCacheEntry(client=client, cookie_jar=cookie_jar)
+        if cache_key is not None:
+            # Close the client being evicted when the LRU is full, to avoid leaking connections.
+            if len(self._client_cache) >= self._client_cache.maxsize:
+                _evicted_key, evicted_entry = next(iter(self._client_cache.items()))
```
Mantisus: If I'm not mistaken, you need to use `popitem` with `cachetools.LRUCache`

src/crawlee/http_clients/_impit.py
```diff
@@ -249,7 +285,13 @@ def _get_client(self, proxy_url: str | None, cookie_jar: CookieJar | None) -> As
...
+                asyncio.get_running_loop().create_task(self._close_client(evicted_entry['client']))
```
Mantisus: The task may be garbage collected before it completes.

src/crawlee/http_clients/_impit.py
```diff
@@ -222,18 +245,31 @@ async def stream(
...
 
+    @staticmethod
+    def _make_cache_key(proxy_url: str | None, cookie_jar: CookieJar | None) -> _ClientCacheKey:
+        return (proxy_url, id(cookie_jar) if cookie_jar is not None else None)
+
+    async def _close_client(self, client: AsyncClient) -> None:
+        # Impit exposes cleanup via the async context manager protocol.
+        result = client.__aexit__(None, None, None)
+        if hasattr(result, '__await__'):
+            await result  # type: ignore[misc]
```
Mantisus: It's a good guard, but overall, `impit` should handle potential leaks well thanks to its Rust implementation

Mantisus (changes requested):
Thank you for your contributions!

When `persist_cookies_per_session` is False, `_resolve_cookie_jar` builds a fresh jar for every request, and since the cache key includes the jar identity, that means a new `AsyncClient` per request. Consider building the `Cookie` header directly instead of passing a `CookieJar` to `AsyncClient`, then the client can stay cached and shared.

For example:

```python
import urllib.request
from http.cookiejar import CookieJar


def get_cookie_header(jar: CookieJar, url: str, headers: HttpHeaders | None = None) -> str:
    request = urllib.request.Request(url, headers=dict(headers) if headers else {})
    jar.add_cookie_header(request)
    return request.get_header('Cookie', '')
```

Ayush7614:
Thanks for the review @Mantisus!

Addressed in `dccf1e7`:

- When `persist_cookies_per_session=False`, cookies are now sent via a `Cookie` header (using `CookieJar.add_cookie_header`) instead of attaching a fresh jar — so the Impit client stays cached/shared by proxy.
- LRU eviction now uses `cachetools.LRUCache.popitem()`.
- Removed the fire-and-forget `create_task(close...)` path (and the `__aexit__` close helper) since Impit’s Rust side already handles resource cleanup well.

src/crawlee/http_clients/_impit.py
Ayush7614: Superseded by the Cookie-header approach for `persist_cookies_per_session=False` — no jar copy is needed anymore.

src/crawlee/http_clients/_impit.py
Ayush7614: Fixed — eviction now uses `self._client_cache.popitem()`.

src/crawlee/http_clients/_impit.py
Ayush7614: Removed the fire-and-forget close task entirely.

src/crawlee/http_clients/_impit.py
```diff
@@ -116,7 +120,43 @@ def __init__(
...
+
+        return None, headers or None
+
+    @staticmethod
+    def _get_cookie_header(jar: CookieJar, url: str, headers: HttpHeaders | None = None) -> str:
+        """Build a Cookie request header from a jar without attaching the jar to the client."""
+        # UrllibRequest is only used to format Cookie headers via CookieJar; it never opens a connection.
+        request = UrllibRequest(url, headers=dict(headers) if headers else {})  # noqa: S310
+        jar.add_cookie_header(request)
+        return request.get_header('Cookie') or ''
```
Mantisus:
```suggestion
        return request.get_header('Cookie', '')
```

src/crawlee/http_clients/_impit.py
```diff
@@ -116,7 +120,43 @@ def __init__(
...
+
+        if session is None:
+            return None, headers or None
+
+        if self._persist_cookies_per_session:
+            return session.cookies.jar, headers or None
+
+        cookie_header = self._get_cookie_header(session.cookies.jar, url, headers)
+        if cookie_header and 'cookie' not in headers:
+            headers = headers | HttpHeaders({'Cookie': cookie_header})
```
Mantisus: `_get_cookie_header`correctly handles cases where a `Cookie` header is present in the headers

```suggestion
        if cookie_header := self._get_cookie_header(session.cookies.jar, url, headers):
            headers = headers | HttpHeaders({'Cookie': cookie_header})
```

src/crawlee/http_clients/_impit.py
```diff
@@ -116,7 +120,43 @@ def __init__(
...
+
+        When persistence is enabled, attach the session jar to Impit so response cookies update it.
+        When persistence is disabled, send existing cookies via the `Cookie` header and keep the
+        shared client (no jar) so clients stay cached and reusable.
+        """
+        if isinstance(headers, dict) or headers is None:
+            headers = HttpHeaders(headers or {})
+
+        if session is None:
+            return None, headers or None
```
Mantisus: `or None` is unnecessary after:

```python
if isinstance(headers, dict) or headers is None:
    headers = HttpHeaders(headers or {})
```

src/crawlee/http_clients/_impit.py
```diff
@@ -249,7 +297,11 @@ def _get_client(self, proxy_url: str | None, cookie_jar: CookieJar | None) -> As
 
         client = AsyncClient(**kwargs, cookie_jar=cookie_jar)
 
-        self._client_by_proxy_url[proxy_url] = _ClientCacheEntry(client=client, cookie_jar=cookie_jar)
+        # Evict the least-recently-used entry explicitly before inserting.
+        if cache_key not in self._client_cache and len(self._client_cache) >= self._client_cache.maxsize:
+            self._client_cache.popitem()
```
Mantisus: This is dead code now. The LRU cache automatically evicts the items.

tests/unit/http_clients/test_http_clients.py
```diff
@@ -323,3 +327,98 @@ def test_import_error_handled(optional_module_name: str, import_path: str) -> No
...
+    """When persistence is enabled, response Set-Cookie must update the session jar."""
+    session = Session()
+    request = Request.from_url(str(server_url.with_path('set_cookies').extend_query(a=1)))
+
+    await custom_http_client.crawl(request, session=session)
+
+    assert {cookie['name']: cookie['value'] for cookie in session.cookies.get_cookies_as_dicts()} == {'a': '1'}
+
+
+async def test_httpx_headers_come_from_single_fingerprint() -> None:
```
Mantisus: Tests that verify the behavior of a specific client should be placed in a separate test file for that client

tests/unit/http_clients/test_http_clients.py
```diff
@@ -323,3 +327,98 @@ def test_import_error_handled(optional_module_name: str, import_path: str) -> No
...
+        combined = client._combine_headers(None)
+
+    mocked.assert_called_once_with(header_names={'Accept', 'Accept-Language', 'User-Agent'})
+    assert combined is not None
+    assert combined['accept'] == 'text/html'
+    assert combined['accept-language'] == 'en-US'
+    assert combined['user-agent'] == 'TestAgent/1.0'
+
+
+async def test_impit_cleanup_clears_client_cache(server_url: URL) -> None:
```
Mantisus: Tests that verify the behavior of a specific client should be placed in a separate test file for that client

Mantisus (changes requested):
Thanks for the update, just a few more points

Ayush7614:
Thanks @Mantisus — addressed in `086b115`:

- `get_header('Cookie', '')`
- simplified Cookie-header merge via walrus + `_get_cookie_header`
- dropped redundant `or None` and manual `popitem`
- moved Httpx/Impit-specific tests into `test_httpx.py` / `test_impit.py`

Mantisus (approved):
Please update the PR description to match the current state, otherwise LGTM

Ayush7614:
Updated the PR description to match the current implementation. Thanks @Mantisus!

src/crawlee/http_clients/_httpx.py
```diff
@@ -284,6 +284,7 @@ def _build_request(
             method=method,
             headers=dict(headers) if headers else None,
             content=payload,
+            cookies=session.cookies.jar if session else None,
```
vdusek, line 303: **Note:** this fixes the direct-request case, but httpx still drops session cookies on redirect. `Client._redirect_headers` ends with an unconditional `headers.pop("Cookie", None)` and rebuilds cookies from the client jar, which crawlee deliberately keeps empty (`_HttpxTransport` deletes `Set-Cookie`), so per-request `cookies=` never survives a hop.

Measured against a local `/start` -> 302 -> `/echo` with this branch: impit (both persistence modes) and curl forward `auth=secret`, httpx sends no `Cookie` at all. Not a regression, but it is the same defect this PR targets and auth flows redirect constantly. Either set the `Cookie` header explicitly (reusing `_get_cookie_header`, as the impit path now does) or seed the client jar per request. At minimum, add a redirect case to the shared cookie tests -- the new tests hide this, since `test_send_request_sends_session_cookies` hits `/cookies` directly and `test_persist_cookies_per_session_*` only asserts jar contents.

vdusek, line 303: **Suggestion:** with `cookies=` added here, `_build_request` and the inline request construction in `crawl` (L163-171) are now equivalent apart from the timeout expression, so `crawl` can just call `self._build_request(...)`. Two hand-maintained copies of this wiring is exactly how the original `cookies=` asymmetry crept in.

src/crawlee/http_clients/_httpx.py
```diff
@@ -333,15 +334,19 @@ def _get_client(self, proxy_url: str | None) -> httpx.AsyncClient:
...
+        Generate a final set of request headers by combining default headers from a single fingerprint
+        (Accept, Accept-Language, User-Agent) and any explicitly provided headers. Using one fingerprint
+        avoids mixing Accept headers from one browser profile with a User-Agent from another.
         """
-        common_headers = self._header_generator.get_common_headers() if self._header_generator else HttpHeaders()
-        user_agent_header = (
-            self._header_generator.get_random_user_agent_header() if self._header_generator else HttpHeaders()
-        )
+        if self._header_generator:
+            generated_headers = self._header_generator.get_specific_headers(
```
vdusek, line 360: **Note:** this was the last production caller of `HeaderGenerator.get_common_headers` and `get_random_user_agent_header` (`fingerprint_suite/_header_generator.py:48,57`) -- only `tests/unit/fingerprint_suite/test_header_generator.py` still exercises them. They are public API so removal is out of scope here, but worth marking deprecated rather than leaving as unexplained dead weight.

src/crawlee/http_clients/_impit.py
```diff
@@ -30,6 +31,9 @@
 
 logger = getLogger(__name__)
 
+# Cache key: (proxy_url, id(cookie_jar) or None)
+_ClientCacheKey = tuple[str | None, int | None]
```
vdusek: **Suggestion:** `http.cookiejar.CookieJar` overrides neither `__eq__` nor `__hash__`, so it already hashes by identity -- `(proxy_url, cookie_jar)` is the exact key you want. `_ClientCacheKey`, `_make_cache_key` (L269), the identity recheck (L282) and the `_ClientCacheEntry` TypedDict all exist only to route around `id()`. Lifetime is unchanged either way, since the entry already held a strong ref to the jar. It needs `CookieJar` imported at runtime rather than under `TYPE_CHECKING` (stdlib, cheap), after which `LRUCache[tuple[str | None, CookieJar | None], AsyncClient]` drops the alias, this comment, the helper and the TypedDict in one edit -- and `_get_client` goes back to its pre-PR shape.

src/crawlee/http_clients/_impit.py
```diff
@@ -116,7 +120,42 @@ def __init__(
...
+        *,
+        session: Session | None,
+        url: str,
+        headers: HttpHeaders | dict[str, str] | None,
+    ) -> tuple[CookieJar | None, HttpHeaders]:
+        """Resolve cookie jar / Cookie header based on `persist_cookies_per_session`.
+
+        When persistence is enabled, attach the session jar to Impit so response cookies update it.
+        When persistence is disabled, send existing cookies via the `Cookie` header and keep the
+        shared client (no jar) so clients stay cached and reusable.
```
vdusek: **Nit:** these lines stop at ~96-99 chars against the 120-col limit the rest of the file fills, and the codebase writes `impit` rather than "Impit".

```suggestion
        """Resolve cookie jar / Cookie header based on `persist_cookies_per_session`.

        When persistence is enabled, attach the session jar to `impit` so response cookies update it. When
        persistence is disabled, send existing cookies via the `Cookie` header and keep the shared client (no jar)
        so clients stay cached and reusable.
```

src/crawlee/http_clients/_impit.py
```diff
@@ -116,7 +120,42 @@ def __init__(
...
+        if self._persist_cookies_per_session:
+            return session.cookies.jar, headers
+
+        if cookie_header := self._get_cookie_header(session.cookies.jar, url, headers):
+            headers = headers | HttpHeaders({'Cookie': cookie_header})
+
+        return None, headers
+
+    @staticmethod
+    def _get_cookie_header(jar: CookieJar, url: str, headers: HttpHeaders | None = None) -> str:
```
vdusek: **Nit:** the only caller always passes headers, so the default is dead.

```suggestion
    def _get_cookie_header(jar: CookieJar, url: str, headers: HttpHeaders) -> str:
```

src/crawlee/http_clients/_impit.py
```diff
@@ -222,18 +265,22 @@ async def stream(
         finally:
             response.close()
 
+    @staticmethod
+    def _make_cache_key(proxy_url: str | None, cookie_jar: CookieJar | None) -> _ClientCacheKey:
+        return (proxy_url, id(cookie_jar) if cookie_jar is not None else None)
```
vdusek: **Nit:** a static helper with a single call site that builds a two-element tuple -- inline it, or drop it entirely (see the comment on L34).

src/crawlee/http_clients/_impit.py
```diff
@@ -222,18 +265,22 @@ async def stream(
...
-        if cached_data:
-            client = cached_data['client']
-            client_cookie_jar = cached_data['cookie_jar']
-            if client_cookie_jar is cookie_jar:
-                # If the cookie jar matches, return the existing client.
-                return client
+        cache_key = self._make_cache_key(proxy_url, cookie_jar)
+
+        cached_data = self._client_cache.get(cache_key)
+        if cached_data and cached_data['cookie_jar'] is cookie_jar:
```
vdusek: **Nit:** `cached_data['cookie_jar'] is cookie_jar` can never be `False` -- the key already contains `id(cookie_jar)` and the entry keeps that jar alive, so a live key cannot map to a different object. (That is also why the `id()`-reuse hazard is genuinely safe here, just unnecessary.)

tests/unit/http_clients/test_http_clients.py
```diff
@@ -323,3 +325,65 @@ def test_import_error_handled(optional_module_name: str, import_path: str) -> No
...
+@pytest.mark.parametrize(
+    'custom_http_client',
+    [
+        pytest.param(CurlImpersonateHttpClient(persist_cookies_per_session=False), id='curl'),
+        pytest.param(HttpxHttpClient(persist_cookies_per_session=False), id='httpx'),
+        pytest.param(ImpitHttpClient(persist_cookies_per_session=False), id='impit'),
+    ],
+    indirect=['custom_http_client'],
+)
+async def test_persist_cookies_per_session_false(custom_http_client: HttpClient, server_url: URL) -> None:
```
vdusek, line 363: **Note:** no test enters the new `Cookie`-header branch at `_impit.py:147`, which is the whole point of the impit change. This test starts from a fresh `Session()` with an empty jar, so `_get_cookie_header` returns `''` and the branch is skipped; `test_send_request_sends_session_cookies` uses the default clients (`persist=True`) and takes the jar path instead. I confirmed the branch works, but nothing here would catch it breaking -- one parametrized case with `ImpitHttpClient(persist_cookies_per_session=False)` plus a pre-seeded session, asserting `/cookies` echoes the cookies, closes it.

tests/unit/http_clients/test_http_clients.py
```diff
@@ -323,3 +325,65 @@ def test_import_error_handled(optional_module_name: str, import_path: str) -> No
                 sys.modules.pop(mod_name, None)
         with pytest.raises(ImportError):
             importlib.import_module(import_path)
+
+
+async def test_send_request_sends_session_cookies(http_client: HttpClient, server_url: URL) -> None:
+    """`send_request` must attach existing session cookies (same as `crawl`)."""
+    session = Session()
+    session.cookies.set('auth', 'token-1', domain=server_url.host or '127.0.0.1', path='/')
```
vdusek, line 333: **Nit:** `or '127.0.0.1'` is unreachable for the local test server -- it is only narrowing `URL.host` from `str | None`. Same on L344. Fine to leave, but a single narrowed local would say that more clearly than a fallback that looks like real behaviour.

tests/unit/http_clients/test_httpx.py
```diff
@@ -54,3 +57,19 @@ async def test_common_headers_and_user_agent(server_url: URL, header_network: di
...
+async def test_headers_come_from_single_fingerprint() -> None:
+    """Accept and User-Agent must come from the same generated fingerprint profile."""
+    header_generator = HeaderGenerator()
+    fingerprint = {'Accept': 'text/html', 'Accept-Language': 'en-US', 'User-Agent': 'TestAgent/1.0'}
+
+    with patch.object(header_generator, 'get_specific_headers', return_value=HttpHeaders(fingerprint)) as mocked:
+        client = HttpxHttpClient(header_generator=header_generator)
+        combined = client._combine_headers(None)
+
+    mocked.assert_called_once_with(header_names={'Accept', 'Accept-Language', 'User-Agent'})
```
vdusek: **Note:** this test cannot fail for the reason it exists. It patches `get_specific_headers` to return a fixed dict, then asserts that dict came back -- but the property being claimed (Accept and User-Agent from *one* fingerprint) lives one level down, in how many times `BrowserforgeHeaderGenerator.generate` runs. Re-introducing a separate `generate()` call for the User-Agent would leave this green. Patch `header_generator._generator.generate` instead, assert exactly one call, and assert all three headers trace to that one dict. Two side notes: instantiating a real `HeaderGenerator()` loads the whole browserforge network just to have its method replaced (a `Mock` suffices), and project style prefers `monkeypatch`/`Mock` over `unittest.mock.patch`.

vdusek (changes requested):

Ayush7614:
Thanks @vdusek 

Addressed in d4f5fae / 6aaadc4:

- Httpx: attach session cookie jar on the client so cookies survive redirects; `crawl` reuses `_build_request`; shared redirect cookie coverage in `test_session_cookies_survive_redirect`.
- Impit: cache keyed by `(proxy_url, CookieJar)` identity; Cookie-header path when `persist_cookies_per_session=False`, covered by `test_persist_false_still_sends_session_cookies`.
- Deprecated `HeaderGenerator.get_common_headers` / `get_random_user_agent_header`.
- Fingerprint test now mocks `_generator.generate` and asserts a single call (so a dual-`generate()` regression would fail).

src/crawlee/http_clients/_httpx.py
```diff
@@ -145,7 +146,7 @@ def __init__(
 
         self._transport: _HttpxTransport | None = None
 
-        self._client_by_proxy_url = dict[str | None, httpx.AsyncClient]()
+        self._client_cache = dict[tuple[str | None, int | None], httpx.AsyncClient]()
```
vdusek: **Important:** this cache was bounded by the number of proxies; keyed per session jar it is now unbounded, one `AsyncClient` per distinct session for the lifetime of the crawl. Each client also pins its jar (`httpx.Cookies(jar).jar is jar`), so the cookies of retired sessions are never freed either.

Measured on this branch: 200 sessions leave 200 entries, all still pinning their jar after the sessions are dropped, while `ImpitHttpClient` stays at 10. This is reachable in a normal crawl, not just synthetically -- a `SessionPool(max_pool_size=3)` mints a fresh jar every time a session retires, 14 distinct jars over 40 checkouts, so the count grows without bound over a long run.

Use a bounded cache keyed by the jar itself, as `_impit.py` now does -- `CookieJar` hashes by identity, so that also removes the need for `id()` and the `int | None` in the key:

```suggestion
        self._client_cache = LRUCache[tuple[str | None, CookieJar | None], httpx.AsyncClient](maxsize=10)
```

That needs `from cachetools import LRUCache` plus `CookieJar` imported at runtime rather than under `TYPE_CHECKING`, and `_get_client` keying on `(proxy_url, cookie_jar)`.

One caveat: do not `aclose()` evicted clients. They all share `self._transport`, and `AsyncClient.aclose()` closes it -- on httpx 0.28.1 / httpcore 1.0.9 that empties the shared connection pool, so the remaining clients lose every keep-alive connection and any in-flight request on them fails with `ReadError`. They do recover on the next request, so it is not fatal, but dropping evicted clients without closing them is both correct and cheaper: the shared transport is the only resource they own, and `cleanup()` already closes it.

tests/unit/http_clients/test_http_clients.py
```diff
@@ -323,3 +325,88 @@ def test_import_error_handled(optional_module_name: str, import_path: str) -> No
...
+        pytest.param(CurlImpersonateHttpClient(persist_cookies_per_session=True), id='curl'),
+        pytest.param(HttpxHttpClient(persist_cookies_per_session=True), id='httpx'),
+        pytest.param(ImpitHttpClient(persist_cookies_per_session=True), id='impit'),
+    ],
+    indirect=['custom_http_client'],
+)
+async def test_session_cookies_survive_redirect(custom_http_client: HttpClient, server_url: URL) -> None:
+    """Pre-seeded session cookies must be present after a redirect (not stripped on the second hop)."""
+    session = Session()
+    session.cookies.set('tracker', 'abc', domain=server_url.host or '127.0.0.1', path='/')
```
vdusek, line 404: **Nit:** same `or '127.0.0.1'` as the still-open thread on L333 -- it is unreachable for the local test server and only narrows `URL.host` from `str | None`, so it reads like real fallback behaviour. This commit added two more (here and `test_impit.py:34`), making four. A single narrowed local at the top of each test would say it more clearly.

vdusek (changes requested):

src/crawlee/http_clients/_httpx.py
Mantisus: This update will significantly affect the performance of `HttpxHttpClient`. At `impit`, this was a necessary measure due to the specific nature of working with cookies.

Ayush7614: Agreed — Impit needs the jar on the client, but Httpx should keep one `AsyncClient` per proxy.

Fixed in 6b2c8c8: reverted to proxy-only client sharing, and re-apply cookies in `_HttpxTransport` on every hop (extensions survive redirects, while httpx strips `Cookie` and rebuilds from the empty client jar). Redirect cookie coverage still passes, and `test_client_cache_is_shared_across_sessions` asserts the cache stays at one entry across distinct sessions.

Ayush7614:
Type-check failures on the previous run came from `test_client_cache_is_shared_across_sessions`: `async with HttpxHttpClient() as client` widens to `HttpClient`, so accessing `_client_by_proxy_url` failed under `ty`.

Fixed in bf9a8d0 by binding the `HttpxHttpClient` first (same pattern as the Impit cache test). Local `ty check` is clean.

Ayush7614:
cc: @vdusek

src/crawlee/http_clients/_httpx.py
```diff
@@ -284,17 +300,22 @@ def _build_request(
             method=method,
             headers=dict(headers) if headers else None,
             content=payload,
-            extensions={'crawlee_session': session if self._persist_cookies_per_session else None},
+            cookies=session.cookies.jar if session else None,
```
Mantisus, line 303: With this new approach that uses transport, `cookies=session.cookies.jar` doesn't make sense

Also, the session is already being passed to the transport. Use the session's `jar` no need to pass it separately.

Mantisus (changes requested):

src/crawlee/http_clients/_impit.py
```diff
@@ -116,7 +110,41 @@ def __init__(
 
         self._async_client_kwargs = async_client_kwargs
 
-        self._client_by_proxy_url = LRUCache[str | None, _ClientCacheEntry](maxsize=10)
+        self._client_cache = LRUCache[tuple[str | None, CookieJar | None], AsyncClient](maxsize=10)
```
vdusek, line 113: Why rename it? All HTTP clients name this attribute `_client_by_proxy_url`, no?

tests/unit/http_clients/test_httpx.py
```diff
@@ -50,7 +53,40 @@ async def test_common_headers_and_user_agent(server_url: URL, header_network: di
...
+    assert host is not None
+
+    client = HttpxHttpClient(http2=False)
+    async with client:
+        for i in range(5):
+            session = Session()
+            session.cookies.set(f'k{i}', f'v{i}', domain=host, path='/')
+            await client.send_request(str(server_url / 'cookies'), session=session)
+
+        assert len(client._client_by_proxy_url) == 1
```
vdusek, line 92: **Note:** the load-bearing new invariant is that one client per proxy is shared across sessions, so session A's cookies must never reach session B. This asserts the sharing but not the isolation. Worth adding: crawl `set_cookies` with session A, then request `/cookies` with session B and with no session, asserting `{}` for both.

tests/unit/http_clients/test_httpx.py
```diff
@@ -50,7 +53,40 @@ async def test_common_headers_and_user_agent(server_url: URL, header_network: di
...
+
+    header_generator._generator.generate.assert_called_once()
+    assert combined is not None
+    assert combined['accept'] == 'text/html'
+    assert combined['accept-language'] == 'en-US'
+    assert combined['user-agent'] == 'TestAgent/1.0'
+
+
+async def test_client_cache_is_shared_across_sessions(server_url: URL) -> None:
+    """Distinct sessions must reuse one AsyncClient per proxy, not one client per cookie jar."""
```
vdusek, line 81: **Suggestion:** `_client_by_proxy_url` was already a per-proxy `dict` on master, so this test passes there too - the docstring implies it guards this change. Also the per-iteration cookies (`k{i}=v{i}`) are set but never asserted.

src/crawlee/http_clients/_httpx.py
```diff
@@ -63,26 +65,41 @@ async def read_stream(self) -> AsyncIterator[bytes]:
...
+    @staticmethod
+    def _apply_cookie_header(request: httpx.Request, jar: CookieJar) -> None:
+        """Set the Cookie header from a jar for the current request URL."""
+        urllib_request = UrllibRequest(str(request.url), headers=dict(request.headers))  # noqa: S310
+        jar.add_cookie_header(urllib_request)
+        cookie_header = urllib_request.get_header('Cookie')
+        if cookie_header:
+            request.headers['cookie'] = cookie_header
+        else:
+            request.headers.pop('cookie', None)
```
vdusek, line 101: **Nit:** the `else` branch is unreachable. When a `Cookie` header is present, `add_cookie_header` short-circuits on `has_header("Cookie")` and `get_header` returns it unchanged, so the `if` always wins; otherwise `pop` is a no-op.

```suggestion
        if cookie_header:
            request.headers['cookie'] = cookie_header
```

tests/unit/http_clients/test_httpx.py
```diff
@@ -50,7 +53,40 @@ async def test_common_headers_and_user_agent(server_url: URL, header_network: di
...
+
+
+def test_headers_come_from_single_fingerprint() -> None:
+    """Accept and User-Agent must come from one `generate()` call, not mixed profiles."""
+    fingerprint = {'Accept': 'text/html', 'Accept-Language': 'en-US', 'User-Agent': 'TestAgent/1.0'}
+
+    # Avoid HeaderGenerator.__init__ loading browserforge; only exercise get_specific_headers.
+    header_generator = HeaderGenerator.__new__(HeaderGenerator)
+    header_generator._generator = Mock()
+    header_generator._generator.generate = Mock(return_value=fingerprint)
```
vdusek, line 68: The comment is not accurate - `_httpx.py:124` (`_DEFAULT_HEADER_GENERATOR = HeaderGenerator()`) evaluates at class-body time, so importing `HttpxHttpClient` in this module already loads browserforge. A plain instance works and drops the `__new__` bypass.

```suggestion
    header_generator = HeaderGenerator()
    header_generator._generator = Mock(generate=Mock(return_value=fingerprint))
```

src/crawlee/http_clients/_impit.py
```diff
@@ -223,19 +255,17 @@ async def stream(
             response.close()
 
     def _get_client(self, proxy_url: str | None, cookie_jar: CookieJar | None) -> AsyncClient:
-        """Retrieve or create an HTTP client for the given proxy URL.
+        """Retrieve or create an HTTP client for the given proxy URL and cookie jar.
 
-        If a client for the specified proxy URL does not exist, create and store a new one.
+        Clients are cached by `(proxy_url, cookie_jar)` — CookieJar hashes by identity so sessions with different
```
vdusek, line 260: Not em dashes please - `src/` is ASCII-only.

```suggestion
        Clients are cached by `(proxy_url, cookie_jar)`: `CookieJar` hashes by identity so sessions with different
```

tests/unit/http_clients/test_httpx.py
```diff
@@ -50,7 +53,40 @@ async def test_common_headers_and_user_agent(server_url: URL, header_network: di
     assert 'accept-language' in response_headers
     assert response_headers['accept-language'] == COMMON_ACCEPT_LANGUAGE
 
-    # By default, HTTPX uses its own User-Agent, which should be replaced by the one from the header generator.
     assert 'user-agent' in response_headers
```
vdusek, line 56: Why deleting this comment? It was the only explanation of why the assertion below checks for `python-httpx`. Also the deletion is unrelated to the fix.

src/crawlee/http_clients/_httpx.py
```diff
@@ -284,17 +300,22 @@ def _build_request(
             method=method,
             headers=dict(headers) if headers else None,
             content=payload,
-            extensions={'crawlee_session': session if self._persist_cookies_per_session else None},
+            cookies=session.cookies.jar if session else None,
+            extensions={
+                # Used by the transport to re-apply cookies on every hop (httpx strips Cookie on redirect).
+                'crawlee_cookie_jar': session.cookies.jar if session else None,
+                'crawlee_session': session if self._persist_cookies_per_session else None,
+            },
```
vdusek, line 308: **Note:** building on Mantisus' open thread here -- the transport can't just read the jar off `crawlee_session`, because that key is `None` whenever `persist_cookies_per_session` is False, so it doubles as both the session and the persist flag. That conflation is the only reason `crawlee_cookie_jar` has to exist.

Pass `persist_cookies_per_session` into `_HttpxTransport.__init__` (the transport is already per-client, L162), send `extensions={'crawlee_session': session}` unconditionally, and derive the jar inside the transport -- that drops both `cookies=` and `crawlee_cookie_jar`, leaving one channel with one meaning.

src/crawlee/http_clients/_httpx.py
vdusek, line 101: **Suggestion:** this `UrllibRequest` + `add_cookie_header` + `get_header('Cookie')` idiom is now hand-rolled twice -- here and in `ImpitHttpClient._get_cookie_header` (`_impit.py:142`) -- and the two copies already diverge: this one pops the header when the jar yields nothing, impit returns `''` and lets the caller skip.

`add_cookie_header` appeared nowhere in `src/` before this PR. `SessionCookies` is what owns cookie state, so a `get_cookie_header(url, headers)` there would give both clients (and curl, the third latent site) one implementation instead of three.

src/crawlee/http_clients/_httpx.py
```diff
@@ -63,26 +65,41 @@ async def read_stream(self) -> AsyncIterator[bytes]:
...
+            self._apply_cookie_header(request, cookie_jar)
+
         response = await super().handle_async_request(request)
         response.request = request
 
-        if session := cast('Session', request.extensions.get('crawlee_session')):
+        if session := cast('Session | None', request.extensions.get('crawlee_session')):
             session.cookies.store_cookies(list(response.cookies.jar))
 
         if 'Set-Cookie' in response.headers:
```
vdusek, line 87: **Note:** httpx is the only client that strips this -- impit (`_impit.py:50-51`) and curl (`_curl_impersonate.py:101-102`) pass response headers through untouched. So the public `HttpResponse.headers` carries `set-cookie` under two clients and never under the third.

Pre-existing, but this PR promotes the deletion from implementation detail to stated mechanism in the new docstring, and it interacts with the semantics just given to `persist_cookies_per_session=False`: under httpx with persistence off, a `Set-Cookie` is now observable nowhere -- not in the session, not in the response -- while impit and curl users can still read it. The new cross-client cookie tests are the natural place to pin that divergence, or one line on `HttpResponse.headers` to document it.

vdusek:
no response from @Ayush7614, closing

PR #2105, opened 2026-07-30
fix: honor error_handler replacement requests on SessionError and preserve forefront on retries 

Ayush7614:
## Summary
- **BasicCrawler `SessionError` path**: honor `error_handler` replacement requests only while session rotations remain; treat same `unique_key` as no replacement (avoids marking the in-progress request handled while `add_request` is a no-op); wrap handler exceptions in `UserDefinedErrorHandlerError` and set `RequestState.ERROR` when the handler raises or rotations are exhausted.
- Shared `_handle_error_handler_replacement` helper for the SessionError and normal retry paths: inherits `retry_count` / `session_rotation_count`, passes `forefront=` to `add_request`, records `record_request_processing_finish` on the replaced original, and retires blocked sessions before awaits on the SessionError path.
- **Retries**: `reclaim_request` passes `forefront=request.forefront` so tiered-proxy priority retries stay at the front of the queue.
- `session.retire()` stays on the retry/rotation path only (preserves long-lived sessions when `max_session_rotations=0`).

## Why
`error_handler` could replace a request on normal failures, but its return value was discarded for `SessionError`. A field-wise `new_request != request` check treated same-URL replacements as new work and could drop the original. Tiered proxies set `request.forefront = True` on retry, but reclaim always used the default `forefront=False`.

## Test plan
- [x] `test_session_error_handler_can_replace_request`
- [x] `test_session_error_handler_same_unique_key_rotates`
- [x] `test_session_error_handler_replacement_ignored_when_rotations_exhausted`
- [x] `test_session_error_handler_exception_is_wrapped`
- [x] `test_reclaim_uses_request_forefront_flag`
- [x] Existing session rotation / error_handler tests still pass

### DIFF
--- a/src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
+++ b/src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
@@ -30,6 +30,7 @@ from crawlee.crawlers import (
 from crawlee.crawlers._beautifulsoup._beautifulsoup_parser import BeautifulSoupParser
 from crawlee.crawlers._parsel._parsel_parser import ParselParser
 from crawlee.crawlers._playwright._playwright_crawler import _PlaywrightCrawlerAdditionalOptions
+from crawlee.errors import SessionError
 from crawlee.statistics import Statistics, StatisticsState
 
 from ._adaptive_playwright_crawler_statistics import AdaptivePlaywrightCrawlerStatisticState
@@ -396,6 +397,10 @@ class AdaptivePlaywrightCrawler(
                     self._context_result_map[context] = static_run.result
                     return
                 if static_run.exception:
+                    # SessionError means the session is blocked. Falling through to the browser with the same
+                    # session would keep using a bad session; re-raise so BasicCrawler can rotate/retire it.
+                    if isinstance(static_run.exception, SessionError):
+                        raise static_run.exception
                     context.log.error(
                         msg=f'Static crawler: failed for {context.request.url}', exc_info=static_run.exception
                     )
--- a/src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
+++ b/src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
@@ -30,7 +30,6 @@ from crawlee.crawlers import (
 from crawlee.crawlers._beautifulsoup._beautifulsoup_parser import BeautifulSoupParser
 from crawlee.crawlers._parsel._parsel_parser import ParselParser
 from crawlee.crawlers._playwright._playwright_crawler import _PlaywrightCrawlerAdditionalOptions
-from crawlee.errors import SessionError
 from crawlee.statistics import Statistics, StatisticsState
 
 from ._adaptive_playwright_crawler_statistics import AdaptivePlaywrightCrawlerStatisticState
@@ -397,10 +396,6 @@ class AdaptivePlaywrightCrawler(
                     self._context_result_map[context] = static_run.result
                     return
                 if static_run.exception:
-                    # SessionError means the session is blocked. Falling through to the browser with the same
-                    # session would keep using a bad session; re-raise so BasicCrawler can rotate/retire it.
-                    if isinstance(static_run.exception, SessionError):
-                        raise static_run.exception
                     context.log.error(
                         msg=f'Static crawler: failed for {context.request.url}', exc_info=static_run.exception
                     )
--- a/src/crawlee/crawlers/_basic/_basic_crawler.py
+++ b/src/crawlee/crawlers/_basic/_basic_crawler.py
@@ -1180,7 +1180,7 @@ class BasicCrawler(Generic[TCrawlingContext, TStatisticsState]):
                         await self._mark_request_as_handled(request)
                         return
 
-            await request_manager.reclaim_request(request)
+            await request_manager.reclaim_request(request, forefront=request.forefront)
         else:
             request.state = RequestState.ERROR
             await self._mark_request_as_handled(request)
@@ -1474,22 +1474,33 @@ class BasicCrawler(Generic[TCrawlingContext, TStatisticsState]):
             if not session:
                 raise RuntimeError('SessionError raised in a crawling context without a session') from session_error
 
+            new_request = None
             if self._error_handler:
-                await self._error_handler(context, session_error)
+                try:
+                    new_request = await self._error_handler(context, session_error)
+                except Exception as e:
+                    raise UserDefinedErrorHandlerError('Exception thrown in user-defined request error handler') from e
+
+            if new_request is not None and new_request != request:
+                await request_manager.add_request(new_request)
+                await self._mark_request_as_handled(request)
+                session.retire()
+                return
 
             if self._should_retry_request(context, session_error):
                 exc_only = ''.join(traceback.format_exception_only(session_error)).strip()
                 self._logger.warning('Encountered "%s", rotating session and retrying...', exc_only)
 
-                if session:
-                    session.retire()
+                session.retire()
 
                 # Increment session rotation count.
                 request.session_rotation_count = (request.session_rotation_count or 0) + 1
 
-                await request_manager.reclaim_request(request)
+                await request_manager.reclaim_request(request, forefront=request.forefront)
                 await self._statistics.error_tracker_retry.add(error=session_error, context=context)
             else:
+                # Exhausted rotations: retire the blocked session so it is not reused from the pool.
+                session.retire()
                 await self._mark_request_as_handled(request)
 
                 await self._handle_failed_request(context, session_error)
--- a/src/crawlee/crawlers/_basic/_basic_crawler.py
+++ b/src/crawlee/crawlers/_basic/_basic_crawler.py
@@ -1148,6 +1148,56 @@ class BasicCrawler(Generic[TCrawlingContext, TStatisticsState]):
         # The URL does not match any `include` pattern - reject it
         return False
 
+    async def _handle_error_handler_replacement(
+        self,
+        context: TCrawlingContext | BasicCrawlingContext,
+        error: Exception,
+        *,
+        request_manager: RequestManager,
+        retire_session: bool = False,
+    ) -> bool:
+        """Invoke the user `error_handler` and enqueue a replacement request when appropriate.
+
+        A returned request with the same `unique_key` as the original is treated as no
+        replacement: `add_request` would be a no-op while the original is still in progress,
+        and marking the original handled would drop the work.
+
+        Args:
+            context: Current crawling context.
+            error: Exception that triggered the error handler.
+            request_manager: Request manager used to enqueue the replacement.
+            retire_session: When True, retire `context.session` before awaiting I/O so a
+                known-blocked session cannot stay in the pool if a later await fails.
+
+        Returns:
+            True if the original request was replaced and marked handled; False to continue
+            with the normal retry or failure path.
+        """
+        if not self._error_handler:
+            return False
+
+        try:
+            new_request = await self._error_handler(context, error)
+        except Exception as e:
+            context.request.state = RequestState.ERROR
+            raise UserDefinedErrorHandlerError('Exception thrown in user-defined request error handler') from e
+
+        if new_request is None or new_request.unique_key == context.request.unique_key:
+            return False
+
+        # Preserve counters so a handler minting a fresh unique_key each time cannot loop forever.
+        new_request.retry_count = context.request.retry_count
+        if context.request.session_rotation_count is not None:
+            new_request.session_rotation_count = context.request.session_rotation_count
+
+        if retire_session and context.session:
+            context.session.retire()
+
+        await request_manager.add_request(new_request, forefront=new_request.forefront)
+        await self._mark_request_as_handled(context.request)
+        self._statistics.record_request_processing_finish(context.request.unique_key)
+        return True
+
     async def _handle_request_retries(
         self,
         context: TCrawlingContext | BasicCrawlingContext,
@@ -1169,16 +1219,8 @@ class BasicCrawler(Generic[TCrawlingContext, TStatisticsState]):
             )
             await self._statistics.error_tracker.add(error=error, context=context)
 
-            if self._error_handler:
-                try:
-                    new_request = await self._error_handler(context, error)
-                except Exception as e:
-                    raise UserDefinedErrorHandlerError('Exception thrown in user-defined request error handler') from e
-                else:
-                    if new_request is not None and new_request != request:
-                        await request_manager.add_request(new_request)
-                        await self._mark_request_as_handled(request)
-                        return
+            if await self._handle_error_handler_replacement(context, error, request_manager=request_manager):
+                return
 
             await request_manager.reclaim_request(request, forefront=request.forefront)
         else:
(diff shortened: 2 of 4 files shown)


src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
```diff
@@ -396,6 +397,10 @@ async def _run_request_handler(self, context: BasicCrawlingContext) -> None:
                     self._context_result_map[context] = static_run.result
                     return
                 if static_run.exception:
+                    # SessionError means the session is blocked. Falling through to the browser with the same
+                    # session would keep using a bad session; re-raise so BasicCrawler can rotate/retire it.
+                    if isinstance(static_run.exception, SessionError):
+                        raise static_run.exception
```
Mantisus: The old behavior is correct here. A site can block the HTTP client even when the same session still works in a browser, for example, by requiring JavaScript execution before granting access.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1474,22 +1474,33 @@ async def __run_task_function(self) -> None:
...
+                try:
+                    new_request = await self._error_handler(context, session_error)
+                except Exception as e:
+                    raise UserDefinedErrorHandlerError('Exception thrown in user-defined request error handler') from e
+
+            if new_request is not None and new_request != request:
+                await request_manager.add_request(new_request)
+                await self._mark_request_as_handled(request)
+                session.retire()
+                return
```
Mantisus: Replacement with `new_request` should only apply in the retry branch. Please drop the early return here because it also triggers when session rotations are exhausted, so the request gets replaced instead of failed and `failed_request_handler` never runs.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1474,22 +1474,33 @@ async def __run_task_function(self) -> None:
...
 
                 # Increment session rotation count.
                 request.session_rotation_count = (request.session_rotation_count or 0) + 1
 
-                await request_manager.reclaim_request(request)
+                await request_manager.reclaim_request(request, forefront=request.forefront)
                 await self._statistics.error_tracker_retry.add(error=session_error, context=context)
             else:
+                # Exhausted rotations: retire the blocked session so it is not reused from the pool.
+                session.retire()
```
Mantisus: Please keep `session.retire()` only in the `_should_retry_request` branch. Otherwise, it breaks the long-lived session use case when `max_session_rotations=0`.

tests/unit/crawlers/_basic/test_basic_crawler.py
```diff
@@ -274,6 +274,89 @@ async def error_handler(context: BasicCrawlingContext, error: Exception) -> None
...
+    @crawler.router.default_handler
+    async def handler(context: BasicCrawlingContext) -> None:
+        context.request.forefront = True
+        raise RuntimeError('retry me')
+
+    await crawler.run([Request.from_url('https://a.placeholder.com')])
+
+    assert reclaim_calls == [True]
+
+    await queue.drop()
```
Mantisus: Let's use the mock
```suggestion
async def test_reclaim_uses_request_forefront_flag() -> None:
    """Retries must reclaim with `request.forefront` so tiered-proxy priority retries stay at the front."""
    queue = await RequestQueue.open()
    crawler = BasicCrawler(request_manager=queue, max_request_retries=1)

    @crawler.router.default_handler
    async def handler(context: BasicCrawlingContext) -> None:
        context.request.forefront = True
        raise RuntimeError('Arbitrary crash for testing purposes')

    with patch.object(queue, 'reclaim_request', wraps=queue.reclaim_request) as reclaim_mock:
        await crawler.run(['https://a.placeholder.com'])

    reclaim_mock.assert_awaited_once()

    (reclaimed_request,), reclaim_kwargs = reclaim_mock.await_args_list[0]
    assert reclaimed_request.url == 'https://a.placeholder.com'
    assert reclaim_kwargs == {'forefront': True}

    await queue.drop()
```

Mantisus (changes requested):
Hey, @Ayush7614. Thank you for your contribution!

Ayush7614:
Thanks for the review @Mantisus!

Addressed in `3a44e62`:

- Reverted the AdaptivePlaywright static `SessionError` re-raise — browser fallback with the same session is intentional when HTTP is blocked but JS may still work.
- `error_handler` replacement requests are honored only inside the `_should_retry_request` branch, so exhausted rotations still hit `failed_request_handler`.
- `session.retire()` is only called on the retry/rotation path (preserves `max_session_rotations=0` long-lived sessions).
- Exhausted path now sets `RequestState.ERROR` before marking handled.
- Forefront reclaim test updated to use `patch.object(..., wraps=...)`.

src/crawlee/crawlers/_adaptive_playwright/_adaptive_playwright_crawler.py
Ayush7614: Reverted — static `SessionError` again falls through to the browser path as before.

src/crawlee/crawlers/_basic/_basic_crawler.py
Ayush7614: Fixed — replacement is only applied inside the `_should_retry_request` branch.

src/crawlee/crawlers/_basic/_basic_crawler.py
Ayush7614: Fixed — `session.retire()` is only called on the retry/rotation path now.

tests/unit/crawlers/_basic/test_basic_crawler.py
Ayush7614: Updated the test to use `patch.object(queue, 'reclaim_request', wraps=...)` as suggested.

tests/unit/crawlers/_basic/test_basic_crawler.py
```diff
@@ -274,6 +274,98 @@ async def error_handler(context: BasicCrawlingContext, error: Exception) -> None
...
+
+    @crawler.router.default_handler
+    async def handler(context: BasicCrawlingContext) -> None:
+        if '|recovered' in context.request.unique_key:
+            return
+        raise SessionError('blocked')
+
+    @crawler.error_handler
+    async def error_handler(context: BasicCrawlingContext, error: Exception) -> Request | None:
+        assert isinstance(error, SessionError)
```
Mantisus: This `assert` is caught by the crawler and re-raised as `UserDefinedErrorHandlerError`.

Mantisus (approved):
One nit, and please update the PR description to match the current state. Otherwise LGTM, thanks for contributing!

Ayush7614:
Thanks @Mantisus — addressed in `20ccb2b`:

- removed the `assert isinstance(error, SessionError)` from the error_handler test
- updated the PR description to match the current scope

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1474,22 +1474,37 @@ async def __run_task_function(self) -> None:
...
-                await self._error_handler(context, session_error)
+                try:
+                    new_request = await self._error_handler(context, session_error)
+                except Exception as e:
+                    raise UserDefinedErrorHandlerError('Exception thrown in user-defined request error handler') from e
 
             if self._should_retry_request(context, session_error):
+                # Replacement requests are only honored while rotations remain, so exhausted sessions
+                # still go through failed_request_handler instead of being silently replaced.
+                if new_request is not None and new_request != request:
```
vdusek: This is a bug.

Comparison `new_request != request` is a field-wise pydantic comparison, so a freshly built `Request` for the same URL is always "different" - and the new `request.state = RequestState.ERROR_HANDLER` on line 1477 guarantees the mismatch. But `add_request` is a no-op while that `unique_key` is in progress, so line 1492 marks the original handled and the request vanishes: no rotation, no `failed_request_handler`, no statistics entry.

Reproduced with an `error_handler` returning `Request.from_url(context.request.url)` and `max_session_rotations=3`: this branch gives `handler_calls=1, failed_calls=0, requests_failed=0`, `master` gives `handler_calls=3, failed_calls=1, requests_failed=1`. Suggest treating `new_request.unique_key == request.unique_key` as "no replacement" and falling through to rotation, or inspecting `ProcessedRequest.was_already_present` from `add_request` and reclaiming instead of marking handled.

`master` already behaves this way in `_handle_request_retries` (lines 1172-1181), so this PR is not inventing the hazard - it extends it to the `SessionError` path, which previously ignored the return value and therefore rotated safely. Worth guarding in both places.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1474,22 +1474,37 @@ async def __run_task_function(self) -> None:
             if not session:
                 raise RuntimeError('SessionError raised in a crawling context without a session') from session_error
 
+            request.state = RequestState.ERROR_HANDLER
+
+            new_request = None
             if self._error_handler:
-                await self._error_handler(context, session_error)
+                try:
```
vdusek: Possible test coverage gap.

On `master` a raising `error_handler` propagated the raw user exception out of `__run_task_function`; now `crawler.run()` raises `UserDefinedErrorHandlerError`. That is the right consistency fix, but nothing covers it - `test_handles_error_in_error_handler` only exercises the non-session path - and the PR description does not mention the changed exception type. A `pytest.raises(UserDefinedErrorHandlerError)` test with a `SessionError` plus a raising handler is about ten lines.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1474,22 +1474,37 @@ async def __run_task_function(self) -> None:
...
+                except Exception as e:
+                    raise UserDefinedErrorHandlerError('Exception thrown in user-defined request error handler') from e
 
             if self._should_retry_request(context, session_error):
+                # Replacement requests are only honored while rotations remain, so exhausted sessions
+                # still go through failed_request_handler instead of being silently replaced.
+                if new_request is not None and new_request != request:
+                    await self._statistics.error_tracker_retry.add(error=session_error, context=context)
+                    await request_manager.add_request(new_request)
+                    await self._mark_request_as_handled(request)
```
vdusek: The replaced original received `record_request_processing_start` but never `..._finish` or `..._failure`, so it leaves a stale entry in `Statistics._requests_in_progress` and is missing from `requests_finished`/`requests_failed` - and therefore from `requests_total`, which also feeds `max_requests_per_crawl`. The same gap exists in the pre-existing sibling path, so this is consistency rather than a new defect.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1474,22 +1474,37 @@ async def __run_task_function(self) -> None:
...
+                    new_request = await self._error_handler(context, session_error)
+                except Exception as e:
+                    raise UserDefinedErrorHandlerError('Exception thrown in user-defined request error handler') from e
 
             if self._should_retry_request(context, session_error):
+                # Replacement requests are only honored while rotations remain, so exhausted sessions
+                # still go through failed_request_handler instead of being silently replaced.
+                if new_request is not None and new_request != request:
+                    await self._statistics.error_tracker_retry.add(error=session_error, context=context)
+                    await request_manager.add_request(new_request)
```
vdusek: Method `add_request(new_request)` ignores `new_request.forefront`, so in a PR whose stated goal is honoring `forefront`, a replacement request asking for the forefront still lands at the back. The replacement also inherits neither `session_rotation_count` nor `retry_count`, so an `error_handler` minting a fresh `unique_key` per failure can loop indefinitely against a permanently blocked site.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1474,22 +1474,37 @@ async def __run_task_function(self) -> None:
             if not session:
                 raise RuntimeError('SessionError raised in a crawling context without a session') from session_error
 
+            request.state = RequestState.ERROR_HANDLER
+
+            new_request = None
```
vdusek: Here is a space for a small shared helper.

This error-handler-plus-replacement block is a near-verbatim copy of `_handle_request_retries` (lines 1172-1181) that already diverges - `error_tracker_retry` vs `error_tracker`, an extra `session.retire()`, no `retry_count` increment. Two copies of this will drift; a small shared helper would keep the semantics in one place.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1474,22 +1474,37 @@ async def __run_task_function(self) -> None:
...
             if self._error_handler:
-                await self._error_handler(context, session_error)
+                try:
+                    new_request = await self._error_handler(context, session_error)
+                except Exception as e:
+                    raise UserDefinedErrorHandlerError('Exception thrown in user-defined request error handler') from e
```
vdusek: Possible inconsistency.

When the handler raises here, `request.state` stays `ERROR_HANDLER`, whereas `_handle_request_error` sets `RequestState.ERROR` before re-raising the same error. Same event, two persisted states.

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1474,22 +1474,37 @@ async def __run_task_function(self) -> None:
...
 
             if self._should_retry_request(context, session_error):
+                # Replacement requests are only honored while rotations remain, so exhausted sessions
+                # still go through failed_request_handler instead of being silently replaced.
```
vdusek: **Nit:** it is exhausted *rotations*, not exhausted sessions - the session is retired either way.

```suggestion
                # Replacement requests are only honored while rotations remain, so exhausted rotations
                # still go through failed_request_handler instead of being silently replaced.
```

src/crawlee/crawlers/_basic/_basic_crawler.py
```diff
@@ -1474,22 +1474,37 @@ async def __run_task_function(self) -> None:
...
+                if new_request is not None and new_request != request:
+                    await self._statistics.error_tracker_retry.add(error=session_error, context=context)
+                    await request_manager.add_request(new_request)
+                    await self._mark_request_as_handled(request)
+                    session.retire()
```
vdusek: **Nit:** `session.retire()` sits after two awaits that can raise; if either does, a known-blocked session stays in the pool. The rotation branch below retires first.

tests/unit/crawlers/_basic/test_basic_crawler.py
```diff
@@ -274,6 +274,97 @@ async def error_handler(context: BasicCrawlingContext, error: Exception) -> None
...
+    async def handler(context: BasicCrawlingContext) -> None:
+        context.request.forefront = True
+        raise RuntimeError('Arbitrary crash for testing purposes')
+
+    with patch.object(queue, 'reclaim_request', wraps=queue.reclaim_request) as reclaim_mock:
+        await crawler.run(['https://a.placeholder.com'])
+
+    reclaim_mock.assert_awaited_once()
+
+    (reclaimed_request,), reclaim_kwargs = reclaim_mock.await_args_list[0]
```
vdusek: **Nit:** the manual `await_args_list` unpacking does by hand what `reclaim_mock.assert_awaited_once_with(ANY, forefront=True)` states directly (`ANY` from `unittest.mock`, added to the import on line 18) - or `reclaim_mock.await_args.kwargs` if you want to keep the URL assertion.

tests/unit/crawlers/_basic/test_basic_crawler.py
```diff
@@ -274,6 +274,97 @@ async def error_handler(context: BasicCrawlingContext, error: Exception) -> None
...
+    original_request = await queue.get_request(request.unique_key)
+    recovered_request = await queue.get_request(f'{request.unique_key}|recovered')
+
+    assert original_request is not None
+    assert original_request.was_already_handled
+    assert recovered_request is not None
+    assert recovered_request.state == RequestState.DONE
+    assert recovered_request.was_already_handled
+
+    await queue.drop()
```
vdusek: **Nit:** `await queue.drop()` is unnecessary here and in the two tests below (lines 343 and 365) - the autouse `_isolate_test_environment` fixture in `tests/unit/conftest.py` gives each test a fresh `CRAWLEE_STORAGE_DIR` and clears the storage instance cache. Most `RequestQueue.open()` tests in this file skip it.

vdusek (changes requested):

Ayush7614:
Thanks @vdusek 

Addressed in a138011:

- Treat `new_request.unique_key == request.unique_key` as no replacement (same for the normal retry path via a shared `_handle_error_handler_replacement` helper).
- Helper inherits `retry_count` / `session_rotation_count`, uses `add_request(..., forefront=)`, calls `record_request_processing_finish` on the replaced original, sets `RequestState.ERROR` when the handler raises, and retires the session before awaits on the SessionError path.
- Added `test_session_error_handler_same_unique_key_rotates` and `test_session_error_handler_exception_is_wrapped`.
- Simplified the forefront reclaim assertion and dropped unnecessary `queue.drop()` calls.

vdusek (approved):
LGTM

PR #2089, opened 2026-07-24
docs: document Playwright browser plugin extensions

shixi-li:
### Description

The Playwright crawler guide linked a custom implementation without explaining the extension contract, making it unclear when configuration is sufficient and when a different launch API needs a plugin subclass.

- distinguish built-in configuration from custom browser launch integrations
- document the `new_browser()` controller contract and `BrowserPool` lifecycle responsibilities
- call out context, incognito, page-limit, fingerprint, header, and user-data behavior that custom integrations must preserve or replace
- link the existing Camoufox implementation as a complete example

### Issues

Refs https://github.com/apify/crawlee-python/issues/1936

### Testing

- `uv run --locked poe lint`
- `cd website && corepack pnpm@11.15.1 format:check`
- `cd website && corepack pnpm@11.15.1 lint:code`
- `cd website && corepack pnpm@11.15.1 lint:md`
- `cd website && uv run --locked corepack pnpm@11.15.1 build`

### Checklist

- [ ] CI passed

Issue #1936, opened 2026-06-03
docs: Add guide on building custom extensions (e.g. browser plugins)


vdusek:
## Context

In https://github.com/apify/crawlee-python/pull/1921 (and discussion https://github.com/apify/crawlee-python/discussions/1920), a third-party stealth-browser integration example was proposed for the "Avoid getting blocked" guide. We decided not to feature fresh third-party tools directly in our docs, since a docs example is effectively an endorsement and we want to keep the bar at proven maturity and adoption.

Internal discussion: https://apify.slack.com/archives/C06ENBV1B9Q/p1779877268101609

## Proposal

- Add a docs page describing how to build custom extensions for Crawlee — e.g. subclassing `PlaywrightBrowserPlugin` to integrate alternative browser backends, custom HTTP clients, etc.
- Third-party projects can then host their own integration guides on their side and we can link to selected ones from this page, instead of maintaining tool-specific examples in our docs.

vedika333:
"Hi, I'd love to work on this! I build Apify Actors and am familiar with PlaywrightBrowserPlugin. I can write the guide covering subclassing, a working code example, and the wiring with BrowserPool. Could you assign this to me or point me to the docs folder structure I should follow?"

vdusek:
Hi @vedika333, we do not assign issues to the community. However, you can try to come up with something. Currently, there are 4 extension points in Crawlee: Crawlers, HTTP clients, Storage clients, and Browser plugins. With the current docs structure, the part about extending the browser plugins should probably live in "Playwright crawler" guide.

PR #2106, opened 2026-08-01
docs: add guide on building custom extensions

shixi-li:
### Description

Refs #1936. Adds `docs/guides/extending_crawlee.mdx`, a page that maps Crawlee's extension points.

The four extension points you listed in the issue each already document their own contract in the guide that owns them: crawlers in the [HTTP crawlers guide](https://crawlee.dev/python/docs/guides/http-crawlers), HTTP clients and storage clients in theirs, and browser plugins in the Playwright crawler guide (#2089). What is still missing is the map, so someone who wants to extend Crawlee has to already know which guide to open, and a third-party project has no single page to point its users at for "here is the interface this integration implements".

This page is that map. For each extension point it states when to subclass rather than configure, names the base class and what its contract covers, and links to the guide that goes deep. It deliberately does not restate those guides.

### Contents

- **Extension points** — a class diagram plus one section each for crawlers (`AbstractHttpCrawler`), HTTP clients (`HttpClient`), storage clients (`StorageClient`), and browser plugins (`PlaywrightBrowserPlugin`), each linking to its detailed guide.
- **Choosing an extension point** — matches the layer to what actually differs in an integration, and says to prefer configuration over a subclass where the built-in class already exposes the knob.
- Framing for third-party integrations, per the issue's rationale: they host their own guide and reference the contract here.

### Notes

- Placed as its own guide rather than a section inside an existing one, since it spans all four extension points. The sidebar picks it up automatically (`autogenerated` over `docs/guides`).
- No new code examples: each linked guide already carries a runnable one for its own extension point, so duplicating them here would mean two copies to keep in sync.
- Happy to adjust the placement, depth, or wording, especially if you would rather this live under a different section or carry a worked example of its own.

### Testing

- `uv run poe build-docs` — build succeeds, the page renders at `/docs/next/guides/extending-crawlee`, and the build reports no broken links or anchors for it (the broken anchors in the log are pre-existing ones on API pages).
- Verified every link target resolves before building: the five doc ids (`http-crawlers`, `http-clients`, `storage-clients`, `playwright-crawler`, `architecture-overview`), the `playwright-crawler-with-camoufox` example, and each ApiLink class (`AbstractHttpCrawler`, `BasicCrawler`, `HttpClient`, `StorageClient`, `PlaywrightBrowserPlugin`, `PlaywrightBrowserController`, `BrowserPool`) against `src/crawlee/`.

### AI assistance

Written with AI assistance (Claude). The scoping decision was mine to check first what each extension point already documents, which is why this is a map rather than four new sections; the browser-plugin quarter was covered by my earlier #2089.

### DIFF
--- a/docs/guides/extending_crawlee.mdx
+++ b/docs/guides/extending_crawlee.mdx
@@ -1,18 +1,18 @@
 ---
 id: extending-crawlee
 title: Extending Crawlee
-description: Learn which parts of Crawlee are designed to be extended, what contract each extension point defines, and where to find the detailed guide for each one.
+description: The extension points Crawlee exposes, the contract each one defines, and how to choose between them.
 ---
 
 import ApiLink from '@site/src/components/ApiLink';
 
-Crawlee is built around a small number of abstract base classes that you can subclass to plug in your own behavior. This guide is the map: it lists every extension point, states the contract each one defines, and links to the guide that covers it in depth.
+Crawlee covers the common cases out of the box, but sooner or later you'll hit something it doesn't do: a parser for a format no built-in crawler understands, an HTTP backend your company mandates, a database you want your data in, or a browser that isn't launched the standard Playwright way. Rather than forking, you can plug your own implementation into the layer that differs and keep everything else.
 
-If you maintain a third-party integration, such as an alternative browser backend or a storage adapter, you can build it against these contracts and host the integration guide in your own project. This page gives your users a stable reference for the interface your integration implements.
+This guide is the map. It covers the extension points, the contract each one defines, and links to the guide that goes deeper on it. If you maintain a third-party integration, you can build it against these contracts and host your own guide for it.
 
 ## Extension points
 
-Crawlee currently has four extension points.
+The four extension points below are the main ones, and they're where most integrations plug in. They aren't the only abstract classes you can subclass - <ApiLink to="class/RequestLoader">`RequestLoader`</ApiLink>, <ApiLink to="class/FingerprintGenerator">`FingerprintGenerator`</ApiLink>, and <ApiLink to="class/RenderingTypePredictor">`RenderingTypePredictor`</ApiLink> are extensible too.
 
 ```mermaid
 ---
@@ -27,6 +27,10 @@ class BasicCrawler {
     <<abstract>>
 }
 
+class AbstractHttpCrawler {
+    <<abstract>>
+}
+
 class PlaywrightCrawler
 
 class HttpClient {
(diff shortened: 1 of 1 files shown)


docs/guides/extending_crawlee.mdx
```diff
@@ -0,0 +1,92 @@
...
+    <<abstract>>
+}
+
+class BrowserPlugin {
+    <<abstract>>
+}
+
+BasicCrawler --> HttpClient : uses
+BasicCrawler --> StorageClient : uses
+BasicCrawler --> BrowserPlugin : uses
```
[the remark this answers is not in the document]
shixi-li: Good catch — fixed in 83a258b. `BasicCrawler` itself has no browser knowledge; the diagram now shows `PlaywrightCrawler` (a `BasicCrawler` subclass) using `BrowserPool`, which initializes and manages the `BrowserPlugin` instances, matching the description in the Browser plugins section below.

docs/guides/extending_crawlee.mdx
```diff
@@ -0,0 +1,98 @@
...
+id: extending-crawlee
+title: Extending Crawlee
+description: Learn which parts of Crawlee are designed to be extended, what contract each extension point defines, and where to find the detailed guide for each one.
+---
+
+import ApiLink from '@site/src/components/ApiLink';
+
+Crawlee is built around a small number of abstract base classes that you can subclass to plug in your own behavior. This guide is the map: it lists every extension point, states the contract each one defines, and links to the guide that covers it in depth.
+
+If you maintain a third-party integration, such as an alternative browser backend or a storage adapter, you can build it against these contracts and host the integration guide in your own project. This page gives your users a stable reference for the interface your integration implements.
```
Mantisus: Hard to understand: "you" is the integration author, then "your users" are someone else, and there are three possessives in a row. I'd also drop "stable", that's a promise of our versioning policy, not of this page.

docs/guides/extending_crawlee.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+## Extension points
+
+Crawlee currently has four extension points.
```
Mantisus: This reads as exhaustive, and there are other ABCs users do subclass, e.g. `RequestLoader`, `FingerprintGenerator`, `RenderingTypePredictor`. These four are the main ones but not the only ones.

docs/guides/extending_crawlee.mdx
```diff
@@ -0,0 +1,98 @@
+---
+id: extending-crawlee
+title: Extending Crawlee
+description: Learn which parts of Crawlee are designed to be extended, what contract each extension point defines, and where to find the detailed guide for each one.
+---
+
+import ApiLink from '@site/src/components/ApiLink';
+
+Crawlee is built around a small number of abstract base classes that you can subclass to plug in your own behavior. This guide is the map: it lists every extension point, states the contract each one defines, and links to the guide that covers it in depth.
```
Mantisus: The intro starts with the mechanics, but it never says why the reader is here in the first place. I think it would help to lead with the motivation.

docs/guides/extending_crawlee.mdx
```diff
@@ -0,0 +1,98 @@
...
+class BrowserPlugin {
+    <<abstract>>
+}
+
+BasicCrawler --> HttpClient : uses
+BasicCrawler --> StorageClient : uses
+BasicCrawler --|> PlaywrightCrawler
+PlaywrightCrawler --> BrowserPool : uses
+BrowserPool --> BrowserPlugin : manages
+```
```
Mantisus, line 111: Three of the four sections describe a relation the diagram doesn't have: `AbstractHttpCrawler`, the per-storage clients `StorageClient` opens, and the controller `new_browser()` returns.

The diagram shows relations between components, but nothing about what is extensible, so it doesn't add any information for the reader.

docs/guides/extending_crawlee.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+### Crawlers
+
+Subclass a crawler when you need a parsing strategy or a request-handler context that the built-in crawlers do not provide.
+
+For HTTP-based crawling, <ApiLink to="class/AbstractHttpCrawler">`AbstractHttpCrawler`</ApiLink> is the base class. A custom crawler supplies a parser that turns an HTTP response into your parsed type, a context type that exposes that parsed data to request handlers, and the crawler class that ties the two together. Everything else, including retries, concurrency, session management, and storage, is inherited from <ApiLink to="class/BasicCrawler">`BasicCrawler`</ApiLink>.
+
+See [HTTP crawlers guide](./http-crawlers) for a worked example built on `selectolax`, and [Architecture overview](./architecture-overview) for how crawlers relate to the other components.
```
Mantisus: I'd expect the section to start with what the component is and what it does, then move on to what you subclass. It opens with "subclass a crawler when...", so it assumes the reader already knows.

The section is focused on the HTTP crawler and parsers, and `BasicCrawler` is mentioned only in passing, though it's the key class for building crawlers.

docs/guides/extending_crawlee.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+See [HTTP clients guide](./http-clients) for the full interface and the built-in implementations.
+
+### Storage clients
+
+Subclass a storage client when you want Crawlee's storages to be backed by a system that is not covered by the built-in memory, file system, SQL, and Redis clients.
+
+<ApiLink to="class/StorageClient">`StorageClient`</ApiLink> is the base class. It is a factory: it opens the per-storage clients for datasets, key-value stores, and request queues, and those clients implement the actual create, read, update, and delete operations.
+
+See [Storage clients guide](./storage-clients) for the interface, a custom client example, and how clients are registered and resolved.
```
Mantisus: Same as with the crawlers, I'd expect the section to start with what the component is and what it does, then move on to what you subclass.

he built-in clients are enumerated in prose, they should be `ApiLink`s to the corresponding classes. Same on L73: `DatasetClient`, `KeyValueStoreClient` and `RequestQueueClient` have neither links nor descriptions, though that's where the actual work is. `StorageClient` itself is only three methods that return them.

Mantisus (changes requested):
Thanks for the PR and the contribution. The direction is right, but the page needs further refinement.

In its current state the guide doesn't match the documentation style of the project, and that's what most of my comments are about.

A few things that run across the page rather than sitting in one place:

- Sections open with "subclass X when...", so they assume the reader already knows what the component is. Our guides describe the component first, then get to the details.
- Class names are given in prose instead of `ApiLink`, which we use everywhere else.
- The contracts aren't named anywhere, even though the page description promises them.
- Style details: we use contractions, we don't glue independent clauses with a semicolon, and we avoid both repeating the same opening across parallel list items and using bold to carry meaning.

The rest is inline. I think this needs another pass over the structure rather than line edits, happy to take another look once it's reworked.

shixi-li:
Thanks for the detailed review — that was exactly the kind of feedback the page needed. Reworked the structure rather than line-editing, as you suggested. Every section now opens with what the component is and what it does, then names its contract, then gets to what you subclass.

Point by point:

**Intro (L9/L11).** Now leads with the motivation — the cases where Crawlee doesn't do what you need and forking is the wrong answer — before any mechanics. Dropped the possessive pile-up and the "stable" promise; the third-party paragraph is one sentence and no longer mixes "you" and "your users".

**"Four extension points" (L15).** Reworded to say these are the main ones and explicitly names `RequestLoader`, `FingerprintGenerator`, and `RenderingTypePredictor` as also extensible.

**The diagram (L51).** Redrawn to show what the sections actually describe: `AbstractHttpCrawler` and `PlaywrightCrawler` under `BasicCrawler`, the three per-storage clients `StorageClient` opens, and the `BrowserController` that `new_browser()` returns. The extensible classes carry `<<abstract>>`, and the contract methods that matter (`create_dataset_client` / `create_kvs_client` / `create_rq_client`, `new_browser`) are on the boxes, so the diagram now says something about extensibility rather than just relations.

**Crawlers (L59).** Starts from `BasicCrawler` and what it owns (queue, fetch, context, retries, concurrency, sessions, storage) and why that makes it the base. The parser contract is named method by method: `parse`, `parse_text`, `select`, `is_matching_selector`, `find_links`.

**Storage clients (L75).** Starts with what a storage client is and why swapping it is useful. `StorageClient` is described as the three factory methods it is, and `DatasetClient`, `KeyValueStoreClient`, `RequestQueueClient` now have `ApiLink`s plus a line each on what they actually do, since that's where the work is. Built-in clients are `ApiLink`s throughout.

**Contracts.** Each section now names its contract explicitly — that was promised in the page description and missing before.

**Style.** Class names go through `ApiLink` everywhere. Contractions used. No semicolons joining independent clauses. No bold carrying meaning. The "Choosing an extension point" list items no longer share an opening — each starts from the thing that differs in your integration.

Verified locally: all 23 `ApiLink` targets resolve to real classes in `src/crawlee`, the Mermaid diagram renders, and every internal doc link points at an existing page.

docs/guides/extending_crawlee.mdx
```diff
@@ -0,0 +1,129 @@
...
+BasicCrawler --|> PlaywrightCrawler
+AbstractHttpCrawler --> HttpClient : uses
+BasicCrawler --> StorageClient : uses
+StorageClient --> DatasetClient : opens
+StorageClient --> KeyValueStoreClient : opens
+StorageClient --> RequestQueueClient : opens
+PlaywrightCrawler --> BrowserPool : uses
+BrowserPool --> BrowserPlugin : manages
+BrowserPlugin --> BrowserController : returns
+```
```
Mantisus, line 111: The diagram is inconsistent. You are specifying methods only for `StorageClient` and `BrowserPlugin`.
Also, `HttpClient` is actually implemented in `BasicCrawler`, not in `AbstractHttpCrawler`.
The text mentions `AbstractHttpParser`, but it is not shown in the diagram.

docs/guides/extending_crawlee.mdx
```diff
@@ -0,0 +1,129 @@
...
+- The response format is one no built-in crawler parses - subclass <ApiLink to="class/AbstractHttpCrawler">`AbstractHttpCrawler`</ApiLink> with your own parser.
+- The transport differs, but parsing doesn't - implement <ApiLink to="class/HttpClient">`HttpClient`</ApiLink> and pass it to any HTTP crawler.
+- Data needs to live somewhere Crawlee doesn't support yet - implement <ApiLink to="class/StorageClient">`StorageClient`</ApiLink> and its three per-storage clients.
+- Browsers need to be launched through a different API - implement <ApiLink to="class/BrowserPlugin">`BrowserPlugin`</ApiLink> and hand it to <ApiLink to="class/BrowserPool">`BrowserPool`</ApiLink>.
+
+When more than one fits, pick the narrowest. A custom HTTP client works with every HTTP crawler, so it's easier to maintain than a crawler subclass that hard-codes the same transport.
+
+## Conclusion
+
+Each extension point is a small abstract class with a documented contract, and everything above it keeps working once you implement one. If you're building a third-party integration, these contracts are also the stable surface to write your own documentation against.
```
Mantisus: "stable surface" on its own is just an assertion. If we say it, it should say what backs it: the contracts are these abstract classes, and they only change with a major release.

docs/guides/extending_crawlee.mdx
```diff
@@ -0,0 +1,129 @@
...
+
+If anything here is unclear or you hit a case the contracts don't cover, open an issue on [GitHub](https://github.com/apify/crawlee-python) or join our [Discord](https://discord.com/invite/jyEM2PRvMU).
```
Mantisus:
```suggestion
If you have questions or need assistance, feel free to reach out on our [GitHub](https://github.com/apify/crawlee-python) or join our [Discord community](https://discord.com/invite/jyEM2PRvMU). Happy scraping!
```

docs/guides/extending_crawlee.mdx
```diff
@@ -0,0 +1,129 @@
...
+
+import ApiLink from '@site/src/components/ApiLink';
+
+Crawlee covers the common cases out of the box, but sooner or later you'll hit something it doesn't do: a parser for a format no built-in crawler understands, an HTTP backend your company mandates, a database you want your data in, or a browser that isn't launched the standard Playwright way. Rather than forking, you can plug your own implementation into the layer that differs and keep everything else.
+
+This guide is the map. It covers the extension points, the contract each one defines, and links to the guide that goes deeper on it. If you maintain a third-party integration, you can build it against these contracts and host your own guide for it.
+
+## Extension points
+
+The four extension points below are the main ones, and they're where most integrations plug in. They aren't the only abstract classes you can subclass - <ApiLink to="class/RequestLoader">`RequestLoader`</ApiLink>, <ApiLink to="class/FingerprintGenerator">`FingerprintGenerator`</ApiLink>, and <ApiLink to="class/RenderingTypePredictor">`RenderingTypePredictor`</ApiLink> are extensible too.
```
Mantisus: The first half says these aren't the only ones, then the list itself reads closed: three items joined with "and", nothing marking them as examples. `AbstractHttpParser` and `BrowserController` are subclassable too and aren't there, so it can't be read as complete. I'd expect something that presents them as examples.

docs/guides/extending_crawlee.mdx
```diff
@@ -0,0 +1,129 @@
...
+
+For HTTP-based crawling, <ApiLink to="class/AbstractHttpCrawler">`AbstractHttpCrawler`</ApiLink> adds the fetch-and-parse layer on top. Extending it means supplying a parser that implements <ApiLink to="class/AbstractHttpParser">`AbstractHttpParser`</ApiLink>: `parse` turns an <ApiLink to="class/HttpResponse">`HttpResponse`</ApiLink> into your parsed type, `parse_text` does the same for a string, `select` and `is_matching_selector` apply selectors to it, and `find_links` extracts the URLs used for link enqueuing. You then pair that parser with a context type that exposes the parsed data to handlers, and a crawler class that ties the two together.
+
+See the [HTTP crawlers guide](./http-crawlers) for a worked example built on `selectolax`, and the [Architecture overview](./architecture-overview) for how crawlers relate to the other components.
```
Mantisus: Browser crawlers still aren't mentioned. The section goes from `BasicCrawler` straight to `AbstractHttpCrawler`, and the diagram shows `PlaywrightCrawler` under `BasicCrawler` with nothing in the text about it. Extending a browser crawler is a real case, we do it ourselves with `StagehandCrawler`.

docs/guides/extending_crawlee.mdx
```diff
@@ -0,0 +1,129 @@
...
+
+See the [Storage clients guide](./storage-clients) for the built-in implementations and a custom client example.
+
+### Browser plugins
+
+A browser plugin is what launches browsers for <ApiLink to="class/PlaywrightCrawler">`PlaywrightCrawler`</ApiLink>. The crawler never launches one itself: it goes through <ApiLink to="class/BrowserPool">`BrowserPool`</ApiLink>, which initializes the plugins it's given, forwards browser context options when creating pages, and manages each browser's lifecycle.
+
+The contract is <ApiLink to="class/BrowserPlugin">`BrowserPlugin`</ApiLink>. Its `new_browser` launches a browser and returns a <ApiLink to="class/BrowserController">`BrowserController`</ApiLink>, which is what the pool then drives to open pages and tear things down. Reach for a subclass when the launch path itself differs from the standard Playwright one, since <ApiLink to="class/PlaywrightBrowserPlugin">`PlaywrightBrowserPlugin`</ApiLink>'s configuration options already cover the cases where it doesn't.
```
Mantisus: I'd keep both levels here. `BrowserPlugin` is the contract when the browser's launch and lifecycle are too specific for `PlaywrightBrowserPlugin` to cover, and `PlaywrightBrowserPlugin` is what most integrations should extend. Right now the page names only the base, so the common case is missing and it isn't clear why the two exist.

docs/guides/extending_crawlee.mdx
```diff
@@ -0,0 +1,129 @@
...
+## Choosing an extension point
+
+Match the layer to what actually differs in your integration.
+
+- The response format is one no built-in crawler parses - subclass <ApiLink to="class/AbstractHttpCrawler">`AbstractHttpCrawler`</ApiLink> with your own parser.
+- The transport differs, but parsing doesn't - implement <ApiLink to="class/HttpClient">`HttpClient`</ApiLink> and pass it to any HTTP crawler.
+- Data needs to live somewhere Crawlee doesn't support yet - implement <ApiLink to="class/StorageClient">`StorageClient`</ApiLink> and its three per-storage clients.
+- Browsers need to be launched through a different API - implement <ApiLink to="class/BrowserPlugin">`BrowserPlugin`</ApiLink> and hand it to <ApiLink to="class/BrowserPool">`BrowserPool`</ApiLink>.
+
+When more than one fits, pick the narrowest. A custom HTTP client works with every HTTP crawler, so it's easier to maintain than a crawler subclass that hard-codes the same transport.
```
Mantisus, line 185: "pick the narrowest" is good, but the narrowest choice is often no subclass at all, and the page no longer says that. Configuration covers a lot of these cases: `PlaywrightBrowserPlugin`'s options, passing an `http_client` to any crawler, or parsing with a third-party library inside an `HttpCrawler` handler.

docs/guides/extending_crawlee.mdx
```diff
@@ -0,0 +1,129 @@
...
+See the [Playwright crawler guide](./playwright-crawler) for the responsibilities a subclass has to preserve, and the [Camoufox example](../examples/playwright-crawler-with-camoufox) for a complete integration.
+
+## Choosing an extension point
+
+Match the layer to what actually differs in your integration.
+
+- The response format is one no built-in crawler parses - subclass <ApiLink to="class/AbstractHttpCrawler">`AbstractHttpCrawler`</ApiLink> with your own parser.
+- The transport differs, but parsing doesn't - implement <ApiLink to="class/HttpClient">`HttpClient`</ApiLink> and pass it to any HTTP crawler.
+- Data needs to live somewhere Crawlee doesn't support yet - implement <ApiLink to="class/StorageClient">`StorageClient`</ApiLink> and its three per-storage clients.
+- Browsers need to be launched through a different API - implement <ApiLink to="class/BrowserPlugin">`BrowserPlugin`</ApiLink> and hand it to <ApiLink to="class/BrowserPool">`BrowserPool`</ApiLink>.
```
Mantisus: Two things beyond the style.
The first bullet answers only for HTTP, so someone who wants a custom browser crawler has no line here.
The conditions aren't parallel either: two describe a state, two describe a requirement.

Mantisus (changes requested):
Thanks for the rework, this is a big step up.

One thing that runs across the page. We don't use a spaced hyphen to join clauses. We split them into separate sentences instead.

docs/guides/extending_crawlee.mdx
shixi-li: Addressed in 1c16751c. Every checklist item now starts with a parallel `If ... needs to change` condition, and there is a separate browser-crawler choice for extending `PlaywrightCrawler`.

docs/guides/extending_crawlee.mdx
shixi-li: Reworked in 1c16751c. The prose now explicitly presents the additional classes as non-exhaustive examples, while the diagram includes the lower-level parser and controller contracts.

docs/guides/extending_crawlee.mdx
shixi-li: Updated in 1c16751c. The conclusion now ties the stable surface to the public abstract class contracts and states that those contracts change only with a major release.

docs/guides/extending_crawlee.mdx
shixi-li: Applied the suggested closing in 1c16751c.

docs/guides/extending_crawlee.mdx
shixi-li: Restored in 1c16751c. The decision section now starts with configuration and gives the `HttpCrawler` handler, `http_client`, and `PlaywrightBrowserPlugin` examples before any subclass choices.

docs/guides/extending_crawlee.mdx
shixi-li: Clarified in 1c16751c. The guide now recommends configuring `PlaywrightBrowserPlugin` first, extending it for the common custom Playwright path, and implementing `BrowserPlugin` directly only when launch and lifecycle need a different contract.

docs/guides/extending_crawlee.mdx
shixi-li: Added in 1c16751c. The crawler section now covers extending `PlaywrightCrawler`, uses `StagehandCrawler` as the concrete example, and distinguishes crawler-level changes from the narrower browser-plugin path.

docs/guides/extending_crawlee.mdx
shixi-li: Addressed in 1c16751c. The diagram now marks extension points and shows `AbstractHttpCrawler` using `AbstractHttpParser`, `StorageClient` opening each per-storage client, and `BrowserPlugin.new_browser()` returning `BrowserController`.

docs/guides/extending_crawlee.mdx
shixi-li: Addressed in 1c16751c. I removed the one-off method lists, moved the `HttpClient` relation to `BasicCrawler`, and added `AbstractHttpParser` to the diagram.

shixi-li:
@Mantisus, thanks for the second pass. I’ve rebased onto the latest `master` and pushed 1c16751c. The update redesigns the diagram around the actual extension contracts, adds the browser-crawler path, distinguishes `BrowserPlugin` from the common `PlaywrightBrowserPlugin` path, restores configure-before-subclass guidance, makes the checklist parallel, and applies the contract-stability, closing, and style changes.

Validation is green:
- `uv run poe build-docs`
- `pnpm lint:md`
- `pnpm format:check`

I’ve replied to each of the nine actionable threads. Could you take another look?

Mantisus (approved):
Thanks for the update. It looks good!

vdusek (approved):
LGTM, thank you

PR #2110, opened 2026-08-03
fix(fingerprints): enforce the user agent allow list so screen constraints are respected

### LINKED ISSUE #2108: Screen constraints are not respected with apify-fingerprint-datapoints 0.14.0
`ScreenOptions` are not respected - with `strict=True` fingerprint generation fails, otherwise the screen
constraints are silently dropped and the fingerprint gets a screen outside of the requested range.

`browserforge` passes the screen-compatible user agents to the header generator only as a hint, so the
generated user agent is often outside of them and the fingerprint network then finds no consistent sample.
The 0.14.0 datapoints made that deterministic, so the version is capped in `pyproject.toml` (#2107).

Fix: enforce the allow list in `PatchedHeaderGenerator` - pin a retry to an allowed user agent with the same
browser, operating system and device.

Failing tests: `test_fingerprint_generator_all_options`, `test_custom_fingerprint_uses_generator_options`.

## Acceptance

The `apify_fingerprint_datapoints<0.14.0` caps removed and the tests above passing with the latest version.

*✍️ Drafted by Claude Code*

vdusek:
`ScreenOptions` were not respected - with `strict=True` fingerprint generation failed, otherwise the screen
constraints were silently dropped and the fingerprint got a screen outside of the requested range.

`browserforge` turns the screen constraints into an allow list of user agents and passes it to the header
generator, but the generator only derives browser, operating system and device from it and then samples the
user agent freely - so the result is often outside the list, and the fingerprint network finds no consistent
sample. `apify-fingerprint-datapoints` 0.14.0 made that deterministic for some inputs, hence the cap in #2107.

`PatchedHeaderGenerator` now enforces the allow list: a user agent outside it pins the next attempt to an
allowed one with the same browser, operating system and device. If nothing is interchangeable or the attempts
run out, the wider header is returned, so inputs that worked before keep working.

Measured on 0.14.0, before -> after:

| Case | Before | After |
| --- | --- | --- |
| `strict` Firefox + Windows, screen `600-1800 x 400-1200` | 299/300 fail, 480 ms each | 0/500 fail, 90 ms each |
| Firefox + Android, screen `300-600 x 500-1200` | ~40% of screens out of range | 0/2000 out of range |

Also drops the `apify-fingerprint-datapoints<0.14.0` caps from #2107.

`test_fingerprint_generator_respects_screen_options_without_strict` covers the non-strict path; it and
`test_fingerprint_generator_all_options` fail on `master` with the 0.14.0 datapoints.

Closes: #2108

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/crawlee/fingerprint_suite/_browserforge_adapter.py
+++ b/src/crawlee/fingerprint_suite/_browserforge_adapter.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import random
 from collections.abc import Iterable
 from copy import deepcopy
-from functools import lru_cache, reduce
+from functools import cache, reduce
 from operator import or_
 from typing import TYPE_CHECKING, Any, Literal
 
@@ -292,7 +292,8 @@ def _pick_interchangeable_user_agent(generated_user_agent: str, allowed_user_age
     return random.choice(candidates) if candidates else None
 
 
-@lru_cache
+# Unbounded - only a few hundred user agents exist, and the default 128 entries would thrash on the allow list scan.
+@cache
 def _get_user_agent_traits(user_agent: str) -> tuple[frozenset[str], frozenset[str], frozenset[str]]:
     """Get browser names, operating systems and devices the header network links to the `user_agent`.
 
--- a/src/crawlee/fingerprint_suite/_browserforge_adapter.py
+++ b/src/crawlee/fingerprint_suite/_browserforge_adapter.py
@@ -3,12 +3,12 @@ from __future__ import annotations
 import random
 from collections.abc import Iterable
 from copy import deepcopy
-from functools import reduce
+from functools import lru_cache, reduce
 from operator import or_
 from typing import TYPE_CHECKING, Any, Literal
 
 import apify_fingerprint_datapoints
-from browserforge.bayesian_network import extract_json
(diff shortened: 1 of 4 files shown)


src/crawlee/fingerprint_suite/_browserforge_adapter.py
```diff
@@ -250,6 +278,43 @@ def generate(self, browser_type: SupportedBrowserType = 'chrome') -> dict[str, s
...
+
+    candidates = [
+        allowed_user_agent
+        for allowed_user_agent in allowed_user_agents
+        if _get_user_agent_traits(allowed_user_agent) == generated_traits
+    ]
+    return random.choice(candidates) if candidates else None
+
+
+@lru_cache
```
Mantisus: Won't the default size of `lru_cache` (128 items) be too small?

Mantisus (approved):
LGTM

src/crawlee/fingerprint_suite/_browserforge_adapter.py
vdusek: switched to unbounded `@cache` in

janbuchar (approved):

PR #2120, opened 2026-08-04
debt: remove async-timeout in favor of asyncio.timeout

### LINKED ISSUE #2052: Remove `async-timeout` dependency in favor of stdlib `asyncio.timeout`
## Summary

Drop the [`async-timeout`](https://github.com/aio-libs-abandoned/async-timeout) dependency (`async-timeout>=5.0.1` in `pyproject.toml`) and replace it with the standard library equivalent.

## Why

`async-timeout` is deprecated and no longer actively maintained. Its functionality has been upstreamed into Python 3.11+ as [`asyncio.timeout`](https://docs.python.org/3/library/asyncio-task.html#asyncio.timeout) and `asyncio.Timeout`. The upstream README states:

> This library has effectively been upstreamed into Python 3.11+. Therefore this library is considered deprecated and no longer actively supported.

Every dependency we can drop is one fewer thing to keep patched and audited, so removing an abandoned package is worth doing.

## Where it's used

The only usage is in `src/crawlee/_utils/time.py`:

```python
from async_timeout import Timeout, timeout
```

used by the `SharedTimeout` context manager.

## Blocker / plan

`asyncio.timeout` was added in Python 3.11, but Crawlee still supports Python 3.10 (`requires-python = ">=3.10"`). Options:

1. **Drop Python 3.10 support** and switch to `asyncio.timeout` / `asyncio.Timeout` directly — fits the 2.0 milestone.
2. Keep 3.10 support via a version-guarded import (`sys.version_info >= (3, 11)`) until 3.10 is dropped.

Targeting the 2.0 milestone since dropping Python 3.10 is the clean path.

RohithPariki:
Fixes #2052 

## Description
This PR drops the deprecated `async-timeout` dependency in favor of the standard library's `asyncio.timeout`, as the functionality was upstreamed into Python 3.11+. 

As discussed in the issue (and targeting the 2.0 milestone), this PR drops support for Python 3.10 to allow for the clean path of using the standard library directly without version-guarded imports.

## Changes Made
- **pyproject.toml**: Removed `async-timeout` from dependencies and bumped `requires-python` (and related classifiers) to `>=3.11`.
- **src/crawlee/_utils/time.py**: Replaced `async_timeout.timeout` with `asyncio.timeout` inside the `SharedTimeout` class. Retained the existing context manager logic as `asyncio.timeout` also serves as a drop-in async context manager in Python 3.11+.

## Verification
- Verified that `SharedTimeout` successfully catches operations exceeding the budget and properly propagates `TimeoutError`.
- Ran `ruff` to ensure formatting and linting rules are upheld.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,13 +10,12 @@ authors = [{ name = "Apify Technologies s.r.o.", email = "support@apify.com" }]
 license = "Apache-2.0"
 license-files = ["LICENSE"]
 readme = "README.md"
-requires-python = ">=3.10"
+requires-python = ">=3.11"
 classifiers = [
     "Development Status :: 5 - Production/Stable",
     "Environment :: Console",
     "Intended Audience :: Developers",
     "Operating System :: OS Independent",
-    "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
@@ -34,7 +33,6 @@ keywords = [
     "scraping",
 ]
 dependencies = [
-    "async-timeout>=5.0.1",
     "cachetools>=5.5.0",
     "colorama>=0.4.0",
     "impit>=0.8.0",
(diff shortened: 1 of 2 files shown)


vdusek:
breaking change, v2 milestone, closing

PR #2121, opened 2026-08-04
debt: use alias_generator=to_camel on Pydantic models

### LINKED ISSUE #1954: Use `alias_generator=to_camel` on Pydantic models instead of manual per-field aliases

RohithPariki:
### Description

This PR cleans up the manual camelCase aliasing across our Pydantic models. 

Instead of adding `Field(alias="camelCaseName")` everywhere, I've updated the `ConfigDict` on our models to use `alias_generator=to_camel` and `populate_by_name=True`. I then stripped out all the redundant `alias=` strings from the individual fields.

A few notes:
- I left specific internal overrides (like `alias='__crawlee'`) untouched so we don't break any core logic.
- Ran `ruff` to clean up the formatting after removing the fields.

### Issues

- Closes: #1954

### Testing

- Verified locally that Pydantic models with `alias_generator=to_camel` and `populate_by_name=True` correctly map incoming camelCase JSON keys into snake_case Python attributes, perfectly mimicking the behavior of the removed manual aliases.

### DIFF
--- a/src/crawlee/_request.py
+++ b/src/crawlee/_request.py
@@ -6,6 +6,7 @@ from enum import IntEnum
 from typing import TYPE_CHECKING, Annotated, Any, TypedDict, cast
 
 from pydantic import BaseModel, BeforeValidator, ConfigDict, Field, PlainSerializer, PlainValidator, TypeAdapter
+from pydantic.alias_generators import to_camel
 from yarl import URL
 
 from crawlee._types import EnqueueStrategy, HttpHeaders, HttpMethod, HttpPayload, JsonSerializable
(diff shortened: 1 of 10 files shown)


vdusek:
2.0 issue, closing

PR #2123, opened 2026-08-06
fix(sitemaps): validate HTTP status when fetching sitemaps

anxkhn:
### Description

- `_fetch_and_process_sitemap` never looked at `response.status_code`. It opened the stream, picked a parser from the content type, parsed whatever body came back, and then broke out of the retry loop. A 503 with an empty body was therefore accepted as a valid, empty sitemap.
- The practical effect is that a transient outage on a sitemap endpoint silently produces zero URLs. `SitemapRequestLoader` then marks the sitemap processed and reports itself finished, so the crawl looks successful while nothing was crawled. Per RFC 9110 section 15.6, a 5xx response means the server failed to fulfil the request, not that a representation was returned.
- This change validates the status before a parser is selected. 408, 429 and 5xx go through the retry loop that already exists in this function, and other non-success statuses are raised immediately without parsing the body.
- `SitemapRequestLoader.is_finished` now retrieves the result of the background loading task, so a load that failed is surfaced to the caller instead of being swallowed as a clean completion. Cancellation is still treated as before.

Behaviour change worth flagging: fetch failures already aborted `parse_sitemap` (a connection error after the retries are exhausted propagates today), and terminal statuses now join that class. That means a stale 404 entry inside a sitemap index will fail the load rather than being parsed as an empty sitemap. If you would prefer terminal statuses to be logged and skipped instead, I am happy to change it.

### Issues

- No related issue filed. Happy to open one if you prefer to track it there.

### Testing

- `tests/unit/_utils/test_sitemap.py`: two 503 responses followed by a valid sitemap are retried and the URLs are returned; a persistent 503 raises once the retries are exhausted; a 404 raises immediately without a retry and without parsing the body it carried.
- `tests/unit/request_loaders/test_sitemap_request_loader.py`: the loader recovers from transient 503s and loads the URLs, and it surfaces the failure instead of finishing empty when the retries are exhausted.
- The existing stream mocks in these files needed a `status_code` on the mocked response, since the code now reads it.
- Ran `uv run pytest tests/unit/_utils/test_sitemap.py tests/unit/request_loaders/test_sitemap_request_loader.py`.

### Checklist

- [ ] CI passed

### DIFF
--- a/src/crawlee/_utils/sitemap.py
+++ b/src/crawlee/_utils/sitemap.py
@@ -9,6 +9,7 @@ from contextlib import suppress
 from dataclasses import dataclass
 from datetime import datetime, timedelta
 from hashlib import sha256
+from http import HTTPStatus
 from logging import getLogger
 from typing import TYPE_CHECKING, Literal, TypedDict
 from xml.sax import SAXParseException
@@ -20,7 +21,9 @@ from yarl import URL
 
 from crawlee._utils.urls import filter_url
 from crawlee._utils.web import is_status_code_successful
-from crawlee.errors import ProxyError
+from crawlee.errors import HttpStatusCodeError, ProxyError
+
+_HTTP_STATUS_UPPER_BOUND = 600
 
 if TYPE_CHECKING:
     from collections.abc import AsyncGenerator
@@ -29,6 +32,12 @@ if TYPE_CHECKING:
     from crawlee import EnqueueStrategy
     from crawlee.http_clients import HttpClient
     from crawlee.proxy_configuration import ProxyInfo
+
+
+def _raise_for_sitemap_status(status_code: int) -> None:
+    if not is_status_code_successful(status_code):
+        raise HttpStatusCodeError('Error status code returned while fetching sitemap', status_code)
+
 
 logger = getLogger(__name__)
 
@@ -372,6 +381,8 @@ async def _fetch_and_process_sitemap(
             async with http_client.stream(
                 sitemap_url, method='GET', headers=SITEMAP_HEADERS, proxy_info=proxy_info, timeout=timeout
             ) as response:
+                _raise_for_sitemap_status(response.status_code)
+
                 # Determine content type and compression
                 content_type = response.headers.get('content-type', '')
 
@@ -455,6 +466,11 @@ async def _fetch_and_process_sitemap(
             break
 
         except Exception as e:
+            if isinstance(e, HttpStatusCodeError) and not (
+                e.status_code in (HTTPStatus.REQUEST_TIMEOUT, HTTPStatus.TOO_MANY_REQUESTS)
+                or HTTPStatus.INTERNAL_SERVER_ERROR <= e.status_code < _HTTP_STATUS_UPPER_BOUND
+            ):
+                raise
             if retries_left > 0:
                 logger.warning(f'Error fetching sitemap {sitemap_url}: {e}. Retries left: {retries_left}')
                 await asyncio.sleep(1)  # Brief pause before retry
--- a/src/crawlee/_utils/sitemap.py
+++ b/src/crawlee/_utils/sitemap.py
@@ -56,14 +56,21 @@ def _is_successful_sitemap_status(status_code: int) -> bool:
     return HTTPStatus.OK <= status_code < HTTPStatus.MULTIPLE_CHOICES
 
 
-def _is_retryable_sitemap_status(status_code: int) -> bool:
-    """Return whether a sitemap response status should be retried."""
-    return status_code == HTTPStatus.REQUEST_TIMEOUT or is_status_code_server_error(status_code)
+class _RetryableSitemapStatusError(Exception):
+    """Internal signal that a sitemap fetch returned a retryable HTTP status.
+
+    Raised inside the response context manager so the stream is closed before the retry delay.
+    """
+
+    def __init__(self, status_code: int) -> None:
+        super().__init__(f'HTTP status {status_code}')
+        self.status_code = status_code
 
 
-async def _sleep_before_sitemap_retry() -> None:
-    """Pause briefly before retrying a failed sitemap fetch."""
-    await asyncio.sleep(SITEMAP_RETRY_DELAY)
+def _raise_for_retryable_sitemap_status(status_code: int) -> None:
+    """Raise `_RetryableSitemapStatusError` for HTTP statuses that warrant a fetch retry (408 and 5xx)."""
+    if status_code == HTTPStatus.REQUEST_TIMEOUT or is_status_code_server_error(status_code):
+        raise _RetryableSitemapStatusError(status_code)
 
 
 @dataclass()
@@ -398,20 +405,10 @@ async def _fetch_and_process_sitemap(
             ) as response:
                 status_code = response.status_code
                 if not _is_successful_sitemap_status(status_code):
-                    if not _is_retryable_sitemap_status(status_code):
-                        logger.warning(f'Skipping sitemap {sitemap_url} due to HTTP status code {status_code}.')
-                        return
-                    if retries_left > 0:
-                        logger.warning(
-                            f'Error fetching sitemap {sitemap_url}: HTTP status {status_code}. '
-                            f'Retries left: {retries_left}'
-                        )
-                        await _sleep_before_sitemap_retry()
-                    else:
-                        logger.warning(
-                            f'Failed to fetch sitemap {sitemap_url}, no retries left: HTTP status {status_code}'
-                        )
-                    continue
+                    # Retryable statuses raise and route to the retry handler below; the rest skip this source.
+                    _raise_for_retryable_sitemap_status(status_code)
+                    logger.warning(f'Skipping sitemap {sitemap_url} due to HTTP status code {status_code}.')
+                    return
 
                 # Determine content type and compression
                 content_type = response.headers.get('content-type', '')
@@ -495,10 +492,18 @@ async def _fetch_and_process_sitemap(
                     parser.close()
             parsed_ok.append(True)
 
+        except _RetryableSitemapStatusError as e:
+            if retries_left > 0:
+                logger.warning(
+                    f'Error fetching sitemap {sitemap_url}: HTTP status {e.status_code}. Retries left: {retries_left}'
+                )
+                await asyncio.sleep(SITEMAP_RETRY_DELAY)
+            else:
+                logger.warning(f'Failed to fetch sitemap {sitemap_url}, no retries left: HTTP status {e.status_code}.')
         except Exception as e:
             if retries_left > 0:
                 logger.warning(f'Error fetching sitemap {sitemap_url}: {e}. Retries left: {retries_left}')
-                await _sleep_before_sitemap_retry()
+                await asyncio.sleep(SITEMAP_RETRY_DELAY)
             else:
                 logger.exception(f'Failed to fetch sitemap {sitemap_url}, no retries left.')
                 raise
@@ -564,6 +569,10 @@ async def parse_sitemap(
 
     Default `ParseSitemapOptions.enqueue_strategy` is `same-hostname` which will skip cross-host URLs.
     Use strategy `all` to process all links.
+
+    URL sources answering with a non-2xx status are skipped: retryable statuses (408 and 5xx) are retried first,
+    others are skipped immediately. If at least one source fails with a fetch error after all retries and no source
+    succeeds, the first such error is raised once all sources have been processed.
     """
     # Set default options
     options = options or {}
@@ -600,6 +609,8 @@ async def parse_sitemap(
                 enqueue_strategy=enqueue_strategy,
             ):
                 yield result
+            # Raw sources cannot fetch-fail, so they always count as successful for the all-sources-failed check.
+            successful_sources += 1
 
         elif source['type'] == 'url' and 'url' in source:
             # Add to visited set before processing to avoid duplicates
@@ -627,8 +638,8 @@ async def parse_sitemap(
                 if parsed_ok:
                     successful_sources += 1
             except Exception as e:
+                # Already logged by `_fetch_and_process_sitemap`; raised below only if no source succeeds.
                 source_errors.append(e)
-                logger.warning(f'Failed to process sitemap source {source["url"]}: {e}')
         else:
             logger.warning(f'Invalid source configuration: {source}')
 
--- a/src/crawlee/_utils/sitemap.py
+++ b/src/crawlee/_utils/sitemap.py
@@ -20,10 +20,8 @@ from typing_extensions import NotRequired, override
 from yarl import URL
 
 from crawlee._utils.urls import filter_url
-from crawlee._utils.web import is_status_code_successful
+from crawlee._utils.web import is_status_code_server_error, is_status_code_successful
 from crawlee.errors import HttpStatusCodeError, ProxyError
-
-_HTTP_STATUS_UPPER_BOUND = 600
 
 if TYPE_CHECKING:
     from collections.abc import AsyncGenerator
@@ -32,12 +30,6 @@ if TYPE_CHECKING:
     from crawlee import EnqueueStrategy
     from crawlee.http_clients import HttpClient
     from crawlee.proxy_configuration import ProxyInfo
-
-
-def _raise_for_sitemap_status(status_code: int) -> None:
-    if not is_status_code_successful(status_code):
-        raise HttpStatusCodeError('Error status code returned while fetching sitemap', status_code)
-
 
 logger = getLogger(__name__)
 
@@ -54,6 +46,21 @@ CONTENT_CHUNK_SIZE = 1024 * 1024
 
 DEFAULT_MAX_DEPTH = 10
 """Default maximum depth of nested sitemaps to follow, guarding against malicious infinite sitemap chains."""
+
+
+def _raise_for_sitemap_status(status_code: int) -> None:
+    """Raise `HttpStatusCodeError` if the sitemap response status is not 2xx."""
+    if not HTTPStatus.OK <= status_code < HTTPStatus.MULTIPLE_CHOICES:
+        raise HttpStatusCodeError('Error status code returned while fetching sitemap', status_code)
+
+
+def _is_retryable_sitemap_status(status_code: int) -> bool:
+    """Return whether a sitemap response status should be retried."""
+    return (
+        HTTPStatus.MULTIPLE_CHOICES <= status_code < HTTPStatus.BAD_REQUEST
+        or status_code in (HTTPStatus.REQUEST_TIMEOUT, HTTPStatus.TOO_MANY_REQUESTS)
+        or is_status_code_server_error(status_code)
+    )
 
 
 @dataclass()
@@ -470,14 +477,14 @@ async def _fetch_and_process_sitemap(
             break
 
         except Exception as e:
-            if isinstance(e, HttpStatusCodeError) and not (
-                e.status_code in (HTTPStatus.REQUEST_TIMEOUT, HTTPStatus.TOO_MANY_REQUESTS)
-                or HTTPStatus.INTERNAL_SERVER_ERROR <= e.status_code < _HTTP_STATUS_UPPER_BOUND
-            ):
-                raise
+            if isinstance(e, HttpStatusCodeError) and not _is_retryable_sitemap_status(e.status_code):
+                logger.warning(f'Skipping sitemap {sitemap_url} due to HTTP status code {e.status_code}.')
+                break
             if retries_left > 0:
                 logger.warning(f'Error fetching sitemap {sitemap_url}: {e}. Retries left: {retries_left}')
                 await asyncio.sleep(1)  # Brief pause before retry
+            elif isinstance(e, HttpStatusCodeError):
+                logger.warning(f'Failed to fetch sitemap {sitemap_url}, no retries left: {e}')
             else:
                 logger.exception(f'Failed to fetch sitemap {sitemap_url}, no retries left.')
                 raise
@@ -553,6 +560,8 @@ async def parse_sitemap(
     # Setup working state
     sources = list(initial_sources)
     visited_sitemap_urls: set[str] = set()
+    successful_sources = 0
+    source_errors: list[Exception] = []
 
     # Process sources until the queue is empty
     while sources:
@@ -575,6 +584,7 @@ async def parse_sitemap(
                 enqueue_strategy=enqueue_strategy,
             ):
                 yield result
+            successful_sources += 1
 
         elif source['type'] == 'url' and 'url' in source:
             # Add to visited set before processing to avoid duplicates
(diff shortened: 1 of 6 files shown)


Mantisus (changes requested):
> That means a stale 404 entry inside a sitemap index will fail the load rather than being parsed as an empty sitemap. If you would prefer terminal statuses to be logged and skipped instead, I am happy to change it.

Yes. I suggest:

- Server errors trigger a retry. Once all retries are exhausted, log a warning instead of raising.
- The same applies to 429 and 408.
- Other client errors skip the sitemap, treating it as empty.

Also, updating `is_finished` interrupts the loader's operation, preventing it from handing out URLs that have already been loaded. Checking `url_queue` and `in_progress` before retrieving the task result would be enough.

src/crawlee/_utils/sitemap.py
```diff
@@ -20,7 +21,9 @@
 
 from crawlee._utils.urls import filter_url
 from crawlee._utils.web import is_status_code_successful
-from crawlee.errors import ProxyError
+from crawlee.errors import HttpStatusCodeError, ProxyError
+
+_HTTP_STATUS_UPPER_BOUND = 600
```
vdusek: **Suggestion:** `is_status_code_server_error` in `crawlee._utils.web` already expresses this range, and this module imports `is_status_code_successful` from there already (`_basic_crawler.py:1619` and `robots.py:64` use it). Dropping the constant leaves the `HTTPStatus` import needed only for 408/429.

src/crawlee/_utils/sitemap.py
```diff
@@ -30,6 +33,12 @@
     from crawlee.http_clients import HttpClient
     from crawlee.proxy_configuration import ProxyInfo
 
+
+def _raise_for_sitemap_status(status_code: int) -> None:
```
vdusek: **Nit:** placement and a missing docstring. The file's order elsewhere is imports -> `TYPE_CHECKING` -> `logger` -> constants -> classes -> functions, but this helper sits between `TYPE_CHECKING` and `logger`, and `_HTTP_STATUS_UPPER_BOUND` above the `TYPE_CHECKING` block. Every other module-level function here (`_get_parser`, `_get_origin_url`, `_process_sitemap_item`) has a docstring.

```suggestion
def _raise_for_sitemap_status(status_code: int) -> None:
    """Raise `HttpStatusCodeError` if the sitemap response status is not successful."""
```

src/crawlee/_utils/sitemap.py
```diff
@@ -30,6 +33,12 @@
...
+    if not is_status_code_successful(status_code):
```
vdusek: **Note:** `is_status_code_successful` accepts 200-399, so a 3xx body that reaches the parser (redirects not followed) still yields nothing silently. crawlee JS requires 2xx, retries all non-2xx, and then warns and skips that source rather than aborting the parse.

src/crawlee/_utils/sitemap.py
```diff
@@ -372,6 +381,8 @@ async def _fetch_and_process_sitemap(
             async with http_client.stream(
                 sitemap_url, method='GET', headers=SITEMAP_HEADERS, proxy_info=proxy_info, timeout=timeout
             ) as response:
+                _raise_for_sitemap_status(response.status_code)
```
vdusek: **Important:** raising here aborts the entire `parse_sitemap` run, not just this source. The `while sources:` loop at `_utils/sitemap.py:582-594` has no per-source containment, so one dead sitemap discards the URLs already yielded from healthy ones - verified against the base commit for `Sitemap.try_common_names`, `RobotsTxtFile.parse_urls_from_sitemaps`, and the two-URL example in `docs/examples/code_examples/using_sitemap_request_loader.py`. Containing the error per source and raising only when every source failed keeps the fix and removes the regression; note `RobotsTxtFile.load` (`_utils/robots.py:57-76`) already degrades gracefully on 4xx/5xx.

src/crawlee/_utils/sitemap.py
```diff
@@ -455,6 +466,11 @@ async def _fetch_and_process_sitemap(
             break
 
         except Exception as e:
+            if isinstance(e, HttpStatusCodeError) and not (
+                e.status_code in (HTTPStatus.REQUEST_TIMEOUT, HTTPStatus.TOO_MANY_REQUESTS)
+                or HTTPStatus.INTERNAL_SERVER_ERROR <= e.status_code < _HTTP_STATUS_UPPER_BOUND
+            ):
+                raise
```
vdusek: **Suggestion:** the retry predicate deserves a name next to `_raise_for_sitemap_status` (e.g. `_is_retryable_sitemap_status(status_code)`), which reduces this to `if isinstance(e, HttpStatusCodeError) and not _is_retryable_sitemap_status(e.status_code): raise` and lets the server-error range use `is_status_code_server_error`.

src/crawlee/request_loaders/_sitemap_request_loader.py
```diff
@@ -214,6 +214,8 @@ async def is_empty(self) -> bool:
     async def is_finished(self) -> bool:
         """Check if all URLs have been processed."""
         state = await self._get_state()
+        if self._loading_task.done() and not self._loading_task.cancelled():
+            self._loading_task.result()
```
vdusek: **Important:** this drops already-loaded URLs while the crawl still reports success. With `BasicCrawler` + `to_tandem()`, sitemap #1 healthy (5 URLs) and sitemap #2 returning 503, `crawler.run()` returned `requests_finished=2, requests_failed=0` while loader state still held `url_queue=['/c','/d','/e']`; pre-PR all five were crawled. The raise also never reaches the caller here: `AutoscaledPool._worker_task_orchestrator` (`_autoscaling/autoscaled_pool.py:221`) evaluates `is_finished_function` in its `while` condition and its `finally` completes `run.result` anyway, so the exception is only logged - and `_load_sitemaps` already logs `Error loading sitemaps` without this change. So on the crawler path this hunk costs URLs and adds no new signal, while `fetch_next_request` does propagate (which is what the new loader test exercises). Suggest dropping this hunk, or at minimum returning `False` while `state.url_queue` or `state.in_progress` is non-empty so a late failure cannot discard work already in hand.

src/crawlee/request_loaders/_sitemap_request_loader.py
```diff
@@ -214,6 +214,8 @@ async def is_empty(self) -> bool:
...
         return not state.url_queue and len(state.in_progress) == 0 and self._loading_task.done()
```
vdusek, line 217: **Note:** the base contract documents this as a pure predicate - `RequestLoader.is_finished` ("Return True if all requests have been handled", `_request_loader.py:41-42`) and `fetch_next_request`'s docstring, which states it returns `None` iff `is_finished` is `True`. If the raise stays, both docstrings need updating; `RequestManagerTandem.is_finished` and `AutoscaledPool` both call this as a predicate.

tests/unit/_utils/test_sitemap.py
```diff
@@ -88,6 +91,30 @@ async def read_stream() -> 'AsyncIterator[bytes]':
     return client, fetched
 
 
+def _make_status_stream_client(responses: list[tuple[int, bytes]]) -> tuple[AsyncMock, list[int]]:
```
vdusek: **Suggestion:** this helper is duplicated byte-for-byte in `tests/unit/request_loaders/test_sitemap_request_loader.py:34`; both files already import shared helpers from `tests/unit/utils.py`.

tests/unit/_utils/test_sitemap.py
```diff
@@ -339,6 +366,36 @@ async def test_sitemap_fetch_raises_after_retries_exhausted() -> None:
     assert len(attempts) == 3
 
 
+async def test_sitemap_fetch_retries_retryable_http_status() -> None:
```
vdusek: **Suggestion:** these tests spend real time in the retry sleep - measured 2.01s each, ~8s across the four new tests. `monkeypatch.setattr('crawlee._utils.sitemap.asyncio.sleep', ...)` keeps every assertion identical at ~0s. Two pre-existing tests have the same cost, so this is a pattern worth breaking rather than following.

tests/unit/_utils/test_sitemap.py
```diff
@@ -339,6 +366,36 @@ async def test_sitemap_fetch_raises_after_retries_exhausted() -> None:
...
+    """Retryable HTTP errors are retried before parsing a successful response."""
+    client, attempts = _make_status_stream_client([(503, b''), (503, b''), (200, get_basic_sitemap().encode())])
+
+    items = [item async for item in parse_sitemap([{'type': 'url', 'url': f'{DEFAULT_URL}sitemap.xml'}], client)]
+
+    assert attempts == [503, 503, 200]
+    assert {item.loc for item in items} == get_basic_results()
+
+
+async def test_sitemap_fetch_rejects_http_error_after_retries_exhausted() -> None:
```
vdusek: **Nit:** "rejects" reads like different behavior next to `test_sitemap_fetch_raises_after_retries_exhausted` just above.

```suggestion
async def test_sitemap_fetch_raises_after_http_error_retries_exhausted() -> None:
```

tests/unit/_utils/test_sitemap.py
```diff
@@ -339,6 +366,36 @@ async def test_sitemap_fetch_raises_after_retries_exhausted() -> None:
...
+    """A persistent retryable HTTP error is raised once retries are exhausted."""
+    client, attempts = _make_status_stream_client([(503, b'')])
+
+    with pytest.raises(HttpStatusCodeError, match='503'):
+        _ = [item async for item in parse_sitemap([{'type': 'url', 'url': f'{DEFAULT_URL}sitemap.xml'}], client)]
+
+    assert attempts == [503, 503, 503]
+
+
+async def test_sitemap_fetch_does_not_retry_terminal_http_status() -> None:
```
vdusek, line 396: **Note:** all three new tests are single-source, so the multi-source behavior change is invisible to the suite. A case with two sources where the first returns 200 and the second 404 would pin down whichever semantics you settle on.

vdusek (changes requested):

anxkhn:
@mantisus @vdusek addressed the review feedback: retryable failures are contained per sitemap, exhausted and non-retryable responses are skipped with warnings, successful sources continue draining, and partial-failure coverage was added with the requested cleanup. rebased onto the latest master; 119 focused tests and ruff pass. could you please take another look?

src/crawlee/request_loaders/_sitemap_request_loader.py
```diff
@@ -214,7 +214,11 @@ async def is_empty(self) -> bool:
     async def is_finished(self) -> bool:
         """Check if all URLs have been processed."""
         state = await self._get_state()
-        return not state.url_queue and len(state.in_progress) == 0 and self._loading_task.done()
+        if state.url_queue or state.in_progress:
+            return False
+        if self._loading_task.done() and not self._loading_task.cancelled():
+            self._loading_task.result()
+        return self._loading_task.done()
```
vdusek: **Important:** the guard only sees this loader's own buffer, so it does not protect the requests the crawl actually owns. `RequestManagerTandem` consults the loader first and outside any `try` - in `is_finished` (`_request_manager_tandem.py:49`) and again in `fetch_next_request` (`_request_manager_tandem.py:77`) - so once this raises, `_read_write_manager` is never reached.

With one unreachable sitemap plus two requests seeded via `crawler.run([...])`, `master` returns normally and crawls both seeded requests, while this branch raises `ConnectionError` out of `crawler.run()` and handles nothing. Same for requests enqueued mid-crawl through `context.add_requests()`.

Either drop the raise and surface the sitemap failure some other way, or have the tandem consult `_read_write_manager` first so its pending work short-circuits the loader.

src/crawlee/request_loaders/_sitemap_request_loader.py
```diff
@@ -214,7 +214,11 @@ async def is_empty(self) -> bool:
     async def is_finished(self) -> bool:
         """Check if all URLs have been processed."""
         state = await self._get_state()
-        return not state.url_queue and len(state.in_progress) == 0 and self._loading_task.done()
+        if state.url_queue or state.in_progress:
+            return False
+        if self._loading_task.done() and not self._loading_task.cancelled():
+            self._loading_task.result()
```
vdusek: **Important:** on the other timing path the same failure is swallowed entirely, so the scenario this PR opens by describing is still a clean empty success. When the loading task fails with no worker parked in the poll loop (unreachable sitemap, no other work), the raise leaves via `AutoscaledPool._worker_task_orchestrator`, whose `finally` completes `run.result` anyway - so `crawler.run()` returns `finished=0 failed=0` and only logs, exactly as on `master`.

So the same failure either aborts the crawl with data loss or reports success, decided purely by timing. Whichever way this is resolved, it needs a `BasicCrawler` + `to_tandem()` test - the current loader-in-isolation test cannot see either path.

src/crawlee/_utils/sitemap.py
```diff
@@ -567,22 +593,30 @@ async def parse_sitemap(
...
+                ):
+                    yield result
+                successful_sources += 1
+            except Exception as e:
+                source_errors.append(e)
+                logger.warning(f'Failed to process sitemap source {source["url"]}: {e}')
         else:
             logger.warning(f'Invalid source configuration: {source}')
 
+    if source_errors and successful_sources == 0:
```
vdusek, line 646: **Important:** `successful_sources` counts "did not raise", so a skipped sibling silences a real failure. Terminal 4xx and retry-exhausted statuses `break` out of `_fetch_and_process_sitemap` rather than raising, and `raw` sources can never fail at all, so all of them count as successes here:

```
Sitemap.load(['broken.xml'])                -> RAISED ConnectionError
Sitemap.load(['broken.xml', 'missing.xml']) -> NO RAISE, urls=[]    # missing.xml = 404
```

The same unreachable sitemap raises or not depending on its siblings, so callers cannot program against it. `Sitemap.try_common_names` hits this directly. Counting only a real 2xx parse as a success - letting the skip path be neither a success nor a `source_error` - keeps a lone 404 quiet while `404 + ConnectionError` still raises.

src/crawlee/_utils/sitemap.py
```diff
@@ -567,22 +593,30 @@ async def parse_sitemap(
...
+        raise source_errors[-1]
```
vdusek: **Nit:** this discards every other failure and flips which source raises - before this PR the *first* failing source propagated.

```suggestion
        raise source_errors[0]
```

tests/unit/utils.py
```diff
@@ -4,20 +4,54 @@
...
+from crawlee.http_clients._base import HttpClient, HttpResponse
+
 T = TypeVar('T')
 
 run_alone_on_mac = pytest.mark.run_alone if sys.platform == 'darwin' else lambda x: x
 
 
+async def sleep_without_delay(_delay: float) -> None:
+    """Yield to the event loop without waiting for a requested test delay."""
+    await asyncio_sleep(0)
```
vdusek: **Note:** two things worth pinning down here. `sitemap.py` does a plain `import asyncio` with no module-local alias (`crawlee._utils.sitemap.asyncio is asyncio` is `True`), so patching `crawlee._utils.sitemap.asyncio.sleep` replaces `asyncio.sleep` process-wide for the test - it also collapses `fetch_next_request`'s 0.1s poll and `poll_until_condition`'s interval into busy-spins.

And this only terminates because it calls the import-time alias `asyncio_sleep`; a later tidy-up to the obvious `asyncio.sleep(0)` would recurse forever. Giving the retry pause a named constant in `sitemap.py` and patching that instead would scope the patch to exactly the delay under test and remove the trap.

tests/unit/_utils/test_sitemap.py
```diff
@@ -357,6 +368,78 @@ async def test_sitemap_fetch_raises_after_retries_exhausted() -> None:
...
+    client, attempts = make_status_stream_client([(200, get_basic_sitemap().encode()), (404, b'')])
+    sources: list[SitemapSource] = [
+        {'type': 'url', 'url': f'{DEFAULT_URL}sitemap.xml'},
+        {'type': 'url', 'url': f'{DEFAULT_URL}missing.xml'},
+    ]
+
+    items = [item async for item in parse_sitemap(sources, client)]
+
+    assert attempts == [200, 404]
+    assert {item.loc for item in items} == get_basic_results()
```
vdusek, line 447: **Suggestion:** this does not exercise what the name claims - a 404 source never raises out of `_fetch_and_process_sitemap`, so this test passes identically with the new `try`/`except` containment removed. Only `test_sitemap_partial_fetch_failure_keeps_healthy_source` covers it.

`make_status_stream_client` also dispatches by call order rather than by URL, so the assertions silently depend on source-iteration order. Keying the mock by URL and adding a skipped-source + raising-source case would cover the gap.

tests/unit/request_loaders/test_sitemap_request_loader.py
```diff
@@ -77,6 +85,71 @@ async def test_sitemap_traversal(server_url: URL, http_client: HttpClient) -> No
...
+    loader = SitemapRequestLoader([f'{DEFAULT_URL}sitemap.xml'], http_client=client)
+
+    assert await loader.fetch_next_request() is None
+
+    assert attempts == [503, 503, 503]
+
+
+async def test_sitemap_loader_drains_requests_before_propagating_failure(monkeypatch: pytest.MonkeyPatch) -> None:
+    """A later sitemap failure is exposed only after requests from healthy sources drain."""
+    monkeypatch.setattr('crawlee._utils.sitemap.asyncio.sleep', AsyncMock(side_effect=sleep_without_delay))
```
vdusek: **Note:** this asserts `pytest.raises` around a bare `is_finished()` and reaches into the private `loader._loading_task`, so it only pins the loader in isolation - where the behaviour is as designed. The failures show up one layer out, in `RequestManagerTandem` and `AutoscaledPool`, which is why neither is visible to the suite today.

src/crawlee/request_loaders/_request_loader.py
```diff
@@ -39,14 +39,14 @@ async def is_empty(self) -> bool:
...
-        """Return True if all requests have been handled."""
+        """Return True if all requests have been handled, or raise if loading failed after pending requests drain."""
 
     @abstractmethod
     async def fetch_next_request(self) -> Request | None:
         """Return the next request to be processed, or `None` if there are no more pending requests.
 
         The method should return `None` if and only if `is_finished` would return `True`. In other cases, the method
-        should wait until a request appears.
+        should wait until a request appears. It can raise a loading error after all pending requests have been handled.
```
vdusek: **Note:** the retained first sentence is now false rather than merely incomplete - "return `None` if and only if `is_finished` would return `True`" no longer holds, because there is a third state where `is_finished` neither returns `True` nor `False`. The appended sentence papers over the contradiction instead of resolving it. `SitemapRequestLoader.is_finished` also still documents itself as just "Check if all URLs have been processed.", and the `while not await loader.is_finished():` pattern in `docs/guides/request_loaders.mdx` can now throw from the loop condition.

src/crawlee/_utils/sitemap.py
```diff
@@ -47,6 +48,21 @@
...
+        raise HttpStatusCodeError('Error status code returned while fetching sitemap', status_code)
+
+
+def _is_retryable_sitemap_status(status_code: int) -> bool:
+    """Return whether a sitemap response status should be retried."""
+    return (
+        HTTPStatus.MULTIPLE_CHOICES <= status_code < HTTPStatus.BAD_REQUEST
+        or status_code in (HTTPStatus.REQUEST_TIMEOUT, HTTPStatus.TOO_MANY_REQUESTS)
+        or is_status_code_server_error(status_code)
+    )
```
vdusek: **Note:** retryability is settled, so this is just the cost - all three clients follow redirects by default and raise `TooManyRedirects` past the limit, so a 3xx reaching here is a 300, a 304, or a redirect with no `Location`, none of which a retry fixes. That is 2 wasted requests and 2s per such sitemap. Separately, `discover_valid_sitemaps` still gates on `is_status_code_successful`, which is 2xx **or** 3xx, so discovery can hand this function a URL it will retry three times and then skip.

tests/unit/utils.py
```diff
@@ -4,20 +4,54 @@
...
 
 import pytest
 
 if TYPE_CHECKING:
-    from collections.abc import Awaitable, Callable
+    from collections.abc import AsyncIterator, Awaitable, Callable
 
     from yarl import URL
 
+from crawlee.http_clients._base import HttpClient, HttpResponse
```
vdusek: **Nit:** this runtime import sits below the `TYPE_CHECKING` block instead of with the other imports at the top of the file.

tests/unit/utils.py
```diff
@@ -4,20 +4,54 @@
...
 run_alone_on_mac = pytest.mark.run_alone if sys.platform == 'darwin' else lambda x: x
 
 
+async def sleep_without_delay(_delay: float) -> None:
+    """Yield to the event loop without waiting for a requested test delay."""
+    await asyncio_sleep(0)
+
+
+def make_status_stream_client(responses: list[tuple[int, bytes]]) -> tuple[AsyncMock, list[int]]:
+    """Create a mock client returning the provided status and body sequence."""
```
vdusek: **Nit:** there are now four near-identical mock HTTP client factories - this one, `_make_flaky_stream_client` and `_make_stream_client` in `test_sitemap.py`, and the inline one in `test_sitemap_request_loader.py`. The PR added `status_code = 200` to each instead of folding them into this shared helper.

src/crawlee/_utils/sitemap.py
```diff
@@ -459,9 +477,14 @@ async def _fetch_and_process_sitemap(
             break
 
         except Exception as e:
+            if isinstance(e, HttpStatusCodeError) and not _is_retryable_sitemap_status(e.status_code):
```
vdusek: **Note:** `_raise_for_sitemap_status` raises an exception that only its own function catches, 85 lines later, in a handler that then has to re-discriminate it from genuine transport errors via `isinstance` - twice. The status decision needs no exception at all; it is a local branch inside the `stream` block. That round trip is also why three of this handler's four outcomes return normally, leaving the caller unable to tell "parsed 0 URLs" from "skipped" - which is the mechanical root cause of the `successful_sources` accounting further down.

src/crawlee/_utils/sitemap.py
```diff
@@ -459,9 +477,14 @@ async def _fetch_and_process_sitemap(
...
+                logger.warning(f'Skipping sitemap {sitemap_url} due to HTTP status code {e.status_code}.')
+                break
```
vdusek: **Nit:** "4xx is terminal, everything else retries" is a policy hardcoded where `ParseSitemapOptions` exists for exactly this - it already carries `sitemap_retries`, `max_depth`, `timeout` and `enqueue_strategy`, and `BasicCrawler` already has the vocabulary (`additional_http_error_status_codes` / `ignore_http_error_status_codes`). `ParseSitemapOptions` is `total=False`, so this is not lock-in and a key can be added later without breaking anyone - worth a deliberate "not now" rather than arriving there by omission.

src/crawlee/_utils/sitemap.py
```diff
@@ -459,9 +477,14 @@ async def _fetch_and_process_sitemap(
...
             if retries_left > 0:
                 logger.warning(f'Error fetching sitemap {sitemap_url}: {e}. Retries left: {retries_left}')
                 await asyncio.sleep(1)  # Brief pause before retry
```
vdusek: **Note:** 429 now reaches this flat 1s sleep. It is in `Session._DEFAULT_BLOCKED_STATUS_CODES` (`_session.py:33`) and the crawler path routes it through `parse_retry_after_header` plus session rotation, but `sitemap.py` takes no `session` at all - so a rate-limiting sitemap endpoint now gets 3 hits at 1s intervals from an unrotated identity, where `master` sent one. Either honour `Retry-After` here, or leave 429 out of the retryable set on this session-less path.

src/crawlee/_utils/sitemap.py
```diff
@@ -567,22 +593,30 @@ async def parse_sitemap(
...
+                    emit_nested_sitemaps=emit_nested_sitemaps,
+                    enqueue_strategy=enqueue_strategy,
+                    proxy_info=proxy_info,
+                    timeout=timeout,
+                ):
+                    yield result
+                successful_sources += 1
+            except Exception as e:
+                source_errors.append(e)
+                logger.warning(f'Failed to process sitemap source {source["url"]}: {e}')
```
vdusek: **Question:** this reinstates the silent-empty path that #1943 (`76927d74`, "retry sitemap fetching on error and raise when retries are exhausted") explicitly removed - its own summary was that a sitemap which keeps failing after all retries now raises to the caller instead of silently producing empty results. That is a partial revert of a shipped fix, and the PR body does not mention it. Worth settling fail-vs-skip once and stating the decision, especially since crawlee JS went the other way at all three layers and stayed there: `sitemap.ts:325-345` warns and gives up without throwing, `Sitemap.parse` (`:485-494`) catches everything and returns `new Sitemap([])`, and `sitemap_request_list.ts:412-414` catches per sitemap URL and continues, keeping `isFinished()` a pure predicate. As written this is a fourth semantic, matching neither #1943 nor JS.

anxkhn:
@vdusek addressed the latest semantics: is_finished stays a predicate, skipped sources are not counted as successes, exhausted 5xx/timeouts warn and skip, 429 is not retried on this path, and a tandem crawler test keeps seeded requests. rebased onto the latest master.

vdusek (approved):
LGTM, let's wait for @Mantisus

Mantisus (approved):
LGTM

PR #2128, opened 2026-08-06
fix: Respect the memory and CPU limits that apply to the process

### LINKED ISSUE #1535: Crawlee increasing concurrency until it dies
I am deploying crawlee in a kubernetes pod. It gets recurrently OOMKilled because crawlee increases the desired concurrency continuously. I don't want to decrease the max_concurrency because I am crawling domains that are super lightweight while crawling others that aren't, and I'd like crawlee to maximize the throughput.  I also could increase the RAM for the pod, but I think there is an underlying issue that would come up later (or I would just underuse my resources)

I am seeing this log which makes me suspicious crawlee doesn't actually know the memory and cpu it is using:
`current_concurrency = 21; desired_concurrency = 21; cpu = 0.0; mem = 0.0; event_loop = 0.148; client_info = 0.0 `

Cpu is not a big problem because the kernel throttles cpu for this pod, but mem is a hardlimit and kubernetes kills the pod.

For more context I am using the Playwright adaptative crawler with beautiful soup and headless firefox and my concurrency settings are:
`        concurrency_settings = ConcurrencySettings(max_concurrency=45, desired_concurrency=10, min_concurrency=10)
`

and I am giving the pod 
```
resources:
    limits:
      cpu: "12"
      memory: "12Gi"
    requests:
      cpu: "8"
      memory: "8Gi"
```

### LINKED ISSUE #2095: Memory metrics ignore container memory limits (no cgroup awareness)
`get_memory_info()` in `src/crawlee/_utils/system.py` takes the system totals from `psutil.virtual_memory()`, which reads `/proc/meminfo`. That file is not namespaced, so inside a container it reports the **host's** memory rather than the cgroup limit:

```python
vm = psutil.virtual_memory()

return MemoryInfo(
    total_size=ByteSize(vm.total),
    current_size=ByteSize(current_size_bytes),
    system_wide_used_size=ByteSize(vm.total - vm.available),
)
```

Consequences for a crawler in a memory-limited container:

- `MemoryInfo.total_size` is the host's total RAM.
- `Snapshotter` derives the memory budget from it - with `memory_mbytes` unset, `max_memory_size = total_size * available_memory_ratio` (`src/crawlee/_autoscaling/snapshotter.py:122-128`). So the autoscaler can scale past the container limit and get OOM-killed instead of throttling.
- `system_wide_used_size` reflects the host's usage, including processes unrelated to the crawler.

Crawlee for JS handles this - `getMemoryInfoV2()` reads the cgroup files directly (`packages/utils/src/internals/systemInfoV2/memory-info.ts`), with cgroups v1/v2 detection, the `max` sentinel for "unlimited", and a separate AWS Lambda path:

```text
TOTAL: v1 /sys/fs/cgroup/memory/memory.limit_in_bytes, v2 /sys/fs/cgroup/memory.max
USED:  v1 /sys/fs/cgroup/memory/memory.usage_in_bytes,  v2 /sys/fs/cgroup/memory.current
```

Suggested direction: mirror the JS behavior - detect containerization, read the cgroup limit when present, and fall back to `psutil.virtual_memory()` otherwise. Worth deciding whether it should be automatic or opt-in via `Configuration` (JS gates it behind a `systemInfoV2` flag plus a `containerized` option).

Notes:

- Not a regression - crawlee-python has never had cgroup awareness (`grep -rn cgroup src/` finds nothing).
- Setting `memory_mbytes` (or `CRAWLEE_MEMORY_MBYTES`) is the current workaround, since it bypasses `total_size` entirely.
- Out of scope for #2078, which hardens the *per-process* side of the same function against unreadable metrics. The container-limit side is untouched by it.

*✍️ Drafted by Claude Code*

Mantisus:
### Description

Crawlee now scales against the memory and CPU limits that apply to the process rather than the resources of the host machine. This covers Docker containers, Kubernetes pods, systemd slices and Windows job objects, with nothing to configure. Without a limit, the host values are used as before.

A crawler under a limit below the host memory gets a smaller `max_memory_size`, so it runs fewer tasks at once. `memory_mbytes` still sets the budget in absolute terms.

### Issues

- Closes: #2095
- Closes: #1535 

### Testing

- Add new tests with a fake cgroup filesystem


src/crawlee/_utils/cgroup.py
```diff
@@ -0,0 +1,434 @@
...
+            # The same hierarchy can be bind-mounted a second time, for instance by an agent that watches the host
+            # from inside a container. Mounts are listed in the order they were made, so the first one is ours.
+            unified = unified or _Hierarchy(point=Path(mount_point), root=mount_root, own_path=unified_path)
+        elif filesystem == 'cgroup':
+            # A cgroup v1 mount names the controllers it carries among its super options, e.g. `rw,cpu,cpuacct`.
+            for option in super_options.split(','):
+                own_path = controller_paths.get(option)
+                if option in _V1_CONTROLLER_NAMES and own_path is not None:
+                    hierarchy = _Hierarchy(point=Path(mount_point), root=mount_root, own_path=own_path)
+                    controllers.setdefault(option, hierarchy)
```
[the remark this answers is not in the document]
Mantisus: Solved

src/crawlee/_utils/cgroup.py
```diff
@@ -0,0 +1,434 @@
...
+
+
+def _count_cpu_list(cpu_list: str) -> int | None:
+    """Count the CPUs a control file lists as a mix of ranges and single numbers, e.g. `0-3,8`."""
+    count = 0
+
+    try:
+        for part in cpu_list.split(','):
+            first, separator, last = part.partition('-')
+            count += int(last) - int(first) + 1 if separator else 1
```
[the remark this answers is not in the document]
Mantisus: Solved

vdusek:
If I am correct, all the tests run against a fake cgroup filesystem. Is there a way to cover this end-to-end? e.g., running a crawler in a container with `--memory` and/or `--cpus` set, and asserting `get_memory_info()` and `get_cpu_info()`? Or have you tested it manually? Also, we should make sure we cover both `--cgroupns=private` and `--cgroupns=host`.

Mantisus:
> If I am correct, all the tests run against a fake cgroup filesystem. Is there a way to cover this end-to-end? 

That's right. To add the test, we'll probably need to add a new job to CI. With my permissions in the repository, I won't be able to test it properly.

> Or have you tested it manually?

Yes, I tested it locally with `systemd` and `docker`

```bash
(crawlee) mantisus@Kleimor:~/repos/apify/crawlee-python$ docker run --rm $MOUNTS $IMG $RUN
configuration
  available_memory_ratio : 0.25
  max_used_memory_ratio  : 0.9
  max_used_cpu_ratio     : 0.95
  memory_mbytes          : None
  system_info_interval   : 0:00:01
limits read from the cgroup
  memory      : None
  cpu quota   : None
  cpu set     : 20
latest memory snapshot
  current_size            : 60.27 MB
  max_memory_size         : 4.88 GB
  system_wide_used_size   : 2.63 GB
  system_wide_memory_size : 19.53 GB
  is_overloaded           : False
latest cpu snapshot
  used_ratio              : 0.216
  max_used_ratio          : 0.95
  is_overloaded           : False
system status over the last 6 cpu samples
  cpu        : 0.00 of 0.40 allowed   ok
  memory     : 0.00 of 0.20 allowed   ok
  event_loop : 0.00 of 0.60 allowed   ok
  client     : 0.00 of 0.30 allowed   ok
  is_system_idle: True
```

```bash
(crawlee) mantisus@Kleimor:~/repos/apify/crawlee-python$ docker run --rm -m 512m --cpus 1 $MOUNTS $IMG $RUN
configuration
  available_memory_ratio : 0.25
  max_used_memory_ratio  : 0.9
  max_used_cpu_ratio     : 0.95
  memory_mbytes          : None
  system_info_interval   : 0:00:01
limits read from the cgroup
  memory      : MemoryLimit(limit=536870912, working_set=84348928)
  cpu quota   : 1.0
  cpu set     : 20
latest memory snapshot
  current_size            : 60.19 MB
  max_memory_size         : 128.00 MB
  system_wide_used_size   : 80.98 MB
  system_wide_memory_size : 512.00 MB
  is_overloaded           : False
latest cpu snapshot
  used_ratio              : 0.996
  max_used_ratio          : 0.95
  is_overloaded           : True
system status over the last 6 cpu samples
  cpu        : 0.74 of 0.40 allowed   OVERLOADED
  memory     : 0.00 of 0.20 allowed   ok
  event_loop : 0.00 of 0.60 allowed   ok
  client     : 0.00 of 0.30 allowed   ok
  is_system_idle: False
```
```bash
(crawlee) mantisus@Kleimor:~/repos/apify/crawlee-python$ docker run --rm -m 512m --cpus 6 $MOUNTS $IMG $RUN
configuration
  available_memory_ratio : 0.25
  max_used_memory_ratio  : 0.9
  max_used_cpu_ratio     : 0.95
  memory_mbytes          : None
  system_info_interval   : 0:00:01
limits read from the cgroup
  memory      : MemoryLimit(limit=536870912, working_set=85512192)
  cpu quota   : 6.0
  cpu set     : 20
latest memory snapshot
  current_size            : 60.32 MB
  max_memory_size         : 128.00 MB
  system_wide_used_size   : 82.48 MB
  system_wide_memory_size : 512.00 MB
  is_overloaded           : False
latest cpu snapshot
  used_ratio              : 0.723
  max_used_ratio          : 0.95
  is_overloaded           : False
system status over the last 6 cpu samples
  cpu        : 0.00 of 0.40 allowed   ok
  memory     : 0.00 of 0.20 allowed   ok
  event_loop : 0.00 of 0.60 allowed   ok
  client     : 0.00 of 0.30 allowed   ok
  is_system_idle: True
```

```bash
(crawlee) mantisus@Kleimor:~/repos/apify/crawlee-python$ docker run --rm -m 512m --cgroupns=host $MOUNTS $IMG $RUN 
configuration
  available_memory_ratio : 0.25
  max_used_memory_ratio  : 0.9
  max_used_cpu_ratio     : 0.95
  memory_mbytes          : None
  system_info_interval   : 0:00:01
limits read from the cgroup
  memory      : MemoryLimit(limit=536870912, working_set=89137152)
  cpu quota   : None
  cpu set     : 20
latest memory snapshot
  current_size            : 60.34 MB
  max_memory_size         : 128.00 MB
  system_wide_used_size   : 85.98 MB
  system_wide_memory_size : 512.00 MB
  is_overloaded           : False
latest cpu snapshot
  used_ratio              : 0.200
  max_used_ratio          : 0.95
  is_overloaded           : False
system status over the last 6 cpu samples
  cpu        : 0.00 of 0.40 allowed   ok
  memory     : 0.00 of 0.20 allowed   ok
  event_loop : 0.00 of 0.60 allowed   ok
  client     : 0.00 of 0.30 allowed   ok
  is_system_idle: True
```

Pijukatel:
Great idea. But let's focus on testing this in different environment setups. This could be a standalone repo, as it can be useful outside of Crawlee and it has quite different testing requirements.

So I would suggest having it in this [placeholder repo](https://github.com/apify/cgroups-sensor) as an experiment, and once we are sure it works in different environments and setups, we can publish it and integrate it into Crawlee.

src/crawlee/_utils/system.py
```diff
@@ -185,35 +189,83 @@ class MemoryInfo(MemoryUsageInfo):
...
+    if limits.memory_budget is None and cores is None:
+        return
 
-def get_cpu_info() -> CpuInfo:
+    memory = str(ByteSize(limits.memory_budget.limit)) if limits.memory_budget else 'unrestricted'
+    cpu = f'{cores:g} core{"" if cores == 1 else "s"}' if cores is not None else 'unrestricted'
+    logger.info(f'Resource limits applying to this process: memory {memory}, CPU {cpu}.')
+
+
+def get_cpu_info(cpu_load: proclimits.CpuLoad) -> CpuInfo:
```
Pijukatel: `cpu_load` seems to be optional. The first branch can return without using it.

src/crawlee/_utils/system.py
```diff
@@ -185,35 +189,83 @@ class MemoryInfo(MemoryUsageInfo):
...
+
+    Args:
+        cpu_load: The sampler owned by the caller. Two callers sharing one would measure each other's windows.
     """
     logger.debug('Calling get_cpu_info()...')
-    cpu_percent = psutil.cpu_percent(interval=0.1)
-    return CpuInfo(used_ratio=cpu_percent / 100)
+
+    # Read on every sample rather than latched, because a limit can be resized while the process runs.
+    if proclimits.get_cpu_limit() is None:
```
Pijukatel, line 248: If made optional:
`or cpu_load is None`

src/crawlee/configuration.py
```diff
@@ -186,9 +186,9 @@ class Configuration(BaseSettings):
             le=1.0,
         ),
     ] = 0.25
```
Pijukatel, line 188: Now, the 0.25 will be far more restrictive in some environments, and users might not even be aware of it.

Would it make sense to log a one-time warning in some scenarios where 0.25 is most likely not the desired value?

Example scenario:
Running inside a container that is restricted to 0.25 of the machine's memory. Now you get 0.25*0.25 of the machine's memory as the default, while previously you would get 0.25 regardless of the container limit.

I think these limits previously served as protection for the machine from being completely consumed by the crawler. But when the user already applies some limits outside of Crawlee, then this safety limit becomes not only redundant, but most likely a second hidden limit that the user might not even be aware of.

So I would log a one-time warning each time someone is using the default value while also using `LocalEventManager` and non-machine limits apply, and reason that if the user already applied some limits to the process, then this 0.25 stacked on top of the other limit is most likely not a desired outcome.

Pijukatel (commented):
Really nice work on the proclimits—just some minor comments.

src/crawlee/_utils/system.py
Mantisus: Made it optional, but without this suggestion. Without a sampler, a process under a limit still gets measured against that limit (a short blocking `proclimits.get_cpu_used_ratio`) rather than falling back to psutil, which would read the whole host.

src/crawlee/configuration.py
Mantisus: Good point, added warning.

Pijukatel (approved):

vdusek (changes requested):
Please give me a little more time to check this.

PR #2140, opened 2026-08-07
perf(impit): optimize client and connection reuse across Sessions

Mantisus:
### Description

- Cookie handling has been moved entirely from `impit` to `ImpitHttpClient`. Clients are now reused per proxy instead of being recreated for every `Session`, so a single client keeps its connection pool alive across all of them. The result is higher performance than that of `HttpxHttpClient`.
- Redirects are followed one hop at a time, which is what makes per-session cookie handling possible. Header handling along the chain now follows the [WHATWG Fetch](https://fetch.spec.whatwg.org/#http-redirect-fetch) algorithm.

### Testing

- Added tests for extracting cookies from Set-Cookie headers into the jar and for building the Cookie header.
- Added tests for redirect handling in ImpitHttpClient.


src/crawlee/http_clients/_impit.py
```diff
@@ -222,34 +267,112 @@ async def stream(
...
 
         # Update the default kwargs with any additional user-provided kwargs.
         kwargs.update(self._async_client_kwargs)
 
-        client = AsyncClient(**kwargs, cookie_jar=cookie_jar)
+        # Redirects are followed hop by hop by this client.
+        client = AsyncClient(**kwargs, follow_redirects=False)
 
-        self._client_by_proxy_url[proxy_url] = _ClientCacheEntry(client=client, cookie_jar=cookie_jar)
+        self._client_by_proxy_url[proxy_url] = client
```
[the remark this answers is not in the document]
Mantisus: `impit.AsyncClient` does not implement a method to close the client, but Rust handles resource cleanup very well.

src/crawlee/http_clients/_impit.py
```diff
@@ -222,34 +267,112 @@ async def stream(
...
+            TooManyRedirectsError: If the number of redirects exceeds `max_redirects`.
+
+        Returns:
+            The final response of the redirect chain.
+        """
+        client = self._get_client(proxy_info.url if proxy_info else None)
+        current_url = URL(url)
+        content = payload
+
+        for _ in range(self._max_redirects + 1):
```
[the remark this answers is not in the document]
Mantisus: The initial request is not counted toward the redirect counter.

src/crawlee/http_clients/_impit.py
```diff
@@ -222,34 +268,112 @@ async def stream(
...
+            stream: Whether the body of the final response should be streamed.
+
+        Raises:
+            TooManyRedirectsError: If the number of redirects exceeds `max_redirects`.
+
+        Returns:
+            The final response of the redirect chain.
+        """
+        client = self._get_client(proxy_info.url if proxy_info else None)
+        current_url = URL(url)
```
vdusek: **Important:** wrapping the URL in `yarl.URL` and sending `str(current_url)` on line 317 re-encodes it -- yarl decodes percent-escapes it considers safe, so `X-Amz-Credential=AKIA%2F20240101%2F...` goes on the wire as `AKIA/20240101/...` and presigned links start failing signature checks. This hits every request through the default client, not only redirect chains; master passed the string through verbatim.

Suggest keeping `url` as a `str` for the wire and building a `yarl.URL` only inside the redirect branch (for `join()` and `_is_cross_origin`).

src/crawlee/http_clients/_impit.py
```diff
@@ -113,10 +154,12 @@ def __init__(
         self._http3 = http3
         self._verify = verify
         self._browser = browser
+        self._follow_redirects = follow_redirects
+        self._max_redirects = max_redirects
 
         self._async_client_kwargs = async_client_kwargs
 
-        self._client_by_proxy_url = LRUCache[str | None, _ClientCacheEntry](maxsize=10)
+        self._client_by_proxy_url = dict[str | None, AsyncClient]()
```
vdusek, line 163: Just noting, this is the same as in https://github.com/apify/crawlee-python/pull/2104#discussion_r3712255447.

I would also appreciate a short docstring here explaining its purpose and why `dict` is being used here (cache without an upper bound).

And we should also unify the naming of this attribute across HTTP clients.

src/crawlee/http_clients/_impit.py
```diff
@@ -222,34 +268,112 @@ async def stream(
...
+        """
+        client = self._get_client(proxy_info.url if proxy_info else None)
+        current_url = URL(url)
+        content = payload
+
+        for _ in range(self._max_redirects + 1):
+            # Header names are normalized to lowercase by `HttpHeaders`, so this replaces any `Cookie` header
+            # passed by the caller instead of adding a second one.
+            if session and (cookie_string := session.cookies.get_cookie_string(str(current_url))):
+                headers['cookie'] = cookie_string
```
vdusek: **Note:** `headers['cookie']` is assigned but never cleared, so on a hop where the jar matches nothing the previous hop's value is sent again -- a cookie stored with `path=/redirect` reaches `/cookies` after the 302. Cross-origin hops are safe (`_CROSS_ORIGIN_HEADERS` strips it), so this violates cookie path scoping within one origin rather than leaking across sites.

Computing per-hop headers instead of mutating the shared dict fixes it while keeping the caller-supplied `Cookie` header that the new tests pin.

src/crawlee/http_clients/_impit.py
```diff
@@ -222,34 +268,112 @@ async def stream(
...
+
+            response = await client.request(
+                method=method,
+                url=str(current_url),
+                content=content,
+                headers=headers or None,
+                timeout=timeout.total_seconds() if timeout else None,
```
vdusek: **Note:** the full `timeout` is now applied per hop, so a chain can take up to `(max_redirects + 1) x timeout`. `crawl()` is bounded by `SharedTimeout` and in-handler `send_request` by `request_handler_timeout`, so the practical blast radius is small -- but it is a change from the single impit call. Worth a deliberate call on whether this should be a deadline for the whole chain.

src/crawlee/errors.py
```diff
@@ -57,6 +58,16 @@ class ProxyError(SessionError):
     """Raised when a proxy is being blocked or malfunctions."""
 
 
+@docs_group('Errors')
+class TooManyRedirectsError(Exception):
```
vdusek: Only `ImpitHttpClient` raises this -- `HttpxHttpClient` surfaces `httpx.TooManyRedirects` and curl-impersonate its own -- and it appears in no public `Raises:` block, while `crawl`/`send_request`/`stream` all document `ProxyError`. Worth reconsideration.

src/crawlee/http_clients/_impit.py
```diff
@@ -222,34 +268,112 @@ async def stream(
...
             'verify': self._verify,
-            'follow_redirects': True,
             'browser': self._browser,
         }
 
         # Update the default kwargs with any additional user-provided kwargs.
         kwargs.update(self._async_client_kwargs)
 
-        client = AsyncClient(**kwargs, cookie_jar=cookie_jar)
+        # Redirects are followed hop by hop by this client.
```
vdusek: **Nit:** "this client" reads as the `AsyncClient` constructed on the next line, which is exactly the thing that no longer follows redirects.

```suggestion
        # Redirects are followed hop by hop by `_request_with_redirects`.
```

src/crawlee/http_clients/_impit.py
```diff
@@ -222,34 +268,112 @@ async def stream(
...
+                return response
+
+            next_method = _redirect_method(response.status_code, method)
+            if next_method != method:
+                method = next_method
+                content = None
+                headers = {key: value for key, value in headers.items() if key.lower() not in _REQUEST_BODY_HEADERS}
+
+            if _is_cross_origin(current_url, next_url):
+                headers = {key: value for key, value in headers.items() if key.lower() not in _CROSS_ORIGIN_HEADERS}
```
vdusek: **Nit:** `key.lower()` is redundant here -- the comment on line 310 states header names are already lowercase, and both call sites guarantee it (`crawl` uses `HttpHeaders`, the others normalize first).

```suggestion
                headers = {key: value for key, value in headers.items() if key not in _REQUEST_BODY_HEADERS}

            if _is_cross_origin(current_url, next_url):
                headers = {key: value for key, value in headers.items() if key not in _CROSS_ORIGIN_HEADERS}
```

src/crawlee/sessions/_cookies.py
```diff
@@ -217,6 +231,36 @@ def set_cookies(self, cookie_dicts: list[CookieParam]) -> None:
...
+
+        Returns:
+            The `Cookie` header value, or an empty string if no stored cookie matches the URL.
+        """
+        # `UrlRequest` is only used as a carrier of the URL and headers for the jar, it never opens a connection.
+        url_request = UrlRequest(url)  # noqa: S310
+        self._jar.add_cookie_header(url_request)
+        return url_request.get_header('Cookie', '')
+
+    def extract_cookie_from_header(self, url: str, set_cookie_headers: list[str]) -> None:
```
vdusek: **Nit:** this takes a list of headers and extracts many cookies, so the plural reads more accurately. Cheap to rename now, breaking once released (the only call site is `_impit.py:325`).

```suggestion
    def extract_cookies_from_headers(self, url: str, set_cookie_headers: list[str]) -> None:
```

tests/unit/server.py
```diff
@@ -270,6 +272,35 @@ async def echo_headers(scope: dict[str, Any], _receive: Receive, send: Send) ->
     await send_json_response(send, headers)
 
 
+async def echo_method(scope: dict[str, Any], receive: Receive, send: Send) -> None:
```
vdusek, line 286: **Nit:** this body-reading loop is a verbatim copy of `post_echo`'s, and both spin forever if `receive()` yields `http.disconnect` instead of `http.request`. A small shared `read_body(receive)` helper would cover both.

tests/unit/http_clients/test_impit.py
```diff
@@ -0,0 +1,223 @@
...
+async def test_too_many_redirects(server_url: URL) -> None:
+    """Test that an endless redirect chain is cut off by `max_redirects`."""
+    async with ImpitHttpClient(max_redirects=2) as client:
+        with pytest.raises(TooManyRedirectsError) as exc_info:
+            await client.send_request(str(server_url / 'redirect_loop'))
+
+    assert exc_info.value.max_redirects == 2
+
+
+async def test_stream_follows_redirects(http_client: ImpitHttpClient, server_url: URL) -> None:
```
vdusek, line 219: **Note:** the suite exercises `send_request` and `stream`, but nothing covers `crawl()` with a redirect chain plus session cookies -- the path the crawler actually takes. A test pinning the on-the-wire target for a URL with `%2F`/`%3A` in the query would also lock down the encoding regression flagged in `_impit.py`.

tests/unit/http_clients/test_impit.py
```diff
@@ -0,0 +1,223 @@
...
+async def test_explicit_cookie_header_kept_within_origin(http_client: ImpitHttpClient, server_url: URL) -> None:
+    """Test that a `Cookie` header set by the caller survives a redirect within the origin."""
+    redirect_url = (server_url / 'redirect').with_query(url=str(server_url / 'cookies'), status=302)
+
+    response = await http_client.send_request(str(redirect_url), headers={'cookie': 'manual=value'})
+
+    assert (await read_json(response))['cookies'] == {'manual': 'value'}
+
+
+async def test_session_cookies_replace_explicit_cookie_header(http_client: ImpitHttpClient, server_url: URL) -> None:
```
vdusek: **Question:** this inverts the precedence master had -- an explicit `Cookie` header used to beat the impit cookie jar. Sessions are on by default, so a header hand-set on a `Request` is now silently dropped as soon as the jar holds anything for that domain. The new behavior looks like the right call; could it be spelled out in the PR description so it reaches the changelog?

pyproject.toml
```diff
@@ -37,7 +37,7 @@ dependencies = [
     "async-timeout>=5.0.1",
     "cachetools>=5.5.0",
     "colorama>=0.4.0",
-    "impit>=0.8.0",
+    "impit>=0.13.2",
```
vdusek, line 40: Do we really need to bump the constraint to exactly this version?

src/crawlee/http_clients/_impit.py
Mantisus: `LRUCache` was used in `ImpitHttpClient` because of caching that involves `CookieJar`. If we used the approach taken by other clients (a `dict` with no upper bound), the number of `AsyncClient` instances created would constantly increase (session pool size + number of replaced sessions). A size of 10 elements was chosen to match JS.

Now caching depends only on the proxy, just as in other clients, so the behavior has been unified to a `dict` with no upper limit.

If we decide to change this, it should be done in all three clients.

> And we should also unify the naming of this attribute across HTTP clients.

In all three cases, it is `self._client_by_proxy_url`

pyproject.toml
Mantisus: Yes, the necessary fix for handling the `set-cookie` header was implemented in the PR https://github.com/apify/impit/pull/492

This is the first release to include it.

vdusek: got it, thanks

tests/unit/http_clients/test_impit.py
Mantisus: It's my bad. I rechecked the standard behavior of `impit` and `httpx`; `jar` shouldn't overwrite explicitly provided headers.

The behavior is different in `curl_cffi` because it merges the explicitly provided cookie header with data from `jar`.

src/crawlee/http_clients/_impit.py
Mantisus: Fixed with the `encoded=True` flag for `yarl.URL`

src/crawlee/http_clients/_impit.py
Mantisus: Added internal timeout logic to maintain a consistent timeout during the entire redirect chain.

tests/unit/server.py
Mantisus: Good point, updated.

tests/unit/http_clients/test_impit.py
Mantisus: Added new test.

src/crawlee/errors.py
Mantisus: It's not the most elegant solution, but I wrapped it in `impit.TooManyRedirects`. This way, the behavior remains consistent across all clients.

vdusek (approved):
LGTM

vdusek (approved):
LGTM

PR #2141, opened 2026-08-07
fix(sitemap): guard the unavailable `ExpatParser.flush` on older `CPython` patch releases

### LINKED ISSUE #2119: Unguarded `ExpatParser.flush()` logs a spurious warning on Python < 3.10.14
[`_XmlSitemapParser.flush`](https://github.com/apify/crawlee-python/blob/5882c6b51a0134097615c6d4e12a3765d963fc28/src/crawlee/_utils/sitemap.py#L201-L212) calls `self._parser.flush()` unguarded.

- `ExpatParser.flush` isn't part of the documented [`IncrementalParser`](https://docs.python.org/3/library/xml.sax.reader.html#incrementalparser-objects) interface and doesn't exist before 3.10.14 - verified `False` on 3.10.0 / 3.10.8 / 3.10.12 / 3.10.13, `True` on 3.10.14+, all of which satisfy our `requires-python = ">=3.10"`
- consequence on those interpreters: **every** sitemap parse logs `Failed to parse remaining XML data: 'ExpatParser' object has no attribute 'flush'` at WARNING
- not data loss - `flush()` contributes 0 items in every scenario I tried (single chunk, 3 chunks, truncated document, 20k `<url>` entries fed in 1 KiB slices cutting tokens mid-token). Expat's reparse deferral landed in the same release as `flush()`, so on interpreters without `flush()` nothing is ever left buffered
- so: pure log noise, low priority. Guard with `suppress(AttributeError)` / `getattr`, or drop the call
- separately, the `except` swallows the `yield` loop below it - if `flush()` ever *did* raise with items buffered, they'd be discarded. Cheap to make robust by yielding `self._handler.items` in a `finally`

Mantisus:
### Description

- `_XmlSitemapParser.flush` called `ExpatParser.flush()` unguarded. The method was added in CPython 3.10.14, 3.11.9, and 3.12.3, so on earlier patch releases every sitemap parse logged a spurious WARNING. It's now called only when it exists.
- Moved the `yield` loop into a `finally`, so a raising `flush()` no longer discards the items the handler has already collected.

### Issues

- Closes: #2119

### Testing

- Added tests for a parser without `flush` and for a failing `flush()`.

### DIFF
--- a/src/crawlee/_utils/sitemap.py
+++ b/src/crawlee/_utils/sitemap.py
@@ -201,15 +201,19 @@ class _XmlSitemapParser:
     async def flush(self) -> AsyncGenerator[_SitemapItem, None]:
         """Process any remaining data in the buffer, yielding items one by one."""
         try:
-            self._parser.flush()
+            # `ExpatParser.flush` isn't part of the `IncrementalParser` interface and is missing before CPython
+            # 3.10.14, 3.11.9 and 3.12.3. Those interpreters don't defer reparsing either, so nothing stays buffered.
+            if (flush := getattr(self._parser, 'flush', None)) is not None:
+                flush()
 
+        except Exception as e:
+            logger.warning(f'Failed to parse remaining XML data: {e}')
+
+        finally:
             for item in self._handler.items:
                 yield item
 
             self._handler.items.clear()
-
-        except Exception as e:
-            logger.warning(f'Failed to parse remaining XML data: {e}')
 
     def close(self) -> None:
         """Clean up resources."""
--- a/src/crawlee/_utils/sitemap.py
+++ b/src/crawlee/_utils/sitemap.py
@@ -202,7 +202,7 @@ class _XmlSitemapParser:
         """Process any remaining data in the buffer, yielding items one by one."""
         try:
             # `ExpatParser.flush` isn't part of the `IncrementalParser` interface and is missing before CPython
-            # 3.10.14, 3.11.9 and 3.12.3. Those interpreters don't defer reparsing either, so nothing stays buffered.
+            # 3.10.14, 3.11.9 and 3.12.3, whose bundled expat predates reparse deferral, so nothing stays buffered.
             if (flush := getattr(self._parser, 'flush', None)) is not None:
                 flush()
 
--- a/tests/unit/_utils/test_sitemap.py
+++ b/tests/unit/_utils/test_sitemap.py
@@ -321,6 +321,22 @@ async def test_sitemap_from_string() -> None:
     assert set(sitemap.urls) == get_basic_results()
 
 
+async def test_malformed_sitemap_keeps_urls() -> None:
+    """A parse error must not discard the URLs collected before it."""
+    malformed = (
+        '<?xml version="1.0" encoding="UTF-8"?>\n'
+        '<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">\n'
+        f'<url><loc>{DEFAULT_URL}first</loc></url>\n'
+        f'<url><loc>{DEFAULT_URL}second</loc></url>\n'
+        f'<url><loc>{DEFAULT_URL}third</loc></mismatched>\n'
+        '</urlset>'
+    )
+
+    sitemap = await Sitemap.from_xml_string(malformed)
+
+    assert sitemap.urls == [f'{DEFAULT_URL}first', f'{DEFAULT_URL}second']
+
+
 async def test_sitemap_fetch_retries_on_transient_error() -> None:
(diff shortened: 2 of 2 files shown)


src/crawlee/_utils/sitemap.py
```diff
@@ -201,16 +201,20 @@ async def process_chunk(self, chunk: str) -> AsyncGenerator[_SitemapItem, None]:
...
-            self._parser.flush()
+            # `ExpatParser.flush` isn't part of the `IncrementalParser` interface and is missing before CPython
+            # 3.10.14, 3.11.9 and 3.12.3. Those interpreters don't defer reparsing either, so nothing stays buffered.
+            if (flush := getattr(self._parser, 'flush', None)) is not None:
+                flush()
 
+        except Exception as e:
+            logger.warning(f'Failed to parse remaining XML data: {e}')
+
+        finally:
```
vdusek, line 212: Regarding the test coverage

> **Note:** the `finally` does more than "no longer discards the items the handler has already collected". A malformed document now yields the URLs parsed before the error, where `master` dropped all of them - A/B on a mismatched close tag gives 0 URLs on `master` vs 5 here. That is the right behavior and matches the already-partial truncated-document case, but it is also the part users will notice, and neither new test covers it (both drive a deleted/mocked `flush`). Worth one test feeding a malformed document through `Sitemap.from_xml_string` and asserting the pre-error URLs come out, plus a line in the PR description - broken sitemaps go from all-or-nothing to best-effort.

tests/unit/_utils/test_sitemap.py
```diff
@@ -562,6 +564,34 @@ async def test_txt_parser_flush_clears_buffer() -> None:
...
+    """A failing `flush()` must not discard the items already collected by the handler."""
+    parser = _XmlSitemapParser()
+    parser._handler.items.append({'type': 'url', 'loc': f'{DEFAULT_URL}page'})
+
+    # `create=True` covers interpreters where `ExpatParser` has no `flush` to replace.
+    with patch.object(parser._parser, 'flush', side_effect=RuntimeError('Broken parser'), create=True):
+        items = [item async for item in parser.flush()]
+
+    assert items == [{'type': 'url', 'loc': f'{DEFAULT_URL}page'}]
+    assert parser._handler.items == []
```
vdusek: We could add an assert that the warning is emitted

> **Nit:** the contract is "warn *and* keep the items", but only the second half is pinned - nothing asserts the warning is emitted.

```suggestion
async def test_xml_parser_yields_items_when_flush_fails(caplog: pytest.LogCaptureFixture) -> None:
    """A failing `flush()` must not discard the items already collected by the handler."""
    parser = _XmlSitemapParser()
    parser._handler.items.append({'type': 'url', 'loc': f'{DEFAULT_URL}page'})

    # `create=True` covers interpreters where `ExpatParser` has no `flush` to replace.
    with (
        caplog.at_level(logging.WARNING, logger='crawlee._utils.sitemap'),
        patch.object(parser._parser, 'flush', side_effect=RuntimeError('Broken parser'), create=True),
    ):
        items = [item async for item in parser.flush()]

    assert items == [{'type': 'url', 'loc': f'{DEFAULT_URL}page'}]
    assert parser._handler.items == []
    assert 'Failed to parse remaining XML data: Broken parser' in caplog.text
```

src/crawlee/_utils/sitemap.py
```diff
@@ -201,16 +201,20 @@ async def process_chunk(self, chunk: str) -> AsyncGenerator[_SitemapItem, None]:
     async def flush(self) -> AsyncGenerator[_SitemapItem, None]:
         """Process any remaining data in the buffer, yielding items one by one."""
         try:
-            self._parser.flush()
+            # `ExpatParser.flush` isn't part of the `IncrementalParser` interface and is missing before CPython
+            # 3.10.14, 3.11.9 and 3.12.3. Those interpreters don't defer reparsing either, so nothing stays buffered.
```
vdusek: Just a comment precision note:

> **Nit:** reparse deferral is a libexpat >= 2.6.0 default, not a CPython patch-level feature - `SetReparseDeferralEnabled` and `flush` were added in the same backports, so a `--with-system-expat` build of e.g. 3.11.8 against libexpat 2.6.x does defer while exposing no `flush`. No practical impact (I fed a 200-URL document in 7-char slices and `flush()` contributed 0 items), so this is comment precision only.

```suggestion
            # `ExpatParser.flush` isn't part of the `IncrementalParser` interface and is missing before CPython
            # 3.10.14, 3.11.9 and 3.12.3, whose bundled expat predates reparse deferral, so nothing stays buffered.
```

vdusek (commented):
just nits / details 🙂

vdusek (approved):
LGTM

PR #2152, opened 2026-08-12
fix(dataset): preserve push_data compatibility with the released Apify SDK

vdusek:
The scheduled E2E tests fail on every matrix job (all 90) with:

```
AttributeError: 'ApifyDatasetClient' object has no attribute '_is_sequence_of_items'
```

`DatasetClient._is_sequence_of_items` was removed in #2145. The released Apify SDK (up to 4.0.1) still calls it from `ApifyDatasetClient.push_data`, inheriting it from this base class, so a master wheel paired with the published SDK raises on every `push_data`. The templates E2E test builds exactly that pairing. Published crawlee 1.9.1 predates the removal, so this is not only a CI problem: the next crawlee release would carry the same break to every Actor resolving `apify<=4.0.1`.

This restores the helper on `DatasetClient`, annotated `-> bool` rather than `TypeIs[...]`. The narrowing is what `ty 0.0.69` objected to in the first place, and no `push_data` implementation here uses the helper anymore, they keep the inlined `isinstance`. Drop it in the next major version, once those SDK versions are out of the supported range.

Verified by installing the built wheel alongside `apify==4.0.1` and confirming the SDK call site resolves; swapping in the unpatched base class reproduces the failure above.

Failing run: https://github.com/apify/crawlee-python/actions/runs/31554176158

*✍️ Drafted by Claude Code*


Pijukatel (approved):

Pijukatel (commented):
Stagehand templates are still failing (Maybe another issue)
https://github.com/apify/crawlee-python/actions/runs/31583394515/job/94071500831

vdusek:
Stagehand:

> ImportError: cannot import name 'AsyncStagehand' from 'stagehand'
  (/usr/local/lib/python3.13/site-packages/stagehand/__init__.py)
  
  So it's not related to this one, I'll resolve it as separate PR, and same for the alerting.

Issue #1660, opened 2026-01-06
Test `test_new_page_with_each_plugin` is flaky


vdusek:
https://github.com/apify/crawlee-python/actions/runs/20748731946/job/59572328541?pr=1656

```
FAILED tests/unit/browsers/test_browser_pool.py::test_new_page_with_each_plugin - TimeoutError: Creating a new page with plugin <crawlee.browsers._playwright_browser_plugin.PlaywrightBrowserPlugin object at 0x1290b3ad0> timed out.
============ 1 failed, 1607 passed, 8 skipped in 565.22s (0:09:25) =============
```

vdusek:
still flaky, see https://github.com/apify/crawlee-python/actions/runs/30372128415/job/90318432735

PR #2156, opened 2026-08-14
fix(browsers): do not close a browser as inactive while a page opening is in flight

vdusek:
The `windows-latest / 3.10` unit-test job [failed](https://github.com/apify/crawlee-python/actions/runs/31604546268/job/94140034785) with `TargetClosedError: BrowserContext.new_page: Target page, context or browser has been closed` in `test_new_page_with_each_plugin`. It is not a test flake: the `BrowserPool` inactive-browser reapers judge a browser only by its finished pages (`idle_time` runs from controller construction and `pages` is populated only once `new_page` returns), so on a slow runner they gracefully close a freshly launched browser whose first `new_page()` is still creating its context. Any user on a loaded machine can hit the same error (related: #1660).

The fix exposes the already-tracked in-flight openings as `BrowserController.is_opening_pages` and guards both reapers with it, so such a browser is neither marked inactive nor closed. A new regression test reproduces the race deterministically by gating the first `context.new_page()` call and running one reaper cycle while the opening is pending — it fails with the exact CI error before the fix and passes after.

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/crawlee/browsers/_browser_controller.py
+++ b/src/crawlee/browsers/_browser_controller.py
@@ -55,6 +55,11 @@ class BrowserController(ABC):
         """Return if the browser has free capacity to open a new page."""
 
     @property
+    def is_opening_pages(self) -> bool:
+        """Return if the browser has any `new_page` calls currently in flight."""
+        return False
+
+    @property
     @abstractmethod
     def is_browser_connected(self) -> bool:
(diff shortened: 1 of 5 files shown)


Mantisus (approved):
LGTM

Pijukatel (approved):

PR #2157, opened 2026-08-14
fix(throttling-manager): Normalize configured domains the same way as crawled hostnames

### LINKED ISSUE #2144: `ThrottlingRequestManager` domain matching misses punycode, trailing dots and IPv6 literals
`domains` is normalized with `d.lower()` while lookups go through `URL(url).host`. yarl normalizes
differently, so three configurations never match anything:

- `domains=['xn--hky-ela4t.cz']` — `URL(...).host` decodes it back to `háčky.cz`
- `domains=['example.com']` against a root-dotted `http://example.com./…` — yarl keeps the dot
- `domains=['[::1]']` — yarl strips the brackets

Each is a silently unthrottled domain. Same symptom as #2126, different cause: there the recording
code never runs, here it runs and looks up a key that cannot exist. The TS port puts both sides
through the URL parser so they land on the same form (apify/crawlee#3741).

From #1762.

Mantisus:
### Description

- Configured `domains` now go through the same URL parsing and normalization as crawled URLs, so punycode, root-dotted and IPv6 spellings match. An entry the parser cannot read now raises `ValueError` instead of silently matching nothing.

### Issues

- Closes: #2144 

### Testing

- Added new unit tests


vdusek:
@janbuchar Does https://github.com/apify/crawlee-python/pull/2158#issuecomment-5328334859 apply to this PR as well, or can this one be reviewed and merged?

janbuchar:
Also free to go, also will check later.

vdusek (approved):
LGTM

PR #2158, opened 2026-08-14
fix(throttling-manager): Apply per-domain backoff when `retry_on_blocked` is disabled

### LINKED ISSUE #2126: `ThrottlingRequestManager` silently does nothing in two configurations
Both cases look identical to a working setup unless you're watching request timings. From #1762.

- With `retry_on_blocked=False`, `_raise_for_session_blocked_status_code` is never called, so `record_domain_delay` never runs and every 429 is ignored.
- Behind a `RequestManagerTandem` — e.g. `request_list.to_tandem(throttler)` — the `isinstance(self._request_manager, ThrottlingRequestManager)` checks in `BasicCrawler` all miss. 429 backoff and robots.txt crawl-delay are both dropped, and the "you are not using `ThrottlingRequestManager`" warning fires at you while you are. The TS port works around this by forwarding `record_domain_delay`/`set_crawl_delay` through the tandem.

#1762 added a fair amount of warning machinery for the other misconfigurations; these two slip past all of it.

Mantisus:
### Description
- Recording a 429 is now separate from the session blocking check and runs regardless of `retry_on_blocked`. That configuration previously dropped the 429 backoff and the robots.txt crawl-delay, while warning that no throttler was in use.

### Issues

- Closes: #2126 

### Testing

- Added new unit tests

### DIFF
--- a/src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
+++ b/src/crawlee/crawlers/_abstract_http/_abstract_http_crawler.py
@@ -303,13 +303,13 @@ class AbstractHttpCrawler(
             The original crawling context if no errors are detected.
         """
         status_code = context.http_response.status_code
+        self._record_rate_limit_status_code(
+            status_code,
+            request_url=context.request.url,
+            retry_after_header=context.http_response.headers.get('retry-after'),
+        )
         if self._retry_on_blocked:
-            self._raise_for_session_blocked_status_code(
-                context.session,
-                status_code,
-                request_url=context.request.url,
-                retry_after_header=context.http_response.headers.get('retry-after'),
-            )
+            self._raise_for_session_blocked_status_code(context.session, status_code)
         self._raise_for_error_status_code(status_code)
         yield context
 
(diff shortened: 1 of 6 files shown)


janbuchar:
Hi @Mantisus and thanks for the PR! I did a bunch more research in the JS port and figured out a different way around this. It involves changes to the request manager, so it will presumably take more time to pass code review.

Just letting you know so that you don't end up thinking I forgot about this :slightly_smiling_face:

janbuchar:
@Mantisus feel free to check https://github.com/apify/crawlee/pull/4061 now — in its current state, it should be pretty clear what direction it's headed. 

However, it's unlikely that it can be replicated here without breaking changes. @vdusek do you think it makes sense to wait for v2, or should we try and think of a backwards-compatible fix?

vdusek:
> @Mantisus feel free to check https://github.com/apify/crawlee/pull/4061 now — in its current state, it should be pretty clear what direction it's headed.
>
> However, it's unlikely that it can be replicated here without breaking changes. @vdusek do you think it makes sense to wait for v2, or should we try and think of a backwards-compatible fix?

If we manage to make it backward-compatible in a reasonable way, I'd go for it. But if that requires introducing some weird quirks/workarounds, I would wait for next major. Especially considering that, nobody has actually reported this issue, if I'm correct.

Mantisus:
I updated the PR. That resolves the issue with `retry_on_blocked` only. This does not require any breaking changes or workarounds.

vdusek (approved):
LGTM

vdusek:
@janbuchar do you want to check this as well?

vdusek:
Merging, this will close #2126, and I opened a follow-up issue for the 2nd case - #2238

PR #2160, opened 2026-08-15
refactor(sessions): move blocked_status_codes from Session to crawler

### LINKED ISSUE #2139: Move `blocked_status_codes` from `Session` to the crawler
Blocked status codes are configured per `Session` (`src/crawlee/sessions/_session.py`), even though deciding whether a response means "blocked" is the crawler's concern - and the crawler already owns `additional_http_error_status_codes` and `ignore_http_error_status_codes`. Move it to a crawler option and keep the handling internal to the crawler.

JS implementation: apify/crawlee#3496

*✍️ Drafted by Claude Code*

aryansk:
Closes #2139

## What

`blocked_status_codes` was configured per `Session` even though deciding whether a response means "blocked" is the crawler's concern — the crawler already owns `additional_http_error_status_codes` and `ignore_http_error_status_codes`. This moves it to a crawler option and keeps the handling internal to the crawler, matching the JS implementation (apify/crawlee#3496).

## Changes

- **`BasicCrawler`** gains a `blocked_status_codes` option, defaulting to `[401, 403, 429]`. The blocked-status check in `_raise_for_session_blocked_status_code` now consults `self._blocked_status_codes - self._ignore_http_error_status_codes` instead of asking the `Session`.
- **`Session`** drops the `blocked_status_codes` parameter, the `is_blocked_status_code()` method, and the internal set; `get_state()` no longer serializes the field.
- **`SessionModel`** drops the `blockedStatusCodes` field.
- **Docs**: the session-management and error-handling examples now configure `blocked_status_codes` on the crawler instead of inside `Session` / `create_session_settings`.
- **Tests**: session and model tests updated for the removed field; new crawler tests cover the default set, a custom set, and the `ignore_http_error_status_codes` interaction.

## Validation

- `ruff check` and `ruff format --check` clean
- `ty check` clean on the touched modules
- Unit suites pass: `tests/unit/sessions/` (29) and `tests/unit/crawlers/_basic/` + `_http/` (196, 1 pre-existing skip)

### DIFF
--- a/docs/guides/code_examples/error_handling/change_handle_error_status.py
+++ b/docs/guides/code_examples/error_handling/change_handle_error_status.py
@@ -4,7 +4,6 @@ import json
 from crawlee import HttpHeaders
 from crawlee.crawlers import HttpCrawler, HttpCrawlingContext
 from crawlee.errors import HttpStatusCodeError
-from crawlee.sessions import SessionPool
 
 # Using a placeholder refresh token for this example
 REFRESH_TOKEN = 'PLACEHOLDER'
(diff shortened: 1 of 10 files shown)


vdusek:
v2 issue, closing

PR #2161, opened 2026-08-15
fix: preserve path and query casing in normalize_url (#2008)

### LINKED ISSUE #2008: `normalize_url` lowercases the entire URL, silently deduplicating case-distinct pages

aryansk:
## Summary

Fixes #2008 — `normalize_url` lowercased the **entire** URL, including the path and query, so any two URLs differing only in path or query casing produced the same unique key:

- `https://example.com/Product/ABC` and `https://example.com/product/abc` → same key
- `https://example.com/?token=SeCrEt` and `https://example.com/?token=secret` → same key

Per [RFC 3986 §6.2.2.1](https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2.1) only the scheme and host are case-insensitive — the path and query are case-sensitive. On sites with case-sensitive paths (base64/hashid identifiers, usernames), case-distinct pages were silently deduplicated with no log message or statistic revealing it.

## Change

`yarl` already lowercases the scheme and host on parse (verified empirically: `URL('HTTP://Example.COM/Product/ABC?Token=SeCrEt')` → `http://example.com/Product/ABC?Token=SeCrEt`). The fix removes the final `.lower()` on the constructed URL so path/query/fragment casing survives normalization, matching browser behavior. The docstring now states precisely which parts are case-folded.

**Breaking change note:** this changes how default unique keys are computed — crawls that relied on case-insensitive dedup will now visit more pages, and keys stored in persisted queues won't match newly computed ones. Per the issue, this should land in a major (2.0) release with the change documented loudly.

## Validation

- New parametrized cases: scheme/host still lowercased (`HTTPS://EXAMPLE.COM/?KEY=VALUE` → `https://example.com/?KEY=VALUE`), path/query/fragment case preserved, trailing-slash removal unaffected.
- New regression test: case-distinct paths normalize to different URLs (dedup no longer collapses them).
- `tests/unit/_utils` — 299 passed, ruff check + format clean.

### DIFF
--- a/src/crawlee/_utils/requests.py
+++ b/src/crawlee/_utils/requests.py
@@ -17,10 +17,11 @@ def normalize_url(url: str, *, keep_url_fragment: bool = False) -> str:
     """Normalize a URL.
 
     This function cleans and standardizes a URL by removing leading and trailing whitespaces,
-    converting the scheme and netloc to lower case, stripping unwanted tracking parameters
+    converting the scheme and host to lower case, stripping unwanted tracking parameters
     (specifically those beginning with 'utm_'), sorting the remaining query parameters alphabetically,
-    and optionally retaining the URL fragment. The goal is to ensure that URLs that are functionally
-    identical but differ in trivial ways (such as parameter order or casing) are treated as the same.
+    and optionally retaining the URL fragment. Per RFC 3986 only the scheme and host are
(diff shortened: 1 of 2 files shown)


vdusek:
v2 issue, closing

PR #2162, opened 2026-08-15
fix(sitemaps): allow Sitemap.from_xml_string to opt into enqueue host filtering

### LINKED ISSUE #2118: `Sitemap.from_xml_string` cannot opt into enqueue host filtering
Found while triaging GHSA-c533-9w4q-667c (which I'm closing as informative - a raw XML string has no origin, so there's no host policy to bypass). It does point at a real API gap though.

- [`Sitemap.from_xml_string`](https://github.com/apify/crawlee-python/blob/5882c6b51a0134097615c6d4e12a3765d963fc28/src/crawlee/_utils/sitemap.py#L494-L496) takes only `content` and hardcodes `SitemapSource(type='raw', content=content)` - no `url`, no `ParseSitemapOptions`
- host filtering is [conditional on `source.get('url')`](https://github.com/apify/crawlee-python/blob/5882c6b51a0134097615c6d4e12a3765d963fc28/src/crawlee/_utils/sitemap.py#L282-L287)
- so if you have raw sitemap XML *and* you know where it came from, there's no way to say so - filtering is silently skipped and the [`same-hostname` default](https://github.com/apify/crawlee-python/blob/5882c6b51a0134097615c6d4e12a3765d963fc28/src/crawlee/_utils/sitemap.py#L531) never applies
- suggestion: `from_xml_string(content, *, sitemap_url: str | None = None, parse_sitemap_options: ParseSitemapOptions | None = None)`, forwarded as `SitemapSource(type='raw', content=content, url=sitemap_url)`
- explicitly *not* worth doing: rejecting `<loc>` values when the origin is unknown - that makes the default behaviour return an empty list
- while in there: [`_get_origin_url`](https://github.com/apify/crawlee-python/blob/5882c6b51a0134097615c6d4e12a3765d963fc28/src/crawlee/_utils/sitemap.py#L231-L238) stamps raw sources with `raw://<sha256>`, so `SitemapUrl.origin_sitemap_url` is a pseudo-URL that would fail scheme validation if anyone fed it back into `filter_url` - harmless today, footgun later

aryansk:
## Summary

`Sitemap.from_xml_string` only accepted `content` and hardcoded `SitemapSource(type='raw', content=content)` with no `url`, so host filtering — conditional on `source.get('url')` — was silently skipped for raw sitemap XML. Even when the caller knew where the content came from, the default `same-hostname` enqueue strategy never applied.

This adds optional `sitemap_url` and `parse_sitemap_options` parameters, forwarded as `SitemapSource(type='raw', content=content, url=sitemap_url)` so raw content participates in the exact same host-filtering and parse options as URL-loaded sitemaps.

Closes #2118

## Changes

- `src/crawlee/_utils/sitemap.py` — `from_xml_string(content, *, sitemap_url=None, parse_sitemap_options=None)`.
- `tests/unit/_utils/test_sitemap.py` — three regression tests: same-host URLs kept with the default strategy, cross-host URLs filtered when `sitemap_url` is given, and `enqueue_strategy='all'` disabling the filter for raw string sitemaps. All three fail on the pre-fix code.

## Validation

- `tests/unit/_utils/test_sitemap.py`: 67 passed (3 new).
- `tests/unit/_utils/`: 297 passed, 6 skipped (Linux-only).
- `tests/unit/request_loaders/test_sitemap_request_loader.py`: 46 passed.
- `ruff check` + `ruff format --check` clean on changed files.

🤖 Generated with Codebuff

### DIFF
--- a/src/crawlee/_utils/sitemap.py
+++ b/src/crawlee/_utils/sitemap.py
@@ -496,8 +496,17 @@ class Sitemap:
         )
 
     @classmethod
-    async def from_xml_string(cls, content: str) -> Sitemap:
-        return await cls.parse([SitemapSource(type='raw', content=content)])
+    async def from_xml_string(
+        cls,
+        content: str,
+        *,
+        sitemap_url: str | None = None,
+        parse_sitemap_options: ParseSitemapOptions | None = None,
+    ) -> Sitemap:
+        source: SitemapSource = {'type': 'raw', 'content': content}
+        if sitemap_url is not None:
+            source['url'] = sitemap_url
+        return await cls.parse([source], parse_sitemap_options=parse_sitemap_options)
 
     @classmethod
     async def parse(
--- a/src/crawlee/_utils/sitemap.py
+++ b/src/crawlee/_utils/sitemap.py
@@ -234,10 +234,10 @@ def _get_parser(content_type: str = '', url: str | None = None) -> _XmlSitemapPa
 
 def _get_origin_url(source: SitemapSource) -> str:
     """Determine the origin URL for a sitemap source."""
-    if source['type'] == 'url' and 'url' in source:
-        return source['url']
+    if url := source.get('url'):
+        return url
     if source['type'] == 'raw' and 'content' in source:
-        # For raw content sources, create a consistent identifier
+        # Raw sources without a known URL get a consistent content-derived identifier.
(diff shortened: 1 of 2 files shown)


vdusek (approved):
LGTM

PR #2165, opened 2026-08-15
fix: make RequestQueueClient.is_finished abstract

### LINKED ISSUE #1985: Make the `is_finished` method abstract for `RequestQueueClient`
In the current implementation, the `is_empty` method is used by default for `is_finished`. This is a workaround designed to avoid breaking changes when the `is_finished` method was added.

aryansk:
Fixes #1985


vdusek:
v2 issue, closing

PR #2166, opened 2026-08-16
fix: align URL pattern matching and limit semantics with crawlee-js

### LINKED ISSUE #2122: Align glob case-sensitivity, regexp anchoring, and limit/transform ordering with crawlee-js
Follow-up from [apify/crawlee#3533](https://github.com/apify/crawlee/pull/3533) (aligning JS `enqueueLinks` with the Python API, #3409). During review, [@vdusek found three remaining behavioral differences](https://github.com/apify/crawlee/pull/3533#pullrequestreview-4721691629) between the two implementations that weren't in scope for the JS-side PR. The team agreed ([1](https://github.com/apify/crawlee/pull/3533#issuecomment-5002323608), [2](https://github.com/apify/crawlee/pull/3533#issuecomment-5002922732), [3](https://github.com/apify/crawlee/pull/3533#pullrequestreview-4722436194)) that these should be evaluated for alignment on the Python side rather than the JS side. Filing as one issue since all three affect the same `enqueue_links`/URL-filtering pipeline.

### 1) Glob case sensitivity

JS uses `Minimatch` with `nocase: true` — glob matching is case-insensitive. Python's `Glob` translates to a regex without `re.IGNORECASE` (`_utils/globs.py`) — case-sensitive. The same glob pattern matches different URLs in each library.

Suggested direction (per @B4nan): make Python's glob matching case-insensitive to match JS.

### 2) Regexp anchoring

Python calls `pattern.match(url)`, which anchors at the start of the string. JS calls `regexp.test(url)`, which searches anywhere in the string. A regex like `/\/products\//` matches `https://x.com/products/1` in JS but not in Python (there it would need to start with `https?://…`).

Suggested direction (per @B4nan, tentative — "curious about others' opinions"): make Python's regex matching unanchored to match JS.

### 3) `limit` vs. `transform` ordering

Python applies `limit` in the filter iterator, before `transform` runs, so requests dropped via a skip action still consume the limit. JS applies `limit` last (after transform), so skipped requests get backfilled by others. With `limit: 5` and a transform skipping 2 of them, Python enqueues 3, JS enqueues 5.

Suggested direction (per @B4nan): apply `limit` last in Python, to match JS.

---

None of these are must-fix — the working assumption in the JS PR discussion was that JS's behavior is the more intuitive one and Python should move toward it, but that's up for debate here. See the [full review comment](https://github.com/apify/crawlee/pull/3533#pullrequestreview-4721691629) for exact line references on the JS side.

atirna:
### Description

Aligns `enqueue_links` / `add_requests` URL filtering with crawlee-js, following up on the differences found in the apify/crawlee#3533 review (thanks @vdusek for the detailed writeup).

- `Glob` patterns now match case-insensitively, like `Minimatch` with `nocase: true` in JS (`_utils/globs.py`)
- regex `include`/`exclude` patterns are matched with `search` instead of `match`, so they match anywhere in the URL like `regexp.test` in JS (crawler + `SitemapRequestLoader`, which mirrors `url.match` on the JS side)
- `limit` now counts requests actually enqueued: applied last, after `transform_request_function` skipping, and duplicates already present in the queue don't consume it (JS decrements its `maxNewRequests` budget only for requests that weren't already there)

One thing worth calling out since @B4nan wanted more opinions on the regexp anchoring: I checked the JS side, and both `enqueueLinks` (`createPatternObjectMatcher` in `packages/core/src/enqueue_links/shared.ts`) and the sitemap loader (`url.match(regexp)` in `packages/core/src/storages/sitemap_request_loader.ts`) search unanchored, so a pattern like `/\/products\//` matching anywhere in the URL is the established cross-library behavior. That's what this implements.

### Issues

- Closes: #2122

### Testing

before, on current `master` (`172699b1`):

- `Glob('https://Someplace.com/**/cats')` did not match `https://someplace.com/blog/category/cats`
- `include=[re.compile(r'/category/cats')]` enqueued nothing (the regex never matched at the start of the URL)
- with `limit=2` and a transform skipping 2 of 3 extracted links, the remaining link was not enqueued (limit burned slots on skipped URLs); URLs already in the queue also consumed the limit

after:

- `uv run pytest tests/unit/_utils/test_globs.py tests/unit/crawlers/_basic tests/unit/crawlers/_beautifulsoup tests/unit/request_loaders/test_sitemap_request_loader.py` — new regression tests for all of the above (each fails on `master`, passes here), existing suite unaffected
- `uv run poe unit-tests` — 2296 passed, 11 skipped
- `uv run poe lint` / `uv run poe type-check` — clean

### Checklist

- [x] CI passed


vdusek:
Hi @atirna, thanks for your work. However, this is planned for v2 (there's a milestone assigned to the issue), as this change introduces breaking behavior. So I'll have to close this PR for now.

PR #2173, opened 2026-08-18
fix(playwright): scope custom request headers to the navigation request

vdusek:
## Description

- `PlaywrightCrawler._navigate()` applied `Request.headers` page-wide via `page.set_extra_http_headers()`, so the browser attached them to every request the page initiated - including cross-origin subresources chosen by the crawled page. Sensitive per-request values (e.g. `Authorization`) could leak to third-party origins.
- Headers are now applied through the route interceptor already used for non-GET requests, matched to the first main-frame navigation request itself rather than to its URL (a URL match breaks on browser URL normalization and on glob metacharacters in the URL). The interceptor merges the custom headers into the headers the browser sends on its own (`User-Agent`, fingerprint headers) instead of replacing them.
- Behavior change: custom headers are sent with the navigation request (and inherited by its redirect hops), but no longer with any page-initiated request afterwards - subresources, XHRs, or later client-side navigations.
- Covers `PlaywrightCrawler`, `AdaptivePlaywrightCrawler`, and `StagehandCrawler`, which all share `_navigate()`. Mirrors how crawlee JS scopes request headers in `gotoExtended`.
- Added regression tests: cross-origin subresources don't receive the headers, browser-normalized URLs still do, and redirect hops inherit them.

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/crawlee/crawlers/_playwright/_playwright_crawler.py
+++ b/src/crawlee/crawlers/_playwright/_playwright_crawler.py
@@ -356,18 +356,18 @@ class PlaywrightCrawler(
         headers: HttpHeaders | dict[str, str] | None = None,
         payload: HttpPayload | None = None,
     ) -> Callable:
-        """Create a request interceptor for Playwright to support non-GET methods with custom parameters.
-
-        The interceptor modifies requests by adding custom headers and payload before they are sent.
+        """Create a request interceptor that applies a custom method, headers, and payload to matching requests.
 
         Args:
             method: HTTP method to use for the request.
-            headers: Custom HTTP headers to send with the request.
+            headers: Custom HTTP headers to send with the request. They are merged into the headers the browser
+                would send on its own (e.g. `User-Agent` or fingerprint headers), with the custom ones winning.
             payload: Request body data for POST/PUT requests.
         """
 
-        async def route_handler(route: Route, _: PlaywrightRequest) -> None:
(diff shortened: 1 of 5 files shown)


src/crawlee/crawlers/_playwright/_playwright_crawler.py
```diff
@@ -411,14 +408,30 @@ async def _navigate(
...
                 headers=context.request.headers,
                 payload=context.request.payload,
             )
-
-            # Set route_handler only for current request
-            await context.page.route(context.request.url, route_handler)
+            applied = False
+
+            async def navigation_route_handler(route: Route, request: PlaywrightRequest) -> None:
+                nonlocal applied
```
Pijukatel: So it adds this function that tracks its state through an outer variable, which it can change. This works, but it makes the already complex function `_navigate` even more complex by showing all these internal details, and that makes it harder to read.

Could we rather extract this new stuff to some dedicated place? Might be even a tiny helper class, since it anyway has a state to track?
I think that it would be sufficient for the readers of  `_navigate` to see one line about `correctly applying headers`, but details of what `correctly applying headers` mean could be in a dedicated single-purpose place.

vdusek: ok, let's transform it into some helper class/method

Pijukatel (approved):

PR #2176, opened 2026-08-19
fix(throttling-manager): Stop holding a concurrency slot while a domain is throttled

### LINKED ISSUE #2127: `ThrottlingRequestManager.fetch_next_request` holds a concurrency slot while waiting out a throttle
- When every configured domain is throttled and the inner manager is empty, `fetch_next_request` sleeps for up to `max_delay` rather than returning. It's called from `__run_task_function`, so that's an autoscaled-pool task slot held open doing nothing — which the snapshotter is liable to read as spare capacity and answer by scaling up.
- The TS port returns `None` and has `is_empty()` report `True` so the task loop idles instead (apify/crawlee#3741). Costs a contract bend — `is_empty()` is `True` while requests still exist — but it stops lying to the autoscaler.
- From #1762.

Mantisus:
### Description

- `fetch_next_request` now returns `None` instead of sleeping out a cooldown, and `is_empty()` counts only domains that are not throttled, so the autoscaled pool idles instead of holding worker slots. `is_finished()` still counts every domain, so the crawl does not end while throttled requests are queued. The old sleep occupied a slot while the system read as idle, which pushed `desired_concurrency` up during the cooldown and burst the moment it expired.

### Issues

- Closes: #2127

### Testing

- Update unit tests


vdusek (approved):
LGTM

PR #2181, opened 2026-08-20
docs: fix code block line number alignment

barjin:
The line numbers were `position: sticky`, so they stayed pinned and overlapped the code.

|before|after|
|---|---|
| image |  image |


vdusek (approved):
thanks

PR #2200, opened 2026-08-28
test: Deflake Redis RQ metadata timestamp test with a controlled clock

vdusek:
`test_metadata_file_updates` in `tests/unit/storage_clients/_redis/test_redis_rq_client.py` failed on Windows / Python 3.12 with `assert accessed_at > accessed_at` on identical timestamps ([run](https://github.com/apify/crawlee-python/actions/runs/33165535500/job/98829928097)).

The Redis `is_empty()` never stamps `accessed_at`, and `get_metadata()` reads the record before stamping its own access. The two values the test compared were therefore the stamps of two back-to-back `get_metadata()` calls (the fixture's purge check and the test's first call), about a millisecond apart, with the `asyncio.sleep(0.01)` sitting after both. On Windows before Python 3.13, `datetime.now()` ticks at 1–15.6 ms, so the stamps collide.

The test now drives the mixin's clock through a `Mock`, uses a read the Redis client does treat as an access (`fetch_next_request()`), and asserts the exact stamps – no sleeps, no wall clock. Removing the `accessed_at` update from `fetch_next_request()` makes it fail, so it still guards the behaviour.

Locally, quantizing the mixin clock to 15.625 ms ticks reproduced the failure 18/20 times (2/40 at 1 ms); the new test passes 20/20 and 40/40 under the same clocks, and the Redis test directory passes 20/20 with `-n auto`.

Note: `is_empty()` bumps `accessed_at` in the memory and file-system clients but not in the Redis one. Left as is here; worth deciding separately whether that's intended.

*✍️ Drafted by Claude Code*

### DIFF
--- a/tests/unit/storage_clients/_redis/test_redis_rq_client.py
+++ b/tests/unit/storage_clients/_redis/test_redis_rq_client.py
@@ -2,14 +2,16 @@ from __future__ import annotations
 
 import asyncio
 import json
+from datetime import timedelta
 from typing import TYPE_CHECKING
-from unittest.mock import AsyncMock, MagicMock, patch
+from unittest.mock import AsyncMock, MagicMock, Mock, patch
 
 import pytest
 from redis.exceptions import RedisError
 
 from crawlee import Request
 from crawlee.storage_clients import RedisStorageClient
+from crawlee.storage_clients._redis import _client_mixin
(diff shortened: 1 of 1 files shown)


Mantisus (approved):
LGTM

Pijukatel (approved):

PR #2204, opened 2026-08-31
fix!: do not count downtime towards crawler_runtime after migration or resurrection

### LINKED ISSUE #1567: Remove deprecated `StatisticsState.crawler_runtime` setter
The setter was added to avoid breaking changes and is no longer doing anything. It can be removed in next breaking release.

See: https://github.com/apify/crawlee-python/pull/1562#discussion_r2528436110

Pijukatel:
### Description

Fixes `crawler_runtime` miscounting when a run is resumed from a persisted state (migration, resurrection, abort), targeting the v2 release. End-user behavior is unchanged except for the runtime-counting fixes.

Guarantees after this change:

- A fresh run's first statistics log reports a runtime of exactly zero.
- A resumed run's first statistics log reports exactly the runtime of the previously persisted state, not the downtime between the runs (previously it included the whole downtime, e.g. 38s, and "corrected" itself later).
- The runtime accumulates across any number of persist/resume cycles: it is restored from the persisted `crawlerRuntimeMillis` value, which every persist writes as the live total, so nothing is lost with repeated resurrections (previously only the last run's segment survived).
- Aborts are handled: a state persisted mid-run carries the total up to that moment, and a resumed run continues from the last persisted total, losing only the tail since the last persist. For states persisted by older versions (no usable `crawlerRuntimeMillis`), the end of an uncleanly finished run is approximated by `stats_persisted_at`, so downtime is not baked into the total.
- Robust loading: an invalid persisted `crawlerRuntimeMillis` (null, garbage) does not fail loading the state - it falls back to the timestamp reconstruction; a negative value is clamped to zero.

### Breaking changes (v2)

- `StatisticsState.crawler_runtime` is a read-only property; the deprecated setter is removed (closes #1567).
- `StatisticsState.crawler_runtime_for_serialization` is removed; the `runtime_offset` field serializes the live runtime under `crawlerRuntimeMillis` and restores it on load.
- `crawlerRuntimeMillis` is serialized as a number of milliseconds (matching the other `*Millis` fields and Crawlee for JavaScript) instead of an ISO 8601 duration string. States persisted by v1.x load correctly.
- The never-populated state fields `errors`, `retry_errors`, `requests_finished_per_minute` and `requests_failed_per_minute` are removed from `StatisticsState`. The `FinalStatistics` fields of the same names and the logged per-minute rates are unaffected.

The changes are documented in the new `docs/upgrading/upgrading_to_v2.md`.

### Issues

- Closes: #1567
- The miscounting itself was reported via a resurrected Actor run log where the initial statistics table showed `crawler_runtime` of 38.01s that later dropped to 542.8ms; no issue filed.

### Testing

- Regression tests in `tests/unit/_statistics/test_persistence.py` cover: fresh run's first log is exactly zero; resumed run's first log equals the persisted runtime without a simulated 2-hour downtime; the total accumulates across multiple resurrections; unclean shutdowns do not bake downtime into the total; v1.x ISO-string states load; null/garbage `crawlerRuntimeMillis` falls back to timestamps; negative values are clamped.
- `poe lint`, `poe type-check`, and the unit test suite pass locally on Python 3.10 and 3.13 (only the pre-existing browser-environment failures unrelated to this change).

### Checklist

- [ ] CI passed

https://claude.ai/code/session_01VMHwDx87jV4rnu5LkbpNxy

### DIFF
--- a/docs/upgrading/upgrading_to_v2.md
+++ b/docs/upgrading/upgrading_to_v2.md
@@ -1,0 +1,15 @@
+---
+id: upgrading-to-v2
+title: Upgrading to v2
+---
+
+This page summarizes the breaking changes between Crawlee for Python v1.x and v2.0.
+
+## Statistics
+
+The runtime tracking in `StatisticsState` was reworked to fix `crawler_runtime` miscounting after migrations and resurrections. The runtime accumulated by previous runs is now restored from the persisted `crawlerRuntimeMillis` value, so it survives any number of interruptions, and the downtime between the runs is no longer counted.
+
+- `StatisticsState.crawler_runtime` is a read-only property now. The deprecated setter was removed; assigning to it raises `AttributeError` instead of emitting a `DeprecationWarning`.
+- `StatisticsState.crawler_runtime_for_serialization` was removed. The persisted state still contains `crawlerRuntimeMillis`, now written by the `runtime_offset` field, which also restores the value when a state is loaded.
+- `crawlerRuntimeMillis` is serialized as a number of milliseconds, consistent with the other `*Millis` fields and with Crawlee for JavaScript, instead of an ISO 8601 duration string. States persisted by v1.x load correctly.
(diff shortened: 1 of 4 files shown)


Pijukatel:
Waiting for V2 as this might be breaking change and will inlcude removal of deprecated fields

PR #2213, opened 2026-09-07
fix: Instrument httpx2 requests in the otel extra

vdusek:
`HttpxHttpClient` runs on `httpx2`, and `opentelemetry-instrumentation-httpx` only gained `HTTPX2ClientInstrumentor` in 0.65b0, so requests through it produced no spans. Every OpenTelemetry floor in the `otel` extra now sits on that release line.

| package | before | after |
| --- | --- | --- |
| `opentelemetry-api` | `>=1.34.1` | `>=1.44.0` |
| `opentelemetry-distro[otlp]` | `>=0.54` | `>=0.65b0` |
| `opentelemetry-instrumentation` | `>=0.54` | `>=0.65b0` |
| `opentelemetry-instrumentation-httpx` | `>=0.54` | `>=0.65b0` |
| `opentelemetry-sdk` | `>=1.34.1` | `>=1.44.0` |
| `opentelemetry-semantic-conventions` | `>=0.54` | `>=0.65b0` |

Verified against the wheels: 0.64b0 registers a single `httpx` entry point, while 0.65b0 adds the class and an `httpx2` entry point. With an in-memory exporter, `HTTPX2ClientInstrumentor` emits a span for an `HttpxHttpClient` request; `HTTPXClientInstrumentor` emits none.

The whole group moves together because these packages release in lockstep and pin each other with `==`. `opentelemetry-instrumentation-httpx==0.65b0` requires `opentelemetry-instrumentation==0.65b0` and `opentelemetry-semantic-conventions==0.65b0`, and the latter requires `opentelemetry-api==1.44.0`; `opentelemetry-sdk` pins `api` to its own version, so it lands on 1.44.0 as well. `uv pip compile --resolution lowest-direct` against the old specifiers resolves to exactly the "after" column, so the previous floors only looked permissive.

Every release of the 0.x packages is a beta, so the floors name the prerelease explicitly. `>=0.65` matches no published version at all, since PEP 440 orders `0.65b0 < 0.65`. `uv.lock` had already resolved 0.65b0 and 1.44.0, so only its `requires-dist` specifiers moved.

The web form example corrections that used to ride along here moved to #2237.

*✍️ Drafted by Claude Code*


Mantisus:
If I'm not mistaken, `opentelemetry-instrumentation-httpx` was added as a dependency when `HttpxHttpClient` was the default client. Maybe we can remove that?

pyproject.toml
```diff
@@ -72,11 +72,12 @@ parsel = ["parsel>=1.10.0"]
 playwright = ["playwright>=1.27.0", "apify_fingerprint_datapoints>=0.0.2", "browserforge>=1.2.3"]
 otel = [
     "opentelemetry-api>=1.34.1",
-    "opentelemetry-distro[otlp]>=0.54",
-    "opentelemetry-instrumentation>=0.54",
-    "opentelemetry-instrumentation-httpx>=0.54",
+    "opentelemetry-distro[otlp]>=0.54b0",
```
Pijukatel: I think you should bump all the instrumentation-related 0.54b0 to 0.65b0.

For unrelated Python packages, this would not be needed, but those instrumentation packages are always bumped all together.

So, for example, this gives the illusion of flexibility
```
opentelemetry-semantic-conventions>=0.54b0
opentelemetry-instrumentation-httpx>=0.65b0
```
, but in reality it resolves as  
```
opentelemetry-semantic-conventions>=0.65b0
opentelemetry-instrumentation-httpx>=0.65b0
```

Mantisus (approved):

vdusek:
> If I'm not mistaken, opentelemetry-instrumentation-httpx was added as a dependency when HttpxHttpClient was the default client. Maybe we can remove that?

cc @Pijukatel

Pijukatel:
> > If I'm not mistaken, opentelemetry-instrumentation-httpx was added as a dependency when HttpxHttpClient was the default client. Maybe we can remove that?
> 
> cc @Pijukatel

I guess we can. It would be useful only if someone uses `httpx` and `otel` extras, and if desired, one can add it on their own.

Pijukatel:
> > If I'm not mistaken, opentelemetry-instrumentation-httpx was added as a dependency when HttpxHttpClient was the default client. Maybe we can remove that?
> 
> cc @Pijukatel

Only useful if someone installs both otel and httpx extras. I guess we can leave it out, and anyone interested can install it manually.

Pijukatel (approved):

PR #2229, opened 2026-09-14
fix: propagate autoscaled pool scheduling errors

Kuang-xianxin:
### Description

When a request manager's `is_empty()` or `is_finished()` raises, `BasicCrawler.run()` currently logs the orchestrator error but returns normally with final statistics. A failed queue status check can therefore look like a successfully completed crawl while requests remain pending.

Retain scheduling exceptions until the existing worker cleanup finishes, then fail the pool's result future instead of resolving it successfully. Preserve a result already set by a worker failure or abort. The regression tests exercise the crawler with a real filesystem queue, inject errors only at its status methods, and verify that a new crawler can process the pending request after recovery. A separate event-controlled test checks that an active worker finishes before the scheduling error reaches the caller.

### Issues

Found during source review; no existing issue linked. This concerns scheduling callbacks, distinct from the worker-task timeout handling in #2009.

### Testing

- Five new regression cases fail on unmodified `0853d79e670cba27f9692919688305d1cc565c6c` because no exception reaches the caller.
- Python 3.10.20: all five regression cases pass.
- Python 3.13.14: `pytest tests/unit/_autoscaling tests/unit/crawlers/_basic tests/unit/_utils/test_recurring_task.py -k 'not test_send_request_works'` passes 157 tests, with 1 skipped and 2 deselected.
- Full Ruff lint and formatting checks pass (599 files); full `ty check --python-platform linux` and changed-file native Windows typing pass.
- Native Windows full typing reports two `BaseContext.Process` diagnostics in unchanged `test_system.py`; the same diagnostics occur on the unmodified base. The two existing `test_send_request_works` cases fail with JSON decoding errors against the local test server on both the patch and the unmodified base; they are excluded from the final related-suite run. The full repository/browser/service matrix was not run locally.

### Checklist

- [ ] CI passed


vdusek (approved):
LGTM. @Pijukatel please take a look as well.

Pijukatel (approved):

PR #2235, opened 2026-09-15
fix(autoscaling): Keep `min_concurrency` tasks running while the system is overloaded

Mantisus:
### Description

- Keep `min_concurrency` tasks running while the system is overloaded instead of dropping to zero concurrency. It fixes `AutoscaledPool` freezing after an overload.

### Testing

- Added a new unit test `test_keeps_min_concurrency_when_overloaded`


vdusek (approved):
LGTM... But @Pijukatel could you take a look as well? I don't think this could cause any harm, but touching the autoscaled pool always makes me nervous 😄

Pijukatel (approved):
It seems reasonable to me, but this would be worth syncing with the JS version of the tooling - especially with the upcoming v4 release.

vdusek:
Issue: https://github.com/apify/crawlee/issues/4129, but it seems a JS version already behaves this way.

Mantisus:
> but it seems a JS version already behaves this way.

Yes, I checked the behavior in TS before making the fix.

PR #2247, opened 2026-09-22
fix: make `extra_name` of `try_import` optional and the install hint reusable

Pijukatel:
`extra_name` landed in #2208 as a required keyword-only argument in a patch release, breaking every external caller: 
`import apify.scrapy` fails with

> TypeError: try_import() missing 1 required keyword-only argument: 'extra_name'

This gives it a default — every call site here passes it explicitly, so user-facing messages are unchanged — and adds `package_name`, since the hint hardcoded `crawlee[<extra>]` and pointed SDK users at an extra that does not exist here.

Also fixes two `IndexError` already in the helper: `e.args[0]` on an argless `ImportError` (the nested-guard case the docstring anticipates), and `extra_name[0]` on an empty list.


janbuchar (approved):
LGTM

